PR #13, opened 2025-04-05
refactor and tests for EvmStateMachine

daiagi:
- Brake EvmStateMachine.ts into easier sub-classes
- Improve test coverage


src/evm/EvmStateMachine.ts
```diff
@@ -1,256 +1,208 @@
...
         await Clock.init(signer.provider!);
 
-        deployedStateChannelContractInstance =
-            deployedStateChannelContractInstance.connect(signer);
-        if (DEBUG_CHANNEL_CONTRACT)
+        // Connect signer to state channel contract
+        deployedStateChannelContractInstance = deployedStateChannelContractInstance.connect(signer);
+
+        // Apply debug proxy if enabled
+        const DEBUG_CHANNEL_CONTRACT = true;
```
MrishoLukamba, line 156: should be false by default

.mocharc.json
```diff
@@ -0,0 +1,5 @@
+{
+    "require": "hardhat/register",
+    "timeout": 40000,
+    "_": ["test/**/*.ts"]
+  }
```
lukastanisic99, line 5: are you testing with mocha directly or hardhat?

contracts/test/SimpleNumberStorage.sol
```diff
@@ -0,0 +1,30 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.8;
+
```
lukastanisic99, line 3: Good idea to have test contracts - just add to .npmignore so they don't become part of the build

src/evm/ContractExecuter.ts
```diff
@@ -0,0 +1,39 @@
+import { EVM, EVMResult, ExecResult } from "@ethereumjs/evm";
+import { Address } from "@ethereumjs/util";
+import { BytesLike, ethers } from "ethers";
+
+
+export class ContractExecuter {
+    private readonly evm: EVM;
+    private readonly contractAddress: Address;
+
```
lukastanisic99, line 9: This file make sense (for now) since we're just executing a single contract (state machine) p2p. Essentially we don't need to isolate the single contract like this, since after deployment we can pass the contract address back to the ethers contract instance and it can target it directly - the only reason we're doing it like this is because we didn't abstract/implement the RPC node that's usually between the EVM and the client(ethers), so like this we can easily push logs and other stuff without worrying about the rest. Long term no need to isolate this.  If we implemented the RPC node, we wouldn't have to be opinionated on ethers either - but for now this is more than enough and in V2 we can generalize it further. 

But in general modularization like this (wrapping) makes sense - we'll do it also for auditing and similar isolated logic that run within the VM.

src/evm/EvmStateMachine.ts
lukastanisic99: Doesn't matter for development you change it to what you need it - it was part of an .env file but removed it since had problems in browsers - however this should be at the top of the file (like in other places) so we don't have to dig to find it 

In production should default to false - maybe instead of a .env file we do a JSON (that works in both nodejs environments and browsers) and for production a build script changes all values to false before building it

src/evm/P2pInteraction.ts
```diff
@@ -0,0 +1,32 @@
+import P2pSigner from "./P2pSigner";
+import { AStateMachine } from "@typechain-types";
+import P2pEventHooks from "@/P2pEventHooks";
+
+/**
+ * Represents a P2P interaction with a state machine contract
+ */
+export class P2pInteraction<T extends AStateMachine> {
+    p2pContractInstance: T;
```
lukastanisic99, line 9: Need a better name for this - this is the dumbest name I came up with in the entire code base and this is the only thing the user of the SDK sees

src/evm/StateMachine.ts
```diff
@@ -0,0 +1,164 @@
+import { ethers, BytesLike, hexlify } from "ethers";
+import { TransactionStruct } from "@typechain-types/contracts/V1/DataTypes";
+import { ContractExecuter } from "./ContractExecuter";
+import { EVM, Log } from "@ethereumjs/evm";
+import { Address } from "@ethereumjs/util"
+/**
+ * Provides a typed interface to interact with a state machine contract
```
lukastanisic99, line 7: Don't see the need for this file and it's confusing since the EvmStateMachine is the state machine that's used polymorphically (implements AStateMachine) so this is just confusing and doesn't do anything special - I do understand that the EvmStateMachine wasn't refactored and that some functionality could be wrapped into a function and reused (that's good) but this file is not needed - I would refactor the few useful functions within the EvmStateMachine and avoid the confusion of multiple stateMachine files (even our contracts have these names)

src/evm/StateMachine.ts
```diff
@@ -0,0 +1,164 @@
...
+    constructor(
+        contractExecuter: ContractExecuter,
+        contractInterface: ethers.Interface
+    ) {
+        this.contractExecuter = contractExecuter;
+        this.contractInterface = contractInterface;
+    }
+
+    private callData(functionName: string, args: any[] = []): Uint8Array {
+        return ethers.getBytes(this.contractInterface.encodeFunctionData(functionName, args));
```
lukastanisic99, line 29: This function makes sense - would name it getEncodedCalldata

src/evm/StateMachine.ts
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    /**
+ * Creates a standardized error with context information
+ * @param methodName The name of the method where the error occurred
+ * @param error The original error
+ * @returns A new Error with standardized format
+ */
+    private createContextError(methodName: string, error: unknown): Error {
+        const errorMessage = error instanceof Error ? error.message : String(error);
```
lukastanisic99, line 39: probably this one also makes sense

lukastanisic99:
What we need and what I just realized is automatic formatting of code on commit so diff doesn't trigger on spaces, new lines etc... The repo uses prettier. I'm fine with that or something else as long as it can be automated on commit so we don't forget to run it (I thin prettier can be automated with husky) - we should a agree on this and the formatting style, automate it, create a PR that does just that and continue from there

.mocharc.json
daiagi: this file allows for nice testing UI in vs code:

image

src/evm/EvmStateMachine.ts
daiagi: moved to top. created issue to handle the envs / jsons

src/evm/StateMachine.ts
daiagi: removed

src/evm/StateMachine.ts
daiagi: done

src/evm/StateMachine.ts
daiagi: done

src/evm/StateMachine.ts
daiagi: renames to StateChannelSession

contracts/test/SimpleNumberStorage.sol
daiagi: added

src/evm/P2pInteraction.ts
daiagi: renamed to StateChannelSession

src/evm/StateChannelSession.ts
```diff
@@ -0,0 +1,22 @@
+import P2pSigner from "./P2pSigner";
+import { AStateMachine } from "@typechain-types";
+import P2pEventHooks from "@/P2pEventHooks";
+
+export class StateChannelSession<T extends AStateMachine> {
+    p2pContractInstance: T;
```
lukastanisic99, line 6: Now on the train so have some time to think -> would rename this to P2pInstance - don't like the word "session" since it reminds me of web2 - everything else looks good

lukastanisic99 (dismissed):

daiagi:
>@[daiagi](https://github.com/daiagi) daiagi dismissed [lukastanisic99](https://github.com/lukastanisic99)’s [stale review](https://github.com/peer3to/state-channels-plus/pull/13#pullrequestreview-2752403450) via [2e26d89](https://github.com/peer3to/state-channels-plus/commit/2e26d89107fb19ba841b1e25872bac7196a82c6e)


@lukastanisic99  i didn't mean to do that. please approve again

lukastanisic99 (approved):

PR #18, opened 2025-04-07
Dispute contract and E2E testing

### LINKED ISSUE #17: Dispute Contract Testing and E2E dispute handler testing
Also re implementing some dispute logic per doc specification

MrishoLukamba:
Closes #17


daiagi:
Next time when opening a PR please assign me and Luka as reviewers, otherwise we don't get notification. Also please assign yourself to it

as for the PR itself
I've got no issue with approving it, as it provides skeleton for testing, but i would not say it closes any issue. 

a. it doesn't (yet) test anything
b. the issue says "Also re implementing some dispute logic per doc specification" -> which this PR definitely does not do

MrishoLukamba:
> Next time when opening a PR please assign me and Luka as reviewers, otherwise we don't get notification. Also please assign yourself to it
> 
> as for the PR itself I've got no issue with approving it, as it provides skeleton for testing, but i would not say it closes any issue.
> 
> a. it doesn't (yet) test anything b. the issue says "Also re implementing some dispute logic per doc specification" -> which this PR definitely does not do

I haven't assigned for review because it isn't ready for review

contracts/V1/DataTypes.sol
```diff
@@ -121,3 +121,34 @@ struct ProcessExit {
...
+    address participant;
+    uint amount;
+    bytes data;
+    bool isPartialExit;
+}
+
+struct ExitChannelBlock {
+    /// @dev no signature requirement for the exitChannel blocks
+    ExitChannel[] exitChannel;
+    /// @dev Hash of the previous exitChannelBlock
```
lukastanisic99, line 136: missing 's' in exitChannels to indicate plural

contracts/V1/DataTypes.sol
```diff
@@ -121,3 +121,34 @@ struct ProcessExit {
...
+
+struct Timeout {
+    /// @dev the participant that is being timed out
+    address participant;
+    /// @dev the block height at which participant is removed from the channel (fork)
+    uint blockHeight;
+    uint minTimeStamp;
+    // ================== optional ==================
+    address previousBlockProducer;
+}
```
lukastanisic99, line 148: Here we're missing 'forkCnt' and 'previousBlockProducerPostedCalldata'

Context:
Only the participant address is needed to actually apply the timeout to the 'latestState' of disputes and the other stuff is there as a safety guard to protect against race condition on-chain. These safety guards are checked on-chain. There are 2 possible race conditions that can happen here:
1) The participant being timedout actually posted their block calldata on-chain and in that case we can't timeout them. This is checked as require(calldata[channelId][participant][forkCnt][blockHeight]==0x00(default value)) (see blockCalldataCommitment in the docs)
2) The participant that produced the block prior to the participant we're trying to timeout posted their block on-chain - this allows the participant more time to produce the block, since the on-chain timestamp 'overrides' the block timestamp set at creation. So if the person initiating the timeout is not aware of it, the timeout (tx) should fail. This is checked as: require(previousBlockProducerPostedCalldata || calldata[channelId][previousBlockProducer][forkCnt][blockHeight-1] == 0x00) and after check if minTimestamp<=block.timestamp 
If previousBlockProducerPostedCalldata is true - the disputer is knowingly aware that the participant has more time and calculated minTimestamp accordingly.

While the 'latestState' property does imply the exact forkCnt and blockHeight we still include them here, since they're not easily accessible on-chain from the 'latestState' and all of this is calldata and is not persisted in storage so it's 'cheap' to include it for easy checking

contracts/V1/DisputeTypes.sol
```diff
@@ -17,23 +17,47 @@ contract DisputeTypes {
...
+    bool selfRemoval;
+}
+
+/// @notice Proof of state finality within a fork
+struct StateProof {
+    /// @notice The state being proven
+    bytes encodedState;
+    
+    /// @notice N/N signatures proving finality of the state
+    bytes[] Signatures;
```
lukastanisic99, line 60: We can't really do this like this :D - Luke was also confused by this, so I'll update the docs later with a specification for this, so it's more clear

contracts/V1/DisputeTypes.sol
```diff
@@ -44,11 +68,22 @@ struct Proof {
...
+    // Timeout related fraud proofs
+    TimeoutThreshold,
+    TimeoutPriorInvalid,
+    TimeoutParticipantNoNext,
+
+    // Dispute fraud proofs
+    FoldRechallenge,
+    IncorrectData, // InvalidOutputState
+    NewerState, // NotLatestState
+    FoldPriorBlock, // challenge a false timeout proof
```
lukastanisic99, line 86: I've renamed some fraud proofs in the docs so there's a naming convention (Block, Dispute,Timeout) - if a fraud proof is missing (some probably are, but will be added) just update the doc and here

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
```diff
@@ -105,6 +105,61 @@ contract DisputeManagerFacet is StateChannelCommon {
...
+         require(
+            isFinalizedAndLatest(
+                channelId,
+                forkCnt,
+                encodedLatestFinalizedState,
+                encodedLatestCorrectState,
+                virtualVotingBlocks,
+                participants
+            ),
+            ErrorLatestFinalizedBlock()
```
lukastanisic99, line 134: In general we don't do any of these checks anymore since they're expensive and don't actually solve the problems since on-chain we don't have the full off-chain context), so we let anyone commit to anything at the risk of slashing - and rely on crypto-economic game theory that the majority will be honest and that few that commit to junk will get slashed anyway by explicitly invoking auditing or generating more fraud proofs - if it's unclear how this should be implemented from the docs either ping me or I don't mind implementing this next week when I'm back from the PBA

There are some checks simple checks that we do as safety guards defined in the doc to protect the disputer from race condition on-chain

contracts/V1/StateChannelDiamondProxy/StateChannelManagerStorage.sol
```diff
@@ -19,8 +19,21 @@ contract StateChannelManagerStorage {
...
-    mapping(bytes32 => Dispute) disputes; // disputes[channelId] => Dispute #only 1 dispute per fork and at a time
+    
+    // ================== Dispute on chain storage ==================
+
+    /// @dev disputes[channelId] => array of dispute commitments
+    /// @dev hash(Dspute Struct, timestamp)
+    mapping(bytes32 => bytes32[]) disputes;
+
+    /// @dev disputesData[channelId] => array of encoded dispute data
+    mapping(bytes32 => bytes32[]) dipsutesData;
```
lukastanisic99, line 30: I forgot to add these time parameters and the stateMachineImplementation to the on-chain state in the doc, but these are static and don't change (unless through governance) - I'll put them under config

As for ever the rest you can do either:
mapping(bytes32 => State) state 
Where State is {
StateSnapshot,
BlockCalldataCommitment,
DisputeData
}
(this is cleaner) 

OR 
map individually:
channelsId -> stateSnapshot
channelsId -> BlockCalldataCommitment
channelsId -> DisputeData

Need to define all the data structures

MrishoLukamba:
And @lukastanisic99 I do think for now Virtual Voting is redundant.
Per video and the doc if a peer refuses to cooperate but post a block call data , thus will be like signing all previous block even though he did not participate in traditional signing of the block.
This brings a lot of issues to the protocol as this type of peer  needs to be handled by submitting a dispute.

Hence I do think for now Virtual Vote is not optimal enough to be in V1

MrishoLukamba:
And there alot of cleaning to do, that's why the PR is in draft. as refactoring and cleaning of implementations are yet to be done am on it.

lukastanisic99:
> And @lukastanisic99 I do think for now Virtual Voting is redundant. Per video and the doc if a peer refuses to cooperate but post a block call data , thus will be like signing all previous block even though he did not participate in traditional signing of the block. This brings a lot of issues to the protocol as this type of peer needs to be handled by submitting a dispute.
> 
> Hence I do think for now Virtual Vote is not optimal enough to be in V1

What issues do you see with it specifically?

You don't nor can't issue a dispute for someone who didn't sign a block - you can't prove it. You can only prove that someone signed, by having the signature. The issue is, if you revoke virtual voting you may never finalize a state if someone doesn't want to cooperate - in the docs it says you finalize the state through virtual voting or disputes, but the dispute may require virtual voting too (stateProof) - probably the stateProof part is what is confusing you guys, so I'll explain that better. 

The idea is you can prove 'milestones' directly or indirectly without having to show the entire history and when you prove the milestone you can link a 'few more blocks' to connect it to the 'latestState' in dispute and that's how you can eventualy finalize the latestState even though you don't have N/N on it directly or indirectly

MrishoLukamba:
> > And @lukastanisic99 I do think for now Virtual Voting is redundant. Per video and the doc if a peer refuses to cooperate but post a block call data , thus will be like signing all previous block even though he did not participate in traditional signing of the block. This brings a lot of issues to the protocol as this type of peer needs to be handled by submitting a dispute.
> > Hence I do think for now Virtual Vote is not optimal enough to be in V1
> 
> What issues do you see with it specifically?
> 
> You don't nor can't issue a dispute for someone who didn't sign a block - you can't prove it. You can only prove that someone signed, by having the signature. The issue is, if you revoke virtual voting you may never finalize a state if someone doesn't want to cooperate - in the docs it says you finalize the state through virtual voting or disputes, but the dispute may require virtual voting too (stateProof) - probably the stateProof part is what is confusing you guys, so I'll explain that better.
> 
> The idea is you can prove 'milestones' directly or indirectly without having to show the entire history and when you prove the milestone you can link a 'few more blocks' to connect it to the 'latestState' in dispute and that's how you can eventualy finalize the latestState even though you don't have N/N on it directly or indirectly

Correct me if I am wrong, BUT if someone did not sign a. block (either by refusal or some other issue) it means that peer will fall into a timeout fraud category. And if we can verify the timeout we can finalize the state.

and non cooperative peer wont submit the virtual vote. The worst will be if we wait for a peer to sign and doesn't sign the leader posts a block call data and proof of time elapse and lack of signature from the so called peer hence a timeout dispute.

So with non cooperative peer the finality reaches after timeout

lukastanisic99:
> > > And @lukastanisic99 I do think for now Virtual Voting is redundant. Per video and the doc if a peer refuses to cooperate but post a block call data , thus will be like signing all previous block even though he did not participate in traditional signing of the block. This brings a lot of issues to the protocol as this type of peer needs to be handled by submitting a dispute.
> > > Hence I do think for now Virtual Vote is not optimal enough to be in V1
> > 
> > 
> > What issues do you see with it specifically?
> > You don't nor can't issue a dispute for someone who didn't sign a block - you can't prove it. You can only prove that someone signed, by having the signature. The issue is, if you revoke virtual voting you may never finalize a state if someone doesn't want to cooperate - in the docs it says you finalize the state through virtual voting or disputes, but the dispute may require virtual voting too (stateProof) - probably the stateProof part is what is confusing you guys, so I'll explain that better.
> > The idea is you can prove 'milestones' directly or indirectly without having to show the entire history and when you prove the milestone you can link a 'few more blocks' to connect it to the 'latestState' in dispute and that's how you can eventualy finalize the latestState even though you don't have N/N on it directly or indirectly
> 
> Correct me if I am wrong, BUT if someone did not sign a. block (either by refusal or some other issue) it means that peer will fall into a timeout fraud category. And if we can verify the timeout we can finalize the state.
> 
> and non cooperative peer wont submit the virtual vote. The worst will be if we wait for a peer to sign and doesn't sign the leader posts a block call data and proof of time elapse and lack of signature from the so called peer hence a timeout dispute.
> 
> So with non cooperative peer the finality reaches after timeout

Timeout is for block authoring, so if you don't author a block when it's your time to author it, you will be timedout after some time. It's not for cosigning blocks from others, just for authoring. 

You refusing to co-sign a block isn't disputable since it's not objective (you can sign it and send it to a peer and the peer can say it didn't receive it, so an independent 3rd party (blockchain) doesn't know the difference between did you refuse to sign it or did the peer refuse to accept it - it can't 'point a finger' who is to blame => no punishment) 

This is entirely subjective, you have your opinion that the other guys is not cooperating and you may refuse to cooperate with them. This will force them to post calldata on-chain and pay fees. 

You can entirely transition (progress) the state on-chain ONLY through calldata without ever cosigning blocks from others. You'll still be able to finalize it through virtual voting indirectly and/or through disputes which kinda use the same thing 'stateProof'. So if you fail to author the block that will count towards virtual voting, you'll be timedout and when you're timedout you're removed from the channel so your signature doesn't matter anymore -> the others can finalize the state without you.

Ping me for a call if needed.

contracts/V1/DataTypes.sol
```diff
@@ -31,13 +31,13 @@ struct ForkDataAvailabilityKey {
 }
 
 struct SignedBlock {
-    bytes encodedBlock; //TODO! change this to bytes
-    bytes signature; //TODO! change this to bytes
+    bytes encodedBlock;
+    bytes[] signatures;
```
lukastanisic99, line 35: First if this is also an array the confirmedBlock bellow is redundant - this was meant to be used to resemble the initial signature from the block author and is used in authoring blocks and in blockConfirmations so even if you cosign the block you always gossip the original signedBlock (signature of the block author) so it's not possible to have other signatures without the block authors signature (unless someone is byzantine) - like this even under normal conditions or race condition it's possible to collect signatures from other peers without having the signature from the original block author which renders the other signatures worthless. The first signature always has to be the block author.

Btw so both of you know (@MrishoLukamba and @daiagi) when you change these struct that are used in typescript and encoded later with ethers you need to change this file too - https://github.com/peer3to/state-channels-plus/blob/ffbfd0f487ae43945eb9d8518aed61a73c8102b7/src/DataTypes.ts#L23

contracts/V1/DisputeTypes.sol
```diff
@@ -17,23 +17,59 @@ contract DisputeTypes {
...
+    // ========================== optional ===============================
+    /// @notice Previous recursive dispute hash
+    bytes32 previousRecursiveDisputeHash;
+    /// @notice Timeout for the dispute
+    Timeout timeout;
+    /// @notice Self removal for the dispute
+    bool selfRemoval;
+}
+
+struct BlockConfirmation {
```
lukastanisic99, line 54: This is ConfirmedBlock from DataTypes

contracts/V1/DataTypes.sol
```diff
@@ -31,13 +31,13 @@ struct ForkDataAvailabilityKey {
...
 }
 
 struct ConfirmedBlock {
     bytes encodedBlock;
```
lukastanisic99, line 39: Let's change encodedBlock to signedBlock, so it's not possible to present a blockConfirmation without the original signature

contracts/V1/DisputeTypes.sol
lukastanisic99: you can keep this name, just move it to DataTypes

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
```diff
@@ -99,121 +99,168 @@ contract DisputeManagerFacet is StateChannelCommon {
...
+        // core specific checks
+        // apply proofs
+        // apply dispute to latest state
+        // set state
+    ) internal {
+
+    }
+
+    function _handleTimeoutDispute(
+
```
lukastanisic99, line 122: Not sure where you're going with this '_handleBlockFraudProofs', '_handleDisputeFraudProofs', '_handleTimeoutDispute' - the pipeline is the same for everything regardless of the fraud proof (each fraud proof caries its specific logic).

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
```diff
@@ -99,121 +99,168 @@ contract DisputeManagerFacet is StateChannelCommon {
...
-        ConfirmedBlock[] memory virtualVotingBlocks,
-        address timedoutParticipant,
-        uint foldedTransactionCnt,
-        Proof[] memory proofs
-    ) public {
-        require(!isDisputeInProgress(channelId), ErrorDisputeInProgrees());
-        require(getForkCnt(channelId) == forkCnt, ErrorDisputeForkMismatch());
+        bytes calldata encodedLatestState,
+        StateProof[] calldata stateProofs,
+        address[] calldata onChainSlashes,
```
lukastanisic99, line 132: This should literally be the Dispute Struct from the docs and it should check msg.sender == dispute.disputer. 

Other way to do it is omit/remove the 'challengeDeadline' from the struct and submit a signedDispute (which would be better) since we want to collect signatures to hopefully finalize early. The timestamp/deadline can be used the same way as with blockCalldata when calculating the commitment (hash(dispute,block.timestamp)).

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
```diff
@@ -99,121 +99,168 @@ contract DisputeManagerFacet is StateChannelCommon {
...
+        Proof[] calldata proofs
+    ) public { 
+        // sanity checks
+        if(onChainSlashes.length != onChainSlashedParticipants.length 
+        || onChainSlashes[onChainSlashes.length - 1] != onChainSlashedParticipants[onChainSlashedParticipants.length - 1]) {
```
lukastanisic99, line 137: this is not correct in general and would fail when when length == 0 since you're trying to access at position -1, either loop through and compare or hash and compare hashes

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
```diff
@@ -99,121 +99,168 @@ contract DisputeManagerFacet is StateChannelCommon {
...
+            revert("onChainSlashes does not match onChainSlashedParticipants");
+        }
+        if(proofs.length == 0) return;
```
lukastanisic99, line 140: also not true - it's possible to just timeout or self remove, but you shouldn't check this as a safety check either way, since we have a fraud proof for this, so if someone commits to nonsense they will get slashed - on-chain we just check the safety guards that can be affected by race condition on-chain (this is not affected by that)

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
```diff
@@ -99,121 +99,168 @@ contract DisputeManagerFacet is StateChannelCommon {
...
-        );
-        require(isParticipant, ErrorNotParticipant());
+        // check if its genesis
+        if(stateProofs.length == 0) {
+           bool isGenesis = isGenesisState(channelId, forkCnt, encodedLatestState); 
+           if(!isGenesis) {
+            revert("latest state is not valid");
+           }
+        }else{
+            _verifyStateProof(encodedLatestState, stateProofs, participants);
```
lukastanisic99, line 151: All of this logic is not done here - it's done in auditing which is 'stateless', but has synchronous composability to read data from state (not modify it) - explained in the doc

Here we just commit to the dispute struct and optionally invoke auditing which returns true/false and other things defined in the docs

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
lukastanisic99: second argument to the function call (dispute is first) should be 'shouldInvokeAuditing' that defaults to false.

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
MrishoLukamba: each fraud proof carries its own specific logic, exactly thts why I have bundled them together for easier reading and separation

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
MrishoLukamba: but we are. checking if the disputer has submitted sensible latest state?

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
MrishoLukamba: this also its a check per doc

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
lukastanisic99: Unless we invoke auditing on-chain (on-chain execution) we don't check anything apart the safety guards. In general we just try to commit and allow others to audit off-chain since it's free and they have the full state and can check. If we want to do on-chain execution we have to upload more than the minimum requirement (e.g. A hash is enough to commit to a state, but you can't transition the state from it. Since peers off-chain have the full state it's enough for them to check, but if you want to invoke on-chain execution you have to upload the state which can be easily verified against the commitment). 

So all of this is part of auditing which requires more data - I'll explain tomorrow in the call 

We should have calls more frequently

MrishoLukamba:
> @MrishoLukamba if everything is clear, can you finish this draft by next week? Trying to have the doc implemented by end of next month - if something is unclear ping me

Pushing the PR today

MrishoLukamba:
Dont review yet, I have pushed an earlier commit

.gitignore
```diff
@@ -13,4 +13,5 @@ artifacts
 typechain-types
 
 # Ignore all .tgz files
-*.tgz
\ No newline at end of file
+*.tgz
+./wake
```
lukastanisic99, line 17: what's this ./wake stuff - is it something useful I should look into?

lukastanisic99:
@MrishoLukamba for the stuff you want us to look at - can you push them to a new branch within this repository, so we don't have to track multiple origins (your fork)

contracts/V1/DataTypes.sol
```diff
@@ -11,33 +11,23 @@ contract DataTypes {
...
 
-struct ConfirmedBlock {
-    bytes encodedBlock;
-    bytes[] signatures; //TODO! change this to bytes
+struct BlockConfirmation {
+    SignedBlock signedBlock;
+    bytes[] signatures;
 }
 
 struct Block {
```
lukastanisic99, line 33: The Block struct needs to be updated:
stateHash -> stateSnapshotHash
previousStateHash -> previousBlockHash

contracts/V1/DataTypes.sol
```diff
@@ -121,3 +116,46 @@ struct ProcessExit {
...
+
+/// @dev a pair consisting of first index (index of the malicious dispute) and last index (last index in the array)
+struct DisputePair {
+    uint firstIndex;
+    uint lastIndex;
+}
+
+/// @dev data for dispute auditing
+struct DisputeAuditingData {
+    bytes genesisStateSnapshot;
```
lukastanisic99, line 157: This, DisputePair and Timeout should be moved to DisputeTypes

contracts/V1/DataTypes.sol
```diff
@@ -84,6 +74,11 @@ struct JoinChannel {
     uint deadlineTimestamp;
     bytes data; //custom data
 }
+
+struct JoinChannelBlock {
+    bytes32 previousBlockHash;
+    JoinChannel[] joinChannels;
+}
 struct SignedJoinChannel {
     bytes encodedJoinChannel;
```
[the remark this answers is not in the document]
lukastanisic99: TransactionType should also be removed (poker stuff)

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
```diff
@@ -54,20 +53,19 @@ contract DisputeManagerFacet is StateChannelCommon {
         DisputeAuditingData memory disputeAudit
     ) public returns (bool isSuccess, address[] memory slashParticipants) {
 
-        address challenger = msg.sender;
         // check if the commitment of dispute is available
         bytes32 memory disputeCommitment = keccak256(abi.encodePacked(
             dispute,
             block.timestamp
```
lukastanisic99, line 59: The block.timestamp is used when making the commitment on-chain. Here you have to provide that timestamp, since it needs to be the same timestamp for the hash (commitment) to match. Feel free to extend AuditingData with it.

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
```diff
@@ -54,20 +53,19 @@ contract DisputeManagerFacet is StateChannelCommon {
...
         ));
         (bool isAvailable, int index) = isDisputeCommitmentAvailable(disputeCommitment);
         if(!isAvailable) {
-            revert("Dispute commitment not available");
+            revert(abi.encode("AUDIT: DISPUTE COMMITMENT NOT AVAILABLE"));
```
lukastanisic99, line 63: This should be revert(string) not revert(abi.encode(string)) - The string is implicitly interpreted by ethers and all/most EVM tooling, but this requires more gas and larger contract size -> better to use custom Errors

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
```diff
@@ -4,9 +4,517 @@ import "./StateChannelCommon.sol";
...
+    function _verifyNotLinkedToLatestState(
+        Proof memory proof
+    ) internal returns (bool isValid, address slashedParticipant) {
+
+    }
+
+    // =============================== State Proofs Verification  ===============================
+
+    function _verifyStateProof(bytes memory encodedLatestState, StateProof memory stateProof, address[] memory participants) internal returns (bool isValid) {
+        // ideal case , no signedBlocks, latestState = lastFinalizedState = MilestoneBlock
```
lukastanisic99, line 400: This should be verified without any special cases, iteratively or recursively. 

Pseudo code:

verifyStateProof(genesisStateSnapshotHash, stateProof, latestStateSnapshotHash){
//all relevant snapshots have to be provided in AuditingData to be able to verify them against the hash
genesisStateSnapshotHash == hash(audtingData.genesisStateSnapshot) 

snapshot = audtingData.genesisStateSnapshot
participants = snapshot.participants

for milestone in ForkProof{
isFinal(milestone.blockConfirmations,participants) //checks if first block in blockConfirmations has N/N threshold from current participant set by taking into account virtual voting of blockConfirmations
block = decodeFirstBlockInMilestone(milestone)
block.stateSnapshotHash == hash(milestone.stateSnapshot) // snapshot should be provided in auditing data - i'm just referring to it as milestone.stateSnapshot so you know it's the snapshot of the first block in blockConfirmations
stateSnapshot = milestone.stateSnapshot
participants = snapshot.participants
}
// now we have proven finality of latest finalized block/snapshot 
lastBlock = lastBlockFromLastMilestone
for signedBlock in signedBlocks{
verifySignature(signedBlock) //checks that the block author signed the block
block = decodeBlock(signedBlock)
checkIsLinked(lastBlock, block)
lastBlock = block
}
lastBlock.stateSnapshotHash  == latestSnapshotHash

//if all this passes the proof is valid, otherwise invalid
}

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
```diff
@@ -4,9 +4,571 @@ import "./StateChannelCommon.sol";
...
+        }
+
+        address disputer = StateChannelUtilLibrary.retriveSignerAddress(dispute.encodedBlock, signature);
+        if(disputer != dispute.disputer) {
+            revert("Invalid signature");
+        }
+        if(proofs.length == 0) return;
+        // commit to dispute struct
+        bytes memory encodedDispute = abi.encode(dispute);
+        bytes32 memory channelIdHash = keccak256(dispute.channelId);
```
lukastanisic99, line 31: channelId is already a hash and the unique identifier - no need to hash it further

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
MrishoLukamba: now i ma first returning the bytes as values and not reverting, if thats make sense
so that the client can inteprete it based on boolean value.

"The main auditing logic on success returns true with the list of slashed participants (useful for [onChainSlashes](https://docs.google.com/document/d/1h1MmOhTgVFVbf3BYc0Di9RdWA-PS4tEGARVOspQv2kE/edit?tab=t.0#bookmark=id.d8irzt63u2xe)) and on failure returns false with an error reason (out of gas, missing oracle, etc..). (Hint: for solidity use bytes and interpret them based on the boolean value)"

MrishoLukamba:
> @MrishoLukamba for the stuff you want us to look at - can you push them to a new branch within this repository, so we don't have to track multiple origins (your fork)

The idea was pushing the commits here, until the PR is final to review then i undraft it

lukastanisic99:
> > @MrishoLukamba for the stuff you want us to look at - can you push them to a new branch within this repository, so we don't have to track multiple origins (your fork)
> 
> The idea was pushing the commits here, until the PR is final to review then i undraft it

Sure, I'm looking at it either way to give some feedback early to save time later

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
MrishoLukamba: What if I return an enum of ErrorType instead of strings So that the function signature to be consistent

returns (bool success, address[], EnumError), hence when handling we just check the success and process

lukastanisic99: The thing is some errors will have to have additional information, for example "blockcalldata commitment is missing for [channelId] [senderAddress] [fork] [blockNumber]"

We should also have an error "AuditingData incorrect" - since the auditor provides the data, so if the auditor provides junk, auditing will fail, but we shouldn't slash the backer for it, but the auditor - even though auditing failed. 

If the backer commits to junk, the auditor should be able to prove it with a separate fraud proof.

The error is mostly useful to the off-chain peer to try and fetch the commitment if its off-chain environment isn't synced with the on-chain one + in the case of wrong 'AuditingData' it's useful on-chain too.

contracts/V1/AStateMachine.sol
```diff
@@ -33,12 +33,12 @@ abstract contract AStateMachine {
     // define the logic that punishes a participant for misbehaving (can also remove the participant from the state channel)
     function _slashParticipant(
         address adr
-    ) internal virtual returns (bool, ProcessExit memory);
+    ) internal virtual returns (bool);
```
lukastanisic99, line 36: instead of ProcessExit it should return ExitChannel

contracts/V1/AStateMachine.sol
```diff
@@ -60,13 +60,13 @@ abstract contract AStateMachine {
 
     function slashParticipant(
         address adr
-    ) external _nonReentrant returns (bool, ProcessExit memory) {
+    ) external _nonReentrant returns (bool) {
```
lukastanisic99, line 63: same

contracts/V1/AStateMachine.sol
```diff
@@ -60,13 +60,13 @@ abstract contract AStateMachine {
...
         return _slashParticipant(adr);
     }
 
     function removeParticipant(
         address adr
-    ) external virtual _nonReentrant returns (bool, ProcessExit memory) {
+    ) external virtual _nonReentrant returns (bool) {
```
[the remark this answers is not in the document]
MrishoLukamba: Yeap all will be updated, the issue with Dispute is most of things needs to be changed as they are dependable in dispute handling, so when implementing dispute , you also need to implement other things recursively.

contracts/V1/DataTypes.sol
```diff
@@ -3,22 +3,21 @@ pragma solidity ^0.8.8;
...
-        SignedJoinChannel memory e,
-        JoinChannelAgreement memory f,
-        ConfirmedJoinChannelAgreement memory g,
-        LeaveChannel memory h,
-        LeaveChannelAgreement memory i
+        Block memory a,
+        SignedBlock memory b,
+        BlockConfirmation memory c,
+        Transaction memory d,
+        TransactionHeader memory e,
```
lukastanisic99, line 10: When a type is contained within a type (TxHeader and TxBody are within Tx) it's enough to only list the top level type (Tx), but this is not important you can list all - this is just so typechain generates the types and they're not skipped by the compiler

contracts/V1/DisputeTypes.sol
```diff
@@ -34,12 +34,14 @@ struct Dispute {
     bytes32 genesisStateSnapshotHash;
     /// @notice encoded latest state (latest on-chain state)
     bytes32 latestStateSnapshotHash;
+    /// @dev the blocknumber/height of the latest state
+    uint latestStateHeight;
```
lukastanisic99, line 38: Though we're not adding this? For the latestState you have to decode the blocks and prove they are and can't rely on this, since an 'attacker' can put here anything and again force you to decode the blocks to prove it's junk - so you're going to decode them anyway -> this doesn't help :D

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
```diff
@@ -14,26 +14,28 @@ contract DisputeManagerFacet is StateChannelCommon {
...
         // sanity checks
-        msg.sender == dispute.disputer;
+        require(msg.sender == dispute.disputer, "CREATE DISPUTE: INVALID DISPUTER");
         // race condition check
-        if(onChainSlashes.length != onChainSlashedParticipants.length 
-        || onChainSlashes[onChainSlashes.length - 1] != onChainSlashedParticipants[onChainSlashedParticipants.length - 1]) {
-            revert("onChainSlashes does not match onChainSlashedParticipants");
+        address[] memory onChainSlashedParticipants = getOnChainSlashedParticipants();
+        if(dispute.onChainSlashes.length != onChainSlashedParticipants.length 
+        || dispute.onChainSlashes[dispute.onChainSlashes.length - 1] != onChainSlashedParticipants[onChainSlashedParticipants.length - 1]) {
```
lukastanisic99, line 21: My original comment on this still holds - you're checking just the last element to be equal instead of the entire array + if both arrays are empty this will break since you're accessing the array at position -1

Either iterate and compare or hash the array and compare

MrishoLukamba: Okey so my idea was this, to save gas, the last element will be a commitment of all element inside hash(all elements), so if its equal meaning we verified all elements and the issue of accessing empty array yaap am fixing it

lukastanisic99: This could work, since onChainSlashed is append only and only cleared on successful submission of the snapshot. So if a peer knows the K-th element it also knows all the K-1 elements prior and since the past elements can't change, this could also work. Just implement it without the [-1] access

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
```diff
@@ -14,26 +14,28 @@ contract DisputeManagerFacet is StateChannelCommon {
...
-        address disputer = StateChannelUtilLibrary.retriveSignerAddress(dispute.encodedBlock, signature);
+        address disputer = StateChannelUtilLibrary.retriveSignerAddress(abi.encode(dispute), signature);
         if(disputer != dispute.disputer) {
-            revert("Invalid signature");
-        }
-        if(proofs.length == 0) return;
+            revert CreateDisputeInvalidSignature();
+        } 
+        // check if fraud proofs or TimoutStruct is provided to process the dispute
+        if(dispute.fraudProofs.length == 0 || isTimeoutSetWithOptional(dispute.timeout, false)) return;
```
lukastanisic99, line 30: You don't have to check this here, since there is no race condition on-chain for this. If the disputer commits to an invalid dispute, someone else will produce a fraud proof to slash them.

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
```diff
@@ -14,26 +14,28 @@ contract DisputeManagerFacet is StateChannelCommon {
...
+
         // commit to dispute struct
         bytes memory encodedDispute = abi.encode(dispute);
-        bytes32 memory channelIdHash = keccak256(dispute.channelId);
         bytes32 memory disputeCommitment = keccak256(abi.encodePacked(
```
lukastanisic99, line 34: We shouldn't use abi.encodePacked - it's cheaper, but can have collision in hashing - in general we just use abi.encode everywhere since it's both collision safe for hashing and it can be easily decoded back and the gas savings don't make a difference

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
```diff
@@ -17,18 +17,16 @@ contract DisputeManagerFacet is StateChannelCommon {
         require(msg.sender == dispute.disputer, "CREATE DISPUTE: INVALID DISPUTER");
         // race condition check
         address[] memory onChainSlashedParticipants = getOnChainSlashedParticipants();
-        if(dispute.onChainSlashes.length != onChainSlashedParticipants.length 
-        || dispute.onChainSlashes[dispute.onChainSlashes.length - 1] != onChainSlashedParticipants[onChainSlashedParticipants.length - 1]) {
+
+        if(keccak256(abi.encodePacked(dispute.onChainSlashes)) != keccak256(abi.encodePacked(onChainSlashedParticipants))) {
```
lukastanisic99, line 21: Please test this - I'm not sure this will compile since storage arrays don't work the same as memory arrays (they're essentially mappings), so I don't think you can abi.encodePacked(storage array)

Your original idea should also work, just add a check to not do [-1]

MrishoLukamba:
the only commits which will be coming is cleaning the rest of codebase(old one) that interferes with new code

lukastanisic99:
@MrishoLukamba it still doesn't compile and I see some of the comments are not yet addressed from above - please fix and resubmit. If you need help or if something is unclear ping me

contracts/V1/StateChannelDiamondProxy/StateChannelManagerStorage.sol
```diff
@@ -14,19 +14,26 @@ contract StateChannelManagerStorage {
...
     AStateMachine stateMachineImplementation;
+
+    // =================== State on chain storage ==================
+
+    /// @dev stateSnapshot Data
+    mapping(bytes32 => StateSnapshot) stateSnapshots;
     /// @dev the fork genesis state
-    mapping(bytes32 => mapping(uint => bytes)) encodedStates; // [channelId][forkCnt] -> encodedState
-    //TODO* - think do we need to store previous timestamps
-    mapping(bytes32 => mapping(uint => uint)) genesisTimestamps; // [channelId][forkCnt] -> encodedState
```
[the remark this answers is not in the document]
MrishoLukamba: Pending in design, genesisTimestamp will it not just be covered in state

lukastanisic99:
image

https://github.com/MrishoLukamba/state-channels-plus/blob/50616e2c098d562c0ad4c65318aba5904bd32572/contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol#L7

There is no ./Events.sol in your directory - The tests are right is shouldn't compile - not sure how it's possible for it to compile on your end

PR #20, opened 2025-04-10
refactor `StateManager` and `AgreementManager`

daiagi:
make `StateManager` a bit more modular, readable 
most notable changes are to the methods:

- onSignedBlock
- onBlockConfirmation
- processExecutionDecision
- processConfirmationDecision

secondary changes: just break  into smaller functions:

- onSuccessCommon
- playTransaction

next step will be to add tests


src/stateManager/BlockValidation.ts
```diff
@@ -0,0 +1,15 @@
...
+export async function runPipeline<
+    TContext,
+    TStep extends (context: TContext) => Promise<ValidationResult>
+>(steps: TStep[], context: TContext): Promise<ValidationResult> {
+    for (const step of steps) {
+        const result = await step(context);
+        if (result.executionFlag !== ExecutionFlags.SUCCESS) {
+            return result; // Stop on first non-SUCCESS
+        }
+    }
```
lukastanisic99, line 13: I really like this idea, but it doesn't have to be exclusive to blockValidation - the dispute system can also use it and in general all pipelines that use checks. These small steps can also isolate fraud proof detection when deviation is objective and is also useful for subjective deviation like detecting spam. Move this to utils or somewhere more general

src/stateManager/validators.ts
```diff
@@ -0,0 +1,328 @@
...
+    }, 0);
+
+    return { executionFlag: ExecutionFlags.SUCCESS };
+};
+
+//====================================================================
+// CONFIRMATION VALIDATION FUNCTIONS
+//====================================================================
+
+/**
```
lukastanisic99, line 256: This file needs to be split into BlockValidationPipeline and BlockConfirmationPipeline

src/stateManager/StateManager.ts
```diff
@@ -0,0 +1,778 @@
...
+                checkForkDisputeStatus,
+                checkDuplicateBlock,
+                checkBlockIsFuture,
+                checkParticipantInFork,
+                checkPastBlockCurrentFork,
+                checkBlockTimestamp,
+                checkEnoughTimeSubjective,
+                checkCorrectBlockProducer,
+                verifyStateTransition
+            ];
```
lukastanisic99, line 268: This is my favorite part of this PR - clearly shows the checks - maybe not the best naming on some, but we can adjust that later. 

We have pipelines and checks everywhere, but the current pipeline logic (as used here) only works due to the way how the underlying pipeline logic is structured - chain on 'success' and pas 'final flags' to some function (processConfirmation/Execution decisions)

Most pipeline steps are 1 or 2 lines and not sure how much a 'step' could be reused and the only benefit to this approach vs just refactoring the validation/confirmation logic in another file into smaller function calls, is that we don't explicitly check 'if flag == success' all the time

Overall I like it, but we'll see how practical it will turnout in the end

lukastanisic99:
Overall I have mixed feelings about this PR - I like the pipeline idea, not sure how practical it is and at at first a bit harder to understand (introduces more complexity) because there has to be a shared 'artificial' context that's a union of all things used in the pipeline (instead of just going sequentially and calling smaller function calls with the specific arguments they require)+ added generics - all for the benefit to not explicitly do 'if flag == success' all the time.

The playTransaction refactors is a good example, it doesn't use the pipeline stuff, but it's refactored with smaller function calls that wrapped the logic -> easier to understand and no added complexity 

@MrishoLukamba what are your thoughts on this

src/stateManager/StateManager.ts
```diff
@@ -0,0 +1,778 @@
...
+            Number(timestamp)
+        );
+        let block = EvmUtils.decodeBlock(signedBlock.encodedBlock);
+        let disputeProof: ProofStruct;
+        if (flag == AgreementFlag.DOUBLE_SIGN) {
+            console.log("StateManager - collectOnChainBlock - double sign");
+            disputeProof = this.disputeHandler.createDoubleSignProof([
+                signedBlock
+            ]);
+            this.disputeHandler.createDispute(
```
MrishoLukamba, line 160: ofcourse here will change as the new structure Is different on creating a dispute

src/stateManager/StateManager.ts
```diff
@@ -0,0 +1,778 @@
...
+                .postBlockCalldata(signedBlock)
+                .then((txResponse) => txResponse.wait())
+                .catch((error) => {
+                    console.log("Error posting block on chain", error);
+                });
+        }
+    }
+
+    // returns participants who haven't signed the block
+    public async getPlayersWhoHaventSignedBlock(
```
MrishoLukamba, line 419: I think the naming should be generic as SDK isn't tied to gaming

test/test_utils/testHelpers.ts
```diff
@@ -0,0 +1,140 @@
...
+    provider: ethers.Provider
+): Promise<number> => {
+    const block = await provider.getBlock("latest");
+    return block!.timestamp;
+};
+export const getCurrentTimeSeconds = (): number => {
+    return Math.floor(Date.now() / 1000);
+};
+
+export async function deployLibraryTestContract(
```
MrishoLukamba, line 50: I think also we should specify deployment on chain and offchain in naming

src/stateManager/StateManager.ts
lukastanisic99: This doesn't change much, maybe some refactor - this just detects a specific fraud proof (double sign) and passes it to the dispute handler - the dispute handler will change to incorporate the 'ELVES' logic

src/stateManager/StateManager.ts
lukastanisic99: Yes - players->participants

test/test_utils/testHelpers.ts
lukastanisic99: Sure - 99% of deployments in tests are on-chain - currently only the specific state machine (Math, Tic-Tac-Toe) is deployed 'off-chain' and we don't deploy it in tests, we just generate a deployment transaction and pass it to the SDK which deploys it internally.

In general the solidity tests are tested normally with hardhat and other Ethereum tooling without knowing about the SDK. Think of it as unit tests and even the stuff that runs locally (off-chain) can be tested 'on-chain' normally, since it's the same EVM. 

For off-chain specifically we can use 'deployLocal'

src/stateManager/StateManager.ts

daiagi:
@lukastanisic99 
please review again
- simplified validators and should be simpler to understand and re-use
- remove artificial context and over the top generics
- validators moved to utils
- moved validation "pipeline" of State Manger into private functions that use the validators in simple, sequential way. no magic

src/utils/validators.ts
```diff
@@ -0,0 +1,337 @@
+import { ExecutionFlags } from "@/DataTypes";
+import { AgreementFlag } from "@/AgreementManager";
+import { ethers } from "hardhat";
```
lukastanisic99, line 3: should be from ethers - hardhat only in tests

src/utils/validators.ts
```diff
@@ -0,0 +1,337 @@
...
+    flag: ExecutionFlags,
+    agreementFlag?: AgreementFlag,
+    message?: string
+): ValidationResult {
+    return { success: false, flag, agreementFlag, message };
+}
+
+export async function isManagerReady(
+    stateManager: StateManager
+): Promise<ValidationResult> {
```
lukastanisic99, line 39: Every function except failure/success takes a stateManager as an argument => all of these functions should be part of the StateManager class since this is equivalent to the implicit argument 'this' of type StateManager. Having this in a file called 'validators' in utils is misleading. In general when I made the comment to put the pipeline in untils it was for a general pipeline that could be re-used across the code base, but since we realized that such a pipeline isn't really reusable no point in doing it in utils. 

Most of the functions here should return a boolean to be re-useable and it's intuative for them to return a boolean given their names start with 'is..'

Some function are just circular implementations like for instance 'isBlockValid' which takes a stateManager argument and calls isValidBlock on the stateManager argument.

In fact most function are circular implementations from stateManager or agreement manager with an artifical return value 'ValidationResult'

I think we're trying too much to refactor and now have added a file with 300+ lines of code of duplicate implementations (wrappers) and all for the reason not to explicitly return execution/agreement flag where these wrappers are called even though they do the exact same thing within them.

src/stateManager/StateManager.ts
```diff
@@ -0,0 +1,827 @@
...
+            );
+
+            successCallback();
+            await this.onSuccessCommon();
+
+            this.scheduleTask(
+                () => this.maybePostBlockOnChain(block, signedBlock),
+                this.timeConfig.agreementTime * 1000,
+                "maybePostBlockOnChain"
+            );
```
lukastanisic99, line 332: @daiagi you didn't need to close the PR - there are good things in it like the playTransaction refactor - everything here is reusable 

The Task scheduling is another useful thing since it wraps the setTimeout in a reusable way that's more clear.

Partitioning this file into a stateManager directory and breaking parts into files like processConfirmation/ExecutionDecision is also good

Any refactor that simplifies, wraps and makes things reusable without adding complexity is good.

daiagi: ok. re-open
 you are right that the validator attempt has become useless
 i'll remove that, and use StateManger.<validator>  that returns plain boolean instead

src/utils/validators.ts
daiagi: validator removed
- validateSignedBlock and  validateBlockConfirmation  use direct validation steps. nothing facny

src/stateManager/StateManager.ts
```diff
@@ -0,0 +1,931 @@
...
+            if (!this.isDisposed) {
+                successCallback();
+            }
+        }, 0);
+
+        return { success: true, flag: ExecutionFlags.SUCCESS };
+    }
+}
+
+export default StateManager;
```
lukastanisic99, line 931: Went over this on the plane so I made comments in notes - I'll dump them here for this entire file - numbers in the notes refer to the line of code

171 - old code checked if (flag == ExecutionFlags.DISPUTE)
184 - same
261/268 - not to decode blocks twice
 A lot of places use Number(BigNumerish) on the forkCnt and transactionCnt from the block - as part of the initial AgreementManager refactor there is ‘numericFields’. The block is a typechain generated type so we can’t add methods to it natively, but we can create a BlockUtils to perform operations on the block and implement this there, so it’s reusable across the codebase. Also we use transactionCnt as blockHeight for now, so this could be a good time to switch this in the code by implemting BlockUtils.getBlockHeight(block) which will still use the transactionCount and later when we make the switch in the blockStruct we just need to update 1 line of code

727 - isManagerReady() - 631
737:748 - isPastFork()
756:761 - isFutureBlock()
766 - isBlockAuthorParticipant()
814 - isNextBlockAuthor()

834 - reuse 727 - 631
844 - reuse 737

921 - why no scheduleTask?

314 - should be same as 727

642 - should reuse 637 or 637 should only exist 

No reason to have 292 (applyTransaction) and 659 (applyTransactionOrThrow) - 292 is enough

363 should be the responsibility of AgreementManager In general the agreement manager should be refactored like the stateManager into its own directory and split into multiple files since there are too many methods - the base file should be simple and list all methods that are available and in general it should be partitioned based on functionality (block methods, signature methods etc etc) so it’s easy to find what we need. State manager should be split into more files as well, following the same logic.  In general our larger Classes should be implemented like this with clear separation and with easy to find methods and where to look, so we don’t end up implementing stuff multiple times

daiagi: working on it
vert well though out review. thank you for that
most of you comments are aleady fixed and pushed
i'm 100% agree that smaller more focused classes is better
i am now working on splitting it to smaller units

src/utils/BlockUtils.ts
```diff
@@ -0,0 +1,53 @@
...
+        },
+        body: {
+            transactionType: Number(block.transaction.body.transactionType),
+            encodedData: block.transaction.body.encodedData as string,
+            data: block.transaction.body.data as string
+        }
+    },
+    stateHash: block.stateHash as string,
+    previousStateHash: block.previousStateHash as string
+});
```
lukastanisic99, line 53: Could be very useful and could be dangerous since this is not updated if the BlockStruct changes with typechain.
I like the idea in general. Maybe better for now to use the 'getters' from utils that normalize while fetching. We could implement some recursive conversion for this since we only have to convert predefined types, but not worth the effort for now I guess

src/utils/BlockUtils.ts
```diff
@@ -0,0 +1,53 @@
...
+    Number(block.transaction.header.forkCnt);
+
+/**
+ * Get the timestamp from a block
+ */
+export const timestampOf = (block: BlockStruct): number =>
+    Number(block.transaction.header.timestamp);
+
+export const participantOf = (block: BlockStruct): AddressLike =>
+    block.transaction.header.participant;
```
lukastanisic99, line 31: Normalize this too and if the "Like" type is needed just access the block directly

src/utils/BlockUtils.ts
```diff
@@ -0,0 +1,53 @@
...
+ */
+export const coordinatesOf = (block: BlockStruct) => ({
+    forkCnt: Number(block.transaction.header.forkCnt),
+    height: Number(block.transaction.header.transactionCnt)
+});
+
+/**
+ * Get the block height (transaction count) from a block
+ */
+export const heightOf = (block: BlockStruct): number =>
```
lukastanisic99, line 15: Maybe better to name everything with 'get' instead of 'of' since we use 'get' across the codebase -> getHeight

daiagi: i like the `of` because it reads nicely
`heightOf(block)`

but you're the boss. if you prefer `get` then `get` it will be

src/utils/BlockUtils.ts
daiagi: removed

daiagi:
oof. this PR is too big

it now closes #25 also

MrishoLukamba (approved):

MrishoLukamba (approved):

lukastanisic99 (changes requested):
I'll make some changes today

lukastanisic99 (approved):
I still want to change things, but have more important stuff to do and things will change here either way when the contract logic is finished. Will incorporate changes later - feel free to merge

PR #21, opened 2025-04-16
AgreementManager Testing

daiagi:
High test coverage for AgreementManager
yes, things will change but we don't want to run after unintential bugs , so in order to change and develop with confidence, it is vital to have good test coverage

### DIFF
--- a/src/AgreementManager.ts
+++ b/src/AgreementManager.ts
@@ -282,54 +282,78 @@ class AgreementManager {
         virtualVotingBlocks: ConfirmedBlockStruct[];
     } {
         const fork = this.forks[Number(forkCnt)];
+        let encodedLatestFinalizedState: string | undefined;
+        let encodedLatestCorrectState: string | undefined;
+        let virtualVotingBlocks: ConfirmedBlockStruct[] = [];
+        const requiredSignatures = new Set(fork.addressesInThreshold);
 
-        const result = fork.agreements.reduceRight(
-            (acc, agreement, index) => {
-                // Only update encodedLatestCorrectState if not found yet and this participant signed
-                if (
-                    !acc.foundSignerState &&
-                    this.didParticipantSign(agreement.block, signerAddress)
-                        .didSign
-                ) {
-                    acc.encodedLatestCorrectState = agreement.encodedState;
-                    acc.foundSignerState = true;
-                }
+        for (let i = fork.agreements.length - 1; i >= 0; i--) {
+            const agreement = fork.agreements[i];
+            const signersAddresses = this.extractSignerAddresses(agreement);
 
-                // Only update encodedLatestFinalizedState if not found yet and everyone signed
-                if (
-                    !acc.foundFinalizedState &&
-                    this.didEveryoneSignBlock(agreement.block)
-                ) {
-                    acc.encodedLatestFinalizedState = agreement.encodedState;
-                    acc.foundFinalizedState = true;
-                }
-
-                // Add to virtualVotingBlocks if the signer signed this or any later block
-                if (acc.foundSignerState) {
-                    acc.virtualVotingBlocks.unshift({
-                        encodedBlock: EvmUtils.encodeBlock(agreement.block),
-                        signatures: agreement.blockSignatures as string[]
-                    });
-                }
-
-                return acc;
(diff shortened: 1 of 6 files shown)


src/AgreementManager.ts
```diff
@@ -308,49 +278,54 @@ class AgreementManager {
...
-                    agreement.block,
-                    signature
-                );
-                if (!encodedLatestCorrectState && adr == signerAddress) {
-                    encodedLatestCorrectState = agreement.encodedState;
+        const fork = this.forks[Number(forkCnt)];
+
+        const result = fork.agreements.reduceRight(
+            (acc, agreement, index) => {
+                // Only update encodedLatestCorrectState if not found yet and this participant signed
```
lukastanisic99, line 285: This runs for every element in the array and does not take into account virtual voting on the signatures. The original implementation even though it starts from the last agreement and it seems it runs to the first, it returns (breaks) the moment a threshold is reached through virtual voting (look how the set is modified) - so if the leader election algorithm is round robin this is at most N runs, where N is the number of participants.

daiagi: OK ill double checked and add test to assert this behivior

src/AgreementManager.ts
```diff
@@ -503,27 +482,14 @@ class AgreementManager {
...
-            EvmUtils.encodeBlock(agreement.block) == EvmUtils.encodeBlock(block)
-        )
-            return true;
-        return false;
+        const agreement = this.getAgreementByBlock(block);
+
+        return agreement
+            ? EvmUtils.encodeBlock(agreement.block) ==
+                  EvmUtils.encodeBlock(block)
+            : false;
```
lukastanisic99, line 492: return agreement && this.areBlocksEqual(agreement.block, block)

src/AgreementManager.ts
daiagi: fixed and tested https://github.com/peer3to/state-channels-plus/pull/21/commits/ca129eff3e78cbc7662b6d25de36a41d4dbbf115

src/AgreementManager.ts
daiagi: your suggestion gives this error:
`Type 'boolean | undefined' is not assignable to type 'boolean'.
  Type 'undefined' is not assignable to type 'boolean'.ts(2322)`
  
  how about 
  ```ts
      public isBlockInChain(block: BlockStruct): boolean {
        const agreement = this.getAgreementByBlock(block);
        return (
            (agreement || false) && this.areBlocksEqual(agreement.block, block)
        );
    }
    ```

src/AgreementManager.ts
```diff
@@ -485,11 +488,9 @@ class AgreementManager {
...
     public isBlockInChain(block: BlockStruct): boolean {
         const agreement = this.getAgreementByBlock(block);
-
-        return agreement
-            ? EvmUtils.encodeBlock(agreement.block) ==
-                  EvmUtils.encodeBlock(block)
-            : false;
+        return (
+            (agreement || false) && this.areBlocksEqual(agreement.block, block)
+        );
```
lukastanisic99, line 493: Not sure why "|| false" is here

src/AgreementManager.ts
lukastanisic99: Not sure what you mean with that. That commit modifies 'didEveryoneSignBlock' - that function doesn't have anything to do with virtual voting and the check you added there (set.size == 0) doesn't change anything (unless signatures from non participants are stored which isn't the case) 

My comment was on this function 'getFinalizedAndLatestWithVotes' and specifically 'reduceRight' - it will always run on all elements of the array and the logic doesn't take into account virtual voting - take a look at the removed code of that function

src/AgreementManager.ts
daiagi: Need to convert undefined to boolean.
Without it we get
`Type 'boolean | undefined' is not assignable to type 'boolean'. Type 'undefined' is not assignable to type 'boolean'.ts(2322)`

lukastanisic99 (approved):
Overall no objections. Didn't look too deep into the tests since parts of the logic will change anyway (not much), but will look deeper when it's all implemented.

PR #23, opened 2025-04-23
refactor and test DisputeHandler

daiagi:
Another PR in the spirit of refactoring and improving testability.

Luka—hope this refactor is more to your taste this time 🙂

There are still no functional changes. The main improvement is the extraction of:

- Proof generation
- Proof encode/decode
- Proof Validation

...into a dedicated class. This helps reduce the cognitive load in DisputeHandler.
I think the change makes sense and does no introduce any "over engineering"

### DIFF
--- a/src/DisputeHandler.ts
+++ b/src/DisputeHandler.ts
@@ -280,7 +280,7 @@ class DisputeHandler {
                 this.signerAddress
             );
             this.p2pEventHooks.onInitiatingDispute?.();
-            await this.stateChannelManagerContract
+            const _txReceipt = await this.stateChannelManagerContract
                 .challengeDispute(
                     this.channelId,
                     dispute.forkCnt,
--- a/src/DisputeHandler.ts
+++ b/src/DisputeHandler.ts
@@ -47,6 +47,7 @@ class DisputeHandler {
         this.stateChannelManagerContract = stateChannelManagerContract;
         this.p2pEventHooks = p2pEventHooks;
         this.proofManager = new ProofManager(agreementManager);
+        return this.self;
     }
 
     public setP2pEventHooks(p2pEventHooks: P2pEventHooks): void {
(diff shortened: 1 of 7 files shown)


src/DisputeHandler.ts
```diff
@@ -443,195 +289,67 @@ class DisputeHandler {
                     encodedLatestFinalizedState,
                     encodedLatestCorrectState,
                     { gasLimit: 2000000 } //TODO! - gas limit
-                );
-            let txReceipt = await txResponse.wait();
+                )
+                .then((txResponse) => txResponse.wait());
```
lukastanisic99, line 293: Why does this have 'await' and '.then'? With txResponse = await challengeDispute() and await txResponse.wait() the single catch block will trigger on error (not that it does anything for now)

src/DisputeHandler.ts
```diff
@@ -53,7 +46,7 @@ class DisputeHandler {
         this.agreementManager = agreementManager;
         this.stateChannelManagerContract = stateChannelManagerContract;
         this.p2pEventHooks = p2pEventHooks;
-        return this.self;
```
lukastanisic99, line 56: Should keep this for debugging - if the debugging flag is set to true the constructor will return the proxy object useful for debugging otherwise it returns the normal this

lukastanisic99:
Overall no objections and everything looks nice.

Most of the logic will change here to account for the new dispute system.

src/DisputeHandler.ts
daiagi: it still triggers on error since there is no `.catch` on the promise
if you like two `awaits` better then so be it. i thought this style is more suitable here since there was no use for the interim `txRecipt` except for turning it in `txResponse`

src/DisputeHandler.ts
daiagi: restored

daiagi:
@lukastanisic99  if no objection please approve and merge
i will need this refactored version to continue breaking apart `stateManager` and `DisputeHandler` into smaller components

src/DisputeHandler.ts
lukastanisic99: I'm not sure it works like that, but I'm not home to test at the moment. Either way this is minor and everything else is good so feel free to merge

lukastanisic99 (approved):

PR #24, opened 2025-04-24
Dispute Functionalities

MrishoLukamba:


lukastanisic99:
Great - I'll take a look over the weekend in more details and push some changes probably. If you're planning on making more changes please push everything till tomorrow night and if you're working on this during the weekend work on tests so there's a lower chance we'll be in conflict since probably I'll be modifying your code. For the next weeks' call let's see how we can partition this more so we can all work on the core in parallel without conflict/breaking/reverting each other's work.

daiagi:
@MrishoLukamba 
Closes #17 ?

MrishoLukamba:
> @MrishoLukamba Closes #17 ?

Yes, its just i opened it here in this branch per Luka instructions

contracts/V1/StateChannelDiamondProxy/StateChannelUtilLibrary.sol
```diff
@@ -109,4 +109,14 @@ library StateChannelUtilLibrary {
         set[currentSetLength] = adr;
         return currentSetLength + 1;
     }
+
+    /// @dev Concatenates two address arrays, but it does not add duplicates
+    function concatAddressArrays(address[] memory array1, address[] memory array2, uint array1Length) internal pure returns (address[] memory) {
```
daiagi, line 114: This does not concat at all

lukastanisic99: I've noticed it too - I'll push code later that accounts for this too

daiagi: here is impl suggestion:

```solidity
    /// @dev Concatenates two address arrays without adding duplicates
    /// @param array1 The first array
    /// @param array2 The second array to append
    /// @return Combined array with no duplicates
    function concatAddressArrays(address[] memory array1, address[] memory array2) internal pure returns (address[] memory) {
        // Create a temporary array with maximum possible size
        address[] memory temp = new address[](array1.length + array2.length);
        
        // Copy all elements from array1
        uint resultLength = array1.length;
        for (uint i = 0; i < array1.length; i++) {
            temp[i] = array1[i];
        }
        
        // Add non-duplicate elements from array2
        for (uint i = 0; i < array2.length; i++) {
            // Check only within the elements we've added so far
            bool isDuplicate = false;
            for (uint j = 0; j < resultLength; j++) {
                if (temp[j] == array2[i]) {
                    isDuplicate = true;
                    break;
                }
            }
            
            if (!isDuplicate) {
                temp[resultLength] = array2[i];
                resultLength++;
            }
        }
        
        // Create final array of exact size
        address[] memory result = new address[](resultLength);
        for (uint i = 0; i < resultLength; i++) {
            result[i] = temp[i];
        }
        
        return result;
    }
    
```

test/V1/DiamondProxy/DisputeManager/Timeout.test.ts
```diff
@@ -63,213 +63,8 @@ describe("DisputeManagerProxy", function () {
     });
 
     describe("Timeout", function () {
-        it("Timeout - transaction 0", async function () {
```
daiagi, line 66: why?
- restore the tests
- use `it.skip` if they are failing, write note about it
- adept the tests to the new impl and make them pass

daiagi (changes requested):

contracts/V1/StateChannelDiamondProxy/StateChannelManagerStorage.sol
```diff
@@ -10,17 +10,48 @@ contract StateChannelManagerStorage {
...
-    mapping(bytes32 => uint) latestFork; // [channelId] -> latestFork
-    //TODO* - this can map to a hash, but it also has to store keys[] and timestamp
-    mapping(bytes32 => mapping(uint => ForkDataAvailability)) postedBlockCalldata; //[channelId][forkCnt].map[transactionCnt][address] -> BlockCalldata
-    //TODO* - enough only to store hash (commitiment)
-    mapping(bytes32 => Dispute) disputes; // disputes[channelId] => Dispute #only 1 dispute per fork and at a time
+
+    // =================== State on chain storage ==================
+
+    /// @dev stateSnapshot Data
+    mapping(bytes32 => StateSnapshot) stateSnapshots;
```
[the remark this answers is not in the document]
lukastanisic99: Every block/dispute commits to a snapshot and majority of them are off-chain or are part of commitments on-chain. However, a single snapshot is persisted on-chain in storage for every channel/rollup and that snapshot is updated when a new one is submitted with a proof. The snapshot itself doesn't store the whole state, it also stores commitments and data that is generally useful to other parts of the system like who are the participants. You're right this can be 'compressed' more into a single hash and every time other parts of the system need to read it, the data is provided and verified against the hash. Will probably update that later, but in general this shouldn't be expensive since the snapshot is a few bytes (depending mostly on the participant count)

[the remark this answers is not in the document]
daiagi: got it

lukastanisic99:
This wasn't done and test were failing, how did the branch protection rules allowed this to merge into master? @daiagi did you have to force merge? 

For now it doesn't matter much since no one is really using the SDK and it's not auto-deploying to npm, but thought the branch protection rules would prevent this

PR #26, opened 2025-04-28
Dispute update

lukastanisic99:
@MrishoLukamba take a look and ask if something is unclear. I've changed a few things - hopefully it's more clear now.  I've moved your fraud proofs to a separate file "FraudProofFacet" and connected it to the main dispute logic. I didn't check your particular fraud proof implementations yet, so I just commented them out. Gave a template to implementing fraud proofs that find and execute the correct fraudproof in O(1) instead of checking everything O(n). This is 'part 1' and I'll implement the rest later. I didn't test or change the typescript code - that will have to follow. Doing the contracts first. Will see to finish everything in a day or 2.

@daiagi also take a look and let me know if something is unclear or could be done in a better way


contracts/V1/StateChannelDiamondProxy/StateChannelUtilLibrary.sol
```diff
@@ -90,33 +90,57 @@ library StateChannelUtilLibrary {
...
+        for (uint i = 0; i < array1.length; i++) {
+            result[i] = array1[i];
+        }
+        for (uint i = 0; i < array2.length; i++) {
+            result[array1.length + i] = array2[i];
+        }
+       return result;
+    }
+
+    function areAddressArraysEqual(
```
daiagi, line 132: can you estimate the typical size of arrays?

maybe hash based approach could be more efficient
```solidity
function areAddressArraysEqual(
    address[] memory array1,
    address[] memory array2
) internal pure returns (bool) {
    if (array1.length != array2.length) {
        return false;
    }
    
    bytes32 hash1 = keccak256(abi.encode(array1));
    bytes32 hash2 = keccak256(abi.encode(array2));
    
    return hash1 == hash2;
}
```

contracts/V1/StateChannelDiamondProxy/StateChannelUtilLibrary.sol
```diff
@@ -90,33 +90,57 @@ library StateChannelUtilLibrary {
...
-        for (uint i = 0; i < currentSetLength; i++) {
-            if (set[i] == adr) return currentSetLength;
-        }
-        set[currentSetLength] = adr;
-        return currentSetLength + 1;
+        return currentThresholdCount;
     }
 
     /// @dev Concatenates two address arrays, but it does not add duplicates
-    function concatAddressArrays(address[] memory array1, address[] memory array2, uint array1Length) internal pure returns (address[] memory) {
```
daiagi, line 114: remove comment above

daiagi (commented):
looks like you adopted my more "functional" style of writing
i like it

contracts/V1/StateChannelDiamondProxy/StateChannelUtilLibrary.sol
lukastanisic99: With hash you still have to hash the entire array even though it's implicit and hashing is more expensive than comparison, but either way this is only used in one place for now, to compare on-chain slashed, so at most N where N is the number of players (so it's small) - we have other places that are less efficient and consume more gas that we can optimize later

lukastanisic99:
> looks like you adopted my more "functional" style of writing i like it

I think this one will be your favorite :D 
https://github.com/peer3to/state-channels-plus/blob/5e7918804b0dba3b1ae9604579fb9b36295efe16/contracts/V1/DisputeTypes.sol#L198

Already made an artificial context that will be part of all fraud proof verifications :D - yet to determine what will be the actual context and if needed actually. Initially wanted to give it the entire dispute context, but realized later if I want to reuse the code also for generating the dispute, the fraud proof stuff needs to run without the dispute context, at least without the posterior stuff like outputState(Snapshot). Mostly I think for now what's useful is the channelId so if you're slashed in 1 channel, it can't be 'replayed' in another channel

MrishoLukamba:
a lot of things changed even in context wise, like even per doc

contracts/V1/DisputeTypes.sol
```diff
@@ -174,9 +176,25 @@ struct DisputePair {
 
 /// @dev data for dispute auditing
 struct DisputeAuditingData {
-    bytes genesisStateSnapshot;
-    bytes latestStateSnapshot;
+    StateSnapshot genesisStateSnapshot;
+    StateSnapshot latestStateSnapshot;
+    StateSnapshot outputStateSnapshot;
+    StateSnapshot[] milestoneSnapshots; //for K milestones there will be K-1 snapshots, since the first milestone is the genesisSnapshot
     bytes latestStateStateMachineState;
```
daiagi, line 183: `latestStateStateMachineState`
need better name for this

daiagi (approved):

daiagi:
:rocket:  merging

contracts/V1/DisputeTypes.sol
```diff
@@ -182,19 +179,20 @@ struct DisputeAuditingData {
...
+    // ========================== optional ===============================
     Dispute previousDispute; // (optional) needed to verify 'this' dispute genesis against the previous dispute outputSnapshot or genesisSnapshot (in the case of a recursive dispute) - if not present, genesis is the latest on-chain Snapshot
     uint previousDisputeTimestamp; // (optional) needed to verify the commitment of the previous dispute
 }
 
-struct DisputeData{
+struct DisputeData {
     DisputePair[] disputePairs;
-    address[] onChainSlashedParticipants;
+    address[] slashedParticipants;
```
lukastanisic99, line 189: In my opinion this (onChainSlashes) should remain as a clear indicator that it's different from slashedParticipants, since populating this actually means that auditing was executed on-chain, while if not executed they would still be slashed 'off-chain' and committed 'on-chain' but this explicit storage array wouldn't reflect it and only when the snapshot is eventually updated would the participant set reflect it.

Since every time we have a fraud proof we actually run auditing on-chain to actually populate this array and prevent the byzantine participant from participating in the 'dispute game' going forward, this is logically correct even if we keep the obfuscated 'slashedParticipants'

lukastanisic99:
> a lot of things changed even in context wise, like even per doc

Mostly I expanded the doc with things that were not there like AuditingData and in some places added 1 or 2 more fields since they we're missing. In general the doc I've written is not a full spec it's guidance on how the protocol 'should' work and if something is missing it can be added and we can all review it

src/DisputeTypes.ts
```diff
@@ -36,19 +36,14 @@ export enum ProofType {
...
-        case ProofType.IncorrectData:
-            return IncorrectDataProofEthersType;
-        case ProofType.NewerState:
-            return NewerStateProofEthersType;
-        case ProofType.FoldPriorBlock:
-            return FoldPriorBlockProofEthersType;
-        case ProofType.BlockTooFarInFuture:
-            return BlockTooFarInFutureProofEthersType;
-    }
+const DISPUTE_PROOF_ETHERS_TYPES: Record<ProofType, string> = {
```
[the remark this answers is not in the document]
lukastanisic99: This is part of Luke's refactor - in general this is also good since it's O(1) lookup, but of course the specific proofs implementations will follow

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
```diff
@@ -127,515 +96,84 @@ contract DisputeManagerFacet is StateChannelCommon {
...
-        // check if the timeout peeer in original dispute is the disputer in recursive dispute
-        if(originalDispute.timeout.participant != recursiveDispute.disputer) {
-            slashParticipants[0] = dispute.disputer;
-            return (false, slashParticipants, abi.encode("TIMEOUT PRIOR INVALID: TIMEOUT PEER MISMATCH"));
-        }
-
-        slashParticipants[0] = recursiveDispute.disputer;
-        return (true, slashParticipants, new bytes(0));
+    // Doesn't do any checks and just applies all slashes, removals and joins to a specific stateMachineState and generates the outputStateMachineState - similar logic to playTransaction in the typescript code - this is done to help the backer generate a correct output state while forging the dispute
+    function playDisputeOutputGeneration(
```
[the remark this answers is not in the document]
lukastanisic99: Yeah I put that there for now so you know it's the same 'reasoning' as playTransaction in the typrescript code - but yes the name is not ideal. We can't name it 'executeDispute' since that too is misleading, this can run before disputes and when you run it locally it helps you generate the output that you will use in disputes. The 'playTransaction' helps you to build the block data structure and this helps you to build the dispute data structure. It's also reused within the dispute system so it has deterministic behavior. Maybe 'generateDisputeOutput'

src/DisputeTypes.ts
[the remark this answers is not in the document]
MrishoLukamba: okey lets merge

PR #30, opened 2025-05-01
posting blockCalldata

lukastanisic99:
This is everything we need for posting a block as calldata on-chain. I've left some comments above the function explaining thigs briefly.
I'll also expand here a bit more:
So in general anyone (even non-participants) can invoke this function and post whatever (junk), but they pay the on-chain fees. In general it's cheap. It's a small function that modifies a single storage slot. Transactions don't get cheaper than this - even a simple transfer is more expensive since it modifies 2 storage slots. The msg.sender takes full responsibility of the data committed. If msg.sender is not a participant, the off-chain clients just ignore the emitted events.

The maxTimestamp prevents race condition. In general there is some time logic that I didn't put in the docs (yet) - I'll explain it here and feel free to update the docs with it.

Time logic:
The off-chain blocks that are linked contain a timestamp and the difference between timestamps in 2 consecutive blocks has to be <= p2pTime. When a peer receives a block, apart from doing all other checks it checks the timestamp and compares it to the **previous block timestamp** . It needs to satisfy <=p2pTime. Apart from that, since peers can put arbitrary timestamps in the block, the timestamp is also compared to the local clock of the peer (subjective perception) and the difference can NOT be more than agreementTime (if it is the peer doesn't cosign the block). From the perspective of the block author, they'll wait agreementTime and expect to collect an N/N threshold. If for any reason that threshold is NOT reached, after agreementTime the block author has chainFallbackTime to post the block on-chain. So the check that needs to hold on-chain eventually is:

**previous block timestamp** + p2pTime + agreementTime + chainFallbackTime >= on-chain block.timetamp

The left part of the equation is 'maxTimestamp' used when posting blockCalldata.

One more thing to keep in mind! 
Since the on-chain timestamp 'overrides' the off-chain timestamp, we need to make sure that we account for that both on-chain and off-chain. So the **previous block timestamp** = previous block commitment posted ? emitted timestamp : original timestamp 

maxTimestamp also takes that into account 

@daiagi @MrishoLukamba take a look and let me know if you have questions

### DIFF
--- a/contracts/V1/StateChannelDiamondProxy/AStateChannelManagerProxy.sol
+++ b/contracts/V1/StateChannelDiamondProxy/AStateChannelManagerProxy.sol
@@ -190,6 +190,13 @@ abstract contract AStateChannelManagerProxy is
         require(block.timestamp <= maxTimestamp, ErrorBlockCalldataTimestampTooLate());
         bytes32 commitment = keccak256(abi.encode(signedBlock,block.timestamp));
         Block memory _block = abi.decode(signedBlock.encodedBlock, (Block));
+        
+        //Don't allow overwriting the blockCalldataCommitment if it already exists
+        require(blockCalldataCommitments[_block.transaction.header.channelId]
+        [msg.sender]
+        [_block.transaction.header.forkCnt]
+        [_block.transaction.header.transactionCnt] == bytes32(0), ErrorBlockCalldataAlreadyPosted());
+
         blockCalldataCommitments
         [_block.transaction.header.channelId]
         [msg.sender]
--- a/contracts/V1/StateChannelDiamondProxy/AStateChannelManagerProxy.sol
+++ b/contracts/V1/StateChannelDiamondProxy/AStateChannelManagerProxy.sol
@@ -177,10 +177,31 @@ abstract contract AStateChannelManagerProxy is
         // return (success, abi.decode(encodedReturnValue, (bytes)));
     }
 
+    /**
+        Posting calldata is lightweight, since it persists a signle hash/commitment. 
+        It's enough to check just the maxTimestamp safety guard that protects against race conditions, since everything else is committed in the block.
+        We also don't allow overwriting the blockCalldataCommitment if it already exists. 
+        We don't even have to check the siganture of the signedBlock, since the msg.sender takes the responsibility of provifing correct data.
+        If the msg.sender provides junk(an invalid SignedBlock), a fraud proof can slash the msg.sender, by verifying the junk data against the committment.
+        If msg.sender is not part of the channel, other peers will ignore emited events and commitments. The sender will still pay tx fees on-chain.
+     */
+    function postBlockCalldata(SignedBlock memory signedBlock, uint maxTimestamp) public override {
+        //Time is the only race condition we need to take into account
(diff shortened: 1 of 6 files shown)


contracts/V1/StateChannelDiamondProxy/AStateChannelManagerProxy.sol
```diff
@@ -177,10 +177,38 @@ abstract contract AStateChannelManagerProxy is
...
+        bytes32 commitment = keccak256(abi.encode(signedBlock,block.timestamp));
+        Block memory _block = abi.decode(signedBlock.encodedBlock, (Block));
+        
+        //Don't allow overwriting the blockCalldataCommitment if it already exists
+        require(blockCalldataCommitments[_block.transaction.header.channelId]
+        [msg.sender]
+        [_block.transaction.header.forkCnt]
+        [_block.transaction.header.transactionCnt] == bytes32(0), ErrorBlockCalldataAlreadyPosted());
+
+        blockCalldataCommitments
```
MrishoLukamba, line 200: also have helper function for this

contracts/V1/StateChannelDiamondProxy/AStateChannelManagerProxy.sol
```diff
@@ -177,10 +177,38 @@ abstract contract AStateChannelManagerProxy is
...
+        If msg.sender is not part of the channel, other peers will ignore emited events and commitments. The sender will still pay tx fees on-chain.
+     */
+    function postBlockCalldata(SignedBlock memory signedBlock, uint maxTimestamp) public override {
+        //Time is the only race condition we need to take into account
+        require(block.timestamp <= maxTimestamp, ErrorBlockCalldataTimestampTooLate());
+        bytes32 commitment = keccak256(abi.encode(signedBlock,block.timestamp));
+        Block memory _block = abi.decode(signedBlock.encodedBlock, (Block));
+        
+        //Don't allow overwriting the blockCalldataCommitment if it already exists
+        require(blockCalldataCommitments[_block.transaction.header.channelId]
```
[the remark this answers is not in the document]
lukastanisic99: Yes i'm aware of getBlockCallDataCommitment - it's not used here since this function is simple and needs to be as cheap as possible since it's one of the few that will actually run even without slashing and the costs for this function are not refundable as with disputes where someone will get slashed and you can refund the winning party. In general solidity is a bit weird since the less modular and less reusable your code is the more gas efficient it is - each function call is an jump instruction and you have a return + copying function arguments - fine to do it in auditing since auditing is very expensive and the few gas costs don't make a difference + the compiler can still inline it

contracts/V1/StateChannelDiamondProxy/AStateChannelManagerProxy.sol
lukastanisic99: Not sure which function you refer for this - this should be the only function that can set the commitment - it shouldn't be set or overwritten anywhere else

MrishoLukamba: a setter for postingBlockCallData

MrishoLukamba (approved):

PR #31, opened 2025-05-04
Balance tracking and operations, invariant & join/exit blockchain verification in auditing - auditing logic should be complete

lukastanisic99:
Hopefully this concludes the auditing logic - what's missing are snapshot updates (Luke is working on that), entry point for joining, recursive dispute logic and to challenge disputes (reuse auditing + allow a new recursive dispute commitment to be created if auditing is successful, taking into account the results from auditing)

Some structural changes also with this PR - from JoinChannel/ExitChannel the 'amount' and 'data' fields have been extracted into a 'balance' field. The aggregated on-chain balance for both deposits and withdrawals is persisted on-chain and is updated exclusively on-chain when deposits/withdrawals are processed (not related to the ones used in the snapshot). They're used with the ones in the snapshot to check the invariant. The AStateMachine is given more interface functions that need to be implemented to interpret balances and do operations with them. This can not be generic and is left to he specific implementation - some may use NFTs as 'balances', by allowing an arbitrary data field (bytes) the concrete implementation is given all the freedom of what a 'balance' is and how to manage it. 

Further, I've included the invariant check in auditing since we've managed to design joining in a way where all the auditing data is gossiped to the joiners for them to run auditing locally and sign off on the result and with their signatures auditing can be skipped on-chain. So auditing will only run if someone will actually get slashed, and the stake will be used to refund the winning party. We still need to test it and benchmark how much gas it actually uses in the worst case so we can estimate limits. I should update the docs with this change, but I'll probably do it tomorrow.


contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
```diff
@@ -49,31 +49,33 @@ contract DisputeManagerFacet is StateChannelCommon {
...
-        //TODO! Verify exitChannelBlocks
+        require(_verifyJoinChannelBlocks(dispute, disputeAuditingData), ErrorDisputeJoinChannelBlocksInvalid());
+        require(_verifyExitChannelBlocks(dispute, disputeAuditingData), ErrorDisputeExitChannelBlocksInvalid());
 
         FraudProofVerificationContext memory poofContext = FraudProofVerificationContext({
             channelId: dispute.channelId
         }); 
-        (bytes memory encodedModifiedState, ExitChannel[] memory exitChannels) = playDisputeOutputGeneration(
+        (bytes memory encodedModifiedState, ExitChannelBlock memory eBlock, 
+        Balance memory totalDeposits, Balance memory totalWithdrawals) = playDisputeOutputGeneration(
```
MrishoLukamba, line 59: lets change the name now, so we don't make it canonical

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
```diff
@@ -49,31 +49,33 @@ contract DisputeManagerFacet is StateChannelCommon {
...
         require(_verifyStateProof(dispute, disputeAuditingData), ErrorDisputeStateProofInvalid());
-        //TODO! Verify exitChannelBlocks
+        require(_verifyJoinChannelBlocks(dispute, disputeAuditingData), ErrorDisputeJoinChannelBlocksInvalid());
+        require(_verifyExitChannelBlocks(dispute, disputeAuditingData), ErrorDisputeExitChannelBlocksInvalid());
 
         FraudProofVerificationContext memory poofContext = FraudProofVerificationContext({
             channelId: dispute.channelId
         }); 
-        (bytes memory encodedModifiedState, ExitChannel[] memory exitChannels) = playDisputeOutputGeneration(
+        (bytes memory encodedModifiedState, ExitChannelBlock memory eBlock,
```
MrishoLukamba, line 58: lets just use naming in full exitBlock

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
```diff
@@ -553,8 +521,56 @@ contract DisputeManagerFacet is StateChannelCommon {
...
+    function _verifyJoinChannelBlocks(Dispute memory dispute, DisputeAuditingData memory disputeAuditingData) internal pure returns (bool) {
+        //check joinChannelBlocks (linked to latestSateSnapshot, chained internally and outputStateSnapshot commits to the head)
+        bytes32 previousJoinChannelBlockHash = disputeAuditingData.latestStateSnapshot.latestJoinChannelBlockHash;
+        for(uint i = 0; i < disputeAuditingData.joinChannelBlocks.length; i++) {
+            if(previousJoinChannelBlockHash != disputeAuditingData.joinChannelBlocks[i].previousBlockHash) {
+                return false;
+            }
+            previousJoinChannelBlockHash = keccak256(abi.encode(disputeAuditingData.joinChannelBlocks[i]));
+        }
+        return previousJoinChannelBlockHash == disputeAuditingData.outputStateSnapshot.latestExitChannelBlockHash;
```
MrishoLukamba, line 535: why is it checking on existBlockChannelBlockHash on return

MrishoLukamba (changes requested):

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
lukastanisic99: changed it to generateDisputeOutputState

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
lukastanisic99: done

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
lukastanisic99: fixed this

lukastanisic99:
While looking at Mrishos latest PR - I've realized that someone posting blockCalldata of a block that 'looks valid', but is not cryptographically linked to the latestState would prevent us from timing them out, so I've added a 'isForce' flag that would allow in such a case to skip the race condition checks for timeout on-chain and force apply the timeout. 

The rational behind this is:
We can't produce a fraud proof and slash a block commitment that looks valid, since there is a non-zero chance the block author is building on a different reality because SOMEONE ELSE performed a double sign and for some reason no one is aware of it yet. However, since timeouts are ignored if slashes happen in a fork, it's safe to timeout the person regardless, since the dispute has to include a 'stateProof' and if the participant being timedout really is building on a wrong reality because someone else produced a doubleSign - the stateProof will provide that participant all the evidence they need to construct a double sign proof against someone which will have a side effect that the timeout is NOT applied in the recursive dispute. Otherwise if the person just committed a random block that doesn't actually build on any reality, the timeout will still hold and they will be removed, since they can't construct a fraud proof against someone else. I've added this in the doc too.

contracts/V1/StateChannelDiamondProxy/AStateChannelManagerProxy.sol
```diff
@@ -92,17 +92,18 @@ abstract contract AStateChannelManagerProxy is
...
         stateMachineImplementation.setState(encodedState);
         for (uint i = 0; i < joinCahnnels.length; i++) {
             bool success = stateMachineImplementation.joinChannel(
                 joinCahnnels[i]
             );
-            // require(success, "JoinChannel failed");
-            if (success) successCnt++;
+            // require(success, "Slash failed");
+            require(success,ErrorDisputeStateMachineJoiningFailed());
+            successCnt++;
```
MrishoLukamba, line 104: why do we need to count number of successes when this function should be atomic

MrishoLukamba (approved):

MrishoLukamba:
> If you spot anything else - let me know, otherwise I'll merge this later today

if its done lets merge it now, so that the TS side can begin on my end

lukastanisic99:
This part is done, but there are a few things that are missing still as mentioned above - I'll see to get them implemented tomorrow, but I'll create a separate PR for that - you can work on the TS in parallel - For auditing locally a contract deployment for the local environment is also needed + a wrapper to just push missing commitments 

Maybe we deploy the same diamond locally and add a facet to inject commitments into the state as the client becomes aware of them from observing the DLT

PR #32, opened 2025-05-04
fraud proofs

MrishoLukamba:


contracts/V1/StateChannelDiamondProxy/FraudProofFacet.sol
```diff
@@ -47,448 +43,251 @@ contract FraudProofFacet is StateChannelCommon {
...
+
+    // ------------------------------- Block Fraud Proofs --------------------------------------- 
+    function _handleBlockDoubleSign(bytes memory encodedProof, FraudProofVerificationContext memory fraudProofVerificationContext) pure internal returns (address) {
+        BlockDoubleSignProof memory blockDoubleSignProof = abi.decode(encodedProof, (BlockDoubleSignProof));
+
+        Block memory block1 = abi.decode(blockDoubleSignProof.block1.encodedBlock, (Block));
+        Block memory block2 = abi.decode(blockDoubleSignProof.block2.encodedBlock, (Block));
+
+        if(fraudProofVerificationContext.channelId != block1.transaction.header.channelId || fraudProofVerificationContext.channelId != block2.transaction.header.channelId) {
+            return address(0);
```
lukastanisic99, line 56: throw an error on failure so the low level function call (delegatecall) will catch it - so we should never return address(0)

lukastanisic99: But still better to throw a specific error so the client can catch the specific error and interpret it (e.g. missing commitment [channelId][adr][fork][height])

MrishoLukamba: so there are conflicting views
1. If we return address(0), we already handle it in verifyfraudProof
2. if we return an error we have to specify error types so that the TS client can handle specific errors.. but that can grow so much as number of errors. (is it sustainable)

lukastanisic99: In general we have many custom errors that don't need any special handling, but help us understand what failed - The few errors that need custom handling are related to commitments and notify the client to try and fetch those commitments on-chain to reconstruct the state/environment off-chain. Everything else is just relevant for logs and debugging and doesn't need custom behavior

contracts/V1/StateChannelDiamondProxy/FraudProofFacet.sol
```diff
@@ -47,448 +43,251 @@ contract FraudProofFacet is StateChannelCommon {
...
+        }
+
+        if(block1.stateSnapshotHash != block2.stateSnapshotHash && block1.previousBlockHash != block2.previousBlockHash) {
```
lukastanisic99, line 59: this comparison is not correct - 2 blocks with the same snapshot (state) and different previousBlockHash would be slashed. 
A simple state machine that does addition and subtraction can have states [2, 4 , 2, 4 ,2, 4] after each state transition. Presenting blocks 1 and 3 for example, the state of the state machine is the same while the previous block hash is different - this is NOT a double sign, but would be slashed under your logic. 

A double sign is when 2 different blocks are authored by the same author (and signed) for the same same fork and blockHeight.

The check is: 
if (block1.fork == block2.fork && block1.blockHeight == block2.blockHeight && hash(block1) != hash(block2)){
    slash
} 

comparing the previous blockHash is not enough since an attacker can issue 2 different blocks for the same block height that have different previousBlockHash

contracts/V1/StateChannelDiamondProxy/FraudProofFacet.sol
```diff
@@ -47,448 +43,251 @@ contract FraudProofFacet is StateChannelCommon {
...
-        // Implementation for handling BlockEmptyBlock proof
-        return address(0); // Replace with actual logic
+    function _handleBlockEmptyBlock(bytes memory encodedProof, FraudProofVerificationContext memory fraudProofVerificationContext) pure internal returns (address) {
+        BlockEmptyProof memory blockEmptyProof = abi.decode(encodedProof, (BlockEmptyProof));
+        Block memory fraudBlock = abi.decode(blockEmptyProof.emptyBlock.encodedBlock, (Block));
+
+        if(fraudProofVerificationContext.channelId != fraudBlock.transaction.header.channelId) {
+            return address(0);
+        }
+        if(fraudBlock.transaction.header.transactionCnt != uint(0)) {
```
lukastanisic99, line 84: transactionCnt is blockHeight - It's not the number of transactions in a block. 
An empty block is a one that has 0 transactions or a one that does NOT transition the state, but no error is thrown, otherwise if an error is thrown during the state transition, it would be an invalid state transition. The simplest check that will always hold is if the prior state == posterior state. If this is true -> slash

Compare the snapshots linked to this block and the previous block - in the case that this is the first block (blockHeight == 0) there is no previous block, but the blocks builds on top of the genesisStateSnapshot for the fork and previousBlockHash == genesisStateSnapshotHash

contracts/V1/StateChannelDiamondProxy/FraudProofFacet.sol
```diff
@@ -47,448 +43,251 @@ contract FraudProofFacet is StateChannelCommon {
...
+            fraudProofVerificationContext.channelId,
+            blockInvalidSTProof.encodedLatestState,
+            fraudBlock.transaction
+        );
+
+        if (isTransitionValid) {
+            return address(0);
+        } 
+
+        if(fraudBlock.stateSnapshotHash == keccak256(encodedModifiedState)){
```
lukastanisic99, line 112: This is also not correct - you can't compare hash of a snapshot to a hash of a state(bytes) - you need to reconstruct the entire snapshot and compare it to the commitment - see how I constructed the snapshot during auditing 

Assume previous snapshot is correct and based on that do the state transition and calculate the output snapshot and compare. 
Deposits stay the same since this can only be changed on-chain, and if exits happen the state transition function needs to return exitChannels [ ]

contracts/V1/StateChannelDiamondProxy/FraudProofFacet.sol
```diff
@@ -47,448 +43,251 @@ contract FraudProofFacet is StateChannelCommon {
...
+        }       
+        address signer = StateChannelUtilLibrary.retriveSignerAddress(
+            blockInvalidSTProof.invalidBlock.encodedBlock,
+            blockInvalidSTProof.invalidBlock.signature
+        );
+        return signer;
     }
 
     function _handleBlockOutOfGas(bytes memory encodedProof, FraudProofVerificationContext memory fraudProofVerificationContext) internal returns (address) {
         // Implementation for handling BlockOutOfGas proof
```
lukastanisic99, line 123: This could also be done in the same way as auditing with a low level function call that has a gas limit (the same limit on the client and on-chain) so if the gas limit is used the execution would fail and this would fall in the category of an 'invalidStateTransition'

contracts/V1/StateChannelDiamondProxy/FraudProofFacet.sol
```diff
@@ -47,448 +43,251 @@ contract FraudProofFacet is StateChannelCommon {
...
+        if(!isAvailable && commitment != originalDisputeCommitment){
+            return address(0);
+        }
 
-    function _handleTimeoutPriorInvalid(bytes memory encodedProof, FraudProofVerificationContext memory fraudProofVerificationContext) internal returns (address) {
-        // Implementation for handling TimeoutPriorInvalid proof
-        return address(0); // Replace with actual logic
-    }
+        uint latestStateHeight = _getLatestHeight(originalTimedOutDispute.stateProof);
+        if(latestStateHeight != thresholdBlock.transaction.header.transactionCnt) {
```
lukastanisic99, line 146: This is over engineered and not correct - the timedout block is latestStateHeight + 1, but you don't need this at all. 

After checking the dispute commitment, you just need the latestStateSnapshot (the dispute commits to it) so you can provide it as part of the proof - from there you get the participants list and all you have to do is:
check that the block is the correct fork and height as the one registered in Timeout and if so, check the threshold. The UtilityLibrary has a function to check the threshold - for this we don't do virtual voting - so you just provide all the signatures from the BlockConfirmation with the expected participant list given in the latestStateSnapshot

contracts/V1/StateChannelDiamondProxy/FraudProofFacet.sol
```diff
@@ -47,448 +43,251 @@ contract FraudProofFacet is StateChannelCommon {
...
+            originalDispute,
+            timeoutPriorInvalidProof.originalDisputeTimestamp
+        ));
+
+        (bool isAvailable, bytes32 commitment) = getDisputeCommitment(fraudProofVerificationContext.channelId, recursiveDispute.disputeIndex);
+
+        if(!isAvailable && commitment != recursiveDisputeCommitment) {
+            return address(0);
+        }
+        if(recursiveDispute.previousRecursiveDisputeIndex == type(uint256).max){
```
lukastanisic99, line 194: This is ok - you check if it's not a recursive dispute, but if it is you also need to check that recursiveDispute.previousRecursiveDisputeIndex == originalDispute.disputeIndex

contracts/V1/StateChannelDiamondProxy/FraudProofFacet.sol
```diff
@@ -47,448 +43,251 @@ contract FraudProofFacet is StateChannelCommon {
...
+            return address(0);
+        }
+            
+        // check if the original timeout is greater than the recursive timeout
+        if(originalDispute.timeout.blockHeight < recursiveDispute.timeout.blockHeight) {
+            return address(0);
+        }
+        // check if the timeout peeer in original dispute is the disputer in recursive dispute
+        if(originalDispute.timeout.participant != recursiveDispute.disputer) {
+            return address(0);
```
lukastanisic99, line 210: Why is this check needed?

lukastanisic99: Anyone should be able to submit a recursive dispute

contracts/V1/StateChannelDiamondProxy/FraudProofFacet.sol
```diff
@@ -47,448 +43,251 @@ contract FraudProofFacet is StateChannelCommon {
...
+        }
+
+        return recursiveDispute.disputer;
     }
+    
+    // ------------------------------------ Dispute Fraud Proofs ------------------------------------
+    function _handleDisputeNotLatestState(bytes memory encodedProof, FraudProofVerificationContext memory fraudProofVerificationContext) view internal returns (address) {
+        DisputeNotLatestStateProof memory disputeNotLatestStateProof = abi.decode(encodedProof, (DisputeNotLatestStateProof));
+        address[] memory slashParticipants = new address[](1);
```
lukastanisic99, line 219: This is not used

lukastanisic99 (changes requested):
I've made comments above - let's discuss tomorrow in our call

contracts/V1/StateChannelDiamondProxy/FraudProofFacet.sol
MrishoLukamba: ive thought about this, but even on this suggestion check-in  if the prior and posterior state equality is wrong

lukastanisic99: Not sure I understand your comment. Why do you think it's wrong?

The reason why we force non-empty blocks is that the leader election algorithm is a function of the state. If the state doesn't change, the leader is the same. If the leader is the same it's really easy to indefinitely stall by producing empty blocks that don't transition the state.

MrishoLukamba: like from your example of simple addition
[1,3,0,0,9]
the block 3,4 will have same state and no transition

lukastanisic99: Yes and we don't want that - so block 4 (addition +0) should be invalid in your example.

Keep in mind that even the simple state machine can use round robin internally, so even if you do +0 and the sum doesn't change, the counter for the leader election will change so state(i) != state(i+1) even though the sum is the same. 

Empty blocks don't transition the state so state(i) == state(i+1) which we don't want.
Even non-empty blocks may not transition the state if the internal state machine logic is implemented incorrectly, but just having round robin internally will always transition the state with non-empty blocks.

contracts/V1/DisputeTypes.sol
```diff
@@ -110,17 +110,22 @@ enum ProofType {
 // ========================== Block related fraud proofs ==========================
 struct BlockEmptyProof {
     SignedBlock emptyBlock;
+    StateSnapshot latestStateSnapshot;
+    bytes32 previousStateSnapshotHash;
+    bytes previousStateMachineState;
```
lukastanisic99, line 115: struct BlockEmptyProof {
    SignedBlock emptyBlock;
    SignedBlock previousBlock;
    }
    
    The check is: 
    if(emptyBlock.height == 0 && emptyBlock.previousBlockHash (this is the genesis snapshot hash) == emptyBlock.stateSnapshotHash) -> slash
    if(emptyBlock.previousBlockHash == hash(previousBlock) && emptyBlock.stateSnapshot == previousBlock.stateSnapshot) -> slash
    revert ErrorIncorrectEmptyBlockFraudProof

contracts/V1/DisputeTypes.sol
```diff
@@ -110,17 +110,22 @@ enum ProofType {
...
 }
 
 struct BlockInvalidStateTransitionProof {
     SignedBlock invalidBlock;
     /// @notice the latest state being transitioned
-    bytes encodedLatestState;
+    DisputeAuditingData disputeFraudData;
+    Dispute dispute;
```
lukastanisic99, line 122: ????????????????????

contracts/V1/StateChannelDiamondProxy/AStateChannelManagerProxy.sol
```diff
@@ -182,15 +185,22 @@ abstract contract AStateChannelManagerProxy is
      */
     function postBlockCalldata(SignedBlock memory signedBlock, uint maxTimestamp) public override {
         //Time is the only race condition we need to take into account
-        require(block.timestamp <= maxTimestamp, ErrorBlockCalldataTimestampTooLate());
+        
+        if (block.timestamp > maxTimestamp) {
+           revert ErrorBlockCalldataTimestampTooLate();
+        }
```
lukastanisic99, line 191: why change the 'require' to 'if -> revert'

contracts/V1/StateChannelDiamondProxy/FraudProofFacet.sol
```diff
@@ -53,11 +50,11 @@ contract FraudProofFacet is StateChannelCommon {
...
 
         if(fraudProofVerificationContext.channelId != block1.transaction.header.channelId || fraudProofVerificationContext.channelId != block2.transaction.header.channelId) {
-            return address(0);
+            revert ErrorNotSameChannelId();
         }
 
-        if(block1.stateSnapshotHash != block2.stateSnapshotHash && block1.previousBlockHash != block2.previousBlockHash) {
-            return address(0);
+        if (block1.transaction.header.forkCnt != block2.transaction.header.forkCnt && block1.transaction.header.transactionCnt != block2.transaction.header.transactionCnt && keccak256(abi.encode(block1)) != keccak256(abi.encode(block2))){
+            revert ErrorDoubleSignBlocksNotSame();
```
lukastanisic99, line 57: This logic is not correct - please think about this. If you want to write logic like this I'd advice writing first what is a double sign:
if(fork == fork && height == height && hash != hash) -> this is a double sign. So you either negate this all together as '!(double sign)' and use your 'if - revert' syntax or just use require OR pass through with '!' mathematically:
(!(fork == fork && height == height && hash != hash)) <=> (fork != fork || height != height || hash == hash)

contracts/V1/StateChannelDiamondProxy/FraudProofFacet.sol
```diff
@@ -74,15 +71,37 @@ contract FraudProofFacet is StateChannelCommon {
...
+            );
+            if(!isTransitionValid){
+                revert ErrorInvalidBlockStateTransition();
+            }
+            if(latestStateSnapshot.stateMachineStateHash != keccak256(blockTransitionEncodedModifiedState)){
+                revert ErrorInvalidBlockState();
+            }
+            if(previousStateSnapshotHash != keccak256(abi.encode(latestStateSnapshot))){
+                revert ErrorInvalidStateSnapshot();
+            }
```
lukastanisic99, line 104: We'll discuss this and everything else in the call tomorrow

contracts/V1/StateChannelDiamondProxy/FraudProofFacet.sol
```diff
@@ -120,8 +164,33 @@ contract FraudProofFacet is StateChannelCommon {
     }
 
     function _handleBlockOutOfGas(bytes memory encodedProof, FraudProofVerificationContext memory fraudProofVerificationContext) internal returns (address) {
-        // Implementation for handling BlockOutOfGas proof
-        return address(0); // Replace with actual logic
+        BlockOutOfGasProof memory blockOutOfGasProof = abi.decode(encodedProof, (BlockOutOfGasProof));
+        bytes memory latestStateStateMachineState = blockOutOfGasProof.latestStateStateMachineState;
+        Block memory fraudBlock = abi.decode(blockOutOfGasProof.invalidBlock.encodedBlock, (Block));
+
```
lukastanisic99, line 170: if you want to replay the state transition you need to verify everything - the previous state (the previous block is linked to the outOfGasBlock and the snapshot is correct and the state is correct) than you can replay with a gas limits, but the easiest is to set the gas limit directly when doing the STF and if it returns out of gas it's an invalid state transition and fraud proof is not needed explicitly.

contracts/V1/StateChannelDiamondProxy/AStateChannelManagerProxy.sol
MrishoLukamba: So, require is supoosed to return a string error not custom error,

contracts/V1/StateChannelDiamondProxy/FraudProofFacet.sol
MrishoLukamba: Yes the logic follows this
if the fork is same, height is same and the content of the block is same, then its a double sign

contracts/V1/StateChannelDiamondProxy/FraudProofFacet.sol
MrishoLukamba: this the peer will provide the state on which we will use to transition, the prevStateSnapshotHash, the latestSnapShot, so when transitioning, the encodedModified state will be checked against latestSnapShot latestStatehash.

Where is the logic incosistency?

contracts/V1/StateChannelDiamondProxy/FraudProofFacet.sol
```diff
@@ -94,23 +113,48 @@ contract FraudProofFacet is StateChannelCommon {
     function _handleBlockInvalidStateTransition(bytes memory encodedProof, FraudProofVerificationContext memory fraudProofVerificationContext) internal returns (address) {
         BlockInvalidStateTransitionProof memory blockInvalidSTProof = abi.decode(encodedProof, (BlockInvalidStateTransitionProof));
         Block memory fraudBlock = abi.decode(blockInvalidSTProof.invalidBlock.encodedBlock, (Block));
-        
+        DisputeAuditingData memory disputeAuditingData = blockInvalidSTProof.disputeFraudData;
+        Dispute memory dispute = blockInvalidSTProof.dispute;
+
```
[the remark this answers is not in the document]
MrishoLukamba: This is literally in the docs and if a peer doesn't follow the rules and we can prove that isn't that an objective fraud?

contracts/V1/StateChannelDiamondProxy/AStateChannelManagerProxy.sol
lukastanisic99: https://soliditylang.org/blog/2024/05/21/solidity-0.8.26-release-announcement/

I'm probably using the latest compiler since on my end it compiles without errors

contracts/V1/StateChannelDiamondProxy/FraudProofFacet.sol
lukastanisic99: if the content is the same implicitly the fork and height are the same too and that IS NOT a double sign - that's the same block (blockA and blockA) - a double sign is 2 conflicting blocks (blockA and blockB) by the same author for the same fork and block height.

lukastanisic99:
Also add foundry to the readme how to us it and add what needs to be added to .gitignore and .npmignore - probably the lib directory should be included in both files

contracts/V1/StateChannelDiamondProxy/FraudProofFacet.sol
```diff
@@ -71,37 +75,27 @@ contract FraudProofFacet is StateChannelCommon {
...
     }
 
-    function _handleBlockEmptyBlock(bytes memory encodedProof, FraudProofVerificationContext memory fraudProofVerificationContext) internal returns (address) {
+    function _handleBlockEmptyBlock(bytes memory encodedProof, FraudProofVerificationContext memory fraudProofVerificationContext) internal pure returns (address) {
         BlockEmptyProof memory blockEmptyProof = abi.decode(encodedProof, (BlockEmptyProof));
-        StateSnapshot memory latestStateSnapshot = blockEmptyProof.latestStateSnapshot;
-        bytes32 previousStateSnapshotHash = blockEmptyProof.previousStateSnapshotHash;
-        bytes memory previousStateMachineState = blockEmptyProof.previousStateMachineState;
         Block memory fraudBlock = abi.decode(blockEmptyProof.emptyBlock.encodedBlock, (Block));
+        Block memory previousBlock = abi.decode(blockEmptyProof.previousBlock.encodedBlock, (Block));
```
lukastanisic99, line 81: don't decode the previousBlock immediately since there might not be a previous block (if fraudBlock.height == 0), so first check the case blockHeight == 0 and if it's not that than decode previousBlock (previousBlock is optional)

contracts/V1/StateChannelDiamondProxy/FraudProofFacet.sol
```diff
@@ -71,37 +75,27 @@ contract FraudProofFacet is StateChannelCommon {
...
 
         if(fraudProofVerificationContext.channelId != fraudBlock.transaction.header.channelId) {
             revert ErrorNotSameChannelId();
         }
 
+        if(fraudBlock.previousBlockHash != keccak256(abi.encode(previousBlock))){
+            revert ErrorLinkingPreviousBlock();
```
lukastanisic99, line 88: Also check this after the blockHeight == 0 case and don't encode since you already have the encoded previous block in the proof, just reuse it

contracts/V1/StateChannelDiamondProxy/FraudProofFacet.sol
```diff
@@ -113,86 +107,48 @@ contract FraudProofFacet is StateChannelCommon {
     function _handleBlockInvalidStateTransition(bytes memory encodedProof, FraudProofVerificationContext memory fraudProofVerificationContext) internal returns (address) {
         BlockInvalidStateTransitionProof memory blockInvalidSTProof = abi.decode(encodedProof, (BlockInvalidStateTransitionProof));
         Block memory fraudBlock = abi.decode(blockInvalidSTProof.invalidBlock.encodedBlock, (Block));
-        DisputeAuditingData memory disputeAuditingData = blockInvalidSTProof.disputeFraudData;
-        Dispute memory dispute = blockInvalidSTProof.dispute;
+        Block memory previousBlock = abi.decode(blockInvalidSTProof.previousBlock.encodedBlock, (Block));
+        StateSnapshot memory previousBlockStateSnapshot = blockInvalidSTProof.previousBlockStateSnapshot;
+        bytes memory previousStateStateMachineState = blockInvalidSTProof.previousStateStateMachineState;
```
lukastanisic99, line 112: You forgot to check the case blockHeight == 0. If it's not 0, then you have a previous block (and that's when you should decode it - again optional). 

since you may not have a previous block rename 'previousBlockStateSnapshot' into 'previousStateSnapshot', since in the case that height == 0 this is the genesis snapshot

contracts/V1/StateChannelDiamondProxy/FraudProofFacet.sol
```diff
@@ -113,86 +107,48 @@ contract FraudProofFacet is StateChannelCommon {
...
         );
+        if(!isSuccess){
+            return signer;
+        }
+        StateSnapshot memory newStateSnapshot = StateSnapshot({
+            stateMachineStateHash: keccak256(encodedModifiedState),
+            participants: getStatemachineParticipants(encodedModifiedState),
+            forkCnt: previousBlockStateSnapshot.forkCnt,
+            latestJoinChannelBlockHash: previousBlockStateSnapshot.latestJoinChannelBlockHash,
+            latestExitChannelBlockHash: previousBlockStateSnapshot.latestExitChannelBlockHash,
```
lukastanisic99, line 142: Good for now, but the STF needs to return exitChannels[ ] so we can calculate the latestExitChannelBlockHash and totalWithdrawals - feel free to mock it for now or try implementing that too - I'll see tonight when I get back from my meetings to implement that

MrishoLukamba: will check the STF and see

contracts/V1/StateChannelDiamondProxy/FraudProofFacet.sol
```diff
@@ -281,6 +237,71 @@ contract FraudProofFacet is StateChannelCommon {
...
+        if(!isOriginalDisputeAvailable && originalCommitment != originalDisputeCommitment) {
+            revert ErrorDisputeCommitmentNotAvailable();
+        }
+        if(invalidRecursiveDisputeTimestamp <= originalDisputeTimestamp + getChallengeTime()) {
+            revert ErrorWithinChallengePeriod();
+        }
+
+        // check if the disputer used the latest state in recursive dispute
+        
+        if(!_areSignedBlocksLinkedAndVerified(latestStateSignedBlocks, invalidRecursiveDispute.latestStateSnapshotHash) && invalidRecursiveDispute.latestStateSnapshotHash != keccak256(latestStateSnapshot)) {
```
lukastanisic99, line 278: up to here it's good

No need to check stateProof and are blocks linked - that’s part of auditing - assume everything in the disputes in correct (they’ve both passed auditing off-chain at least - if they haven’t they would be challenged and we wouldn’t be here) and just check linkage/relation to originalDispute 
Original and Recursive disputes are valid (they pass auditing) -> data within them is correct.   Recursive dispute needs to do AT LEAST 1 of:
- extend the list of slashed participants (recursiveDispute.fraudproofs.lenght + recursiveDispute.onChainSlashes.lenght > originalDispute.fraudproofs.lenght + originalDispute.onChainSlashes.lenght) : this is enough to check since both arrays are checked in auditing so if they’re not correct and applied successfully atomically auditing would fail and we wouldn’t be here.
- substitute latestState for a newer state - either through blockCalldata being posted on-chain prior (ROOT original dispute timestamp (so if this is used you sadly have to provide all chained/recursive dispute commitments so you can verify the root timestamp OR have a separate fraud proof just for this - other stuff just use the previous dispute as the original which doesn't have to be the root original) is compared agains the timestamp of blockcalldata posting on-chain) OR the original disputer is slashed in the recursiveDispute in that case the recusiveDisputer can set his latestState as he wishes without the calldata being posted prior as a requirement, but it still needs to be a newer state - this also covers when the original disputer posted an older latestState and signed a newer one - the check for this is ‘is original disputer in recursiveDispute.outputSnapshotParticipants’ 
- substitute the timeout struct for an olderOne  to check if recursiveDisputeLatestState >= originalDisputeLatestState (blocks commit to the snapshot which commits to the latest state - check recursiveLatestStateBlock.Height >= originalLatestStateBlock.Height


Few more things:
2 more fraud proofs (feel free to add them to the doc):
- DisputeIncludedInOnChainDisputePairs - the disputers that backed those disputes are slashed
- RecursiveDisputesNotInTotalOrder - FORKS in recursive disputes! (Example: Original dispute A. B builds on A. C comes after B and does NOT build on B, but on A.) -> Require total order (you have to chain) - you have knowledge of all N-1 disputes and if you try and create a fork you’re slashed 

Further I've been thinking to explicitly include a filed in the Disputes Struct - 'previousDispute' which is different than 'previousRecursiveDispute'. For now this was implicit since it was always the last dispute, but when we take into consideration onChainDisputePairs it doesn't have to be the lastDispute. Example disputes[A,B,C] - you challenge B and win. OnChainDispute pairs become [[1,2]] so B and C are invalid and now you're creating dispute D, but it doesn't build on C it builds on A, so you specify A as the genesis. In essence it's the last dispute that's not included in onChainDispute pairs - therefore this can also be implicitly deducted, but maybe not bad to include it explicitly so when we're doing computations on-chain we don't have to calculate the 'target' based on onChainDisputePairs, but assume the disputer did it correctly, otherwise if they want to build on C and target C, they get slashed since C is included in onChainDisputePairs. In theory there could be a potential race condition where the user is not aware of the updated onChainDisputePairs, but in reality it wont happen since for them to create dispute D, they had to audit all previous disputes A, B, C. So they would be aware of B being faulty so they would either check that someone challenged it on-chain or they would explicitly do it. Also they would be aware of C, since we're already checking the i-th dispute index at creation so that means they know all N-1 disputes created so no need for any additional race condition check.

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
```diff
@@ -98,25 +98,29 @@ contract DisputeManagerFacet is StateChannelCommon {
...
-        //     onChainDisputePairs.push(disputePair);
-        //     addOnChainSlashedParticipants(collectedSlashParticipants);
-        //     address[] memory returnedSlashParticipants = getOnChainSlashedParticipants();
-        //     emit DisputeChallengeResultWithDisputePair(dispute.channelId, disputePair, isAllAuditValid, returnedSlashParticipants);
-        // }
-        
+        uint256 gasLimit = getGasLimit();
+        bytes memory data = abi.encodeCall(
+            DisputeManagerFacet.auditDispute,
+            (dispute, disputeAuditingData, block.timestamp)
```
lukastanisic99, line 104: this is not the block.timestamp it's the block.timestamp when the dispute was created/committed. So you have to upload it so it will be checked against the commitment.

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
```diff
@@ -98,25 +98,29 @@ contract DisputeManagerFacet is StateChannelCommon {
...
+            emit DisputeChallengeResult(dispute.channelId, success, returnedSlashParticipants);
+        }else{
+            // slash the challenger
+            address[] memory slashParticipants = abi.decode(returnData,(address[]));
+            addOnChainSlashedParticipants(dispute.channelId, slashParticipants);
+            uint disputeLength = getDisputeLength(dispute.channelId);
+            DisputePair memory disputePair = DisputePair(dispute.disputeIndex, disputeLength-1);
+            disputeData[dispute.channelId].disputePairs.push(disputePair);
+            address[] memory returnedSlashParticipants = getOnChainSlashedParticipants(dispute.channelId);
+            emit DisputeChallengeResultWithDisputePair(dispute.channelId, disputePair, success, returnedSlashParticipants);
```
lukastanisic99, line 122: Overall this function is good - it's just missing a new dispute commitment when the challenge is successful (the auditor can instantly submit a new correct dispute that includes the expected onChainSlashes since it only happens on successful on-chain audit - no special check needed just reuse createDispute and add an additional argument newDispute). 

Maybe also have just one event DisputeChallengeResult(id,success,disputePair, slashes) - so depending on 'success' the client will use or ignore disputePair

.wake/extension/local-local-chain-6d298d71-04ff-4f51-b48a-05a2b4bac802.json
```diff
@@ -0,0 +1,69 @@
...
+            "uri": "ws://127.0.0.1:60466"
+        }
+    },
+    "stateFingerprint": "3a82e3a6d1ac2e4daae286769e6107faa655d321408b0b1f9df23ede1a63b531",
+    "persistence": {
+        "isDirty": false,
+        "isAutosaveEnabled": true,
+        "lastSaveTimestamp": 1745503664374
+    }
+}
```
lukastanisic99, line 69: If this is some foundry stuff please write a readme on how to use it and what benefits does it provide. In general a readme for foundry since this PR introduces it.

contracts/V1/StateChannelDiamondProxy/FraudProofFacet.sol
```diff
@@ -47,448 +40,274 @@ contract FraudProofFacet is StateChannelCommon {
...
-    //         slashParticipants[0] = dispute.disputer;
-    //         return (false, slashParticipants, abi.encode("BLOCK DOUBLE SIGN: CHANNEL ID MISMATCH"));
-    //     }
-
-    //     if(block1.stateSnapshotHash != block2.stateSnapshotHash && block1.previousBlockHash != block2.previousBlockHash) {
-    //         slashParticipants[0] = dispute.disputer;
-    //         return (false, slashParticipants, abi.encode("BLOCK DOUBLE SIGN: STATE HASH MISMATCH"));
-    //     }
+        if(fraudBlock.transaction.header.transactionCnt == 0){
+            require(fraudBlock.previousBlockHash == keccak256(abi.encode(fraudBlock)));
```
lukastanisic99, line 120: for blockHeight 0, the previousBlockHash is the hash of the genesisSnapshot -> hash(previousStateSnapshot)

.wake/extension/local-local-chain-6d298d71-04ff-4f51-b48a-05a2b4bac802.json
MrishoLukamba: this shouldnt be commited and its not related to foundry

lukastanisic99: then add it to .gitignore and remove it so it's not tracked

contracts/V1/StateChannelDiamondProxy/AStateChannelManagerProxy.sol
```diff
@@ -244,35 +244,28 @@ abstract contract AStateChannelManagerProxy is
...
-        // If the call was successful, decode the result
-        if (success) {
-            address[] memory slashedParticipants = abi.decode(returnData, (address[]));
-            //for sure no duplicates, otherwise auditing would fail -> just insert
-            addOnChainSlashedParticipants(dispute.channelId, slashedParticipants);
-        }
-        // if !success and returnData.length == 0 => Auditing ran out of gas
-        return (success, returnData);
+        address[] memory slashedParticipants = abi.decode(returnData, (address[]));
+        return  slashedParticipants;
```
lukastanisic99, line 263: How do you know the low level call will succeed?

lukastanisic99 (approved):
Overall is good - didn't go over it too deep this time, but if things are missing we'll correct them as we go

PR #33, opened 2025-05-06
State-snapshots

daiagi:


contracts/V1/StateChannelDiamondProxy/StateSnapshotFacet.sol
```diff
@@ -0,0 +1,217 @@
...
+
+        bool hasLatestDisputeProof = updateStateSnapshotStruct
+            .latestDisputeProof
+            .hasValue;
+        bool hasMilestoneProof = updateStateSnapshotStruct
+            .milestoneProofs
+            .hasValue;
+
+        // No teleport, no progress - useless case, just validate snapshot
+        if (!hasLatestDisputeProof && !hasMilestoneProof) {
```
lukastanisic99, line 30: hasLatestDisputeProof is not needed nor should it be 'oracalized' (uploaded by the off-chain client) - the condition for requiring a 'latest dispute proof' is fully checkable on-chain:
if(on-chain.snapshot.fork != on-chain.disputes.lenght) -> fork teleportation required otherwise the tx reverts

We use this to determine the genesisStateSnapshot - if a dispute is required, upload the dispute (and timestamp) and the snapshot that that dispute commits to. Verify the dispute commitment and is it final/expired - otherwise revert. If it is, this provides the 'genesis snapshot' otherwise the on-chain snapshot is the 'genesis snapshot'.

Now that we have the genesis snapshot - time to hop to the latest finalized snapshot in the fork (if any). For this you can fully reuse this function https://github.com/peer3to/state-channels-plus/blob/08d287bbeedb2ec943bd028829b48c399e15f111/contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol#L289C14-L289C30
This function is originally written with the full dispute context, but the full dispute context is not needed and it can be refactored. If you take a look what the function does, it takes a genesis snapshot, a list of milestone proofs and a list of snaspshots that each of the milestones commit to, to prove the last milestone in the list, starting from the genesis snapshot. That's all you need here. You can even fully reuse the stateProof struct just ignore the signedBlocks and just use the milestones. 

The last milestone (snapshot) you prove is the snapshot that overrides that snapshot on-chain. 

You don't need hasMilestoneProof either, since if milestones.length is 0 the milestone part is ignored and the snapshot you use to override the on-chain snapshot is the genesisSnapshot.

lukastanisic99: in general no need to write this function as a 4 case condition (A,B),(!A,B),(A,!B),(!A,!B)

It's basically:

if(isDispute required){
check dispute and dispute output snapshot uploaded and verified correctly, otherwise revert.
snasphot = dispute output snapshot 
}else{
snapshot = on-chain snapshot
}
reuse the function linked above with the correct genesisSnapshot (the snapshot we set above), the milestones and snapshots each milestone commits to - all of this has to be uploaded. Feel free to reuse stateProof. Mostly just understand the linked function. 

The result of all this computation is you've proven the last snapshot. Now you can override the last snapshot, but before you do, process withdrawals by applying exitChannelBlocks between the old snapshot and the new. You pass each exit channel to the AStateChannelManagerProxy that the concrete proxy will override and implement the method that knows how to process the withdrawal given an exitChannel (which assets and amounts to send and to whom) -  That's all.

contracts/V1/StateChannelDiamondProxy/StateSnapshotFacet.sol
```diff
@@ -0,0 +1,217 @@
...
+        require(
+            newStateSnapshot.forkCnt == getDisputeLength(channelId) &&
+                stateSnapshots[channelId].forkCnt ==
+                getDisputeLength(channelId),
+            "StateSnapshot is not valid"
+        );
+
+        // for each exit channel block, there must be a milestone proof
+        require(
+            milestoneProofs.length == exitChannelBlocks.length,
```
lukastanisic99, line 104: This check is not needed since the function I linked above will require the correct milestone and snapshot length and in general this relationship doesn't hold for cross fork exits since you can have many exits that happened in a previous fork and are 'carried over' to this fork so new exits from this fork build on top of them but they don't produce new milestones in this fork and we don't verify past milestones - we assume the genesis is correct when the dispute is proven final and latest so we just 'teleport'. 

The relationship 'an exit channel' is linked to a 'milestone' is useful off-chain as a checkpoint where to produce a milestone when the STF returns a non-empty exitChannel[ ]. Here it's not relevant since it's cross fork and we already have proven the latest snapshot that will override the on-chain one, so we just process all exit channels blocks between them and make sure they're all cryptographically linked. [oldSnapshot.latestExitChannelBlockHahs <-[exitChannelBlocksUploaded]<-NewSnapshot]

contracts/V1/StateChannelDiamondProxy/StateSnapshotFacet.sol
```diff
@@ -0,0 +1,217 @@
...
+        require(
+            newStateSnapshot.latestExitChannelBlockHash ==
+                keccak256(
+                    abi.encode(exitChannelBlocks[exitChannelBlocks.length - 1])
+                ),
+            "Last Exit block not linked to new state snapshot"
+        );
+
+        // Get current participants from state snapshot
+        address[] memory currentParticipants = getSnapshotParticipants(
```
lukastanisic99, line 123: Re-use the function linked above - this is duplicating the logic - participants shouldn't be used here at all, not even when processing exits - even the exits are not interpreted here

contracts/V1/StateChannelDiamondProxy/StateSnapshotFacet.sol
```diff
@@ -0,0 +1,217 @@
...
+            );
+            require(isFinal, "Milestone not finalized");
+
+            // Process each exit in the block and update participant set
+            for (uint j = 0; j < exitBlock.exitChannels.length; j++) {
+                ExitChannel memory exit = exitBlock.exitChannels[j];
+
+                // If not a partial exit, remove participant from the set
+                if (!exit.isPartialExit) {
+                    currentParticipants = removeParticipant(
```
lukastanisic99, line 158: Removal doesn't happen here - here we just process the withdrawals on-chain without even interpreting the exit channels - we just pass them to AStateMachineManagerProxy. Even the concrete proxy doesn't do any removal, it just processes the withdrawal to the beneficiary listed in the ExitChannel, by interpreting the ExitChannel.

isPartilExit - I think I've made a mistake including it explicitly, since a normal transaction signals an exit from the state machine and the state machine logic will decide does it want to remove the participant or keep them (a partial exit - is their balance > 0). We can keep this for now and not use it, just as signaling what's going on. In general a partial exit will look like a normal exit, but the participant will remain in the channel. When you call stateMachine.getParticpants() the participant will be in the list and this is what is called when setting the participant list for the snapshot - only the opinion of the stateMachine matters and the stateMachine does the internal accounting. In general the ExitChannel can contain even non participants as 'beneficiaries' - that's how we'll collect fees - the treasury contract address will be the beneficiary for the fees accumulated in the state machine.

contracts/V1/DataTypes.sol
```diff
@@ -121,4 +123,27 @@ struct StateSnapshot {
...
+
+struct OptionalMilestoneProofs {
+    bool hasValue;
+    ForkMilestoneProof[] value;
+}
+
+struct OptionalDisputeProof {
+    bool hasValue;
+    LatestDisputeProof value;
+}
```
lukastanisic99, line 141: these optional data structures are not needed as explained in my other comments

daiagi:
Hey @lukastanisic99 
all of you comments have been implemented

a deviation from the logic you have described is that i spitted the `updateStateSnapshot` solidity function into two:
- with dispute
- without

this is because there are no optional args in solidity

contracts/V1/StateChannelDiamondProxy/StateSnapshotFacet.sol
```diff
@@ -0,0 +1,230 @@
...
+        );
+        require(isStateValid, "Invalid State Proof");
+
+        StateSnapshot memory lastProovenSnapshot = milestoneSnapshots[
+            milestoneSnapshots.length - 1
+        ];
+
+        _validateExitChannelBlocks(
+            exitChannelBlocks,
+            genesisStateSnapshot,
```
lukastanisic99, line 111: for exitChannels 'genesisStateSnapshot' should always be onChainStateSnapshot since we're processing withdrawals between the last onChainSnapshot and the current one that we're updating

contracts/V1/StateChannelDiamondProxy/StateSnapshotFacet.sol
```diff
@@ -0,0 +1,230 @@
...
+
+    function updateStateSnapshotWithoutDispute(
+        bytes32 channelId,
+        ForkMilestoneProof[] memory milestoneProofs,
+        StateSnapshot[] memory milestoneSnapshots,
+        ExitChannelBlock[] memory exitChannelBlocks
+    ) external onlySelf {
+        require(
+            getDisputeLength(channelId) == getSnapshotForkCnt(channelId),
+            "Dispute proof is required"
```
lukastanisic99, line 42: Custom error would be preferred to stay consistent with the rest of the codebase

contracts/V1/StateChannelDiamondProxy/StateSnapshotFacet.sol
```diff
@@ -0,0 +1,230 @@
...
+        bytes32 channelId,
+        DisputeProof memory disputeProof
+    ) internal view returns (StateSnapshot memory) {
+        if (getDisputeLength(channelId) != getSnapshotForkCnt(channelId)) {
+            // dispute proof is required
+            Dispute memory dispute = disputeProof.dispute;
+            uint disputeTimestamp = disputeProof.timestamp;
+
+            require(
+                _isDisputeProofProvided(disputeTimestamp),
```
lukastanisic99, line 65: why is this check needed? If dispute proof is not provided the _isDisputeCommitmentValid bellow would anyway return false

contracts/V1/StateChannelDiamondProxy/StateSnapshotFacet.sol
```diff
@@ -0,0 +1,230 @@
...
+                "Dispute proof is required"
+            );
+            require(
+                _isDisputeCommitmentValid(dispute, disputeTimestamp, channelId),
+                "Dispute proof is not valid"
+            );
+            require(
+                _isDisputeFinalized(disputeTimestamp),
```
lukastanisic99, line 73: This only accounts for isExpired, which is fine for now but we should also add the option for threshold early finalization

contracts/V1/StateChannelDiamondProxy/StateSnapshotFacet.sol
```diff
@@ -0,0 +1,230 @@
...
+                _isStateSnapshotValid(
+                    disputeProof.outputStateSnapshot,
+                    dispute
+                ),
+                "State snapshot is not valid"
+            );
+
+            return disputeProof.outputStateSnapshot;
+        } else {
+            return stateSnapshots[channelId];
```
lukastanisic99, line 86: you can flip the condition in the 'if' and have the single return stateSnapshots[channelId] and instead of the 'else' you just have everything else normally - either way it's the same and not a big difference

contracts/V1/StateChannelDiamondProxy/StateSnapshotFacet.sol
```diff
@@ -0,0 +1,230 @@
...
+                milestoneProofs,
+                milestoneSnapshots,
+                genesisSnapshot
+            );
+        return isValid;
+    }
+
+    function _validateExitChannelBlocks(
+        ExitChannelBlock[] memory exitChannelBlocks,
+        StateSnapshot memory genesisStateSnapshot,
```
lukastanisic99, line 137: rename 'genesisStateSnapshot' to old/currentOnChainStateSnapshot

contracts/V1/examples/MathStateMachine/MathStateChannelManagerProxy.sol
```diff
@@ -13,8 +13,9 @@ contract MathStateChannelManagerProxy is AStateChannelManagerProxy {
     constructor(
         address aStateMaachineAddress,
```
lukastanisic99, line 14: A typo probably from me - 'aStateMaachineAddress'

lukastanisic99:
Finished going over stuff this morning - overall well done! The comments I've made above are mostly style related and nothing serious. The logic seems good to me

contracts/V1/StateChannelDiamondProxy/StateSnapshotFacet.sol
daiagi: yes.
#34

contracts/V1/StateChannelDiamondProxy/StateSnapshotFacet.sol
daiagi: i also prefer early returns over if/else

daiagi:
@lukastanisic99  all comments have been taken care of

contracts/V1/StateChannelManagerEvents.sol
```diff
@@ -19,21 +19,34 @@ interface StateChannelManagerEvents {
     event DisputeUpdated(bytes32 indexed channelId, Dispute dispute);
 
     event DisputeCommited(
-    bytes encodedDispute,
-    uint timestamp
+        bytes encodedDispute,
+        uint timestamp,
+        bytes32 disputeCommitment
```
lukastanisic99, line 24: why is disputeCommitment also emitted? The client can just hash the encodedDispute and timestamp

daiagi: convenience. it is redundant

lukastanisic99:
I'm thinking of some things now - seems it will be more convenient to use a hash for the onChainSnapshot and 'oracleize' it too. Not only to have constant storage costs, but we need to include a timestamp at channel open - we can include it in the snapshot for channel open, but we can not include it when generating the dispute since you don't know the on-chain timestamp to commit to an outputSnapshot so best not in include it within the structure, but like disputes use the commitment in the form hash(snapshot,timestamp) so you can still commit to the snapshot and on-chain you'll commit to the timestamp too. 

Apart from this onChainSlashes needs to be cleared when updating the snapshot successfully. Need to figure out 1 more thing tonight and I'll put it all together and we'll discuss tomorrow in the call.

lukastanisic99 (approved):

Issue #37, opened 2025-05-13
reset pending participants


daiagi:

daiagi:
Closed by #46

PR #38, opened 2025-05-14
Dispute-threshold-finalization

### LINKED ISSUE #35: support dispute is finalized via threshold signatures
[see comment](https://github.com/peer3to/state-channels-plus/pull/33#discussion_r2083627054)

daiagi:
- Closes #35

@lukastanisic99  PR is almost complete, but if you look at the call to `updateStateSnapshotWithDispute` (in StateManger) you will see it will not compile, please teach me how to do optional args in sol functions properly

after that i will align the sol side of updateStateSnapshotWithDispute as well

### DIFF
--- a/contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
+++ b/contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
@@ -8,8 +8,7 @@ import "./Errors.sol";
 contract DisputeManagerFacet is StateChannelCommon {
 
     function createDispute(
-        Dispute memory dispute,
-        bytes calldata disputerSignature
+        Dispute memory dispute
     ) public { 
         require(msg.sender == dispute.disputer, ErrorDisputerNotMsgSender());
         require(_canParticipateInDisputes(dispute.channelId, msg.sender), ErrorCantParticipateInDispute());
@@ -24,7 +23,7 @@ contract DisputeManagerFacet is StateChannelCommon {
             block.timestamp
         ));
         disputeData[dispute.channelId].disputeCommitments.push(disputeCommitment);
-        emit DisputeCommited(encodedDispute,block.timestamp, disputerSignature);
+        emit DisputeCommited(encodedDispute, block.timestamp);
     }
     
     
--- a/contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
+++ b/contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
@@ -8,7 +8,8 @@ import "./Errors.sol";
 contract DisputeManagerFacet is StateChannelCommon {
 
     function createDispute(
-        Dispute memory dispute
+        Dispute memory dispute,
+        bytes calldata disputerSignature
     ) public { 
         require(msg.sender == dispute.disputer, ErrorDisputerNotMsgSender());
         require(_canParticipateInDisputes(dispute.channelId, msg.sender), ErrorCantParticipateInDispute());
@@ -23,7 +24,7 @@ contract DisputeManagerFacet is StateChannelCommon {
             block.timestamp
         ));
         disputeData[dispute.channelId].disputeCommitments.push(disputeCommitment);
-        emit DisputeCommited(encodedDispute,block.timestamp, disputeCommitment);
+        emit DisputeCommited(encodedDispute,block.timestamp, disputerSignature);
     }
     
     
--- a/contracts/V1/StateChannelDiamondProxy/StateSnapshotFacet.sol
+++ b/contracts/V1/StateChannelDiamondProxy/StateSnapshotFacet.sol
@@ -4,6 +4,7 @@ import "./StateChannelCommon.sol";
 import "../DataTypes.sol";
 import "./AStateChannelManagerProxy.sol";
 import "./Errors.sol";
+import "./StateChannelUtilLibrary.sol";
 
 contract StateSnapshotFacet is StateChannelCommon {
     function updateStateSnapshotWithDispute(
@@ -65,8 +66,10 @@ contract StateSnapshotFacet is StateChannelCommon {
             _isDisputeCommitmentValid(dispute, disputeTimestamp, channelId),
             ErrorDisputeProofNotValid()
         );
+        // TODO: add addressesInThreshold
+        // addressesInThreshold are? snapshot.participents - onchainSlasshes?
         require(
-            _isDisputeFinalized(disputeTimestamp),
+            _isDisputeFinalized(disputeProof, ),
             ErrorDisputeNotFinalized()
         );
         require(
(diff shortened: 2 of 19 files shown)


src/stateManager/StateManager.ts
```diff
@@ -438,37 +441,90 @@ class StateManager {
...
-            this.channelId,
-            milestoneProofs,
-            milestoneSnapshots,
-            disputeProof,
-            exitChannelBlocks
+        // Check if dispute is within agreement time
+        const currentTime = Clock.getTimeInSeconds();
+        const timeSinceDispute = currentTime - disputeData.timestamp;
+
+        if (timeSinceDispute > this.timeConfig.agreementTime) {
```
lukastanisic99, line 477: Dispute is finalized after 'disputeChallengeTime'

src/stateManager/StateManager.ts
```diff
@@ -438,37 +441,90 @@ class StateManager {
...
+            );
+        }
+
+        // Check if we have threshold signatures on the dispute
+        const fork = this.agreementManager.forks.latestFork();
+        if (!fork) {
+            throw new Error("No latest fork found");
+        }
+        const requiredAddressesSet = SetUtils.stringSetFromArray(
+            fork.addressesInThreshold
```
lukastanisic99, line 494: this will change - for this in particular the set is 'onChainSnapshot - onChainSlashes'
There won't be addressesInThreshold for a fork, since the threshold can contract due to off-chain exits. Since we 'persist' snapshots, you can always fetch the one that's relevant to you to get the threshold

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
```diff
@@ -8,7 +8,8 @@ import "./Errors.sol";
 contract DisputeManagerFacet is StateChannelCommon {
 
     function createDispute(
-        Dispute memory dispute
+        Dispute memory dispute,
+        bytes calldata disputerSignature
```
lukastanisic99, line 12: This signature is not needed since dispute creation checks msg.sender == dispute.disputer so the creator of the dispute is implicitly counted towards the threshold, so N-1 signatures are enough.  If it's convenient to include it, it's fine since it's not expensive

daiagi: done

src/stateManager/StateManager.ts
daiagi: so the required address are
the participants in the state-snapshot 1 before this dispute minus the on-chain slashes

but is the correct snapshot necessarily  the one that is on-chain?

it could be that there were a few disputes already, but now when submitting the snapshot i am teleporting over them to the latest one

in this case i am not sure the the on-chain state-snapshot is the correct one to get participants from

and if it isn't , then where from?


also, doesn't this mean that the "oraclized" dispute proof will also need to include the genesisStateSnapshot

which we will compare to the `dispute.genesisStateSnapshotHash` and from which we will take the particiapnts for early finalization

contracts/V1/StateChannelDiamondProxy/StateSnapshotFacet.sol
```diff
@@ -65,8 +66,10 @@ contract StateSnapshotFacet is StateChannelCommon {
             _isDisputeCommitmentValid(dispute, disputeTimestamp, channelId),
             ErrorDisputeProofNotValid()
         );
+        // TODO: add addressesInThreshold
+        // addressesInThreshold are? snapshot.participents - onchainSlasshes?
```
daiagi, line 70: @lukastanisic99 
notice code is incomplete here, since i am not sure where to get the participants for early finalization from
- the current on-chain snapshot?
- genesis state snapshot of the dispute? (needs to be oracelized as well)

daiagi:
I filtered out the dispute signature from the set when assembling the state snapshot proof. Which is I now realize is wrong.
Instead the signature should be checked when coming in from gossip, and rejected if it is the disputers signature. This could also be used as double sign fraud proof.
Will fix later today

src/stateManager/StateManager.ts
lukastanisic99: Forgot to take into account pendingParticipants from joins. That should also be taken into account as onChainSlashes. Probably also clear that as onChainSlashes when the snapshot is updated successfully. 

So the threshold set is = onChainSnapshot.participants + pendingParticipants - onChainSlashes 

This threshold includes the largest possible set we should take into account, since off-chain the threshold set can't expand only contract. Yes, it's possible that someone exited off-chain and for some reason didn't submit the snapshot to process their withdrawal on-chain which would also update the participant list to account for that, but that's unlikely (its in their best interest to process the withdrawal immediately and update the snapshot). Otherwise they should wait. We can maybe design a 'milestone' system here too, but no need for it now.

contracts/V1/StateChannelDiamondProxy/StateSnapshotFacet.sol
lukastanisic99: Hopefully my previous comment above also answers this

lukastanisic99:
> I filtered out the dispute signature from the set when assembling the state snapshot proof. Which I now realize is wrong. Instead the signature should be checked when coming in from gossip, and rejected if it is the disputers signature. This could also be used as double sign fraud proof. Will fix later today

Yes on gossip/event do all the checks - You can include the disputers signature as you did with createDispute and have a N/N threshold or exclude it entirely and use N-1 threshold since on-chain msg.sender == dispute.disputer is checked
Whatever is more convenient to you, doesn't make much of a difference.

daiagi:
@lukastanisic99  using state-snapshot - slashes set 
[4f156586](https://github.com/peer3to/state-channels-plus/pull/38/commits/4f15658608146e62472be41947fe9f7d22b86327)

I've read you comment:
> So the threshold set is = onChainSnapshot.participants + pendingParticipants - onChainSlashes


but i didn't code it, because i don't understand it

pending participants are not yet in, so how does it make sense to expect their signatures on a dispute?

lukastanisic99:
> @lukastanisic99 using state-snapshot - slashes set [4f156586](https://github.com/peer3to/state-channels-plus/pull/38/commits/4f15658608146e62472be41947fe9f7d22b86327)
> 
> I've read you comment:
> 
> > So the threshold set is = onChainSnapshot.participants + pendingParticipants - onChainSlashes
> 
> but i didn't code it, because i don't understand it
> 
> pending participants are not yet in, so how does it make sense to expect their signatures on a dispute?

So they are 'in'. The snapshot just doesn't reflect that yet, since the join is async as explained before.
Their deposits are processed and the latestJoinChannelBlock is appended and its hash is updated in  https://github.com/peer3to/state-channels-plus/blob/32ee9477f0f6167376da72f6065f30b12a1182fe/contracts/V1/DisputeTypes.sol#L200

The joining participants are inserted in pendingParticipants (which is part of the same data structure, available on-chain). This forces the dispute game to account for their entrance but still allows the dispute game to play out async to come to an objective truth, but they will be accounted for in the canonical reality that prevails out of the dispute game, so you do take their signature into account since they are actually part of any valid dispute output state and by being in the pendingParticipants array they can participate in disputes (they may not have the data to participate, but we solve that by requiring the peers to feed them the data so they can double check (audit) and sign, so if their signature is provided, auditing of disputes can be skipped on chain, otherwise it has to run on-chain to account for their safety).

daiagi:
@lukastanisic99  added pending
any leftovers regarding joins and pending i will handle in #37

lukastanisic99 (approved):

PR #39, opened 2025-05-14
clear onChainSlashedParticipants

### LINKED ISSUE #36: clear on chain slashes when snapshot is updated on chain

daiagi:
- Closes #36

### DIFF
--- a/contracts/V1/StateChannelDiamondProxy/StateSnapshotFacet.sol
+++ b/contracts/V1/StateChannelDiamondProxy/StateSnapshotFacet.sol
@@ -109,7 +109,7 @@ contract StateSnapshotFacet is StateChannelCommon {
         stateSnapshots[channelId] = lastProovenSnapshot;
 
         // clear onChainSlashedParticipants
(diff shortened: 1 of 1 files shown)


contracts/V1/StateChannelDiamondProxy/StateSnapshotFacet.sol
```diff
@@ -105,7 +105,12 @@ contract StateSnapshotFacet is StateChannelCommon {
 
         _applyExitChannelBlocks(channelId, exitChannelBlocks);
 
+        // Update the state snapshot
         stateSnapshots[channelId] = lastProovenSnapshot;
+
+        // clear onChainSlashedParticipants
+        disputeData[channelId].onChainSlashedParticipants = new address[](0);
```
lukastanisic99, line 112: Think it's better to use 'delete array' - gpt said it also clears the storage slots so you get a gas refund at the end of the tx

daiagi: thanks
done

lukastanisic99 (approved):

PR #40, opened 2025-05-15
Collecting frauds & Dispute handling TS

MrishoLukamba:
This PR introduces collecting of frayd proofs in Block pipeline and dispute handling in TS


lukastanisic99:
@MrishoLukamba why are there like 50 commits here from before - I've approved your previous PR, so please merge everything so this PR only shows the differences what's introduced here so I don't have to look at duplicate stuff

PR #51, opened 2025-05-24
state transistion returns exit channels

daiagi:
state transition returns exitChannels

@Krayt78  welcome to review as well


contracts/V1/DataTypes.sol
```diff
@@ -111,6 +113,19 @@ struct LeaveChannel {
     bytes data; //custom data
 }
 
+
+struct ExitChannel {
+    address participant;
+    bool isPartialExit;
```
Krayt78, line 119: why do we need this one? @lukastanisic99 
We already have the amount so we should already know its its partial or not

lukastanisic99: I commented on this in some other PR also - We don't need it as you noticed, since the state machine does all the accounting internally and 'knows' is it a partial exit or a full exit (should it keep the address in the participant set or not) - Originally we were keeping track about this outside of the state machine, so it would have been useful in that case, but since we can just call 'getParticipants' on the state machine, we don't beed this information

contracts/V1/AStateMachine.sol
```diff
@@ -60,33 +81,43 @@ abstract contract AStateMachine {
...
             if (result.length == 0)
                 revert("AStateMachine - Call failed - result lenght 0");
             assembly {
                 let returndata_size := mload(result)
                 revert(add(32, result), returndata_size)
             }
         }
-        return success;
+
+        return (callSuccess, _exitChannels);
```
lukastanisic99, line 120: We can keep this function as just 'return bool' since both the 'state' and 'exitChannels' can be fetched 'getState' 'getExitChannels' if needed by the calling function/context if needed

lukastanisic99:
Overall looks fine, just change the branch target from 'master' to 'disputes' - Hopefully the file diff with also be updated to account for all the stuff that's already in disputes

contracts/V1/AStateMachine.sol
daiagi: i think this is better API
you make state transition, you get the associated exit channels. clear and simple
no need to "remember" to get them before the next transition

daiagi:
>Overall looks fine, just change the branch target from 'master' to 'disputes' - Hopefully the file diff with also be updated to account for all the stuff that's already in disputes

done

contracts/V1/AStateMachine.sol
```diff
@@ -91,21 +116,27 @@ abstract contract AStateMachine {
...
-    ) external _nonReentrant returns (bool) {
+    )
+        external
+        _nonReentrant
+        returns (bool success, ExitChannel[] memory exitChannels)
+    {
         _tx = transaction;
-        (bool success, bytes memory result) = address(this).call{gas: gasLimit}(
+        _clearExitChannels();
+        (bool callSuccess, bytes memory result) = address(this).call(
```
lukastanisic99, line 126: Why is the gasLimit removed

PR #54, opened 2025-05-24
base storage module

daiagi:
@Krayt78 
starting point for you

just an agnostic key-value storage with optional persist to local storage
no business logic (yet)

### DIFF
--- a/src/storage/AgnosticStorage.ts
+++ b/src/storage/AgnosticStorage.ts
@@ -1,0 +1,170 @@
+import localforage from "localforage";
+
+/**
+ * AgnosticStorage provides a simple key-value storage interface.
+ * Completely wraps localStorage with memory caching for optimal performance.
+ */
+export class AgnosticStorage<T = any> {
+    private memoryCache: Map<string, T> = new Map();
+    private storage?: LocalForage;
+    private isPersistent: boolean;
+    private storeName: string;
+
+    /**
+     * Create a storage instance
+     * @param options.persist - Whether to enable persistence (default: false)
+     * @param options.storeName - Store name for persistence (default: 'default-store')
+     */
+    constructor(
+        options: {
+            persist?: boolean;
+            storeName?: string;
+        } = {}
+    ) {
+        this.isPersistent = options.persist ?? false;
+        this.storeName = options.storeName ?? "default-store";
+
+        if (this.isPersistent) {
+            this.storage = localforage.createInstance({
+                name: "state-channels-plus",
+                storeName: this.storeName,
+                driver: localforage.LOCALSTORAGE
+            });
+        }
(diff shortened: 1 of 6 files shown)


src/storage/AgnosticStorage.ts
```diff
@@ -0,0 +1,170 @@
+import localforage from "localforage";
+
+/**
+ * AgnosticStorage provides a simple key-value storage interface.
+ * Completely wraps localStorage with memory caching for optimal performance.
+ */
+export class AgnosticStorage<T = any> {
+    private memoryCache: Map<string, T> = new Map();
```
lukastanisic99, line 8: 'memoryMap' - in general let's have all maps be named '....map'

src/storage/AgnosticStorage.ts
```diff
@@ -0,0 +1,170 @@
...
+                        key,
+                        error
+                    );
+                });
+            }
+        }
+    }
+
+    async get(key: string): Promise<T | null> {
+        return this.memoryCache.get(key) ?? null;
```
lukastanisic99, line 80: No need to have this function async, when the function body is fully sync and doesn't use async code

src/storage/AgnosticStorage.ts
```diff
@@ -0,0 +1,170 @@
...
+
+        if (this.isPersistent && this.storage) {
+            if (options.wait) {
+                await this.storage.removeItem(key);
+            } else {
+                this.storage.removeItem(key).catch((error) => {
+                    console.error(
+                        "Background remove failed for key:",
+                        key,
+                        error
```
lukastanisic99, line 105: Same as above - why 'wait'

lukastanisic99:
Just some things to take into account - This PR essentially doubles our memory footprint since it stores data both in memory and in 'storage'. Local storage has a limit of 5MB per origin (site) - not much, so in 'storage' we should persist only the minimal set of CRITICAL information needed for safety to reconstruct the latest information (e.g. the latest thing we can finalize + all the blocks/disputes that are inputs to STF + signatures needed to prove all of that. Every thing else can be derived from that). In memory 'cache' we can persist the full data (full states, snapshots etc.. + non critical data) and occasionally prune it. 

In general since we're creating these abstractions, it would be better to separate these two and create a Memory/RAM like abstraction and storage abstraction + a service that coordinates the 2 (decides what stays in memory and what is persisted and how to reconstruct memory from storage when needed)

Krayt78:
for now i will just not use the agnostic storage, and later on once the rest works we can think about the persistant data

daiagi:
OK I'm closing it
@Krayt78 just use regular memory.
We will revisit the topic of persistence later on. Low priority

PR #56, opened 2025-05-25
Basic storage

Krayt78:
## 📋 Description

Implementing basic Storage to store offchain stateSnapshots, join/exit lock hashes and deposits/withdrawals


src/storage/StateSnapshotStorage.ts
Krayt78, line 1: @daiagi you can maybe start looking at it while i gather the data from stateManager

lukastanisic99:
@Krayt78 can you reopen this PR with a local branch, so we don't have to keep track of multiple origins - in general feel free to create your branches in the repo directly (you have access) and PR from them. 

Remerge with the dispute branch and/or setup foundry so the precommit hooks can format solidity, so we don't look at just formatting changes. 

If you're on windows, probably best to use wsl and work in that environment since foundry has limits in windows. 

If you're using VS Code as your default editor it should have full support within wsl - you just navigate to the repo and do 'code .' and it will open it 

If you switch to wsl you'll have to reinstall dependencies (node_module). You can still mount to your disk and use it from the 'windows' environment, but my empirical results show it's better to use the storage partition that's used by wsl by default since for me at least the performance is more or less native and if it's the windows partition it's 50% reduction

PR #57, opened 2025-05-25
Stabilize-tests

daiagi:
many tests are not passing
these PR partly addresses and fixes some failing test

effort in this direction (testing, coverage) to be continued
- #58


MrishoLukamba:
looks fine on my end as even I will remove some tests so that for each new PR will include its new tests @lukastanisic99

src/utils/Codec.ts
```diff
@@ -32,11 +38,51 @@ export class Codec {
...
+                else if ("transaction" in struct && "stateHash" in struct) {
+                    ethersType = BlockEthersType;
+                }
+                // Try Transaction
+                else if ("header" in struct && "body" in struct) {
+                    ethersType = TransactionEthersType;
+                }
+                // Try Dispute
+                else if ("disputeIndex" in struct && "channelId" in struct) {
+                    ethersType = DisputeEthersType;
```
lukastanisic99: Fallback doesn't make sense if you already predefine what 'can be encoded' and also 'explicitType' doesn't make sense if you define what can be encoded -> remove the fallback (let it throw if you try encoding an invalid type) and split the encode into 2 functions that are overloaded:
encode(struct: StructType) -> which has all the info to deduct the correct encoding
encode(struct: any, explicitType: StructTypeName) -> useful if the object is not an instance of the predefined class -> try and encode it to the explicitType - it will fail if it can't encode

lukastanisic99: @MrishoLukamba everything looks good - just split this and you can merge

src/utils/Codec.ts
```diff
@@ -88,6 +55,17 @@ export class Codec {
         return ethers.AbiCoder.defaultAbiCoder().encode([ethersType], [struct]);
     }
 
+    public static encodeWithExplicitType(
```
lukastanisic99, line 58: should be also named 'encode' - typescript supports method overloading

MrishoLukamba:
so should I merge this PR, or update the existing test on this PR  all together

lukastanisic99:
> so should I merge this PR, or update the existing test on this PR  all together

Just change the name as for my previous comment and you can merge

MrishoLukamba:
> > so should I merge this PR, or update the existing test on this PR  all together
> 
> Just change the name as for my previous comment and you can merge

sure, deleting some old test and will include new ones in the another PR for dispute and agreement manager

src/utils/Codec.ts
lukastanisic99: @MrishoLukamba you didn't change this

MrishoLukamba: @daiagi , So am getting linting errors with function overloading and my opinion is in TS we shouldn't do the overload, whats your opinion?

daiagi: @Mrisho
1.  He told you his opinion.

2. "_In my opinion we should not_" 
is not an argument.
This is too basic to be up to opinion. It is a small and pure issue of can do or can't do. 
If you have a issue doing it then post screenshot and backup your claims with link.

3. If you can't overload you could make a function with optional argument that has delegates to two internal functions based on the existence / lack of the argument. 

4. `EncodeWithExplicitType` is way too long  / verbose name.

Kinda OK for internal function. Very not OK for static / public function

lukastanisic99: @MrishoLukamba pls argument your opinions - just saying 'we shouldn't do overloading' without an argument is worthless. And in general opinions without arguments are worthless - I've mentioned this for foundry and viem also. 

Typescript supports overloading and I don't see why we shouldn't use it here. It gives a clear interface that for some predefined types you don't need to pas additional arguments and for the rest another argument is needed. With an optional argument you don't get typescript helping you. Remove the linter part that doesn't allow overloading in the config. 

In general we shouldn't spend too much time on these things. I always argument my opinions, and if you have counter arguments just write them and if they make sense we'll accept them, otherwise we wont.

PR #60, opened 2025-05-25
return exitChannels

### LINKED ISSUE #49: make state transition function return ExitChannel[]
**AStateMachine.sol**

The `stateTransition` function should return an array of `ExitChannel[]`.

The general idea is as follows:

* The `_joinChannel` and `_removeParticipant` functions will no longer return `ExitChannel` directly. Instead, they will append the exit channels to a **transient storage** field — a new kind of field added to the `StateMachine` contract.

* In addition to the existing `getState` function, we will add a `getExits` function that retrieves all the `ExitChannel` instances from this transient storage field.

* Currently, the `getState` function only returns a `bool success` indicating whether the state transition was successful. It should be updated to also return a flag indicating whether a new exit was created.

* A marker will be placed in the TS storage (`agreementManager`), which will represent the "place" where we intend to save a `stateSnapshot` milestone and its proof.

daiagi:
## 📋 Description

get acces to exit channels returned from state transition

## 💪 Related Issues

- Closes #49

## ✅ Checklist

- [x] My code compiles without errors.
- [x] I have performed a self-review of my code.
- [x] I have written tests that cover the new or changed code.
- [x] I have linked relevant issues and/or created follow-up issues.

### DIFF
--- a/contracts/V1/AStateMachine.sol
+++ b/contracts/V1/AStateMachine.sol
@@ -7,6 +7,7 @@ abstract contract AStateMachine {
     address _stateChannelManager;
     bool _nonreentrant;
     uint gasLimit;
+    ExitChannel[] private _exitChannels;
 
     constructor(uint256 _gasLimit) {
         gasLimit = _gasLimit;
@@ -16,6 +17,13 @@ abstract contract AStateMachine {
     // event TxExecutedA(bool success, bytes encodedState);
 
     // ***** DEBUG *****
+
+    modifier _nonReentrant() {
+        require(!_nonreentrant, "ReentrancyGuard: reentrant call");
+        _nonreentrant = true;
+        _;
+        _nonreentrant = false;
+    }
 
     // Restore the state (variables) of the contract by deserializing/decoding the given the encoded state
     function _setState(bytes memory encodedState) internal virtual;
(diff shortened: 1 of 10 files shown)


src/evm/EvmStateMachine.ts
```diff
@@ -90,12 +90,27 @@ class EvmStateMachine extends AStateMachine {
 
         try {
             const result = await this.contractExecuter.executeCall(encodedData);
+            // Decode the return values: (bool success, ExitChannel[] exitChannels)
+            const hexResult = ethers.hexlify(result.returnValue);
+            const [success, exitChannels] =
+                ethers.AbiCoder.defaultAbiCoder().decode(
+                    [
+                        "bool",
+                        "tuple(address participant, bool isPartialExit, uint256 amount, bytes data)[]"
```
lukastanisic99: Use the ethers type defined in @types/ethers so if it changes it only needs to change in the single location

src/evm/EvmStateMachine.ts
```diff
@@ -122,6 +137,18 @@ class EvmStateMachine extends AStateMachine {
         return addresses.toArray();
     }
 
+    async getExitChannels(): Promise<any[]> {
+        const callData = this.getEncodedCalldata("getExitChannels");
+
+        let result = await this.contractExecuter.executeCall(callData);
+        const hexResult = ethers.hexlify(result.returnValue);
+        const [exitChannels] = ethers.AbiCoder.defaultAbiCoder().decode(
+            ["tuple(address participant, uint256 amount, bytes data)[]"],
```
lukastanisic99: same

contracts/V1/AStateMachine.sol
```diff
@@ -17,6 +18,13 @@ abstract contract AStateMachine {
 
     // ***** DEBUG *****
 
+    modifier _nonReentrant() {
+        require(!_nonreentrant, "ReentrancyGuard: reentrant call");
+        _nonreentrant = true;
+        _;
+        _nonreentrant = false;
+    }
+
```
lukastanisic99: Move to bottom

lukastanisic99:
Overall good - just the little changes

PR #69, opened 2025-05-31
config: load once, use anywhere

### LINKED ISSUE #14: create dev / prod envs or json files
Doesn't matter for development you change it to what you need it - it was part of an .env file but removed it since had problems in browsers - however this should be at the top of the file (like in other places) so we don't have to dig to find it 

In production should default to false - maybe instead of a .env file we do a JSON (that works in both nodejs environments and browsers) and for production a build script changes all values to false before building it

_Originally posted by @lukastanisic99 in https://github.com/peer3to/state-channels-plus/pull/13#discussion_r2029970433_

daiagi:
## 📋 Description

- config is being loaded once from file
-  hard-coded default values
- easy to use :

```ts
import { PROVIDER_URL } from "@/utils/config";
```

## 💪 Related Issues

- Closes #14

## ✅ Checklist

- [x] My code compiles without errors.
- [x] I have performed a self-review of my code.
- [x] I have written tests that cover the new or changed code.
- [x] I have added necessary documentation (if appropriate).


lukastanisic99:
@daiagi we had .env before and it doesn't work in browsers and that's why it's changed to .json so it works in all environments

daiagi:
@lukastanisic99  back to json

lukastanisic99:
> @lukastanisic99  back to json

I'll combine some good ideas from this PR to the one already opened from Mrisho since that one already has majority of the changes and other stuff that's good

daiagi:
image

the target branch is Mrisho's PR
if you like it you can just merge

README.md
```diff
@@ -72,10 +72,6 @@ Create a `peer3config.json` file in the root of your project (next to `package.j
 }
 ```
 
-```shell
-yarn && yarn build
```
lukastanisic99: Why is this deleted

README.md
```diff
@@ -58,7 +58,7 @@ The SDK abstracts away most of the complexities of the system and is designed to
 
 ## Configuration
 
-Create a `peer3config.json` file in the root of your project (next to `package.json`) with the following structure and set the values per your configuration:
+Create a `config.json` file in the root of your project (next to `package.json`) with the following structure and set the values per your configuration:
```
lukastanisic99: I'll keep the name peer3config since it's something that can be populated by the consumer of the application

src/utils/config.ts
```diff
@@ -0,0 +1,66 @@
...
+        const raw = fs.readFileSync(configPath, "utf-8");
+        return JSON.parse(raw);
+    } catch (error) {
+        console.warn(`Failed to load ${configPath}:`, error);
+        return {}; // On error, use defaults
+    }
+}
+
+// Load config once at startup
+const fileConfig = loadConfigFromFile();
```
lukastanisic99, line 29: This is a better idea than what I originally had with the singleton - I'll keep this and maybe do just some small structural changes - overall all seems good

README.md
daiagi: mistake. restored

README.md
daiagi: changed back, but to `peer3.config.json`

hope that's OK

lukastanisic99 (approved):
Actually I have no changes to add to this PR - it's all good

PR #80, opened 2025-06-04
Storage (Storage PR 0)

Krayt78:
## 📋 Description

Working on the data storage for the offchain logic.

## 💪 Related Issues

- Closes #issue-number

## ✅ Checklist

## 🔗 Follow-up Issues


src/stateManager/StateManager.ts
```diff
@@ -155,6 +156,26 @@ class StateManager {
...
+        this.p2pEventHooks.onJoinChannel?.(joinChannelBlock);
+    }
+    private handleJoinChannelStorage(joinChannelBlock: JoinChannelBlockStruct) {
+        const forkCnt = this.getForkCnt();
+        const blockHeight = this.getNextBlockHeight();
+        const blockHash = EvmUtils.encodeJoinChannelBlock(joinChannelBlock);
+        this.storageModule.storeJoinChannelBlockHash(
+            forkCnt,
+            blockHeight,
+            blockHash
```
lukastanisic99: JoinChannel isn't really tied to a fork and blockHeight - it has a lifecycle on its own and at time of creation we don't even know will it be applied to that specific [fork][height] since the dispute 'game' needs to play out - so you just persist it in the map

src/stateManager/StateManager.ts
```diff
@@ -349,6 +370,20 @@ class StateManager {
...
 
+        //This is done before the state snapshot is created
+        //This is because the exit channels need to be taken into account when creating the state snapshot
+        this.handleExitChannelsStorage(
+            exitChannels,
+            Number(transaction.header.forkCnt),
+            Number(transaction.header.transactionCnt)
+        );
+
+        this.handleStateSnapshotStorage(
```
lukastanisic99, line 363: same stuff - fork and height are not relevant to store exitChannels and you need to wrap them in an ExitChannelBlock (it points to the previous block and includes all exitChannels that are returned from the STF).

Where [fork][height] are useful here are for 'markers' so we know where to create a milestone if the participant set contracted if (stateMachine.getParticipants().length < same thing before STF) 

Plus you called the handle twice

Krayt78: ok for the first part will change, and the handle twice i called handleExit channel and the handleStateSnapshot to avoid having boilerplate

lukastanisic99: The milestone markers can also be a storage module so for a given fork you retrieve all the heights where you need to produce milestone proofs

src/stateManager/StateManager.ts
```diff
@@ -358,6 +393,41 @@ class StateManager {
...
+        blockHeight: number
+    ) {
+        const previousBlockHash =
+            this.storageModule.getLatestExitChannelBlockHash();
+        const exitChannelBlock: ExitChannelBlockStruct = {
+            exitChannels,
+            previousBlockHash: previousBlockHash as BytesLike
+        };
+        const exitChannelBlockHash =
+            EvmUtils.encodeExitChannelBlock(exitChannelBlock);
```
lukastanisic99: @daiagi we're still using encode like this?

lukastanisic99: you need to hash this not just encode

src/stateManager/StateManager.ts
lukastanisic99: Ok I see the handler wraps them in the block - good

src/stateManager/StateManager.ts
```diff
@@ -358,6 +393,41 @@ class StateManager {
...
+    private async handleStateSnapshotStorage(
+        encodedState: string,
+        forkCnt: number,
+        blockHeight: number
+    ) {
+        const stateSnapshot = await this.createStateSnapshot(
+            encodedState,
+            forkCnt
+        );
+        this.storageModule.storeStateSnapshot(blockHeight, stateSnapshot);
```
lukastanisic99: maybe best to have only hash for Snapshot too since they can come from Blocks (have fork and height) and Disputes (doesn't have height) so you can always fetch them in the context you need (working with blocks -> through blocks | working with disputes -> through disputes)

src/stateManager/StateManager.ts
Krayt78: i have updated the code

src/stateManager/StateManager.ts
Krayt78: what what would i map it from ? cause we get this from an onchain event and it should be applied to the next block right ?

src/storage/IStorageModule.ts
```diff
@@ -19,4 +20,21 @@ export interface IStorageModule {
...
+        forkCnt: number,
+        blockHeight: number,
+        blockHash: string
+    ): void;
+
+    storeExitChannelBlockHash(
+        forkCnt: number,
+        blockHeight: number,
+        blockHash: string
+    ): void;
```
lukastanisic99: Why are you persisting hashes? Hashes are they key and the value is the object hash -> object

If you have the object you can easily compute the hash and if you have the hash you can easily get the object

src/stateManager/StateManager.ts
daiagi: This api is also available, but I think using codec directly and remove the the wrappers in evm utils is the way to go

src/utils/EvmUtils.ts
```diff
@@ -87,4 +89,22 @@ export class EvmUtils {
     public static ethersResultToObjectRecursive(result: ethers.Result) {
         return Codec.ethersResultToObjectRecursive(result);
     }
+
+    public static encodeExitChannelBlock(ecb: ExitChannelBlockStruct): string {
+        return Codec.encode(ecb, Type.ExitChannelBlock);
```
lukastanisic99: @daiagi can we remove all of these explicit functions for specific types which call encode/decode and just have encode/decode

daiagi: Yes. Agree

src/stateManager/StateManager.ts
Krayt78: i went for hash => JoinChannelBlockStruct for now, will see later on if we need more pointers

src/storage/StorageModule.ts
```diff
@@ -1,32 +1,319 @@
...
+    private latestJoinChannelBlockHash: BytesLike;
+
+    //map [blockHash] => ExitChannelBlockStruct
+    private exitChannelBlockMap: Map<string, ExitChannelBlockStruct>;
+    //TODO: check if BytesLike is more efficient than string
+    // we can either store the hash or the key to the exitChannelBlockHashesMap data
+    private latestExitChannelBlockHash: BytesLike;
+
+    private totalDeposits: BalanceStruct;
+    private totalWithdrawals: BalanceStruct;
```
lukastanisic99: Partition this further into smaller modules BlockStorage, SnapshotStorage etc... Then this class can be the StorageManager that exposes them and uses them internally - that way we can clearly see how you do storage for different components instead of this file having all the maps

src/storage/StorageModule.ts
```diff
@@ -1,32 +1,319 @@
...
+    //#region Block confirmation management methods
+    getPreviousBlockHash(
+        forkCnt: number,
+        transactionCnt: number
+    ): string | undefined {
         // TODO
         return ethers.ZeroHash;
     }
 
+    getLatestBlock(): UnrolledSignedBlock {
```
lukastanisic99: I thought we're not using these artificial data structures 'Unrolled' and similar

src/storage/StorageModule.ts
Krayt78: > Partition this further into smaller modules BlockStorage, SnapshotStorage etc... Then this class can be the StorageManager that exposes them and uses them internally - that way we can clearly see how you do storage for different components instead of this file having all the maps

yeah i planned to do a refactoring into smaller modules once i have the logic done

lukastanisic99: Ah ok than I have to look at the logic now - wanted to look at it in isolation

src/storage/StorageModule.ts
Krayt78: > I thought we're not using these artificial data structures 'Unrolled' and similar

yeah i used what was already done in the dispute branch, but i can easily use the usual datatypes instead

src/storage/StorageModule.ts
```diff
@@ -1,32 +1,292 @@
...
+                    amount: BigInt(0),
+                    data: "0x"
+                },
+                totalWithdrawals: {
+                    amount: BigInt(0),
+                    data: "0x"
+                }
+            },
+            timestamp: 0
+        };
```
lukastanisic99: better to have this undefined and set through event triggers and other stuff since this code can run before the channel is even open

src/storage/StorageModule.ts
lukastanisic99: Yes please do

src/stateManager/StateManager.ts
lukastanisic99: yes that's fine since it's a blockchain and it's linked through hashes

src/storage/StorageModule.ts
```diff
@@ -1,32 +1,292 @@
...
+    BlockConfirmationStruct,
+    ExitChannelBlockStruct,
+    JoinChannelBlockStruct
+} from "@typechain-types/contracts/V1/DataTypes";
+import { UnrolledBlock, UnrolledSignedBlock } from "@/types/storage";
 
 export class StorageModule implements IStorageModule {
-    getLatestJoinChannelBlockHash(): string {
+    //BlockConfirmationStruct => SignedBlockStruct => encodedBlock => BlockStruct
+    private blockConfirmationStructsMap: Map<string, BlockConfirmationStruct>;
```
lukastanisic99: best to have a type alias for all these maps -> I assume this is a hash
use the blockHash as the identifier (not the hash of the BlockConfirmation) since you'll be expanding it with signatures when they arrive async
 
type blockHash = string;

src/storage/StorageModule.ts
```diff
@@ -1,32 +1,292 @@
...
     }
 
+    getLatestBlock(): UnrolledSignedBlock {
+        const latestBlockConfirmation = this.getLatestBlockConfirmation();
+        const latestSignedBlock = latestBlockConfirmation.signedBlock;
+        const latestBlock = EvmUtils.decodeBlock(
+            latestSignedBlock.encodedBlock
+        );
+        const stateSnapshot = this.stateSnapshotStructsMap.get(
+            hexlify(latestBlock.stateSnapshotHash)
```
lukastanisic99: you don't need to hexlify. You can use the hash directly - it's a string. Ethers uses Byteslike or something - you can use that too if needed

src/storage/StorageModule.ts
```diff
@@ -1,32 +1,292 @@
...
+        return this.stateSnapshotStructsMap.get(key);
+    }
+
+    /**
+     * Clear all mapped state snapshots and block hashes
+     */
+    clearStateSnapshot() {
+        this.stateSnapshotStructsMap.clear();
+        this.joinChannelBlockMap.clear();
+        this.exitChannelBlockMap.clear();
```
lukastanisic99: why are joins/exits cleared?

lukastanisic99: Where is this function used in general?

src/storage/StorageModule.ts
```diff
@@ -1,32 +1,292 @@
...
+        console.log("Cleared all state snapshots");
+    }
+
+    getLatestOnChainStateSnapshot(): {
```
lukastanisic99: This is misleading since someone will think they're querying the blockchain

src/storage/StorageModule.ts
```diff
@@ -1,32 +1,292 @@
...
+        const forkCnt = parseInt(parts[0], 10);
+        const blockHeight = parseInt(parts[1], 10);
+
+        if (isNaN(forkCnt) || isNaN(blockHeight)) {
+            return [undefined, undefined];
+        }
+
+        return [forkCnt, blockHeight];
     }
+    //#endregion
```
lukastanisic99: best to partition this file early - will be easier for you to work in isolation also

src/storage/StorageModule.ts
daiagi: I created this name — yes, it's not the best name, but here's the rationale:

The `set`call is meant to only be triggered by the **event emitted by the chain** when the state snapshot is updated.

So effectively, this should be equivalent to querying the chain directly — only it's local, faster, and cheaper.

The name `getLatestSnapshot` is also not ideal, because it refers to the snapshot from the latest block, not necessarily the finalized one.

a good alternative name could be 
`getLatestFinalizedStateSnapshot`

but we also use the jargon `finalized` to say all participants signed. which is also not the intention here



```txt
     +-------------+        emits event        +------------+
     |  Blockchain |  --------------------->   |  set()     |
     |   Event     |                          |  (local)    |
     +-------------+                          +------------+
                                                   |
                                                   v
                                          +-------------------+
                                          | Local State Store |
                                          +-------------------+

```

lukastanisic99: Yes I know the intent - In most cases this will be sufficient - there can of course be a race condition on-chain where the snapshot is updated, which causes the tx that's using the 'cached local' snapshot to fail (revert) - we can interpret the revert reason and use the retry module - no need to complicate things for now.

Another naming alternative 'getCachedOnChainSnapshot' - a bit long but shouldn't be ambiguous

src/storage/StorageModule.ts
Krayt78: its not used yet, i can remove it.

src/storage/StorageModule.ts
Krayt78: yeah i am just not sure if i should do bytelikes or strings everytime.

src/utils/EvmUtils.ts
Krayt78: > Yes. Agree

i will remove my changes, do you want to open a seperate PR for this ?

lukastanisic99: Sure

src/storage/StorageModule.ts
lukastanisic99: Better to do Byteslike for the getter/setter and have the underlying map be string (type alias for easy code reading) 

In general keccak returns a string when you analyze the underlying object, but for typescript so you don't have to type cast it all the time use Byteslike and then do .toString() inside the getter/setter to be safe

src/storage/StorageModule.ts
Krayt78: @daiagi in your code its ok if i return the hashes to you with the usual data and then you get the actual value via another get call to the map hash=>data ?

lukastanisic99: We can have helper functions for common operations that need to return multiple things (perform multiple reads) - the StorageManager can implement them, but first let's finish the storage layout (and partition) so it's clear what is persisted and how to CRUD (create, read, update, delete)

src/storage/StorageModule.ts
daiagi: ok @Krayt78  , `getCachedOnChainSnapshot` it is then

src/storage/StorageModule.ts
Krayt78: refactoring done

src/storage/StorageModule.ts
Krayt78: i have added two new types, one for blockhash and one for statesnapshothash

src/storage/StorageModule.ts
Krayt78: yeah now that the refactoring is done these helper functions will be easier to make on the top level storageManager

Krayt78: @daiagi how do you want the data ? giving you a tuple ? Since creating an unrolled typed is nah

Krayt78: Screenshot from 2025-06-05 11-04-37

Krayt78: this ok ?

lukastanisic99:
Would remove the name 'module' from all file and just call it BlockStorage etc 
Also would rename StorageModule -> StorageManager to follow the convention (AgreementManger, P2PManager, StateManager) etc

src/storage/interfaces/IBlockStorageModule.ts
```diff
@@ -0,0 +1,14 @@
...
+import { BlockHash } from "@/types/storage";
+
+export interface IBlockStorageModule {
+    getPreviousBlockHash(
+        forkCnt: number,
+        transactionCnt: number
+    ): BlockHash | undefined;
+    getLatestBlock(): SignedBlockStruct;
+    getLatestBlockConfirmation(): BlockConfirmationStruct;
+}
```
lukastanisic99: In general not sure why all these specific interfaces exist 'IBlock...', 'IExit...' - the point of the interface is to be general - if iStorage is something that all storage components should implement than yes, otherwise no. Then it's just better to have the specific storage component (e.g. BlockStorage) define its interface and implementation

Krayt78: yeah at first i had them inside their respective storage files. I figured i would extract them in a seperate folder just to be cleaner (habit of C#)

src/storage/StorageModule.ts
```diff
@@ -1,32 +1,111 @@
...
+import { IBlockStorageModule } from "./interfaces/IBlockStorageModule";
+import { IExitChannelStorageModule } from "./interfaces/IExitChannelStorageModule";
+import { IJoinChannelStorageModule } from "./interfaces/IJoinChannelStorageModule";
+import { JoinChannelStorageModule } from "./JoinChannelStorageModule";
+import { IStateSnapshotStorageModule } from "./interfaces/IStateSnapshotStorageModule";
+import { ExitChannelStorageModule } from "./ExitChannelStorageModule";
+import { StateSnapshotStorageModule } from "./StateSnapshotStorageModule";
+import { BlockHash } from "@/types/storage";
 
 export class StorageModule implements IStorageModule {
```
lukastanisic99: I understand your logic to 'force' StorageManager to implement iStorage witch internally implements all storage interfaces, but you don't need to do that since this is the only class that gets exposed to the rest of the system so you choose the interface that is exposed and it internally knows how to interact with specific storage components and should it expose some of their APIs and to which degree 

Later if we plan on having different implementations for StorageManger it's worth creating an interface or abstract class so the concrete implementations can be polymorphically substituted - while there is only 1 instance it's just over engineering

src/storage/BlockStorage.ts
```diff
@@ -0,0 +1,43 @@
...
+    getLatestBlockConfirmation(): BlockConfirmationStruct {
+        const confirmation = this.blockConfirmationStructsMap.get(
+            this.latestBlockConfirmationKey
+        );
+        if (!confirmation) {
+            throw new Error("No latest block confirmation found");
+        }
+        return confirmation;
+    }
+}
```
lukastanisic99, line 298: So this file should only care about CRUD operations on blocks (blockConfirmations) it should not care about any business logic or keep track of 'what is latest' - the component that needs the latest should keep track of it.

It should be possible to do CRUD operations both by [hash] and by [fork][height] - a dual interface that maps to the same data

One possible implementation:
2 maps:
[hash] -> blockConfirmation
[fork][height] -> blockConfirmation 
Since blockConfirmation is a complex type this is essentially 'pas by reference' (duplicate data won't be stored, just 2 'pointers') and updating 1 will update the other - in case of CD (create, delete) you just insert/delete both 'pointers' 

That's one possible way of doing it - if you have a better way, sure

C (create) operations:
insertBlock(signedBlock)
insertBlock(hash, fork, height, signedBlock) - same as above, but you don't need to decode and hash to get it - just an optimization
insertBlock(blockConfirmation)
insertBlock(hash, fork, height, blockConfirmation) 
// you're not checking the signatures or the validity of the data - you just persist it
//all require that the entry doesn't exist already - otherwise throw

R(read) operations:
getBlockConfirmation(hash)
getBlockConfirmation(fork,height)

D(delete) operations:
deleteBlock(hash)
deleteBlock(fork, height)

The only 'U' (update) operation that's possible on this structure is InsertSignature - you expose insert and don't care about the business logic is it a duplicate signature, is it from a participant or not etc etc ... some other component will do those checks - this component just unconditionally persists it

src/storage/ExitChannelStorage.ts
```diff
@@ -0,0 +1,73 @@
...
+            amount: this.totalWithdrawals.amount,
+            data: hexlify(this.totalWithdrawals.data)
+        };
+    }
+
+    /**
+     * Add to the total withdrawals
+     * Doesnt handle data yet, just the amount
+     */
+    private addToTotalWithdrawals(balance: BalanceStruct) {
```
lukastanisic99: You can't know this here or at least shouldn't - What a 'Balance' is and how to add/subtract is left to the stateMachine implementation - here you can only 'persist'. The information will be given to you. https://github.com/peer3to/state-channels-plus/blob/e67a351a01b50ae9bdee38443eac06000229e9ee/contracts/V1/AStateMachine.sol#L39

Here you have only CR operations mostly -  D can be implemented for pruning when needed later and in general all 'D' operations are not used at the moment. 

There are no U operations, at least for now

This is just a single mapping - hash(ExitChannelBlock) -> ExitChannelBlock 
There is no dual mapping [fork][height] - at least it's not needed at this stage

src/storage/JoinChannelStorage.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    /**
+     * Add to the total deposits
+     * Doesn't handle data yet, just the amount
+     */
+    private addToTotalDeposits(balance: BalanceStruct) {
+        //TODO: create a method to add two balances together
+        this.totalDeposits.amount += toBeHex(balance.amount);
+        console.log(`Total deposits updated: ${this.totalDeposits}`);
+    }
```
lukastanisic99: same comments as for exits

src/storage/StateSnapshotStorage.ts
```diff
@@ -0,0 +1,81 @@
...
+
+export class StateSnapshotStorageModule implements IStateSnapshotStorageModule {
+    //map [stateSnapshotHash]=> struct
+    // we make the key a string to avoid double mapping
+    private stateSnapshotStructsMap: Map<
+        StateSnapshotHash,
+        StateSnapshotStruct
+    >;
+    //could be replaced with a key, but we would need some computation
+    private cachedOnChainStateSnapshot:
```
lukastanisic99: Similar analogy to 'latestBlockConfirmation' - doesn't need to be here. I agree this information is useful, the snapshot is persisted regardless and some other component can handle the business logic/caching what's the onChainSnapshot - maybe StorageManager can keep the hash and when needed retrieve the specific snapshot from here

src/storage/BlockStorage.ts
Krayt78: why do you want this ? insertBlock(signedBlock)
i thought we only stored 

export type BlockConfirmationStruct = {
  signedBlock: SignedBlockStruct;
  signatures: BytesLike[];
};

with this insert we are missing the signatures from everyone that isnt the one authoring the block right ?

lukastanisic99: Yes exactly.

The block confirmation is a superset of signedBlock. In most cases you'll first receive a signBlock and persist it and eventually receive confirmation signatures and update the structure as signatures arrive. Signatures may never arrive and you may just have signedBlocks which is valid and you'd eventually finalize through virtual voting or disputes.

Best case: the BlockConfirmation has all N signatures confirming the block
Worse case: only the signedBlock exists

src/storage/interfaces/IBlockStorageModule.ts
lukastanisic99: Sure, but my question is why do they even exits? It would make sense to have them if we planned on having multiple storage implementations of the same type (e.g. BlockStorageA, BlockStorageB, BlockStorageC...), but currently I don't foresee this - so in my mind it's overengineering, unless there is another valid reason to include them?

Krayt78: Nah i agree its overengineering, no real reason to have them i wil lremove it

src/storage/BlockStorage.ts
Krayt78: > Yes exactly.
> 
> The block confirmation is a superset of signedBlock. In most cases you'll first receive a signBlock and persist it and eventually receive confirmation signatures and update the structure as signatures arrive. Signatures may never arrive and you may just have signedBlocks which is valid and you'd eventually finalize through virtual voting or disputes.
> 
> Best case: the BlockConfirmation has all N signatures confirming the block Worse case: only the signedBlock exists

so i can just create a BlockConfirmationStruct with the signed block but with the signatures array empty and then when doing gossip the client can then update the storage by adding the signatures 1 by 1 to the array ?

lukastanisic99: Yes

src/storage/ExitChannelStorage.ts
Krayt78: so removing withdrawals/deposits from join/exist storage and just putting them in the storage manager for now with only getter/setter without anything else ?

lukastanisic99: I think you can keep them here, you just can't calculate them here. Expect that they're provided

src/storage/StorageModule.ts
daiagi: yep. this will work

src/storage/StorageManager.ts
```diff
@@ -0,0 +1,264 @@
...
+                this.blockStorageModule.insertBlock(
+                    blockData,
+                    blockHash,
+                    fork,
+                    height
+                );
+            } else {
+                this.blockStorageModule.insertBlock(blockData);
+            }
+        }
```
daiagi: the external `if/else` is not needed. both arms do the same

```suggestion
            if (blockHash && fork && height) {
                this.blockStorageModule.insertBlock(
                    blockData,
                    blockHash,
                    fork,
                    height
                );
            } else {
                this.blockStorageModule.insertBlock(blockData);
            }
 ```

File: src/storage/StorageManager.ts [hunk 22]
```diff
@@ -0,0 +1,264 @@
...
+                blockHashOrFork
+            );
+        } else if (typeof blockHashOrFork === "number" && height) {
+            return this.blockStorageModule.getBlockConfirmation(
+                blockHashOrFork,
+                height
+            );
+        }
+        return undefined;
+    }
```
daiagi:
```suggestion
        if (typeof blockHashOrFork === "string") {
            return this.blockStorageModule.getBlockConfirmation(
                blockHashOrFork
            );
        } 
        if (typeof blockHashOrFork === "number" && height) {
            return this.blockStorageModule.getBlockConfirmation(
                blockHashOrFork,
                height
            );
        }
        return undefined;
    }

File: src/storage/StorageManager.ts [hunk 23]
```diff
@@ -0,0 +1,264 @@
...
+        height?: number
+    ): void {
+        if (typeof blockHashOrFork === "string") {
+            this.blockStorageModule.deleteBlockConfirmation(blockHashOrFork);
+        } else if (typeof blockHashOrFork === "number" && height) {
+            this.blockStorageModule.deleteBlockConfirmation(
+                blockHashOrFork,
+                height
+            );
+        }
```
daiagi:
```suggestion
    if (typeof blockHashOrFork === "string") {
        this.blockStorageModule.deleteBlockConfirmation(blockHashOrFork);
        return;
    }

    if (typeof blockHashOrFork === "number" && height !== undefined) {
        this.blockStorageModule.deleteBlockConfirmation(blockHashOrFork, height);
    }
 ```

File: src/storage/BlockStorage.ts [hunk 24]
```diff
@@ -0,0 +1,192 @@
...
+                    blockData,
+                    blockHash,
+                    fork,
+                    height
+                );
+            } else {
+                this.insertSignedBlock(blockData);
+            }
+        }
+    }
```
daiagi:
```suggestion
    const hasKeys = blockHash !== undefined && fork !== undefined && height !== undefined;

    if (this.isBlockConfirmation(blockData)) {
        return hasKeys
            ? this.insertBlockConfirmationWithKeys(blockData, blockHash, fork, height)
            : this.insertBlockConfirmation(blockData);
    }

    // At this point, it's a SignedBlockStruct
    return hasKeys
        ? this.insertSignedBlockWithKeys(blockData, blockHash, fork, height)
        : this.insertSignedBlock(blockData);
```

src/storage/BlockStorage.ts
```diff
@@ -0,0 +1,192 @@
...
+    getBlockConfirmation(
+        blockHashOrFork: BlockHash | number,
+        height?: number
+    ): BlockConfirmationStruct | undefined {
+        if (typeof blockHashOrFork === "string") {
+            // Called with blockHash
+            return this.blockhashToBlockConfirmationStructsMap.get(
+                blockHashOrFork
+            );
+        } else if (typeof blockHashOrFork === "number" && height) {
```
daiagi: no need for else if
since the previous if is returning

daiagi (changes requested):
just style comments
try to avoid nested if/else if/else

prefer flat  and use early returns

Krayt78:
> just style comments try to avoid nested if/else if/else
> 
> prefer flat and use early returns

roger, i will do a pass on the code and use early returns instead of nested ifs !

src/storage/StorageManager.ts
Krayt78: yeah i will remove the nested if but coding it this way is not possible tho since we need to check blockData's type

Krayt78: for this one i dont think i can remove the first nested if unless i add some redundency to the function

Krayt78: nvm i keep forgetting that ternary operator exist since it doesnt in rust xD

daiagi:
@lukastanisic99  re-review please

any comments you may have will be handled in next PR

changes:
1. style, order, comments - just clean up and arrange code
    
2. commented out  `getPreviousBlockHash` from `Storage` module.
    since disputes are no longer total ordered, this will require more logic / extra maps.  will be handled in next PR (keep this one light)
    
    
3. not included in this PR
    a. Dispute storage
    b. milestone markings

src/storage/BlockStorage.ts
```diff
@@ -0,0 +1,284 @@
...
+
+        this.hashToBlockMap.set(blockHash, blockConfirmation);
+        this.coordinatesToBlockMap.set(coordinateKey, blockConfirmation);
+        return blockHash;
+    }
+
+    private isBlockConfirmation(
+        blockData: SignedBlockStruct | BlockConfirmationStruct
+    ): blockData is BlockConfirmationStruct {
+        return "signedBlock" in blockData && "signatures" in blockData;
```
lukastanisic99: we should include a test for just this, so if `BlockConfirmation` ever changes - the test will fail - so we don't forget to update this if needed

src/storage/ExitChannelBlockStorage.ts
```diff
@@ -0,0 +1,82 @@
...
+    ExitChannelBlockStruct,
+    BalanceStruct
+} from "@typechain-types/contracts/V1/types/DataTypes";
+import { Hash } from "@/types/types";
+import { Codec, Type } from "@/utils";
+
+export class ExitChannelBlockStorage {
+    private blockMap: Map<Hash, ExitChannelBlockStruct>;
+    private _latestBlockHash: Hash;
+    private _totalWithdrawals: BalanceStruct;
```
lukastanisic99: I'll make the same comment as to Ludo b4 - not sure we should keep track of `_latestBlockHash` here. How do you define the latest? Does it take into account reorgs, forks, etc etc..  ? It would require additional 'business logic' that goes beyond just the simple `CRUD` operations. 

As for `_totalWithdrawals` you could do it here, but not globally like this, since it too would require business logic for the same reason as `_latestBlockHash`

What you could do is extend the struct that is mapped internally to include `totalWithdrawals`. `Hash` -> `{ExitChannelBlock, totalWithdrawals}`. You calculate `totalWithdrawals` as `previousTotalWithdrawals` + `currentExitChannelBlock` - you can do that since this is a blockchain in the end + you don't need any business logic - the relation will always hold even if it forks / reorgs

src/storage/BlockStorage.ts
```diff
@@ -0,0 +1,284 @@
...
+    // ====================================
+    // CREATE - throw if entry exists
+    // ====================================
+
+    /*────────────────────────────────────────────────────────────────────────────
+      OVERLOAD SIGNATURES
+    ────────────────────────────────────────────────────────────────────────────*/
+
+    /** [OVERLOAD 1] Insert signed block with auto-computed keys */
+    insertBlock(signedBlock: SignedBlockStruct): Hash;
```
lukastanisic99: would rename to `storeBlock` to be consistent with other 'blockchain like' storage modules that all use `store...`

src/storage/JoinChannelBlockStorage.ts
```diff
@@ -0,0 +1,82 @@
...
+    JoinChannelBlockStruct,
+    BalanceStruct
+} from "@typechain-types/contracts/V1/types/DataTypes";
+import { Hash } from "@/types/types";
+import { Codec, Type } from "@/utils";
+
+export class JoinChannelBlockStorage {
+    private blockMap: Map<Hash, JoinChannelBlockStruct>;
+    private _latestBlockHash: Hash;
+    private _totalDeposits: BalanceStruct;
```
lukastanisic99: same comment as for exitChannel

src/storage/Storage.ts
```diff
@@ -0,0 +1,176 @@
...
+    // R
+    getBlockConfirmation = this.blockStorage.getBlockConfirmation.bind(
+        this.blockStorage
+    );
+
+    // U
+    insertSignature = this.blockStorage.insertSignature.bind(this.blockStorage);
+
+    // D
+    deleteBlock = this.blockStorage.deleteBlock.bind(this.blockStorage);
```
lukastanisic99: maybe better instead of binding all `CRUD` functions for all storage instances, to just have a getter and expose those storage instances - cleans up the code here and if the storage instances change (expose a different interface) you don't have to update anything. 

Then this class can expose additional helper functions and act as the 'singleton'

src/storage/BlockStorage.ts
lukastanisic99: To consider - I know it was my idea to do method overloading for `signedBlock` and `blockConfirmation` and it would make sense in languages that support method overloading normally, but here maybe better to have a split/dual interface and not have the `4 case` overloading - just something to consider, this works regardless

src/storage/BlockStorage.ts
```diff
@@ -0,0 +1,284 @@
...
+    ): BlockConfirmationStruct | undefined {
+        const blockConfirmation =
+            height === undefined
+                ? this.hashToBlockMap.get(hashOrForkId as Hash)
+                : this.coordinatesToBlockMap.get(
+                      this.coordinatesToKey([hashOrForkId as ForkId, height])
+                  );
+
+        if (blockConfirmation) {
+            blockConfirmation.signatures.push(signature);
```
lukastanisic99: Do we want to check for duplicates here? That should be simple without needing external business logic - in reality we don't have to do it since some hire function should do all the checks and this should just persist

It's always the question what should we do here (which checks) and what should we do somewhere else - ideally you want to reuse as much as possible, but also be as simple as possible 

Should some checks be part of the `StorageManager` or what is called `Storage` for now or should consuming modules perform all the checks and just use this `unsafe`

There are pros and cons to both options

src/storage/ExitChannelBlockStorage.ts
```diff
@@ -0,0 +1,82 @@
...
+        blockHash?: Hash
+    ): Hash {
+        const hash =
+            blockHash ??
+            ethers.keccak256(Codec.encode(block, Type.ExitChannelBlock));
+
+        // Check for duplicates
+        if (this.blockMap.has(hash)) {
+            throw new Error(
+                `Exit channel block with hash ${hash} already exists`
```
lukastanisic99: No need to error here - if it exists just return the same hash - it's possible to trigger this for the same `exitChannelBlock` multiple times if a reorg/fork happens - probably same applies for `JoinChannelBlock`

src/storage/Storage.ts
daiagi: if i understand correctly you mean something like:

```ts

export class Storage {
    private blockStorage = new BlockStorage();
...

    get blocks() { return this.blockStorage; }

```


Storage.BlocksStorage..do_something is a very verbose,  "Java like" interface
I wanted a flat interface.

how about this? https://github.com/peer3to/state-channels-plus/pull/80/commits/d8fdfedab8dfd6714713f54177482f743519e518

src/storage/BlockStorage.ts
daiagi: yes, overloading is over engineered and does not work well in ts

src/storage/ExitChannelBlockStorage.ts
daiagi: how do we define `latestExitChannelBlockHash` in state snapshot?

it makes sense to keep it  because you need to upload it for updating state snapshot fro example

if no, then how to get the latest?
pick random entry and search for a block that has the chosen hash as `previousBlockHash`?
move forward like this through the map until you don't find a block (O(n^2))

or , we extend the data structure to include a "nextBlockHash"?

src/storage/BlockStorage.ts
daiagi: if you ask me, yes,  it should
but your comments to ludo made it clear you want un-opinionated storage layer

src/storage/ExitChannelBlockStorage.ts
lukastanisic99: You do have it, but in some higher class like `StateManager` since it checks are you appending the next block -> it knows the `latestBlock` and keeps track of it and accounts for forks/reorgs so it can also keep track of this as a cache optimization and it should, but in reality if you know the previous/latestBlock you know everything since you would fetch the hash from the snapshot

src/storage/Storage.ts
lukastanisic99: Imo it should remain split into small components even though they're mostly maps - easy to see what we're persisting and we don't clog a single file so this file can have additional helper functions that combine elements from the modules.

`Storage.storeBlock(block)` is indeed cleaner than `Storage.blockStorage.storeBlock(block)` so it's probably worth the `bind` or implementing it's own `storeBlock(block)` that reuses the underlying one if additional checks are done here

src/storage/BlockStorage.ts
daiagi: added duplicate prevention

src/storage/BlockStorage.ts
daiagi: broke it down to 2 overloaded functions
deleted `isBlockConfirmation` function

src/storage/BlockStorage.ts
lukastanisic99: Let's discuss in the call later - imo we shouldn't expose additional business logic like `isSignerParticipant` but we should do all checks that are independent of the business logic of the consuming modules, so we don't have inconsistent data. It may lead to some redundancy, since higher modules may perform the same checks, but better to optimize for consistency/safety and if we really hit performance issues later we can see what we can strip when everything is working

src/storage/BlockStorage.ts
```diff
@@ -0,0 +1,295 @@
...
+    ): Hash {
+        const coordinateKey = this.coordinatesToKey(coordinates);
+
+        // Check for duplicates - throw if exists
+        if (this.hashToBlockMap.has(blockHash)) {
+            throw new Error(`Block with hash ${blockHash} already exists`);
+        }
+        if (this.coordinatesToBlockMap.has(coordinateKey)) {
+            throw new Error(
+                `Block at fork ${coordinates.forkId}, height ${coordinates.height} already exists`
```
lukastanisic99: Probably also good to not error here - if the blockConfirmation exists already try and update it with signatures from the provided one and return the same hash regardless

src/storage/JoinChannelBlockStorage.ts
```diff
@@ -0,0 +1,82 @@
...
+        blockHash?: Hash
+    ): Hash {
+        const hash =
+            blockHash ??
+            ethers.keccak256(Codec.encode(block, Type.JoinChannelBlock));
+
+        // Check for duplicates - throw if exists
+        if (this.blockMap.has(hash)) {
+            throw new Error(
+                `Join channel block with hash ${hash} already exists`
```
lukastanisic99: same

src/storage/ExitChannelBlockStorage.ts
lukastanisic99: There's also an option where you can `cache` is here, but you'd have to expose a `setter` that a higher level component e.g. `StateManager` can call to set the latestBlockHash when a reorg/fork happens and you need to remember to actually do this - if it stays decoupled you can't forget since you have to do it in StateManager regardless

src/storage/Storage.ts
```diff
@@ -0,0 +1,99 @@
...
+    private bindMethods(source: any) {
+        Object.getOwnPropertyNames(Object.getPrototypeOf(source))
+            .filter(
+                (name) =>
+                    typeof source[name] === "function" && name !== "constructor"
+            )
+            .forEach((name) => {
+                (this as any)[name] = source[name].bind(source);
+            });
+    }
```
lukastanisic99: This is only runtime - typescript doesn't deduct the interface / method signatures

src/storage/BlockStorage.ts
daiagi: I don't understand why you want  these implicit behaviors

you don't want errors thrown? fine. that's a reasonable API design choice.
return key of prev existing entry? OK i guess, i think return something like `{success: boolean, key: Hash}` is better but maybe we don't need this added complexity.

but implicitly altering an existing entry?! why?

src/storage/Storage.ts
daiagi: ok 

https://github.com/peer3to/state-channels-plus/pull/80/commits/e96c03572fedfd04096203d014a1f7eb701bc6db

src/storage/BlockStorage.ts
lukastanisic99: No reason to error if it's already persisted - the system is not in an inconsistent state, it can progress - it can happen if a higher component has a race condition and is not prevented with something like a `mutex` - which in our case it is for now, but still 

As for my opinion on why to `try` and alter an exiting entry if an entry already exits, we're not breaking anything - we want to collect all the signatures and it shouldn't matter in which order we do it and if there's a race condition in a higher component and `storeBlock` is called multiple times - we can account for that easily with everything we already have here without exposing additional business logic

src/storage/Storage.ts
daiagi: i went your way.
there are some other fancy tricks we can do to flatten the interface, like proxy for example

all will feel somewhat over-engineering

not worth spending time on this, lets move on, we can always can come back and improve interfaces if we realize they are not convenient to use

lukastanisic99 (approved):

PR #82, opened 2025-06-06
Join-channel

daiagi:
Not quite done but good enough for a PR to get feedback early

outstanding gaps:

- implement `prepareStateSnapshotData`
- getting joiner signature over dispute
on this i am not quite sure, this is my proposed idea:
a. through a new rpc endpoint, we gossip the dispute
b. recipient who is not the join requester => re-broadcast
c. join requestor => sign an broadcast

but this is somewhat problematic, we already have 1 gossip process to collect all of the signatures of the current channel participants

so i guess what should happen is:
1. gossip until all signatures are collected
2. make a dispute struct - link it in memory to the `joinChannel`
3. broadcast it again
4. participant that gets a dispute signed by the joiner AND they already have all of the signatures on the linked join channel does all steps for which code is already written i.e
a validate
b. possibly send state snapshot to chain
c. send the join request

also, @lukastanisic99  could you clarify exactly what data is needed to be made available to the joiner? i assumed so far it is only a `dispute` struct, but maybe an `auditDispute` struct as well?


lukastanisic99:
> Not quite done but good enough for a PR to get feedback early
> 
> outstanding gaps:
> 
> * implement `prepareStateSnapshotData`
> * getting joiner signature over dispute
>   on this i am not quite sure, this is my proposed idea:
>   a. through a new rpc endpoint, we gossip the dispute
>   b. recipient who is not the join requester => re-broadcast
>   c. join requestor => sign an broadcast
> 
> but this is somewhat problematic, we already have 1 gossip process to collect all of the signatures of the current channel participants
> 
> so i guess what should happen is:
> 
> 1. gossip until all signatures are collected
> 2. make a dispute struct - link it in memory to the `joinChannel`
> 3. broadcast it again
> 4. participant that gets a dispute signed by the joiner AND they already have all of the signatures on the linked join channel does all steps for which code is already written i.e
>    a validate
>    b. possibly send state snapshot to chain
>    c. send the join request
> 

Mostly it reuses the same dispute pipeline - the only divergence is it assumes that latestJoinChannelBlockHash will be updated in the same atomic transaction prior to the dispute being created (deposits will be processed successfully on-chain and the snapshot updated). Apart from that it's the same dispute process - you don't need to collect any signatures for disputes - if you had N/N on JoinChannel that's enough to deposit funds and force the dispute game to include the new joiner in the new fork. 

The optimization we do for disputes is before you initialize it on-chain, you gossip it and sign it and if a threshold is reached you can finalize it instantly without the challenge period. Otherwise it will playout the dispute game. 

Apart from the optional early finalization through N/N, we force auditing on-chain unless all participants that are in `pendingParticipants` sign the dispute. That means that they agree with the dispute and that somehow they could run auditing (by someone providing them all the data (`dispute` + `auditingData`)). If this is not the case, auditing has to run on-chain to account for their safety until they join (`pendingParticipants` are cleared - snapshot is updated). 

In reality no one will want to run auditing on-chain for some random peers that want to join, so practically it means that they will provide them the correct data for them to audit and they will cooperate and sign - otherwise the joining will be aborted and `JoinChannel` will eventually expire worthless. 

Summary - you don't need any 'special' RPC endpoint and you don't differentiate between participants and pendingParticipants while collecting signatures, you just subjectively decide to abort if pendingParticipants are not cooperating.

contracts/V1/StateChannelDiamondProxy/AStateChannelManagerProxy.sol
```diff
@@ -210,6 +214,20 @@ abstract contract AStateChannelManagerProxy is StateChannelManagerInterface, Sta
         );
     }
 
+    function joinChannel(
+        bytes32 channelId,
+        JoinChannelBlock memory joinChannelBlock,
+        DisputeConfirmation memory disputeConfirmation,
+        bytes[][] memory confirmationSignatures
```
lukastanisic99: Better to do it JoinChannelConfirmation[ ] instead of bytes[ ][ ]

lukastanisic99: Plus the block is formed on-chain - it wraps all successful (good signatures and successful deposits) JoinChannels from the transaction

contracts/V1/StateChannelDiamondProxy/JoinChannelFacet.sol
```diff
@@ -0,0 +1,196 @@
...
+        bytes[][] memory confirmationSignatures // per join channel, per participant
+    ) external {
+        // VERIFICATION PHASE
+
+        // 1. Validate join channel deadlines
+        _validateJoinChannelDeadlines(joinChannelBlock);
+
+        // 2. for each join channel, verify that the joining participant has signed the dispute struct
+        _verifyJoinChannelSignatures(
+            joinChannelBlock, disputeConfirmation.signedDispute.encodedDispute, disputeConfirmation.signatures
```
lukastanisic99: Imo best to do this as:
for(jcc in JoinChannelConfimrations){
require(verifyThreshold(joinChannel, snapshot.participants+jcc.participant, sigantures)) 
}

since this is fully deterministic (no race condition) it's ok to atomically require all to pass (the submitter of the tx should submit only valid N/N  stuff 

deposits can be on a case-by-case basis - some fail some succeed since there can be a race condition (1 peer spent their funds before the tx executed on-chain -> this is something the tx creator can not know, so we tolerate failure here)

contracts/V1/StateChannelDiamondProxy/JoinChannelFacet.sol
```diff
@@ -0,0 +1,196 @@
...
+        );
+
+        // 3. Verify:
+        //    a. threshold signatures from existing participants on each join channel
+        //    b. that the block is properly linked
+
+        _verifyJoinChannelBlock(channelId, joinChannelBlock, confirmationSignatures);
+
+        // 4. Verify balance invariants
+        _verifyJoinChannelBalanceInvariants(channelId);
```
lukastanisic99: This is part of disputes in the end - should already be implemented in auditing

contracts/V1/StateChannelDiamondProxy/JoinChannelFacet.sol
```diff
@@ -0,0 +1,196 @@
...
+
+        // PROCESSING PHASE - Execute state changes after all verifications pass
+
+        // 5. Process deposits
+        _processJoinChannelDeposits(channelId, joinChannelBlock);
```
lukastanisic99, line 105: the JoinChannels that made it to here are processed and only the successful ones are wrapped in the block

contracts/V1/StateChannelDiamondProxy/JoinChannelFacet.sol
```diff
@@ -0,0 +1,196 @@
...
+
+        // 6. Update pending participants (must be done before createDispute)
+        _updatePendingParticipants(channelId, joinChannelBlock);
+
+        // 7. Create dispute - this handles remaining verification and state updates
```
lukastanisic99, line 110: Latest StateSnapshot is required (at least the highest possible fork) - in the current dispute system it's the latest fork since it's total ordered, in the new one that uses CRDTs there's a fork choice rule that picks the best fork and does it recursively - I'll do that logic with disputes 

Here you can assume that there is a function isSnapshotLatest or something similar that returns true/false and in that case you know there are no ongoing disputes.

daiagi:
>The optimization we do for disputes is before you initialize it on-chain, you gossip it and sign it and if a threshold is reached you can finalize it instantly without the challenge period. Otherwise it will playout the dispute game.

you misunderstand, this is not at all what I'm about, The early finalization of dispute is not the topic here.
I'm concerned with the process of  providing the joiner with the required data (which i misunderstood to be only `dispute` no i realized it is `dispute` + `auditingData`)




We need to:
- provide the joiner with `dispute` + `auditingData`. **how?** => RPC / gossip
- they sign and **then what**? is it the joiner responsibility to submit to chain or do they gossip again?

by the way, how can they even gossip if they are not in the channel? 


>In reality no one will want to run auditing on-chain for some random peers that want to join, so practically it means that they will provide them the correct data for them to audit and they will cooperate and sign - otherwise the joining will be aborted and JoinChannel will eventually expire worthless.



put it all together, apart from the concern of how do you gossip with someone not in the channel, the process should look something like

1. via p2psigner, a new joiner sends request to join
2. peers in channel gossip until all signatures are collected
3. post a state-snapshot
4. assemble dispute + audit data
5. gossip it
6. joiner runs auditing on their local evm - singes it , or not
7. wait for joiner signature for some timeout
7.  PROBLEM - by the time we wait for the signature of the joiner to arrive, possibly a newer state snapshot has been submitted

8. post the join request to chain

contracts/V1/StateChannelDiamondProxy/JoinChannelFacet.sol
daiagi: I'm not interested in the signatures of the `snapshot.participants`
I am interested to see that the prospective joiner has signed the data provided to them in order to allow them to audit the dispute

contracts/V1/StateChannelDiamondProxy/AStateChannelManagerProxy.sol
daiagi: done

contracts/V1/StateChannelDiamondProxy/JoinChannelFacet.sol
daiagi: removed

contracts/V1/StateChannelDiamondProxy/JoinChannelFacet.sol
daiagi: done

contracts/V1/StateChannelDiamondProxy/JoinChannelFacet.sol
daiagi: done

contracts/V1/StateChannelDiamondProxy/JoinChannelFacet.sol
lukastanisic99: You're interested in both and you already check that `snapshot.participants` signed with another function call bellow - what i'm saying is that you can do it in one check where the threshold is `snapshot.participants` + `jcc.participant` - This is the minimal requirement, I'd be more strict and require the threshold to be snapshot.participants` + `pendingParticipants` so you can't have collusion and in most cases it would be equivalent since 1 would join per dispute anyway.

The comment after was about atomicity and requiring this signature stuff to be atomic (success/failure) while its execution on-chain is always deterministic/no race condition (the tx creator know in advance will it pass or fail), while actually processing the deposits are by a case by case bases since there can be race condition there. 

Dispute signatures are something else than jcc signatures. Dispute signatures are required in order not to run auditing on-chain, but it shouldn't be the responsibility of this code to enforce it - my PR on disputes already requires that check.

In general when I give feedback I share a bit more around that feedback, not only the single line of code I'm commenting, so you at least have my reasoning why certain parts of the system where designed the way there were and you can think of them in isolation in stead of thinking them as a whole.

lukastanisic99:
> you misunderstand, this is not at all what I'm about, The early finalization of dispute is not the topic here. I'm concerned with the process of providing the joiner with the required data (which i misunderstood to be only `dispute` no i realized it is `dispute` + `auditingData`)

Yes I understood you're asking what data is needed to be gossiped for them to audit and sign and instead of just saying (`dispute` + `auditingData`) I gave you more context around it :D 

> 
> by the way, how can they even gossip if they are not in the channel?
> 

The p2p network comes before this web3 stuff - they connect to one another via Holepunch and can openly communicate - even in the current demo they exchange information and communicate before they decide to commit and open the channel.
E.g. they check that their Clocks (perception of time) is not too skewed. In general they never have to use this web3 stuff they can just use the p2p network for arbitrary messages if they wanted to. 'Being in the channel' is not the requirement for maintaining/having/establishing a p2p connection with anyone. The connection always comes first.

> 
> We need to:
> 
> * provide the joiner with `dispute` + `auditingData`. **how?** => RPC / gossip
> * they sign and **then what**? is it the joiner responsibility to submit to chain or do they gossip again?

Yes RPC and like most RPC it's gossiped. The single Dispute endpoint can always gossip `dispute` + `auditingData` - we have 'infinite bandwidth' off-chain and it's 'free'. Peers gossip to their open connections regardless if the recipient is part of the channel or not. 

There is no 'special case' and in general I've tried to designed the system to not have 'special cases'. This is just a normal dispute. The only difference is instead of first updating the snapshot and processing the deposits to collect the event (get notified) that something happened on-chain and then submitting a dispute to account for the updated information, we make this atomic. You as the tx submitter assume that your transaction will successfully update the snapshot and process the deposits and you account for that in the dispute you submit (normal dispute) in the same transaction that runs after the snapshot update and deposit processing. Like this we have atomic success/failure and in the case that the dispute did not account for their join the tx will fail regardless and the deposits wouldn't be processed so they wouldn't join and would have their money unlocked. 

> 
> put it all together, apart from the concern of how do you gossip with someone not in the channel, the process should look something like
> 
> 1. via p2psigner, a new joiner sends request to join
> 2. peers in channel gossip until all signatures are collected
> 3. post a state-snapshot
> 4. assemble dispute + audit data
> 5. gossip it
> 6. joiner runs auditing on their local evm - singes it , or not
> 7. wait for joiner signature for some timeout
> 8. PROBLEM - by the time we wait for the signature of the joiner to arrive, possibly a newer state snapshot has been submitted
> 9. post the join request to chain

You don't do 3. - at least not there

6. everyone not just joiner
7. everyone not just joiner

8. For joins, you don't stop progressing the state while you wait `AgreementTime` to see if you've collected signatures or not. If signatures don't come, you just abort like the join was never initialized. If you progressed the state more and signatures eventually come in time, you submit 'more evidence' within the dispute 'window/challenge period' and in the end it will reduce to the latest state that accounts for joins. (You can take a look at the draft PRs for the dispute CRDTs - you have `reduce` and you have `uploadDispute` already)

As for the race condition of submitting a 'newer snapshot' that can happen even if `AgreementTime = 0` (you don't wait and submit instantly), but that is not a problem and doesn't affect signatures in any way. If currently the on-chain snapshot is `A` and you want to update it to `Z` even if someone updates do `B`, `C` .... `Y` before you it doesn't matter since `(A,B,C,...Z)` <=> `(A,Z)` and you update to `Z`.
Further all the snapshots that can be updated can only reduce the participant set - only joins can expand it - and if you have `N` signatures and the participant set is reduced to `K`, where `K<N`, `K` is for sure a subset of `N` and our thresholdVerify function already accounts for submitting more signatures than the threshold requirement, so even if you submit `N` signatures when `K` is expected it will return true and pass.

contracts/V1/DisputeTypes.sol
```diff
@@ -192,6 +195,21 @@ struct DisputeAuditingData {
     uint256 previousDisputeTimestamp; // (optional) needed to verify the commitment of the previous dispute
 }
 
+struct DisputeAuditBundle {
+    Dispute dispute;
+    DisputeAuditingData disputeAuditingData;
+}
+
+struct SignedDisputeAuditBundle {
+    bytes encodedDisputeAuditBundle;
```
lukastanisic99: you don't need to sign the bundle, just the `dispute` since the dispute commits to `auditingData` - Not sure why the 'bundle' is useful, just adding more data structures

contracts/V1/DisputeTypes.sol
```diff
@@ -192,6 +195,21 @@ struct DisputeAuditingData {
...
+    bytes signature;
+}
+
+struct DisputeAuditBundleConfirmation {
```
lukastanisic99: Since we have `signedDispute` and `DisputeConfirmation` which are minimal and sufficient all these new data structures that just wrap the same thing are over engineering

contracts/V1/DisputeTypes.sol
daiagi: is unused, i forgot to delete it

contracts/V1/StateChannelDiamondProxy/JoinChannelFacet.sol
```diff
@@ -0,0 +1,152 @@
...
+            if (joinChannels[i].deadlineTimestamp > block.timestamp) {
+                isValid[i] = true;
+                validCount++;
+            }
+        }
+
+        // 2. Filter by threshold signatures from existing participants
+
+        address[] memory allParticipants = StateChannelUtilLibrary.concatenateWithoutDuplicates(
+            getSnapshotParticipants(channelId), getPendingParticipants(channelId)
```
lukastanisic99, line 55: pendingParticipants will be here empty since the deposits are not yet processed. Imo since in the end all of this needs to be atomic (we can't do partial failure as I wanted), since the `dispute` commits to the specific `latestJoinChannelBlockHash` and the dispute happens in the same transaction so it has to assume all joins have ran successfully (if a single join fails the hash won't match and the dispute will fail before being committed). Best to verify the signatures at the end when all deposits are processed, since then it's a clean interface - just need to verify `snapshot.participants` + `pendingParticpants` that contains everyone - so we batch all the signatures together and just call verifyThreshold once in the end 

Ideally later we find a way where it doesn't have to be atomic and we can account for partial failures

contracts/V1/StateChannelDiamondProxy/AStateChannelManagerProxy.sol
```diff
@@ -352,4 +367,18 @@ abstract contract AStateChannelManagerProxy is StateChannelManagerInterface, Sta
         );
         return abi.decode(result, (bool, bytes));
     }
+
+    function multicall(bytes[] calldata calls) external override returns (bytes[] memory results) {
```
lukastanisic99, line 371: where is this used?

src/DisputeFactory.ts
```diff
@@ -0,0 +1,110 @@
...
+import { Storage } from "@/storage";
+import { AStateChannelManagerProxy } from "@typechain-types/index";
+
+export class DisputeFactory {
+    private readonly scmContract: AStateChannelManagerProxy;
+    private readonly storage: Storage;
+
+    constructor(scmContract: AStateChannelManagerProxy) {
+        this.scmContract = scmContract;
+        this.storage = Storage.getInstance();
```
lukastanisic99, line 14: will be storageManager

contracts/V1/StateChannelDiamondProxy/JoinChannelFacet.sol
daiagi: ok I guess i had a terminology misunderstanding

for you the pending participants are the ones that **just joined**

for me pending participants are from previous rounds of joining, and yes it would be empty since we just updated the state snapshot, which also clears the pending

for me the mental image was
- current snapshot participants
- pending (should be empty indeed)
- new comers

for you the image is 
- current snapshot participants
- new joiners => becoming the pending
- verify threshold with the pending aka the new joiners


still, for the sake of quickness and gas efficiency, isn't it better to run as many validation as possible early and fail early?

src/DisputeFactory.ts
daiagi: yes, it will be

src/DisputeFactory.ts
```diff
@@ -0,0 +1,110 @@
...
+                forkCnt: 0,
+                isForced: false,
+                previousBlockProducer: ethers.ZeroAddress,
+                previousBlockProducerPostedCalldata: false
+            },
+
+            selfRemoval: false
+        };
+
+        return dispute;
```
lukastanisic99, line 108: We should have a single function for populating/creating disputes - it's always a function of your 'state'/perception (let's call this dispute related state `evidence`) and there shouldn't be any special cases - if you have a valid joinChannel threshold, at creation it fetches it and populates the appropriate fields to fully reflect what your opinion is on what a `dispute` should be - as you exchange your opinion/evidence with others and as others exchange their opinion/evidence with you -> you all should converge on the same 'evidence state' (state replication) and reduce the same dispute

contracts/V1/StateChannelDiamondProxy/AStateChannelManagerProxy.sol
daiagi: JoinChannelSerive.ts, line 418
```ts

        // 6. Execute multicall

        try {
            const tx = await this.scmContract.multicall(calls);
            const receipt = await tx.wait();
  ```
  
  it is my attempt to create an atomic transaction. quite possibly there are better ways

lukastanisic99: Ok I'll think about it - Imo I was just thinking of a normal `function X` that does `A`,`B`,`C` and requires `A,B,C` to succeed. Simple and clean - this is cool but useful when we kinda have dynamic combinations of function calls and currently we don't or at least I don't see them - I think it's cheaper and easier to just call `function X`, but I like this idea in general

contracts/V1/StateChannelDiamondProxy/JoinChannelFacet.sol
lukastanisic99: Yes and yes, but the signature verification is deterministic - the tx submitter knows are the sigantures valid or not before submitting, what they don't know will the deposits be processed successfully so it's best to fail early on that and on 'time' (is it expired) if all of that succeeds, the rest (signature verification) will resolve deterministically - so for an 'honest' submitter verifying signatures last saves them gas on average, since that's not the part that will fail 

We can discuss today in the call

contracts/V1/StateChannelDiamondProxy/AStateChannelManagerProxy.sol
lukastanisic99: I like this idea a lot - we'll keep it - not necessarily for this since you want A,B,C to happen atomically and not have combinations A,C or C,B possible, just A,B,C, but for code that can have arbitrary combinations this is really good and saves on gas by making it a single function call.

daiagi:
@lukastanisic99  i think we can close this one no?
you used already what you needed  didn't you?

lukastanisic99:
> @lukastanisic99  i think we can close this one no?
> you used already what you needed  didn't you?

Just the solidity part - didn't touch the ts part

lukastanisic99:
Solidity is implemented, ts is missing - maybe RPC parts could be re-used

daiagi:
no longer relevant
too many changes happened since, both to protocol and typescript code
whatever useful code may be here will be easier to re-write rather then to mangle out of this PR
bah bye..

PR #84, opened 2025-06-07
Disputes CRDT - PR 1

lukastanisic99:
## 📋 Description

Reduces a single output given an array of disputes. The final outputSnapshot is calculated on the reducedOutput. 
We don't do much checks on the data within the reduce function, since reduce runs after auditing - if the data gets to reduce it's atomically correct and just 'reduce' it into a single output.

Could be refactored a bit into a more 'functional syntax', so it's a bit more readable, but not important atm and every section is commented.

## 💪 Related Issues

- Closes #issue-number

## ✅ Checklist

## 🔗 Follow-up Issues


contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
```diff
@@ -20,6 +20,100 @@ contract DisputeManagerFacet is StateChannelCommon {
...
+            StateProof memory stateProof = dispute.stateProof;
+            Block memory disputeLatestBlock = stateProof.signedBlocks.length > 0
+                ? abi.decode(stateProof.signedBlocks[stateProof.signedBlocks.length - 1].encodedBlock, (Block))
+                : abi.decode(
+                    stateProof.forkProof.forkMilestoneProofs[stateProof.forkProof.forkMilestoneProofs.length - 1]
+                        .blockConfirmations[stateProof.forkProof.forkMilestoneProofs[stateProof
+                        .forkProof
+                        .forkMilestoneProofs
+                        .length - 1].blockConfirmations.length - 1].signedBlock.encodedBlock,
+                    (Block)
```
daiagi: worth extracting to a function
image

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
```diff
@@ -20,6 +20,100 @@ contract DisputeManagerFacet is StateChannelCommon {
         emit DisputeCommited(encodedDispute, block.timestamp);
     }
 
+    function reduce(Dispute[] memory disputes, uint256 disputeWindowExpirationTimestamp)
```
daiagi: ok two things i don't quite get
1. slashParticipants: you are collecting both the already on chain ones and the new ones from the Dispute array. why? let me answer myself:
so the auditing dispute logic will know to ignore all of these slashed participants, and will not have to query the onChainSlashes storage slot again
correct?

2. reducedOutput.timeout;
keeping the shortest timeout, what does that mean?

daiagi (approved):

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
lukastanisic99: > ok two things i don't quite get
> 
> 1. slashParticipants: you are collecting both the already on chain ones and the new ones from the Dispute array. why? let me answer myself:
>    so the auditing dispute logic will know to ignore all of these slashed participants, and will not have to query the onChainSlashes storage slot again
>    correct?

In the previous dispute design, since you had a total order perception (you knew the previous i-1 disputes when you were creating the i-th dispute) you could also know on-chain slashes, so it was part of the race condition checks and auditing never 'queried' for the persisted on-chain slashes (that could changed), but used the one that was successfully committed in the dispute (that's immutable) so auditing would be fully deterministic regardless if on-chain slashes changed. 

In the new design since you do it in 'parallel' not knowing other adjacent disputes you don't know on-chain slashes at the time of creation, so it's not a hard requirement for them to match (the dispute one and the on-chain one) and the on-chain one can expand during the dispute window.

Here auditing still uses the one committed in disputes and doesn't care about the one in the storage slots, so it's still deterministic.

Where the storage slot one is used:
1) to check 'can you participate in the dispute game' while performing on-chain operations - if you're in the set the tx fails so you can not spam junk
2) At the end of the dispute window to reduce to a single output - that's why the 'storage slot' one is also expanded with timestamps - so this is also deterministic now (but the off-chain environment has to replicate the on-chain one, which we'll do regardless) - this way even if the on-chain one expands during the dispute window, reduce that runs later, will account for it. In most cases, after the dispute window expires on-chain slashes will contain all slashes, so the ones from fraud proofs don't make a difference. 

Not sure does this have gas saving side effects, but it wasn't designed to save gas, but to account for not knowing what the on-chain slashes are during creation + allowing it to expand during the challenge period and accounting for the expansion.

> 2. reducedOutput.timeout;
>    keeping the shortest timeout, what does that mean?

So all disputes that 'survive the challenge period' are correct and you timeout the earliest one - useful when people are colluding and try and timeout e.g. block height 100, but the peer being timedout isn't aware of block height 99 (didn't sign and the calldata is not posted), so they just submit a dispute that timesout block height 99 and reduce will take the earliest one. In other words if you try and timeout finalized one or one that has calldata posed on-chain, you'll be slashed and the dispute killed (so you can't timeout too early) and if you try and timeout too late, other can push to the earliest one possible - so you can't collude on this.

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
lukastanisic99: sure I'll put it in my notes and push it in the next PR - probably easier like that than to rebase all these PR's since they all build one on top of each other?

What are your thoughts on applying changes to these 'blockchain like' PR's @daiagi - threat them as immutable where fixes to a PR K happen in some PR N in the future and in the end they all collapse into a single atomic commitment or update PR K which forces all PRs > K to incorporate the changes too

daiagi: >What are your thoughts on applying changes to these 'blockchain like' PR's

for this case, clever idea, since you wanted to do large amount of work and allow easier reviewing

in general, a bit of re-inventing the wheel
can simply rebase with 'main' (or 'dispute' in our case)
every follow up pr after it's previous one has been merged

lukastanisic99: Yes this is simply for easier reviewing - in the end it's equivalent of just doing dispute/master <- PR N

PR #86, opened 2025-06-12
Disputes CRDT - PR 3

lukastanisic99:
## 📋 Description

DisputeOutput generation, upload and audit, challenge, kill dispute

## 💪 Related Issues

- Closes #issue-number

## ✅ Checklist

## 🔗 Follow-up Issues


contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
```diff
@@ -156,38 +168,35 @@ contract DisputeManagerFacet is StateChannelCommon {
...
 
         FraudProofVerificationContext memory poofContext = FraudProofVerificationContext({channelId: dispute.channelId});
-        (
-            bytes memory encodedModifiedState,
-            ExitChannelBlock memory exitBlock,
-            Balance memory totalDeposits,
-            Balance memory totalWithdrawals,
-            address[] memory slashes
-        ) = generateDisputeOutputState(
+        address[] memory slashes = _verifyFraudProofs(dispute.fraudProofs, poofContext);
```
```suggestion
        address[] memory slashes = _verifyFraudProofs(dispute.fraudProofs, proofContext);
```

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
```diff
@@ -156,38 +168,35 @@ contract DisputeManagerFacet is StateChannelCommon {
...
+        slashes = StateChannelUtilLibrary.concatAddressArrays(slashes, dispute.onChainSlashes);
```
daiagi: 1. back to my question from PR-2, are you querying here the onchainslashes?
2. you proabaly want `concatAddressArraysNoDuplicates`

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
```diff
@@ -200,45 +209,72 @@ contract DisputeManagerFacet is StateChannelCommon {
...
-            DisputePair memory disputePair = DisputePair(dispute.disputeIndex, disputeLength - 1);
-            disputeData[dispute.channelId].disputePairs.push(disputePair);
-            address[] memory returnedSlashParticipants = getOnChainSlashedParticipants(dispute.channelId);
-            emit DisputeChallengeResultWithDisputePair(
-                dispute.channelId, disputePair, success, returnedSlashParticipants
-            );
+            if (_canParticipateInDisputes(dispute.channelId, msg.sender)) {
+                addOnChainSlashedParticipants(dispute.channelId, msg.sender);
+            }
         }
```
daiagi: easier to read i think:

```suggestion
        if (success) {
            // auditing passed - dispute is correct, slash the challenger
            if (_canParticipateInDisputes(dispute.channelId, msg.sender)) {
                addOnChainSlashedParticipants(dispute.channelId, msg.sender);
            }
        } else {
            // auditing failed - dispute is invalid, kill it
            _killDispute(dispute, disputeCreationTimestamp);
        }
```

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
```diff
@@ -653,4 +689,24 @@ contract DisputeManagerFacet is StateChannelCommon {
...
+
+    function _calculateRemovals(Dispute memory dispute) internal view returns (address[] memory removals) {
+        uint256 removalCount = 0;
+        address[] memory _removals = new address[](2);
+        if (dispute.selfRemoval) {
+            _removals[removalCount++] = dispute.disputer;
+        }
+        if (
+            dispute.fraudProofs.length == 0 && dispute.onChainSlashes.length == 0
+                && dispute.timmeout.participant != address(0)
```
daiagi: typo `timmeout`

also worth adding a comment what this logic means

i read as:
"if the dispute is not 'pointing a finger at anyone' and somone is timedout
add the timedout particpants to removals"

why are these two conditions depend on eachother?

contracts/V1/StateChannelDiamondProxy/StateChannelCommon.sol
```diff
@@ -20,10 +20,8 @@ contract StateChannelCommon is StateChannelManagerStorage, StateChannelManagerEv
     }
 
     //This is executed only after sucessful auditing -> can safely add/insert participants without checking for duplicates (otherwise auditing would have failed)
-    function addOnChainSlashedParticipants(bytes32 channelId, address[] memory slashedParticipants) internal virtual {
-        for (uint256 i = 0; i < slashedParticipants.length; i++) {
-            disputeData[channelId].onChainSlashes.push(OnChainSlash(slashedParticipants[i], block.timestamp));
-        }
+    function addOnChainSlashedParticipants(bytes32 channelId, address slashedParticipant) internal virtual {
```
daiagi:
```suggestion
    function addOnChainSlashedParticipant(bytes32 channelId, address slashedParticipant) internal virtual {
```

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
```diff
@@ -200,45 +209,72 @@ contract DisputeManagerFacet is StateChannelCommon {
...
         }
     }
 
+    function _killDispute(Dispute memory dispute, uint256 disputeCreationTimestamp) internal {
+        DisputeData storage disputeData = disputeData[dispute.channelId];
+        DisputeWindow storage disputeWindow = disputeData.disputeWindowMap[dispute.genesisSnapshotDataHash];
+
+        // require that the dispute window exists and is not expired
+        require(
+            disputeWindow.creationTimestamp != 0
```
daiagi: i think this should be validated when creating the dispute


```solidity
            disputeWindow.creationTimestamp != 0
```

daiagi (changes requested):

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
lukastanisic99: I've updated this locally for my next PR - if you prefer updating specific PRs and cascading changes up, a bit more work for me, easier for you to track but I guess it's fine :D

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
lukastanisic99: 1. Nope - hopefully this has the answer https://github.com/peer3to/state-channels-plus/pull/84#discussion_r2156277953

2. Logically you're right - I'll change that just so it's more clear, even though the disputer has to populate `dispute.onChainSlashes` and `dispute.fraudProofs` so we can force him to not include duplicates (otherwise auditing fails when we actually try and apply the slashes)

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
lukastanisic99: Github formats this weird, sure I can flip the IF - I'll add it to my notes and will be applied here on in the next PR

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
lukastanisic99: It is, if you upload a dispute it's set

We can remove this since the other part of the 'if' `block.timestamp < disputeWindow.creationTimestamp + getChallengeTime()` will still fail since the default value for the timestamp is 0 and if we're not deploying to a 'just created/deployed' blockchain `block.timestamp` will always be greater than `getChallengeTime()`

I put it there more of explicit reasoning since that's the check 'is a dispute window created'

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
lukastanisic99: Yeah the type is already fixed in some future PR. 

As for the logic - for now we don't apply timeouts if there are slashes - mostly to prevent colluding. E.g a malicious peer 'tricks' the honest peer to build on a different fork/conflicting block (Block A), by performing a double sign and the peer is not aware of Block B and builds upon Block A and the byzantine peer tries to timeout the peer on Block B and when the peer becomes aware of Block B it constructs a `doubleSignFraudProof` and slashes the byzantine peer, which also has a side effect of not applying the timeout

So you can defer a timeout at a cost of a slash which has higher economic punishment than a timeout - later we can think how we can separate the two and maybe support multiple timeouts in parallel (not easy)

PR #87, opened 2025-06-13
Detect is own network  behind a symmetric NAT

daiagi:
First PR on this topic

## Add NAT Type Detection
**Purpose**: Detect if a peer is behind symmetric NAT to determine if direct P2P connections are possible.
### How it works:
When a device sends packets to external destinations, NAT assigns a public IP/port
**Cone NAT**: Uses the same external port for all destinations
**Symmetric NAT**: Uses different external ports for each destination
### Detection method:
1. Send WebRTC offers to multiple STUN servers
2. Each STUN server reports our external IP/port as it sees us
3. **Same ports across all servers** → Cone NAT (direct P2P possible)
4. **Different ports per server** → Symmetric NAT (need relay fallback)

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,8 @@
     "glob": "10.3.10",
     "hardhat": "^2.22.1",
     "hyperswarm": "4.3.5",
-    "prompt": "1.3.0"
+    "prompt": "1.3.0",
+    "stun": "^2.1.0"
   },
   "engines": {
     "yarn": ">=1.15"
--- a/package.json
+++ b/package.json
@@ -53,8 +53,7 @@
     "glob": "10.3.10",
     "hardhat": "^2.22.1",
     "hyperswarm": "4.3.5",
-    "prompt": "1.3.0",
-    "stun": "^2.1.0"
+    "prompt": "1.3.0"
   },
   "engines": {
     "yarn": ">=1.15"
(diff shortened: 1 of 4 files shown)


lukastanisic99:
Good idea - check if webRTC or holepunch have something we can reuse 

My concerns - we didn't account for IPv6 - if IPv6 is available the peer should be diable/holepunchable. Problem can occur if the other peer only uses IPv4 since that uses NAT.

Since we're supporting browsers as first class citizens, we will have public relayers that are used to establish the initial connection and upgrade to WebRTC if needed, but we have to have resource protection on them. They should only serve to negotiate a WebRTC/holepunch connection and not as actual relayers where they relay unbounded messages. So in cases when a direct p2p connection is not possible (can not be upgraded), we should be able to notify the client - they can run their own relayer or VPN

daiagi:
added support for both browsers and node
> check if webRTC or holepunch have something we can reuse

it is using WebRTC in browser case (see testSTUNBrowser function)

> we didn't account for IPv6

you are right, I've added that

lukastanisic99:
We have WebRTC already that's general (browsers and non browsers) - take a look and use that

daiagi:
assuming you mean `get-webrtc`
then thanks for the good tip. indeed it simplifies the code

removed `stun` dependncy

PR #88, opened 2025-06-14
Disputes CRDT - PR 4

lukastanisic99:
## 📋 Description

cleanup/refactor, commit and challenge reduce result

## 💪 Related Issues

- Closes #issue-number

## ✅ Checklist

## 🔗 Follow-up Issues


contracts/V1/DisputeTypes.sol
```diff
@@ -74,11 +73,22 @@ struct DisputeConfirmation {
...
+}
+
+struct DisputeWindowEvidence {
     uint256 creationTimestamp;
     bytes32[] disputeCommitments;
     mapping(address => bool) hasPosted; // inefficient, occupies a whole storage slot for a single bit - idealy we do a bitmask later as a f(participants) -> makes it also easy to delete the entire bitmask later. For now this is ok.
 }
 
+struct DisputeWindowReducedResult {
+    bytes32 reducedForkId;
```
daiagi: the fields name can be simply forkId, timestamp
calling it will be clear anyway
`disputeWindow,reducuedResult.timsestamp`

is already clear enough

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
```diff
@@ -25,24 +25,27 @@ contract DisputeManagerFacet is StateChannelCommon {
...
                 ErrorDisputeChallengePeriodExpired()
             );
-            require(!disputeWindow.hasPosted(dispute.disputer), ErrorDisputeAlreadyPosted());
+            require(!disputeWindow.evidence.hasPosted[dispute.disputer], ErrorDisputeAlreadyPosted());
         }
 
         if (isThresholdFinal) {
-            disputeWindow.creationTimestamp = block.timestamp - getChallengeTime() - 1;
-            delete disputeWindow.disputeCommitments; //delete all previous commitments
+            disputeWindow.evidence.creationTimestamp = block.timestamp - 2 * getChallengeTime() - 1;
```
daiagi: please move the magic numbers to a constant

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
```diff
@@ -709,4 +737,93 @@ contract DisputeManagerFacet is StateChannelCommon {
...
+        for (uint256 i = 0; i < disputeAuditingData.joinChannelBlocks.length; i++) {
+            if (previousJoinChannelBlockHash != disputeAuditingData.joinChannelBlocks[i].previousBlockHash) {
+                return false;
+            }
+            previousJoinChannelBlockHash = keccak256(abi.encode(disputeAuditingData.joinChannelBlocks[i]));
+        }
+        return previousJoinChannelBlockHash == disputeAuditingData.outputStateSnapshot.latestExitChannelBlockHash;
+    }
+
+    function _commitToDisputeReducedResult(
```
daiagi, line 848: i'm sure you have your reasons, but i find the name misleading
for me  commit means place hash on chain
but this function is validating and replacing fields on the struct

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
```diff
@@ -25,24 +25,27 @@ contract DisputeManagerFacet is StateChannelCommon {
...
         }
 
         if (isThresholdFinal) {
-            disputeWindow.creationTimestamp = block.timestamp - getChallengeTime() - 1;
-            delete disputeWindow.disputeCommitments; //delete all previous commitments
+            disputeWindow.evidence.creationTimestamp = block.timestamp - 2 * getChallengeTime() - 1;
+            delete disputeWindow.evidence.disputeCommitments; //delete all previous commitments
+            _commitToDisputeReducedResult(
+                disputeWindow, dispute.outputSnapshotDataHash, block.timestamp - getChallengeTime() - 1
+            );
```
daiagi: this logic is pretty opaque
some comment explaining the rational would be appreciated

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
```diff
@@ -227,22 +273,96 @@ contract DisputeManagerFacet is StateChannelCommon {
         }
     }
```
daiagi, line 212:
```suggestion
    /**
     * @notice Challenges a dispute reduction by providing disputes and verification data
     * @dev IMPORTANT: The disputes array must be provided in the same order as they were committed 
     *      to the dispute window. The off-chain client is responsible for ensuring disputes are 
     *      ordered correctly to save on gas during verification.
     */
```

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
```diff
@@ -96,16 +107,7 @@ contract DisputeManagerFacet is StateChannelCommon {
...
-                ? abi.decode(stateProof.signedBlocks[stateProof.signedBlocks.length - 1].encodedBlock, (Block))
-                : abi.decode(
-                    stateProof.forkProof.forkMilestoneProofs[stateProof.forkProof.forkMilestoneProofs.length - 1]
-                        .blockConfirmations[stateProof.forkProof.forkMilestoneProofs[stateProof
-                        .forkProof
-                        .forkMilestoneProofs
-                        .length - 1].blockConfirmations.length - 1].signedBlock.encodedBlock,
-                    (Block)
-                );
+            Block memory disputeLatestBlock = _getLatestBlock(stateProof);
```
daiagi, line 89: 😃

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
```diff
@@ -227,22 +273,96 @@ contract DisputeManagerFacet is StateChannelCommon {
...
+        DisputeData storage disputeData = disputeData[channelId];
+        DisputeWindow storage disputeWindow = disputeData.disputeWindowMap[disputes[0].genesisSnapshotDataHash];
+
+        //rquire all disputes are part of commitment
+
+        require(
+            areDisputesCommitted(disputeWindow, disputes, disputeCreationTimestamps),
+            ErrorDisputeCommitmentNotAvailable()
+        );
+        //require reduce challenge period is not expired - this also ashures it's commited
```
```suggestion
        //require reduce challenge period is not expired - this also assures it's commited
```

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
```diff
@@ -227,22 +273,96 @@ contract DisputeManagerFacet is StateChannelCommon {
...
+        SnapshotData memory snapshotData =
+            reduceOutputToSnapshotData(reducedOutput, stateSnapshot, encodedStateMachineState, joinChannelBlocks);
+
+        bytes32 winingForkId = keccak256(abi.encode(snapshotData));
+
+        if (winingForkId != disputeWindow.reducedResult.reducedForkId) {
+            addOnChainSlashedParticipants(channelId, disputeWindow.reducedResult.reducer);
+            disputeWindow.reducedResult.reducedForkId = bytes32(0); // unset
+            _commitToDisputeReducedResult(disputeWindow, winingForkId, block.timestamp - getChallengeTime() - 1);
+        } else {
```
daiagi, line 267: im guessing the `-1` is because somehwere the expiration check uses `<`
could use `<=` instead and remove the `-1`

just a guess

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
```diff
@@ -709,4 +737,93 @@ contract DisputeManagerFacet is StateChannelCommon {
...
+        //check joinChannelBlocks (linked to latestSateSnapshot, chained internally and outputStateSnapshot commits to the head)
+        bytes32 previousJoinChannelBlockHash = disputeAuditingData.latestStateSnapshot.latestJoinChannelBlockHash;
+        for (uint256 i = 0; i < disputeAuditingData.joinChannelBlocks.length; i++) {
+            if (previousJoinChannelBlockHash != disputeAuditingData.joinChannelBlocks[i].previousBlockHash) {
+                return false;
+            }
+            previousJoinChannelBlockHash = keccak256(abi.encode(disputeAuditingData.joinChannelBlocks[i]));
+        }
+        return previousJoinChannelBlockHash == disputeAuditingData.outputStateSnapshot.latestExitChannelBlockHash;
+    }
```
```suggestion
    function _isCorrectAuditingData(
        Dispute memory dispute,
        DisputeAuditingData memory disputeAuditingData
    ) internal view returns (bool) {
        //check dispute commits to disputeData
    require(
        dispute.disputeAuditingDataHash == keccak256(abi.encode(disputeAuditingData)),
        ErrorDisputeWrongAuditingDataHash()
    );
    
        //check dispute commits to genesisStateSnapshot
    require(
        dispute.genesisSnapshotDataHash == 
        keccak256(abi.encode(disputeAuditingData.genesisStateSnapshot.snapshotData)),
        ErrorDisputeGenesisSnapshotMismatch()
    );
        //check latestStateSnapshot
    require(
        dispute.latestStateSnapshotHash == 
        keccak256(abi.encode(disputeAuditingData.latestStateSnapshot)),
        ErrorDisputeLatestStateSnapshotMismatch()
    );
        //check outputStateSnapshot
    require(
        dispute.outputSnapshotDataHash == 
        keccak256(abi.encode(disputeAuditingData.outputStateSnapshot.snapshotData)),
        ErrorDisputeOutputStateSnapshotMismatch()
    );

        //check latestStateStateMachineState
    require(
        disputeAuditingData.latestStateSnapshot.stateMachineStateHash == 
        keccak256(disputeAuditingData.latestStateStateMachineState),
        ErrorDisputeStateMachineStateMismatch()
    );

        //check joinChannelBlocks (linked to latestSateSnapshot, chained internally and outputStateSnapshot commits to the head)
        bytes32 previousJoinChannelBlockHash = disputeAuditingData
            .latestStateSnapshot
            .latestJoinChannelBlockHash;
        for (
            uint256 i = 0;
            i < disputeAuditingData.joinChannelBlocks.length;
            i++
        ) {
        require(
            previousJoinChannelBlockHash == disputeAuditingData.joinChannelBlocks[i].previousBlockHash,
            ErrorDisputeJoinChannelBlocksNotLinked()
        );
            previousJoinChannelBlockHash = keccak256(
                abi.encode(disputeAuditingData.joinChannelBlocks[i])
            );
        }
    require(
        previousJoinChannelBlockHash == 
        disputeAuditingData.outputStateSnapshot.latestExitChannelBlockHash,
        ErrorDisputeJoinChannelBlocksEndMismatch()
    );
    }
```

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
```diff
@@ -709,4 +737,93 @@ contract DisputeManagerFacet is StateChannelCommon {
...
+        disputeWindow.reducedResult.reducer = msg.sender; //calling function should check that msg.sender is part of channel 'can participate'
+    }
+
+    function _isEvedincePeriodExpired(DisputeWindow storage disputeWindow) internal view returns (bool) {
+        return block.timestamp > disputeWindow.evidence.creationTimestamp + getChallengeTime();
+    }
+
+    function _isKillPeriodExpired(DisputeWindow storage disputeWindow) internal view returns (bool) {
+        return block.timestamp > disputeWindow.evidence.creationTimestamp + 2 * getChallengeTime();
+    }
```
daiagi, line 868: could this have a name?
`killingTime` maybe?

daiagi (changes requested):

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
lukastanisic99: yeah I was just lazy with this one - essentially the challenge period has an `evidence submission time` and `kill time` where kill time > evidence submission time or twice the amount - It's already in my notes to add a field for it - will push with next PR

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
lukastanisic99: Yeah the forkId is a hash already that commits to the genesis and you're committing to that forkId in the specific `DisputeWindow` and you may have some time to challenge it depending on the `reductionTimestamp` that's provided and you also persist the msg.sender to know who to slash if the commitment is wrong 

It could also be commitment = hash(forkId, timestamp, msg.sender) and emit the event - You'd save 2 storage slots, but would have to upload/oracleize these things which would have a more clunky interface elsewhere which for now imo is not worth the fixed 2 storage slot savings + when the snapshot is updated to the latest one all of these slots are free creating a gas refund that gives an additional incentive to the snapshot submitter to submit the latest one

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
lukastanisic99: Added to my notes and will add comments in the next PR - in short if a N/N threshold is reached on a dispute, that `DisputeWindow` immediately concludes with the submitted dispute being the outcome. Side effect - delete all the storage slots of previous dispute commitments in this `DisputeWindow` (free up storage). The part `block.timestamp - getChallengeTime() - 1` just sets an old timestamp so when you call is `_isReduceChallengePeriodExpired` it returns true immediately (meaning final)

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
lukastanisic99: Will add this - in reality they don't have to be provided in the same order, but then it's O(n^2) instead of O(n) and I was lazy to write O(n^2) algo + better to do the ordering in JS.

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
lukastanisic99: Honestly didn't think too much about are we going to use `<` or `<=` (allow the same block) so I put -1 which for sure is in the past so it works regardless :D

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
lukastanisic99: Yes it will - I was just lazy to do it then and added it to my notes and just used `challengeTime` and `2*challengeTime` for now :D

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
lukastanisic99: Will take a look at this later - github UI just shows a blob of red and a blob of green code not making it easy to see/deduct the suggested changes

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
daiagi: it is better, no  doubt
just need to be clear of this "hidden" requirement

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
daiagi: simple
i am suggesting to use `require` instead of 
```js
if(cond) {
return false;
}

```

lukastanisic99: Ah ok - Imo it's better to not use require so it can be reused, since returning false doesn't imply that the transaction reverts and with require it does - the calling function can chose does it want to do `requrire(_isCorrectAuditingData(...))` or something like `if(!_isCorrectAuditingData(...)) slash`

PR #90, opened 2025-06-20
Disputes CRDT - PR 6

lukastanisic99:
Split fraud proofs into `normal fraud proofs` and `dispute fraud proofs`. Dispute fraud proofs are similar to 'normal' ones, but they always target an already committed dispute, making them 'stateful', while the 'normal' ones are stateless. Logically and conceptually they are similar/the same. Provide an on-chain interface for just applying the dispute fraud proofs (without submitting new disputes) in order to kill malicious disputes already submitted during the `evidence` collection phase. 

Updated the solidity compiler to the latest one 0.8.30.

Removed `forge git submodules` and added it to .gitingore - not sure why it wasn't done like that in the first place ?! 

Solidity `data types` are refactored into more files. 

Applied review suggestions/comments from previous N-1 PRs related to `Dispute CRDTs` 

Only 'major' thing that's missing is joining the channel and actually populating the correct fraud proofs - probably some minor stuff/checks are needed. 

Code could be refactored more in some places - will do that in the end.


contracts/V1/StateChannelDiamondProxy/AStateChannelManagerProxy.sol
```diff
@@ -230,6 +235,16 @@ abstract contract AStateChannelManagerProxy is StateChannelManagerInterface, Sta
...
 
+    function verifyDisputeFraudProofs(DisputeFraudProof[] memory disputeFraudProofs)
+        public
+        returns (bytes memory maliciousDisputesEncoded)
+    {
+        return _delegatecall(
+            address(disputeFraudProofFacet),
+            abi.encodeCall(disputeFraudProofFacet.verifyDisputeFraudProofs, (disputeFraudProofs))
+        );
+    }
```
lukastanisic99, line 224: @daiagi take a look at this function - It should work, but I spent the entire day yesterday fighting with `stack too deep` and still not sure why - like this it works. Originally it returned directly `Dispute[ ] memory maliciousDisputes` instead of the encoded version of the same `bytes memory maliciousDisputesEncoded`. Doing `abi.decode(maliciousDisputesEncoded, (Dispute[ ]))` within the function results in `stack too deep` during compiling (regardless how you do it). If done in other functions it works. I've read over the docs and chatted with GPT and updated the compiler and used both hardhat and forge to compile and always get the same error. The EVM has a stack size of 1024, but you can only access the top 16 stack elements and I don't see how this function uses more than 16 local variables - all the structures should be in memory with a single pointer located on the stack that points to them in memory.

For some reason `public` functions use more stack variables than `internal` functions, which GPT explains is due to public functions being part of the ABI and having to be encoded, but the logic still doesn't add up and other public functions essentially use more variables and larger structs and still work. Plus in the solidity docs it says that the calldata is copied to memory, so it should still be a single pointer that's stored on the stack. 

Don't waste too much time on this , it's not important since I did a workaround - I just returned bytes and converted it somewhere else, but can wrap my head around and understand why

contracts/V1/StateChannelDiamondProxy/StateChannelCommon.sol
```diff
@@ -212,4 +225,18 @@ contract StateChannelCommon is StateChannelManagerStorage, StateChannelManagerEv
...
     }
+
+    function isDisputeCommitted(Dispute memory dispute) internal view returns (bool) {
+        bytes32 channelId = dispute.channelId;
+        DisputeData storage disputeData = disputeData[channelId];
+        DisputeWindow storage disputeWindow = disputeData.disputeWindowMap[dispute.genesisSnapshotDataHash];
+        bytes32 commitment = keccak256(abi.encode(dispute));
+
+        for (uint256 i = 0; i < disputeWindow.evidence.disputeCommitments.length; i++) {
+            if (disputeWindow.evidence.disputeCommitments[i] != commitment) {
```
daiagi: typo?

don't you mean `==` ?

```suggestion
            if (disputeWindow.evidence.disputeCommitments[i] == commitment) {
```

contracts/V1/StateChannelDiamondProxy/StateChannelCommon.sol
```diff
@@ -73,20 +73,24 @@ contract StateChannelCommon is StateChannelManagerStorage, StateChannelManagerEv
...
 
-    function getAllTimes() public view virtual returns (uint256, uint256, uint256, uint256) {
-        return (p2pTime, agreementTime, chainFallbackTime, challengeTime);
+    function getAllTimes() public view virtual returns (uint256, uint256, uint256, uint256, uint256) {
+        return (p2pTime, agreementTime, chainFallbackTime, evidenceTime, killTime);
     }
 
     function _isReduceChallengePeriodExpired(DisputeWindow storage disputeWindow) internal view returns (bool) {
-        return block.timestamp > disputeWindow.reducedResult.reductionTimestamp + getChallengeTime();
+        return block.timestamp > disputeWindow.reducedResult.timestamp + getEvidenceTime();
```
daiagi:
```suggestion
        return block.timestamp > disputeWindow.reducedResult.timestamp + evidenceTime;
```

contracts/V1/StateChannelDiamondProxy/StateChannelManagerStorage.sol
```diff
@@ -1,15 +1,18 @@
...
 
 contract StateChannelManagerStorage {
     //Config shared across all instances of the state machine
     uint256 public p2pTime;
     uint256 public agreementTime;
     uint256 public chainFallbackTime;
-    uint256 public challengeTime;
+    // Time within more dispute can be submitted during the challenge period
+    uint256 public evidenceTime;
+    /// @dev Time within the dispute cam be killed during the challenge period
```
daiagi: typo:

```suggestion
    /// @dev Time within the dispute can be killed during the challenge period
```

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
```diff
@@ -31,17 +31,20 @@ contract DisputeManagerFacet is StateChannelCommon {
...
                 ErrorDisputeChallengePeriodExpired()
             );
             require(!disputeWindow.evidence.hasPosted[dispute.disputer], ErrorDisputeAlreadyPosted());
         }
 
         if (isThresholdFinal) {
-            disputeWindow.evidence.creationTimestamp = block.timestamp - 2 * getChallengeTime() - 1;
-            delete disputeWindow.evidence.disputeCommitments; //delete all previous commitments
+            //finalize the dispute windown by making the kill period expired
+            disputeWindow.evidence.creationTimestamp = block.timestamp - getKillTime() - 1;
```
daiagi: Not a fan of this.
I would prefer explict boolean feild `isFinalized` on `disputeWindow.evidence`  rather then changing the creationTimestamp

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
```diff
@@ -736,11 +752,14 @@ contract DisputeManagerFacet is StateChannelCommon {
     }
 
     function _calculateRemovals(Dispute memory dispute) internal view returns (address[] memory removals) {
+        //Try and combine timeout and selfRemova -> max 2 removals per dispute
```
```suggestion
        //Try and combine timeout and selfRemoval -> max 2 removals per dispute
```

contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol
```diff
@@ -0,0 +1,190 @@
...
+    function _handleDisputeInvalidPreviousRecursive(bytes memory encodedFraudProof, Dispute memory dispute)
+        internal
+        view
+        returns (address)
+    {
+        //TODO
+        return dispute.disputer;
+    }
+
+    // function _handleTimeoutThreshold(
```
daiagi, line 239: why keep as comments?
just delete it

daiagi (changes requested):
small stuff

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
lukastanisic99: Easy to add, but the specific field is `not auto-updated` when the timestamp expires - this gives a single and easy way to have a `f(block.timestamp)` function that determines is it finalized or not. In all other cases the boolean flag would be set to false so you'd still check the timestamp eventually + in this specific case (the only one where the flag would be set to true) you also instantly set `disputeWindow.reducedResult` - Imo it's not a big cognitive load to think `isFinal(f(timestamp))` which has nice `auto-update` properties in solidity

contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol
lukastanisic99: So I can populate them correctly in PR 9 :D

contracts/V1/StateChannelDiamondProxy/StateChannelCommon.sol
lukastanisic99: yes - changed locally PR 9 will reflect

PR #91, opened 2025-06-26
Disputes CRDT - PR 7

lukastanisic99:
Simplified `JoinChannel` logic. Hopefully this concludes all architectural stuff behind the CRDT implementation and now only some cleanup/refactor and fraud proof population is left and typescript can advance in parallel.

JoinChannel flow/logic explanation: 

1) The new participant connects to any/all participant(s).
2) At least 1 existing participant, shares all the required state data / makes it available to the `joining participant` over the p2p network (RPC). Essentially the joining peer connects as a `spectator` - If the data does not become available/shared, the joining peer `aborts`. 
3) The joining peer validates that data and `syncs` - here it can subjectively perform the `invariant check` and decide does it want to join the channel. It continues replicating state as the new state transitions are shared with it. It also cosigns and gossips blocks like it was already part of the channel - indicating cooperation. The peer can `abort` at any step at its own discretion without affecting any existing peers. 
4) If the peer wants to join, it sends a `JoinChannel` request over the p2p network (RPC) indicating it wants to join the channel.
5) Existing peers analyze the request and if they agree with it (want to allow the peer to join), they cosign it and gossip it over the p2p network. All of the signing is done async in parallel without affecting other existing operations (e.g. their state replication/progression). If for any reason the joining is `aborted` the existing peers remain unaffected like it never was initiated.
6) When/If a N/N threshold is reached on the `JoinChannel` the joining peer can at its own discretion initiate the joining `on-chain` by calling `StateChannelManagerProxy.joinChannel` and providing the `JoinChannelConfirmation` proving that existing peers agreed to let the new peer in. This performs all the lockups and deposits of funds. If it fails, the joining is aborted without affecting other peers. At its own discretion the peer has a `multicall` endpoint that allows it to `update the snapshot (optional)` if needed prior (only useful to reduce the participant set for the N/N is some cases, since the peer at this point has all the data, otherwise it wouldn't be at this step, so it's not needed for security) and `upload a dispute (optional)` that triggers its joining immediately in the `StateMachine` - this is optional since any future dispute will trigger the join through `reduce` and it will be accounted for in the state machine. In most cases peers will upload a dispute immediately and not delay their own joining.
7) Uploading the dispute creates the `dispute window` for submitting more evidence and `playing out the dispute game` - this is the trigger for other peers that the join operation happened successfully on-chain (N/N and deposits) and that it `can't be aborted` anymore. Peers, like with other disputes, stop their state progression for the disputed `forkId` and try and submit the latest `evidence` in most cases they just submit the latest state, so the final `reduction` accounts for the latest state when applying the join.
8) The resulting fork contains the joining peers as participants.

`Game Theory` behind this design: 
Up until the last point the joining operation can abort randomly without affecting any peers. It's in everyone's best interest to cooperate or abort early. The original peers should share the full/latest state data with all joining peers, so when the joining peer actually performs the join operation and creates the `dispute` it uses the `latest state` which will save the existing peers gas fees, by potentially not having to upload more `evidence` during the `dispute window`. Since the cost of joining are always passed on to the joining peer, its not in their interest to perform double spend on their deposits, since their transactions will fail and they'll just pay gas fees on-chain without affecting the existing participant set.  

Continuation on point 7) above - Existing peer can during the `dispute window` submit more `evidence` in most cases used to push the state to the `latest` and they will try to get signatures from all pending participants, but in the case that they don't get those signatures they're forced to either accept the older state or upload the `AuditingData` as calldata so it's verifiable. Joining peers may not be synced to what's actually the `latest state` since existing peers can withhold that data, but in such cases they gain no advantage since they'll be forced to pay gas fees on-chain and the data will eventually be provided and made available or forfeited - so it's better for them to do it for free over the p2p network than on-chain through calldata where fees have to be paid. The joining peer, since it also cosigns blocks even when `spectating` (`step 3`) (otherwise other peers don't cosign its `JoinChannel` request (`step 5`)) it is also forced to use `the latest state provided to it` during the initial `dispute` or like other peers a `dispute fraud proof` for enforcing `latest signed state` will slash the joining peer and kill the dispute.

### DIFF
--- a/contracts/V1/StateChannelDiamondProxy/JoinChannelFacet.sol
+++ b/contracts/V1/StateChannelDiamondProxy/JoinChannelFacet.sol
@@ -37,12 +37,6 @@ contract JoinChannelFacet is StateChannelCommon {
         // Deposit funds
         isValid = _processJoinChannelDeposits(jc);
         require(isValid, ErrorJoinChannelFailed());
-
-        // Create and apply JoinChannelBlock
-        _insertJoinChannelBlock(jc);
-
-        // Uupdate pendingParticipants
-        _updatePendingParticipants(jc);
     }
 
     // ############### Processing Functions ###############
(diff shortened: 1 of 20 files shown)


contracts/V1/StateChannelDiamondProxy/JoinChannelFacet.sol
```diff
@@ -0,0 +1,93 @@
...
+
+        // Process the deposit for the specific StateChannelManager
+        success = AStateChannelManagerProxy(address(this)).addParticipantComposable(jc);
+
+        // Update on-chain total deposits
+        if (success) {
+            totalOnChainProcessedDeposits[channelId] =
+                stateMachineImplementation.addBalance(totalOnChainProcessedDeposits[channelId], jc.balance);
+        }
+        return success;
```
daiagi:
```suggestion
        // Process the deposit for the specific StateChannelManager
        if (!AStateChannelManagerProxy(address(this)).addParticipantComposable(jc)) {
            return false;
        }
        // Update on-chain total deposits
        totalOnChainProcessedDeposits[channelId] =
            stateMachineImplementation.addBalance(totalOnChainProcessedDeposits[channelId], jc.balance);
        
        return true;
```

daiagi (approved):

contracts/V1/StateChannelDiamondProxy/JoinChannelFacet.sol
lukastanisic99: PR8 should already have this change

PR #92, opened 2025-06-28
Disputes CRDT - PR 8

lukastanisic99:
Simplified deposits/withdrawals and invariant check - more cleanup and refactor

### DIFF
--- a/contracts/V1/StateChannelDiamondProxy/AStateChannelManagerProxy.sol
+++ b/contracts/V1/StateChannelDiamondProxy/AStateChannelManagerProxy.sol
@@ -40,56 +40,6 @@ abstract contract AStateChannelManagerProxy is StateChannelManagerInterface, Sta
 
     // ********** public/external functions **********
 
-    function removeParticipantComposable(bytes32 channelId, ExitChannel memory exitChannel)
-        public
-        onlySelf
-        returns (bool)
-    {
-        return _removeParticipantComposable(channelId, exitChannel);
-    }
-
-    function applyJoinChannelToStateMachine(bytes memory encodedState, JoinChannel[] memory joinCahnnels)
-        public
-        override
-        onlySelf
-        returns (bytes memory encodedModifiedState)
-    {
-        return _applyJoinChannelToStateMachine(encodedState, joinCahnnels);
-    }
-
-    function applySlashesToStateMachine(bytes memory encodedState, address[] memory slashedParticipants)
-        public
-        onlySelf
-        returns (bytes memory encodedModifiedState, ExitChannel[] memory)
-    {
-        return _applySlashesToStateMachine(encodedState, slashedParticipants);
-    }
-
-    function removeParticipantsFromStateMachine(bytes memory encodedState, address[] memory participants)
-        public
-        onlySelf
-        returns (bytes memory encodedModifiedState, ExitChannel[] memory)
-    {
-        return _removeParticipantsFromStateMachine(encodedState, participants);
-    }
-
-    function addParticipantComposable(JoinChannel memory joinChannel) public onlySelf returns (bool) {
-        return _addParticipantComposable(joinChannel);
-    }
-
-    function executeStateTransitionOnState(bytes32 channelId, bytes memory encodedState, Transaction memory _tx)
-        public
-        override
-        returns (bool, bytes memory)
-    {
-        //channelId not used currenlty since all channels have the same SM - later they can be mapped to different ones
-        stateMachineImplementation.setState(encodedState);
-        (bool success,) =
-            address(stateMachineImplementation).call(abi.encodeCall(stateMachineImplementation.stateTransition, _tx));
-        return (success, stateMachineImplementation.getState());
-        // return (success, abi.decode(encodedReturnValue, (bytes)));
-    }
-
     /**
      * Posting calldata is lightweight, since it persists a signle hash/commitment.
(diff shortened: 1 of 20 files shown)


contracts/V1/StateChannelDiamondProxy/StateSnapshotFacet.sol
```diff
@@ -59,14 +59,14 @@ contract StateSnapshotFacet is StateChannelCommon {
         ExitChannelBlock[] memory exitChannelBlocks
     ) internal {
         _validateExitChannelBlocks(exitChannelBlocks, currentOnChainSnapshot, newSnapshot);
-        _applyExitChannelBlocks(channelId, exitChannelBlocks);
+        _applyExitChannelBlocks(channelId, exitChannelBlocks, newSnapshot.snapshotData.latestJoinChannelBlockHash);
```
daiagi, line 62: a bit strange to pass `latestJoinChannelBlockHash` into `applyExitChannelBlocks`

how about pass `totalDeposits` directly instead?

or do the validation before? outside of `apply` function

contracts/V1/StateChannelDiamondProxy/AStateChannelManagerProxy.sol
```diff
@@ -227,6 +113,78 @@ abstract contract AStateChannelManagerProxy is StateChannelManagerInterface, Sta
...
+
+    function removeParticipantsFromStateMachine(bytes memory encodedState, address[] memory participants)
+        public
+        onlySelf
+        returns (bytes memory encodedModifiedState, ExitChannel[] memory)
+    {
+        return _removeParticipantsFromStateMachine(encodedState, participants);
+    }
+
+    function executeStateTransitionOnState(bytes32 channelId, bytes memory encodedState, Transaction memory _tx)
```
daiagi: state transition is always on state and it's clear enough from the function signature


```suggestion
    function executeStateTransition(bytes32 channelId, bytes memory encodedState, Transaction memory _tx)
```

contracts/V1/StateChannelDiamondProxy/AStateChannelManagerProxy.sol
```diff
@@ -385,4 +303,56 @@ abstract contract AStateChannelManagerProxy is StateChannelManagerInterface, Sta
...
+        internal
+        override
+        returns (bytes memory encodedModifiedState, ExitChannel[] memory exitChannels)
+    {
+        exitChannels = new ExitChannel[](slashedParticipants.length);
+        stateMachineImplementation.setState(encodedState);
+        for (uint256 i = 0; i < slashedParticipants.length; i++) {
+            bool success;
+            (success, exitChannels[i]) = stateMachineImplementation.slashParticipant(slashedParticipants[i]);
+            // require(success, "Slash failed");
```
daiagi:
```suggestion

```

daiagi (approved):
small stuff

contracts/V1/StateChannelDiamondProxy/StateSnapshotFacet.sol
lukastanisic99: Yeah it's strange and not intuitive - The validation needs to be done inside since it's done for each `exitChannel` as it's applied - `totalDeposits` can be passed, that would work, but I don't like that in the main function or if it gets reused I have to always fetch it from the `nested map` and this way I pass the `key` and fetch it inside while already fetching adjacent fields - maybe change the name to `key???`

Long answer:

The root problem / the ugly code here comes from the nested map how we track deposits which is a byproduct of the CRDT design - previously we tracked just `totalDeposits`, same way as we track `totalWithdrawals` a single field. That was enough since everything was totally ordered and more importantly you had to have full knowledge of the past when you were doing something in the present, which is not the case here. 

In general the mental model for our CRDT is that it's still a 'blockchain' where 'blocks' (`disputeWindow`) are totally ordered, but what's contained within the dispute window can happen in parallel and eventually it too becomes part of a higher total order structure, since the ordering of disputes within the window doesn't matter. 

Taking that into account, while applying the join / interacting with the `disputeWindow` we don't have `perfect` information, so we anchor / milestone to parts where we have perfect information (genesisState of each fork) since it was derived from `reduce` after the `parallel window` expires/concludes giving everyone perfect information on the outcome. Still deposits happen during the window and we let the blockchain order them and progress their state (build the on-chain blockchain or what I call the `nested map`) and eventually you become aware of it. So in most cases you're calculating on some 'checkpoint' in the past and eventually applying the total order given by the blockchain. This is mostly needed for the invariant check. So instead of checking the invariant on the 'latest deposit' you check it on the 'latest checkpoint' and if that is correct, the eventual result will also be correct since the dispute game forces all inputs to be correct so everything 'correct' gets reduced eventually. It has some other nice properties too, but I guess for now this is enough.

So what we'll do here to make the code 'nicer' I'm not sure, since there isn't any 'elegant' solution and if we get rid of the nested map this will also go away as a byproduct, but need to finish other stuff before thinking about this + for now this code is called just here

lukastanisic99:
Most suggestions I've already applied locally and will push with PR9

contracts/V1/StateChannelDiamondProxy/StateSnapshotFacet.sol
daiagi: not a big deal
it can remain as is

PR #93, opened 2025-06-29
add a wasm precompile

daiagi:
tests example of a simple add(a,b) wasm precompile

example can be used to "plug-in" the actual wasm blob pretty straight forward

had to update some dependencies


daiagi:
added a tiny bit more complex example where the wasm expose more then 1 function

there is a choice
a. entire wasm under a single address / pre-compile => requires function selector
b. each function in a separate address


for this example option a made more sense to me
for the real pre-compile of Jeff's crypto we may choose otherwise

main benefit of the second approach is cleaner functions signautre

src/evm/CustomEVM.ts
```diff
@@ -0,0 +1,22 @@
...
+            chain: Mainnet,
+            hardfork: Hardfork.Prague
+        });
+
+        const mathPrecompile = await createMathPrecompile();
+        const addPrecompile = await createAddPrecompile();
+
+        const evm = await createEVM({
+            common,
+            customPrecompiles: [mathPrecompile, addPrecompile]
```
lukastanisic99: Can we parametrize this so we can reuse this in our EVM and just pass a list of precompiles and expose it during `setup` so everyone just plug in their own without having to change the actual source code

src/evm/precompiles/add.wasm
lukastanisic99, line 1: How is this generated? From rust, typescript or? Maybe include the source code for this also

src/evm/CustomEVM.ts
daiagi: who is everyone?
i thought this is meant only for us to add jeff's crypto

src/evm/precompiles/add.wasm
daiagi: rust

daiagi: https://github.com/peer3to/math-precompile-example

src/evm/CustomEVM.ts
lukastanisic99: That's why we're interested and exploring this, yes. Jeff's precompiles are for poker and the SDK should be general, so if someone needs custom precompiles they should be able to use them easily - they may or may not need Jeff's precompile and his precompiles are not part of the SDK - if you're not doing poker or card games you probably don't need them. 

So in the Poker repo we'll have something like `setup(....., [jeff's precompiles])`
In the TicTacToe repo we just have `setup(....)`

src/evm/precompiles/add-precompile.ts
```diff
@@ -0,0 +1,55 @@
...
+import { to32Bytes } from "./util";
+
+export const ADD_PRECOMPILE_ADDRESS = createAddressFromString(
+    "0x0000000000000000000000000000000000000123"
+);
+
+export async function initAddWasm(): Promise<Wasm> {
+    const wasmPath = path.resolve(__dirname, "./add.wasm");
+    try {
+        const wasmSource = fs.readFileSync(wasmPath);
```
lukastanisic99: This should work for browsers too - I assume we have to use `fetch()` or something

src/evm/CustomEVM.ts
daiagi: ok. ill see what i can do

src/evm/precompiles/add-precompile.ts
daiagi: good point. thanks

src/evm/CustomEVM.ts
lukastanisic99: Imo I don't see a problem of just adding arguments to `create()` -> `create(precompiles)` and funneling it up to the `setup` we use for the SDK

src/evm/precompiles/add-precompile.ts
```diff
@@ -0,0 +1,55 @@
...
+
+            const result = add(
+                Number(a & 0xffffffffn),
+                Number(b & 0xffffffffn)
+            );
+
+            return {
+                returnValue: to32Bytes(result),
+                gasUsed: 1000n,
+                executionGasUsed: 1000n
```
lukastanisic99: What's the difference between `gasUsed` and `executionGasUsed`

src/evm/precompiles/math-precompile.ts
```diff
@@ -0,0 +1,74 @@
...
+            const selector = Number(
+                BigInt("0x" + Buffer.from(data.slice(0, 32)).toString("hex"))
+            );
+
+            // Map selector to WASM function name
+            const funcName = {
+                0: "add",
+                1: "multiply",
+                2: "divide"
+            }[selector];
```
lukastanisic99: Not sure we can do this or at least not sure we can do this here since the precompile that runs `on-chain` doesn't execute any `.ts` code to deduct the selector. I think every precompile for the EVM has to have a single entry point and if you want to do these tricks with selectors you'd have to do them in rust by interpreting the `bytes` provided and calling the appropriate function

daiagi: you are mixing thing up i think
this `Precompile` conforms to the type `CustomPrecompile` from ethereumjs evm

it's not a trick, the precompile supplied to the customEVM is typescipt function


or ask it another way
if this is a trick then  all ts code is a trick. what should the function field in `CustomPrecompile` should contain?

and why do the tests work?

idk. maybe i am missing something 

https://github.com/ethereumjs/ethereumjs-monorepo/blob/29b5472f6f61b5f8bd3fd95307d35f0724aed332/packages/evm/src/precompiles/index.ts#L247

src/evm/precompiles/add-precompile.ts
daiagi: https://github.com/ethereumjs/ethereumjs-monorepo/blob/29b5472f6f61b5f8bd3fd95307d35f0724aed332/packages/evm/src/types.ts#L402

src/evm/precompiles/math-precompile.ts
daiagi: ignore comment above

i'm starting to see what you mean

src/evm/precompiles/add-precompile.ts
lukastanisic99: there isn't any `gasUsed` maybe you meant `gas` the amount of gas that's left while `executionGasUsed` is how much is consumed

src/evm/precompiles/math-precompile.ts
lukastanisic99: Ok I'll share more of my thoughts on this - 

I'm not surprised that typescript tests work and that it works in `ethereumJs` since the EVM is in JS so the precompile just calls a custom js function most likely - that's probably how ethereumJs implemented precompiles.

My concern is when we run the precompile `on-chain` (blockchain production node) - no typescript - most likely rust code. It doesn't call this typescript function to deduct the selector and what it does do it passes the bytes from the call directly to the wasm blob 

My concern is, this precompile like this would not act the same in ethereumJS and on-chain where there are no javascript functions.

Imo we should have the same interface for precompiles `bytes -> wasm -> bytes` so it will act the same `on-chain` and in `ethereumJS`

Maybe I'm mistaken, I didn't spend too much time on the actual implementation of precompiles in node software - I just assume this is how it works

daiagi: it's impossible to run an app specific pre-compile on a public chain

depending on our intended deployment chain:

### Ethereum mainnet:

to be able to run a pre-compile "on-chain" means that every client has implemented it and "shipped" it in a pre-agreed upon address
each client in their own language 

here is the list of precompile mandated by the protocol:
https://www.evm.codes/precompiled

and you can see them in the clients' source code:

* geth https://github.com/ethereum/go-ethereum/blob/master/core/vm/contracts.go
* ethereumjs https://github.com/ethereumjs/ethereumjs-monorepo/tree/29b5472f6f61b5f8bd3fd95307d35f0724aed332/packages/evm/src/precompiles 
* besu https://github.com/hyperledger/besu/tree/a65e02e707aacbaa0eb1902ecac49e92620f3bf6/evm/src/main/java/org/hyperledger/besu/evm/precompile

this is obviously not going to happen for an app specific pre-compile like ours

### Polkadot parachain

Parachains execute WASM smart contracts (e.g. ink!); you cannot deploy arbitrary WASM blobs and have them run as free‑standing native functions

so you still can't just take Jeff's rust code, compile to wasm, deploy it to chain and expect it to run 

so the options are
1. as a pre-compile -> only in our own custom EVM -> only "off-chain"
2. rewrite as smart-contract

lukastanisic99: Yes we're aware that you can't just deploy this to a public EVM network - it has pros and cons - cons that it's not pluggable everywhere and pros it's not easy to copy since you can't just deploy it on Ethereum. You can do it on Polkadot however, not as wasm smart contracts (!ink) - we're not doing that - It's the EVM pallet, but we're not operating the chain. Tanssi already confirmed they can support our precompiles and they'll operate the chain 

Both the on-chain and off-chain environment must execute the STF deterministically equivalent, so both environments have to have the same precompile with the same interface and the code is called directly from solidity by passing the arguments to the predefined address that holds the precompile

lukastanisic99: And that's also the reason why we want the precompiles as an option and passed during `setup` so our Poker game may be limited to environments where the precompile is available for now (later we may get rid of the precompile, but easier for Jeff to initially do it like that), but other state machines that don't use custom precompiles can deploy anywhere

daiagi: ok but
in order to have a custom pre-compile added to the vm in etherum js i still have to write some js function
even the most minimal function

```js
function(bytes) {
    return wasm(bytes)
}
```

is still js code

and it will still not run 'on-chain'

>Tanssi already confirmed they can support our precompiles

what does that mean exactly?
they will add a function under an address we agree on that will call our supplied wasm?

lukastanisic99: > ok but in order to have a custom pre-compile added to the vm in etherum js i still have to write some js function even the most minimal function
> 
> ```js
> function(bytes) {
>     return wasm(bytes)
> }
> ```
> 
> is still js code
> 
> and it will still not run 'on-chain'

yes yes that's fine - that's what I had in mind. I know that typescript code won't run on-chain, but it's simple it basically just passes bytes to the wasm blob which should run deterministically in both environments. So even if the on-chain node is in rust and it uses a rust function as a hook, there shouldn't be any problems. 

Of course we can have duplicate implementations in `js` and `rust` but then we can have bugs - same idea as deploying the diamond in our local EVM - to reuse the same logic -> less chance of bugs 
> 
> > Tanssi already confirmed they can support our precompiles
> 
> what does that mean exactly? they will add a function under an address we agree on that will call our supplied wasm?

Not sure how they'll implement it on their end, but yes we'll have a dedicated address like other precompiles that are already implemented in the original EVM

daiagi:
all comments addressed

please review again
@lukastanisic99 


https://github.com/peer3to/math-precompile-example/blob/main/src/lib.rs

src/evm/wasm.ts
```diff
@@ -0,0 +1,68 @@
...
+                if (!response.ok) {
+                    throw new Error(
+                        `Failed to fetch WASM file: ${response.statusText}`
+                    );
+                }
+                const buffer = await response.arrayBuffer();
+                wasmBytes = new Uint8Array(buffer);
+            } else {
+                const resolvedPath = path.resolve(wasmPathOrUrl);
+                wasmBytes = fs.readFileSync(resolvedPath);
```
lukastanisic99: I know you check the environment and this conde won't run in browsers, but it could still be problematic for the bundler that's building the UI application - there are probably some npm packages that make this polymorphic and bundler safe

daiagi: let me offer an alternative approach, see the latest commit
`fs` and `path` are now only used in a build step that embeds the WASM as a base64 string directly into the SDK source code

the FE bundler knows nothing about it

- no runtime dependency on node modules
- will load fast at runtime
- no network dependency

package.json
```diff
@@ -65,7 +68,8 @@
     "testc": "yarn compile && yarn hardhat test",
     "test": "yarn hardhat test",
     "test:coverage": "nyc --reporter=html yarn testts",
-    "build": "yarn compile && rimraf dist && tsc && tsc-alias --project tsconfig.json && npm pack && node rename-pack.js",
+    "embed-wasm": "node wasm/embed_wasm.js",
+    "build": "yarn embed-wasm && yarn compile && rimraf dist && tsc && tsc-alias --project tsconfig.json && npm pack && node rename-pack.js",
```
lukastanisic99: Imo still better to use polymorphic import/loading since this is an extra step for the consumer of the SDK and they have their own package.json

What's best in my opinion, is to not be opinionated at all - not even on `wasm`. We will use wasm for our precompiles, but we shouldn't force that opinion on others and it simplifies our life here too. The calling context, the UI application, has to define a precompile function `CustomPrecompile` - if it wants to use wasm let it be its own responsibility to load that wasm for its environment and in a way it wants to load it (not our opinion like in `wasm.ts`) and like that our SDK doesn't have these problems and it's clean - it can run any precompile wasm or not wasm. 

Also we can remove the `precompile` dir - those examples don't have to be part of the codebase - they can be moved to `tests` or `precompile examples`.

daiagi: this reduces this pr to nothing more then another parameter on p2psetup

if it is on them to write the `cusomPrecomile` however they want, then the sdk need no polymorphic import/loading

and the math wasm example is also useless because the sole point was to give infra to load wasm, if this is not needed then there is nothing to do here until we need to onboard the actual crypto code
and anyways the handling of Jeff's code will need to be in the poker app, not here.

lukastanisic99: Yes true - I didn't know it was that simple to add a precompile to `ethereumJS` so this was useful regardless and we don't have to complicate thing since they already accounted for it - originally it was more of a R&D to figure out do we have to fork or what we have to do so we can easily add custom stuff - we can keep this open for now, but yes it's mostly just adding a parameter to `setup`

daiagi: ill drop the rest, move the wasm stuff to example to serve as learning tool for future us or others

daiagi: image


moved.
changes to SDK are minimal, as agreed

lukastanisic99: all looks good - maybe just remove the `embed-wasm` from the main `package.json` and `build` command and just keep it in the examples.

daiagi: :see_no_evil: 

removed

lukastanisic99 (approved):

PR #95, opened 2025-07-03
add exitPoints record (Storage PR 1)

daiagi:
## 📋 Description

recording of exit point for milestone generation


src/storage/Storage.ts
```diff
@@ -14,11 +14,19 @@ import { Codec } from "@/utils";
...
+};
+
+type ExitPoint = Coordinate;
+
 export class Storage {
     private blockStorage = new BlockStorage();
     private joinChannelBlockStorage = new JoinChannelBlockStorage();
     private exitChannelBlockStorage = new ExitChannelBlockStorage();
     private stateSnapshotStorage = new StateSnapshotStorage();
+    private exitPoints: ExitPoint[] = [];
```
lukastanisic99, line 27: better to do a mapping of `forkId` -> `blockHeights [ ]` - so when you're constructing the state proof for a given fork you fetch all the milestones/blockHeights for it immediately - like this you would have to iterate and check every time is `forkId == forkId` and collect `blockHeight` if true

lukastanisic99: this can be its own storage module too

daiagi:
it got messed up in a rebase
ill create a new one

PR #96, opened 2025-07-03
block as a class (Order TS  PR 0)

daiagi:
Block as  a class
removed block utils

https://trello.com/c/px7L905K


src/Block.ts
```diff
@@ -0,0 +1,112 @@
...
+    Hash,
+    Signature
+} from "@/types/types";
+import { SignedBlockStruct } from "@typechain-types/contracts/V1/types/DataTypes";
+
+export class Block {
+    private constructor(private readonly block: BlockStruct) {}
+
+    static from(block: BlockStruct): Block {
+        return new Block(block);
```
lukastanisic99: Any reason/benefit of having this and not just doing `new Block(blockStruct))`

daiagi: Nicer syntax. Flows better.

src/agreementManager/QueueService.ts
```diff
@@ -1,10 +1,7 @@
...
-    BlockStruct
-} from "@typechain-types/contracts/V1/types/DataTypes";
+import { SignedBlockStruct } from "@typechain-types/contracts/V1/types/DataTypes";
 import { BlockConfirmation } from "./types";
-import { BlockUtils, Codec, EvmUtils, Type } from "@/utils";
 import { BytesLike } from "ethers";
+import { Block } from "@/Block";
 
 type forkId = BytesLike;
 type Height = number;
```
lukastanisic99: import from `types` so we don't define this in multiple places

daiagi: Settled in pr 2

src/Block.ts
```diff
@@ -0,0 +1,112 @@
...
+    get transaction() {
+        return this.block.transaction;
+    }
+
+    equals(other: Block): boolean {
+        return this.encode() === other.encode();
+    }
+
+    getSignerAddress(signature: SignatureLike): Address {
+        return ethers.verifyMessage(ethers.getBytes(this.hash), signature);
```
lukastanisic99: this has to be equivalent to `author` so not sure what's the use of this function - in general when you see something like `getSignerAddress` and pass it a signature you don't tie it to `block` logic, so better to keep this general and if you want to retrieve the underlying address from a signature just use the general `getSignerAddress` in `SingatureUtils`

lukastanisic99: Ah ok I understand what you wanted to do here - multiple people sign the same block so you want to retrieve their address - Ok, but still as you see I was confused, so maybe better to keep this elsewhere, but let me think until I finish the review - also taking of now so will be offline for a few hours

lukastanisic99: It's all fine - there's a lot of benefits to having these crypto utils here and the mental overhead is minimal and well worth the benefits

lukastanisic99 (approved):

PR #99, opened 2025-07-08
Dispute-storage

daiagi:
## 📋 Description

Dispute storage


src/storage/BlockStorage.ts
```diff
@@ -279,14 +279,20 @@ export class BlockStorage {
...
             const signaturesSet = new Set(existingBlock.signatures);
             for (const newSignature of blockConfirmation.signatures) {
                 signaturesSet.add(newSignature);
             }
-            existingBlock.signatures = Array.from(signaturesSet);
 
+            const mergedBlock: BlockConfirmationStruct = {
+                signedBlock: existingBlock.signedBlock,
+                signatures: Array.from(signaturesSet)
+            };
```
lukastanisic99: hmm what was wrong with the previous implementation - too me it looks the same. Was updating through the reference not being propagated?

src/storage/Storage.ts
```diff
@@ -24,6 +27,17 @@ export class Storage {
...
 
+    private constructor() {
+        // Private constructor for singleton pattern
+    }
+
+    public static getInstance(): Storage {
+        if (!Storage.instance) {
+            Storage.instance = new Storage();
+        }
+        return Storage.instance;
```
lukastanisic99: so we're doing the singleton in the end 😜

daiagi: indeed. I'm jugling a few PR's at the same time
you will see it resolved in next one

src/storage/BlockStorage.ts
daiagi: born in this commit
https://github.com/peer3to/state-channels-plus/pull/99/commits/89c544f6be483b344e744b6463e86849823b0d2a

due to a bug spotted in the tests

imagine this pseodcode

```ts
const blockConfirmation = blockConfirmationwith 3 sigs
storage.block.store(blockConfirmation)

const otherBlockConifmation = blockConifrmation with 2 sigs
// under the same hash as the first one
storage.block.store(otherBlockConifmation)

blockConfirmation.signaures.lenth == 5 // true

```


updating the number of signatures IN the storage affects the object outside the storage (stored by ref)
this is undesirable

hence, the fix

lukastanisic99: Ah yes - but in that case better to do an object copy - `deep copy` since this just partially solves the problem - it's still a `shallow copy` the internal references are the same still like `signedBlock`

Best to do a `deep copy` at creation do decouple the persisted object from the one used `outside` and then "in storage" you can use the deep copied one safely as reference, but keep in mind if you're pushing complex types into it they should also be deep copied so they're decoupled from the outside. 

Reading from storage should also return a `deep copy` so the outside doesn't change what's in storage. 

Maybe easiest to do this once with a `proxy` that `serializes` and `deserializes` the arguments when interacting with `storage modules` from the outside (you interact with the proxy) and internally you interact with the regular object that uses references - that will implicitly perform a `deep copy` everywhere implicitly without you having to think about it here and in other storage modules 

Take a look at our `RPC proxy` for inspiration and our `Debug proxy` and where it's used which with 1 line of code does a polymorphic substitution of the object with the proxy

lukastanisic99: you basically inject the proxy in just 1 place and forget about this totally - you do it in the constructor of `StorageManager` or what you call `Storage` since everything from the outside is interacting with it, so you have something like:

```typescript
StorageManager constructor(...){

return DeepCopyProxy(this);
}

Comment (daiagi) [reply src/storage/BlockStorage.ts]: excellent idea!

done

File: src/utils/DeepCopyProxy.ts [hunk 3]
```diff
@@ -0,0 +1,39 @@
...
+                    // Call original method
+                    const result = originalValue.apply(target, copiedArgs);
+
+                    // Deep copy result
+                    return cloneDeep(result);
+                };
+            }
+            // For class instances , apply recursive proxy
+            if (isClassInstance(originalValue)) {
+                return createDeepCopyProxy(originalValue as T);
```
lukastanisic99: I see the need for this since we're mostly using `storage.someSpecificStorage` and without this we don't actually serialize/deserialize if we access it that way - good idea, but if we do this recursively the proxy will cascade on all properties of `someSpecificStorage` and further if it has nested objects, which we don't want - we want just to proxy the first layer of interaction. Also doing it here like this creates a new object/proxy every time we access a nested class instance which can bloat memory even though eventually the garbage collector will dispose of the object.

Better to proxy just `functions` that way it doesn't cascade and maybe in the constructor of `StorageManager` you can substitute all instances with proxies - maybe there's a more clever way of doing this without cascading and creating new objects all the time

daiagi: What's the issue?
First layer is the class instances in storage, second layer is functions

Why does it create a new object every time we access?

It creates exactly as many proxies as the storage sub classes plus the storage class itself.

lukastanisic99: There are 2 issues:
1) Fully recursive proxy:
All nested class instances (not just instances in `StorageManager`) will be 'proxified', if you try accessing them
2) Every time you access `storage.blocks` or similar (not just the first time) you're creating a new proxy of the same type (e.g. BlockStorage or similar) which is compute and memory inefficient

daiagi: fixed

changed for direct proxying in `Storage` constructor

src/storage/Storage.ts
```diff
@@ -4,17 +4,22 @@ import { ExitChannelBlockStorage } from "./ExitChannelBlockStorage";
...
+    public readonly exitPoints: ExitPointsStorage;
 
     constructor() {
-        return createDeepCopyProxy(this);
+        this.blocks = deepCopyProxy(new BlockStorage());
+        this.joinChannelBlocks = deepCopyProxy(new JoinChannelBlockStorage());
+        this.exitChannelBlocks = deepCopyProxy(new ExitChannelBlockStorage());
+        this.stateSnapshots = deepCopyProxy(new StateSnapshotStorage());
+        this.exitPoints = deepCopyProxy(new ExitPointsStorage());
+        return deepCopyProxy(this);
```
lukastanisic99, line 22: Fine for now - Has a drawback when used with helper functions defined in `Storage` it will perform the deep copy again even though calling the function already performed the deep copy - But still better than cascading the proxy recursively and always creating new instances - I'll put an optimization card for this at the end of our backlog - this is enough for now

lukastanisic99 (approved):

PR #100, opened 2025-07-08
exit points storage

daiagi:


src/storage/ExitPointsStorage.ts
```diff
@@ -0,0 +1,57 @@
...
+
+        if (start === undefined && end === undefined) {
+            return list;
+        }
+
+        const actualStart = start ?? list[0];
+        const actualEnd = end ?? list[list.length - 1] + 1;
+
+        const filteredExitPoints = list.filter(
+            (height) => height >= actualStart && height < actualEnd
```
lukastanisic99: To consider maybe best to do `[low,high]` instead of `[low,high)` even though most conventions do [low,high) since we'll be calling this mostly in the form `[firstBlock.height, latestBlock.height]` where if the lastBlock also produced an exit we want to account for it

Either way even if we have to do a +1 it's fine

lukastanisic99 (approved):

src/storage/ExitPointsStorage.ts
daiagi: done

PR #103, opened 2025-07-09
Store-join/exit-block

### LINKED ISSUE #101: exitChannel -> ExitChannelBlok
You do have it, but in some higher class like `StateManager` since it checks are you appending the next block -> it knows the `latestBlock` and keeps track of it and accounts for forks/reorgs so it can also keep track of this as a cache optimization and it should, but in reality if you know the previous/latestBlock you know everything since you would fetch the hash from the snapshot

_Originally posted by @lukastanisic99 in https://github.com/peer3to/state-channels-plus/pull/80#discussion_r2191475597_

daiagi:
- Closes #101 

attempted to implement storing of join and exit blocks

hope i am not too far off the mark

tests will follow if the logic is correct @lukastanisic99

### DIFF
--- a/src/stateManager/StateManager.ts
+++ b/src/stateManager/StateManager.ts
@@ -76,7 +76,7 @@ import { BalanceStruct } from "@typechain-types/contracts/V1/AStateMachine";
 
 let DEBUG_STATE_MANAGER = false;
 
-const EMPTY_ADDRESS: Address = "0x00";
+const NULL = "0x00";
 class StateManager {
     stateMachine: AStateMachine;
     p2pEventHooks: P2pEventHooks;
@@ -87,7 +87,7 @@ class StateManager {
     stateChannelManagerContract: AStateChannelManagerProxy;
     p2pManager: P2PManager;
     timeConfig: TimeConfig;
-    channelId: ChannelId = "0x00";
+    channelId: ChannelId = NULL;
     mutex: Mutex = new Mutex();
     self = DEBUG_STATE_MANAGER ? DebugProxy.createProxy(this) : this;
     isDisposed: boolean = false;
@@ -206,7 +206,7 @@ class StateManager {
                 this.disputeHandler.proofManager.createDoubleSignProof([
                     signedBlock
                 ]);
-            this.disputeHandler.createDispute(block.forkId, EMPTY_ADDRESS, 0, [
+            this.disputeHandler.createDispute(block.forkId, NULL, 0, [
                 disputeProof
             ]);
         } else if (flag == AgreementFlag.INCORRECT_DATA) {
@@ -215,7 +215,7 @@ class StateManager {
                 this.disputeHandler.proofManager.createIncorrectDataProof(
                     signedBlock
                 );
-            this.disputeHandler.createDispute(block.forkId, EMPTY_ADDRESS, 0, [
+            this.disputeHandler.createDispute(block.forkId, NULL, 0, [
                 disputeProof
             ]);
         }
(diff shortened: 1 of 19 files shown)


src/stateManager/StateManager.ts
```diff
@@ -71,7 +92,7 @@ class StateManager {
     self = DEBUG_STATE_MANAGER ? DebugProxy.createProxy(this) : this;
     isDisposed: boolean = false;
     validationService: ValidationService;
-    private storageModule: IStorageModule = new StorageModule();
+    private storage: Storage = new Storage();
```
daiagi, line 95: should be a singleton, already handled in another PR

src/stateManager/StateManager.ts
```diff
@@ -566,29 +588,70 @@ class StateManager {
         return this.getEncodedState().then(ethers.keccak256);
     }
 
-    private handleJoinChannelStorage(joinChannelBlock: JoinChannelBlockStruct) {
-        const blockHash = ethers.keccak256(
-            Codec.encode(joinChannelBlock, Type.JoinChannelBlock)
+    private async calculateTotalBalance(
+        channels: { balance: BalanceStruct }[],
```
lukastanisic99: would call this still `balances` since you're only interested in that property - was confusing on first read

src/stateManager/StateManager.ts
```diff
@@ -566,29 +588,70 @@ class StateManager {
...
+        joinChannelBlock: JoinChannelBlockStruct
+    ) {
+        const prevBlock =
+            this.storage.joinChannelBlocks.getJoinChannelBlockEntry(
+                joinChannelBlock.previousBlockHash
+            );
+
+        const totalDeposits = await this.calculateTotalBalance(
+            joinChannelBlock.joinChannels,
+            prevBlock?.totalDeposits
```
lukastanisic99: For joins we calculate this `on-chain`, so we could emit it as part of the event - it's fine to calculate it off-chain too, since it would sum up to the same result, unless the `previousJoinChannelEntry` is missing in storage -> we should add a simple check here and in the exitChannel processing code to:
```typescript
if(prevBlock == undefined &&  joinChannelBlock.previousBlockHash != "0x00") throw Error(" previous JoinChannel missing in storage ${hash}"); 
``` 
This prevents a peer from committing to an invalid snapshot due to it lacking the previous balance -> later we can implement recovery

src/stateManager/StateManager.ts
```diff
@@ -566,29 +588,70 @@ class StateManager {
...
         };
-        const exitChannelBlockHash = ethers.keccak256(
-            Codec.encode(exitChannelBlock, Type.ExitChannelBlock)
+
+        // Get previous block's total withdrawals or zero balance if first block
+        const prevBlock =
+            this.storage.exitChannelBlocks.getExitChannelBlockEntry(
+                previousBlockHash
+            );
+
```
lukastanisic99, line 629: same check here - overall the logic is good

src/stateManager/StateManager.ts
daiagi: let's simplify and keep it on chain only and via the event

lukastanisic99: Sure, but you can only do it for joins, for exits you need this anyway

cyony95: why would you need to calculate it both on chain and off chain for exits? isn't it enough to do it just on chain?

lukastanisic99: For exits we don't do it on-chain :D You can exit off-chain and eventually when a snapshot is submitted you process that on-chain 

Joining has to happen on-chain since you have to lockup funds

daiagi: no real need. Luka is just saying we could
just a detail i missed and therefore wrote redundant code

daiagi:
@lukastanisic99 
all done
added  a bit of testing and found some bugs

* test for `Block` model => revealed a bug in `Block.getSignersSet`
* test for `StateSnapshot` model => bug in encoding (`StateSnapshotEthersType`)
* test for the added map in `StateSnapshotStorage` => all good
* test for `Storage.getStateSnapshot` => all good

lukastanisic99 (approved):

PR #105, opened 2025-07-10
Move-queues

daiagi:
Move queues to storage https://trello.com/c/cEheyzUR
Then i got carried away and did larger refactoring then i expected

@lukastanisic99  if you approve of the logic in `QueueStorage` then i will add tests

also this PR includes storage as  passed singleton

i.e p2psetup creates the new storage instance and from there it is passed

p2psetup -> StateManager -> AgreementManager -> (OnChainTracker, BlockValidator)
                             |
                             |> (ExecutionDecisionProcessor, ConfirmationDecisionProcessor)

this is A LOT

 i am thinking of middle ground between direct dependency injection and global singleton

I'll propose something in next PR


lukastanisic99:
> Move queues to storage Then i got carried away and did larger refactoring then i expected
> 
> @lukastanisic99 if you approve of the logic in `QueueStorage` then i will add tests
> 
> also this PR includes storage as passed singleton
> 
> i.e p2psetup creates the new storage instance and from there it is passed
> 
> p2psetup -> StateManager -> AgreementManager -> (OnChainTracker, BlockValidator) | |> (ExecutionDecisionProcessor, ConfirmationDecisionProcessor)

Yes you see where the Singleton idea came from, but keep in mind that 90% of agreement manager is getting deleted - OnChainTracker, BlockValidator also - I've made Trello tasks with descriptions for many things - I'm mostly doing that today going over everything 

As for `QueueStorage` I didn't look at the code yet (will do now), but from its name it seems it's a special storage module - it shouldn't be a special storage module - you basically just persist the block in the map. The `StateManager` will try and retrieve the next block from storage when it wants - it will either get `undefined` or some block that it will apply across the pipeline and perform checks - so even if the block is junk it can sit in block storage until we get there

Maybe for `pruning` we'd have a problem if the spam is too much and we have random blocks in the map at random locations.

What we could always do when it comes to pruning:
 - take the `last N valid blocks` - `StateManager` keeps track of the latest valid
 - create new map with them
 - delete the previous map entirely - will delete even the random storage locations

daiagi:
>Yes you see where the Singleton idea came from, but keep in mind that 90% of agreement manager is getting deleted - OnChainTracker, BlockValidator also - I've made Trello tasks with descriptions for many things - I'm mostly doing that today going over everything

yes of course i see, just trying to balance it with testabilty

anyway I'll sit on my hands for a bit while you rampage through agreement manager
then ill try and propose a middle ground for this issue

daiagi:
>it shouldn't be a special storage module

haaa ok.  ill change

lukastanisic99:
> > Yes you see where the Singleton idea came from, but keep in mind that 90% of agreement manager is getting deleted - OnChainTracker, BlockValidator also - I've made Trello tasks with descriptions for many things - I'm mostly doing that today going over everything
> 
> yes of course i see, just trying to balance it with testabilty
> 
> anyway I'll sit on my hands for a bit while you rampage through agreement manager then ill try and propose a middle ground for this issue

I'm done with `AgreementManager` in Trello - left descriptions on all cards (hopefully) - AgreementManager was just an ugly storage module, but served its purpose for the w3f grant

daiagi:
@lukastanisic99 
remove queue storage

closer to target?

lukastanisic99:
> @lukastanisic99 remove queue storage
> 
> closer to target?

Yes something like that - didn't look too deep 

Thinks to consider - while we can store any block by `hash`, we can't store multiple by `forkId:height` 

Should we make it possible to store multiple by `forkId:height` too? 

Let's discuss this in the call today - easier for me to share my thoughts there than to write blobs of text and I need to finish some other stuff - in general if I forget, remind me to go over my thoughts for `StateManger` refactor and mental model and this queue stuff

lukastanisic99:
> > @lukastanisic99 remove queue storage
> > closer to target?
> 
> Yes something like that - didn't look too deep
> 
> Thinks to consider - while we can store any block by `hash`, we can't store multiple by `forkId:height`
> 
> Should we make it possible to store multiple by `forkId:height` too?
> 
> Let's discuss this in the call today - easier for me to share my thoughts there than to write blobs of text and I need to finish some other stuff - in general if I forget, remind me to go over my thoughts for `StateManger` refactor and mental model and this queue stuff

Maybe the explicit queue idea is not a bad idea - in that case what's in blockStorage is always `consistent` and `correct` - it has passed all pipeline checks and has been applied successfully to the `state` - we'll discuss

src/storage/QueueStorage.ts
```diff
@@ -0,0 +1,112 @@
...
+            signatures: []
+        };
+
+        this.queuedBlocks.set(block.hash, blockConfirmation);
+
+        // Update coordinate index
+        this.addHashToCoordinateIndex(block.hash, block.forkId, block.height);
+
+        return block.hash;
+    }
```
lukastanisic99, line 31: Thinking we should just use `BlockConfirmation` everywhere and not have separate pipelines for `SignedBlock` - that would cut a lot of the code in `StateManager` and other places (`BlockStorage`) and in general to do this for every `struct` that has its `Confirmation` equivalent 

As for the return value of these operation imo we should just return `bool success/failure` - if we need the hash we already have it

src/storage/QueueStorage.ts
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+        // Remove from queue
+        this.blocksByCoordinates.delete(coordinateKey);
+
+        return blockConfirmations;
+    }
+
+    isBlockQueued(hash: Hash): boolean {
+        return this.queuedBlocks.has(hash);
```
lukastanisic99: this is good, but in general where it's used better to still try and persist the `BlockConfirmation` so it potentially expands the `signature set`

lukastanisic99: to consider 
``` typescript
isBlockQueues(blockConfirmation, option{hash?}){
if(hash not in hash map) return false;
else try expand signatures;
return true;
}
```

src/storage/QueueStorage.ts
daiagi: it would not cut code
it will create burden of converting everywhere

when the ts code matures, we may find that because the return value from storage module is always  `...Confirmation`, that the underlying struct isn't used anywhere anymore
that would be the point where you would cut down this code
for now, better off like this

as for the hash...
you see  that the methods return just `Hash`? and not `Hash| undefined`?

this means that they will return true always => useless

at least the hash saves you double computation time
put something in => get hash out => useful

src/storage/QueueStorage.ts
daiagi: actually not even used => deleted

daiagi: i made a mistake
it is used
restored it

it is used in `BlockValidator`
```ts
    /** In chain OR parked in the “future queue” */
    isBlockDuplicate(block: Block): boolean {
        return (
            this.isBlockInChain(block) ||
            this.storage.queues.isBlockQueued(block.hash)
        );
    }

```
yes it will be removed, is this duplicate question will still need answer

daiagi:
@cyony95  pay attention to this commit https://github.com/peer3to/state-channels-plus/pull/105/commits/cab1a8817db382bcb1b0c183740f3d15b0775fcc


it touches on what you are doing now

daiagi:
@lukastanisic99  look again plz

lukastanisic99 (approved):

PR #106, opened 2025-07-11
Block storage - options based interface

daiagi:
Block storage interface is annoying
it is mixing up optional hash with optional coordinates

so if want to provide optional hashkey you also need to provide coordinates and vice-versa
not intuitive and not convenient

this PR offers `options` based interface

i think it's better, easier and clear to use

as side bonus, it cuts down lines of code


lukastanisic99 (approved):
Feel free to do the same in other components

PR #107, opened 2025-07-11
fix: outdated tests

daiagi:
Nothing interesting here

just updating failing tests that i failed to update in previous PR's

no new code


daiagi:
I'm assuming there  will be no objection, just merging

PR #110, opened 2025-07-14
on-chain-timestamp

daiagi:
## 💪 Related Issues

- Closes https://trello.com/c/IjXacXoO

### DIFF
--- a/src/models/Block.ts
+++ b/src/models/Block.ts
@@ -67,7 +67,9 @@ export default class Block {
         return this._onChainTimestamp;
     }
     get relevantTimestamp(): Timestamp {
-        return this._onChainTimestamp ?? this.timestamp;
+        return this._onChainTimestamp
+            ? Math.max(this._onChainTimestamp, this.timestamp)
+            : this.timestamp;
     }
 
     set onChainTimestamp(onChainTimestamp: Timestamp) {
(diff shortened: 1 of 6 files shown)


src/models/Block.ts
```diff
@@ -19,6 +19,8 @@ export type BlockCoordinates = {
 };
 
 export default class Block {
+    private _onChainTimestamp?: Timestamp;
+
     private constructor(private readonly block: BlockStruct) {}
```
lukastanisic99: add an optional argument to the constructor too? 

To consider should we expand the `Block` into `BlockConfirmation` and add `authorSignature?` and `signatures[ ]?` - there are pros and cons

daiagi: >add an optional argument to the constructor too?

will it make sense to cerate block with this timetamp? i think it only makes ense to set afterwards. no?

>To consider should we expand the Block into BlockConfirmation and add authorSignature? and signatures[ ]? - there are pros and cons

i believe we agreed not to do this in out call, at least not for now

lukastanisic99: > > add an optional argument to the constructor too?
> 
> will it make sense to cerate block with this timetamp? i think it only makes ense to set afterwards. no?
> 
`Theoretically:` If the block is only posted on-chain and not gossiped. `Practically:` in more places than that, since every time a block is posted on-chain, you update your storage with that timestamp and later if you're reconstructing a block from storage you'll set it at creation. 

> > To consider should we expand the Block into BlockConfirmation and add authorSignature? and signatures[ ]? - there are pros and cons
> 
> i believe we agreed not to do this in out call, at least not for now

Yes let's not do it for now

daiagi: done

PR #112, opened 2025-07-15
Cleanup BlockValidator from agreement manager - keeping it for a while to use as reference - will delete later

cyony95:
Closes https://trello.com/c/FfHsdq1d


cyony95:
Regarding the ValidationService: 

1. should it still do validateSignedBlock at all or just without the BlockValidator logic? (or maybe it will use StateManager logic)
2. should it still do validateBlockConfirmation at all or just without the BlockValidator logic? (or maybe it will use StateManager logic)
3. 2. I think isGoodTimestamp should be gone.

daiagi:
I'm not seeing where you replaced the API with new calls to storage

* `isBlockInChain` => `getBlockConfirmation(hash) !== undefined` (or create a `hasBlock` method in `blockStorage`
* `isBlockDuplicate` => method on `storage` that checks both `blockStorage` and `queue`
* `latestBlockTimestamp` => get all the blocks on a fork (a new method is needed , i guess on `BlockStorage`) and iter / reduce to  find max, 
* `latestRelevantTimestamp` is handled by me in #110 , you can skip


I didn't go through `check`

As to you question:
Yes, it should do validations, but it should realy and use the storage module rather then whatever is on `AgreementManager`
see this card https://trello.com/c/s6mJlIdH

cyony95:
> I'm not seeing where you replaced the API with new calls to storage
> 
> * `isBlockInChain` => `getBlockConfirmation(hash) !== undefined` (or create a `hasBlock` method in `blockStorage`
> * `isBlockDuplicate` => method on `storage` that checks both `blockStorage` and `queue`
> * `latestBlockTimestamp` => get all the blocks on a fork (a new method is needed , i guess on `BlockStorage`) and iter / reduce to  find max,
> * `latestRelevantTimestamp` is handled by me in [on-chain-timestamp #110](https://github.com/peer3to/state-channels-plus/pull/110) , you can skip
> 
> I didn't go through `check`
> 
> As to you question: Yes, it should do validations, but it should realy and use the storage module rather then whatever is on `AgreementManager` see this card https://trello.com/c/s6mJlIdH

you're right, I made it a PR instead of a draft, changing it now.

PR #113, opened 2025-07-15
remove `processStateTransition`

daiagi:
#### Problem

`processStateTransition` is misleading to be in `ValidationService`

Also it is plain wrong
transaction is applied once "to validate"
and second time (on already transitioned state)
to "actaully apply"

#### Solution

- Removed `processStateTransition` , apply transaction once, in `StateManager`
- moved storage ops out of `applyTransaction` and into `onSignedBlock`

### DIFF
--- a/src/stateManager/StateManager.ts
+++ b/src/stateManager/StateManager.ts
@@ -277,22 +277,53 @@ class StateManager {
     // Passes the signedBlock through a verification pipeline and returns an execution flag based on the outcome
     public async onSignedBlock(
         signedBlock: SignedBlockStruct,
-        block?: Block
+        blk?: Block
     ): Promise<ExecutionFlags> {
         // Default everything to SUCCESS + no AgreementFlag
         let finalExecutionFlag: ExecutionFlags = ExecutionFlags.SUCCESS;
         let finalAgreementFlag: AgreementFlag | undefined = undefined;
-        const decodedBlock = block ?? Block.decode(signedBlock.encodedBlock);
+        const block = blk ?? Block.decode(signedBlock.encodedBlock);
 
         try {
             await this.mutex.lock();
             const result = await this.validationService.validateSignedBlock(
                 signedBlock,
-                decodedBlock
+                block
             );
 
             finalExecutionFlag = result.flag;
             finalAgreementFlag = result.agreementFlag;
+
+            // If validation passed, apply the transaction
+            if (result.success && result.flag === ExecutionFlags.SUCCESS) {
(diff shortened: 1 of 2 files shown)


src/stateManager/StateManager.ts
```diff
@@ -277,23 +277,54 @@ class StateManager {
...
+                    this.storage.blocks.storeBlock(signedBlock);
+
+                    // Schedule the success callback
+                    scheduleTask(
+                        applyResult.successCallback,
+                        0,
+                        "stateTransitionSuccessCallback"
+                    );
+                }
+            }
```
lukastanisic99, line 326: This should be part of `processExecutionDecision` or renamed to `processValidationDecision`- The mental model for this is: 
1) apply `block validation pipeline` which has some `result`
2) based on `result` do something 

if `flag == success` all checks have passed and block should be applied to current state, co-sign and gossip
if `flag == dispute` stop progressing current fork -> create a `dispute struct` as a function of `storage` (FraudProof storage, ...) and submit on-chain - a Trello card with have an algo
... other flags ... 

return the execution flag so the calling component can take adequate action if needed

daiagi: Luka i can't do both
on the one hand you want small digestible PR's 
on the other you want to see fully done code

many things are not  where they should and i can't possibly put it all in place in 1 PR
while allowing you to review and Adrian to follow


this PR was born because i wanted to do `Time-logic` trello card
and things were too messy around here so i tidy up a bit


or if you prefer. no new features until the ts code is sensible, time logic is new feature

lukastanisic99: Not sure I follow where the problem is - `Time logic` is just a check part of the `pipeline` - you don't need to restructure things for that (or I'm not following something). If this being here helps with `time logic`, sure - we can place it later in the correct place. I'm mostly sharing my thoughts as I'm reviewing code

daiagi: Yes it isn't absolutely needed. But it was bothering me that there was a state transition in validation 

So I removed it, because it doesn't belong there. 
And Just moved storage stuff out of execute transaction, so the function has a single responsibility 

Also, if you could please, Use the "request changes" option 

This way I'll will know if your comments are general pointers / ideas / thoughts or or if things in this pr need changing

Let's say in jamixir it is common that we approve AND write comments

So the other person knows, OK things are OK / good enough. Can be improved so and so.

lukastanisic99 (approved):
Fine for now - will be moved to the appropriate place later - will create a Trello card

PR #114, opened 2025-07-16
Time-logic

daiagi:
- Closes https://trello.com/c/8cLJYFfm

### DIFF
--- a/src/stateManager/ValidationService.ts
+++ b/src/stateManager/ValidationService.ts
@@ -6,11 +6,11 @@ import DisputeHandler from "@/DisputeHandler";
 import { ethers } from "ethers";
 import { AStateChannelManagerProxy } from "@typechain-types/contracts/V1/StateChannelDiamondProxy";
 import {
-    subjectiveTimingFlag,
     SignatureUtils,
     getActiveParticipants,
     Codec,
-    Type
+    Type,
+    hash
 } from "@/utils";
 import AStateMachine from "@/AStateMachine";
 import { Clock } from "..";
@@ -23,6 +23,7 @@ import {
     Timestamp
 } from "@/types/types";
 import { Block } from "@/models";
+import Storage from "@/storage";
 
 interface ValidationResult {
     success: boolean;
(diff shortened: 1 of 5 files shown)


src/stateManager/ValidationService.ts
```diff
@@ -350,25 +350,29 @@ export default class ValidationService {
                 return undefined;
             }
 
-            // Query recent logs (last 10 blocks)
```
cyony95: @daiagi what if the block is not in the most recent 3 blocks?

daiagi: I don't see a scenario where this happens.
Really it should be just the previous block.

So yes , query by exact block number would be better, but idk how to do that, 
And it would be more complex.
This seems to me like a reasonable practical solution.
Open to suggestions

daiagi: Altough this makes me wonder..
What is the relation, if any, between our blocks and the the network blocks.
Whenwe I say it should be previous block I mean on our blocks, but I'm not sure this means anything about the network blocks

cyony95: When you say network blocks, you mean what s on chain or what the other peers have?

daiagi: Nvm. I think I got myself confused for no reason.
We run the evm. We determine when a state transition happens.
There is no "mainnet" blocks

src/stateManager/ValidationService.ts
```diff
@@ -94,16 +96,83 @@ export default class ValidationService {
...
+                return dispute(AgreementFlag.INCORRECT_DATA);
+            }
+            // If valid, continue with rest of validation
         }
 
+        // Subjective timestamp validation (requireS)
+        if (
+            Math.abs(Clock.getTimeInSeconds() - blk.timestamp) >=
+            this.timeCfg.agreementTime
+        )
```
lukastanisic99: Commented on Trello - this one should be ignored if the on-chain timestamp is set

lukastanisic99 (approved):
Overall everything is good - just ignore the subjective check if the on-chain timestamp is present 

The `IncorrectData` fraud proof will be renamed later to  `invalidTimestamp` since it's the only thing that can be objectively invalid that's not covered by the `invalid state transition proof` 

Hopefully I'll be faster to look and review PRs

src/stateManager/ValidationService.ts
daiagi: done

PR #115, opened 2025-07-17
cut down agreementManager

daiagi:
Once we agree on the logic, ill add / update the tests

## 💪 Related Issues

- Closes https://trello.com/c/s6mJlIdH


cyony95 (approved):

src/agreementManager/AgreementManager.ts
```diff
@@ -3,237 +3,108 @@ import {
...
+    public getLatestSignedBlockByParticipant(
         forkId: ForkId,
-        genesisTimestamp: Timestamp
-    ) {
-        this.forks.newFork(
-            forkGenesisStateEncoded,
-            addressesInThreshold,
+        participantAdr: Address
+    ): { block: Block; signature: Signature } | undefined {
+        const blockEntries = this.storage.blocks.getBlocksByForkId(
```
lukastanisic99, line 24: Fine for now but retrieving everything is expensive so it should be an iterator and let's use an enum or bool or something instead of string `desc` so we don't end up with misspells (me :D) 

In general we should know the latest block from which to start iterating either from `storage` or what's the case for now `StateManager`

src/agreementManager/AgreementManager.ts
```diff
@@ -3,237 +3,108 @@ import {
...
-        if (!agreement || !agreement.block.equals(block))
-            return { didSign: false, signature: undefined };
+        const blockEntry = this.storage.blocks.getBlockEntry(block.hash);
+        if (!blockEntry) return { didSign: false, signature: undefined };
+
+        // Check if participant is the author
+        if (block.author === participant) {
+            return {
+                didSign: true,
+                signature: this.storage.blocks.getOriginalSignature(block)
```
lukastanisic99: you already have the signature in the `signedBlock` of the blockEntry

src/agreementManager/AgreementManager.ts
```diff
@@ -3,237 +3,108 @@ import {
...
+            block.coordinates
+        );
 
-    public isParticipantInLatestFork(participant: Address): boolean {
-        const fork = this.forks.latestFork();
-        if (!fork) return false;
-        return new Set(fork.addressesInThreshold).has(participant);
-    }
+        const signersSet = block.getSignerAddresses(
+            this.getAllSignatures(blockEntry.blockConfirmation)
```
lukastanisic99, line 100: just a reminder to account for authors signature

lukastanisic99: ah ok it does it already - ignore

src/agreementManager/AgreementManager.ts
daiagi: Iterator is great idea. Thanks

src/agreementManager/AgreementManager.ts
```diff
@@ -3,237 +3,108 @@ import {
...
     }
+
     /**
-     * Gets the latest finalized state (ecnoded) and the latest signed/confirmed state (encoded) from the signer with virtual votes proving it
-     * @param forkId
-     * @param signerAddress
-     * @returns
+     * Get the latest finalized state and latest signed state (by signer) with virtual voting blocks
      */
     public getFinalizedAndLatestWithVotes(
```
lukastanisic99, line 110: This has to be changed for milestone logic - can't be a direct 1:1 mapping from our old logic -> fine for now so Adrian is not blocked - I'll crate a Trello card which I'll populate later

src/agreementManager/AgreementManager.ts
```diff
@@ -243,176 +114,195 @@ class AgreementManager {
...
-        forkId: ForkId,
-        maxTransactionCnt: BlockHeight
-    ): Timestamp {
-        return this.chain.latestTimestamp(forkId, maxTransactionCnt);
-    }
-
+    /**
+     * Check if a participant has posted a block on-chain
+     */
     public didParticipantPostOnChain(
```
lukastanisic99: didParticipantPostOnChainLocal -> something like that so we know we're not actually checking on-chain here, since in the case that this fails we might need to check on-chain to double check

src/storage/BlockStorage.ts
```diff
@@ -244,6 +244,116 @@ export class BlockStorage {
         return true;
     }
 
+    /*────────────────────────────────────────────────────────────────────────────
+      GET ALL BLOCKS BY FORK ID
+    ────────────────────────────────────────────────────────────────────────────*/
+    getBlocksByForkId(
+        forkId: ForkId,
+        sortOrder?: "asc" | "desc"
+    ): BlockEntry[] {
```
lukastanisic99: some iteration logic is needed, but not in the form of fetch everything and sort -> in most cases we're interested only in the `last N` blocks

lukastanisic99 (approved):
Fine for now just left some comments

src/agreementManager/AgreementManager.ts
daiagi: done already

PR #116, opened 2025-07-17
Dual mappings double-sign protection

cyony95:
Closes https://trello.com/c/ZUnAgxcE/134-blockstorage-and-other-dual-mappings-doublesign-protection

### DIFF
--- a/src/storage/BlockStorage.ts
+++ b/src/storage/BlockStorage.ts
@@ -43,7 +43,10 @@ export class BlockStorage {
     /*────────────────────────────────────────────────────────────────────────────
       STORE  BLOCK - IMPLEMENTATION
     ────────────────────────────────────────────────────────────────────────────*/
-    storeBlock(signedBlock: SignedBlockStruct, options?: StoreOptions): Hash | undefined {
+    storeBlock(
+        signedBlock: SignedBlockStruct,
+        options?: StoreOptions
+    ): Hash | undefined {
         // Convert SignedBlock to BlockConfirmation (empty signatures)
         const blockConfirmation: BlockConfirmationStruct = {
             signedBlock: signedBlock,
@@ -271,39 +274,37 @@ export class BlockStorage {
 
         // Store the block entry
         const coordinateKey = this.coordinatesToKey(coordinates);
-        const existingEntryByHash = this.hashToBlockMap.get(blockHash);
-        const existingEntryByCoordinates = this.coordinatesToBlockMap.get(coordinateKey);
+        const existingEntry = this.coordinatesToBlockMap.get(coordinateKey);
 
-        // If the same block already exists, merge signatures and return hash - existing functionality
-        if (existingEntryByHash && existingEntryByCoordinates) {
-            const existingBlockDecoded = Block.decode(existingEntryByHash.blockConfirmation.signedBlock.encodedBlock);
-            if (existingBlockDecoded.coordinates.forkId === coordinates.forkId &&
-                existingBlockDecoded.coordinates.height === coordinates.height) {
-                // Same block, merge signatures and return hash
+        if (existingEntry) {
+            // Compare incomingBlock.encodedBlock == existingBlock.encodedBlock
+            const incomingEncodedBlock =
+                blockEntry.blockConfirmation.signedBlock.encodedBlock;
+            const existingEncodedBlock =
+                existingEntry.blockConfirmation.signedBlock.encodedBlock;
+
+            if (incomingEncodedBlock === existingEncodedBlock) {
+                // They are equal => merge signatures
                 const signaturesSet = new Set(
-                    existingEntryByHash.blockConfirmation.signatures
+                    existingEntry.blockConfirmation.signatures
                 );
-                for (const newSignature of blockEntry.blockConfirmation.
-                    signatures) {
+                for (const newSignature of blockEntry.blockConfirmation
+                    .signatures) {
                     signaturesSet.add(newSignature);
                 }
-                existingEntryByHash.blockConfirmation.signatures = 
+                existingEntry.blockConfirmation.signatures =
                     Array.from(signaturesSet);
 
                 // Update on-chain timestamp if provided
                 if (blockEntry.onChainTimestamp !== undefined) {
-                    existingEntryByHash.onChainTimestamp = blockEntry.onChainTimestamp;
+                    existingEntry.onChainTimestamp =
+                        blockEntry.onChainTimestamp;
                 }
-                return blockHash;
-            }
-        }
 
-        // Conflict: different blocks with the same coordinates
-        if (existingEntryByCoordinates) {
-            const existingBlockHash = ethers.keccak256(
-                existingEntryByCoordinates.blockConfirmation.signedBlock.encodedBlock
-            );
-            if (existingBlockHash !== blockHash) {
+                // Return the hash (same object in both maps)
+                return blockHash;
+            } else {
+                // Not equal => abort
                 return undefined;
             }
         }
--- a/src/storage/BlockStorage.ts
+++ b/src/storage/BlockStorage.ts
@@ -43,7 +43,7 @@ export class BlockStorage {
     /*────────────────────────────────────────────────────────────────────────────
       STORE  BLOCK - IMPLEMENTATION
     ────────────────────────────────────────────────────────────────────────────*/
-    storeBlock(signedBlock: SignedBlockStruct, options?: StoreOptions): Hash {
+    storeBlock(signedBlock: SignedBlockStruct, options?: StoreOptions): Hash | undefined {
         // Convert SignedBlock to BlockConfirmation (empty signatures)
         const blockConfirmation: BlockConfirmationStruct = {
             signedBlock: signedBlock,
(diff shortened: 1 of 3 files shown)


src/storage/StateSnapshotStorage.ts
```diff
@@ -24,12 +24,29 @@ export class StateSnapshotStorage {
...
+                existingSnapshot.snapshotData.stateMachineStateHash === snapshot.snapshotData.stateMachineStateHash) {
+                return hash;
+            }
+            // Conflict
+            return undefined;
+        }
+
         this.snapshotsByHash.set(hash, snapshot);
 
         if (snapshot.isGenesis) {
```
cyony95: @daiagi @lukastanisic99 does a check for the genesis makes sense here similar to how the check for the existing snapshot is made? I think the existingSnapshot check is enough, a genesis snapshot is still a snapshot

src/storage/BlockStorage.ts
```diff
@@ -271,29 +271,44 @@ export class BlockStorage {
...
-
-        if (existingEntry !== undefined) {
-            // Merge signatures
-            const signaturesSet = new Set(
-                existingEntry.blockConfirmation.signatures
-            );
-            for (const newSignature of blockEntry.blockConfirmation
-                .signatures) {
-                signaturesSet.add(newSignature);
+        const existingEntryByHash = this.hashToBlockMap.get(blockHash);
```
daiagi: logic seems correct
but code is overly complex

here is a simpler path suggestion

1. get the coordinates of incoming block (supplied or decoded)
2. get the block from the map at that coordinate
3. compare `incomingBlock.encodedBlock == existingBlock.encodedBlock` 
4. if they are equal => merge signatures, not equal => abort

you don't  need to touch the `by hash` map (object in both map are the same object, by reference)

Don't forget to add tests please
By the way this is  a good test case to add:

- Change a block by coordinate
- Get that block by hash
- Assert that the changes are also applied on  the `by hash` block

this will protect us from someone in the future braking the ref equality on which we rely

src/storage/StateSnapshotStorage.ts
daiagi: snapshot storage doesn't have a  dual mapping,  i don't think you need to change anything here
I think this issue is only about `BlockStorage`

daiagi (changes requested):

src/storage/BlockStorage.ts
cyony95: you are absolutely right, thanks

src/storage/StateSnapshotStorage.ts
cyony95: I removed the changes to it.

src/storage/BlockStorage.ts
```diff
@@ -271,29 +274,42 @@ export class BlockStorage {
...
+                    existingEntry.blockConfirmation.signatures
+                );
+                for (const newSignature of blockEntry.blockConfirmation
+                    .signatures) {
+                    signaturesSet.add(newSignature);
+                }
+                existingEntry.blockConfirmation.signatures =
+                    Array.from(signaturesSet);
+
+                // Update on-chain timestamp if provided
```
cyony95: @daiagi @lukastanisic99  should I update this?

test/storage/BlockStorage.test.ts
```diff
@@ -420,4 +420,190 @@ describe("BlockStorage", () => {
...
+                const hash1 = storage.storeBlockConfirmation(blockConf1);
+                const hash2 = storage.storeBlockConfirmation(blockConf2);
+
+                expect(hash1).to.equal(hash2);
+
+                const stored = storage.getBlockEntry(hash1);
+                expect(stored?.blockConfirmation.signatures).to.have.lengthOf(
+                    3
+                );
+            });
```
daiagi: these two cases are covered already
```suggestion
```

test/storage/BlockStorage.test.ts
```diff
@@ -420,4 +420,190 @@ describe("BlockStorage", () => {
...
+        });
+
+        describe("Different blocks with same coordinates", () => {
+            it("should return undefined when storing different blocks with same coordinates", () => {
+                // Store first block
+                const hash1 = storage.storeBlockConfirmation(
+                    mockBlockConfirmation
+                );
+                expect(hash1).to.equal(mockBlockHash);
```
daiagi: redundant, tested before
```suggestion
```

test/storage/BlockStorage.test.ts
```diff
@@ -420,4 +420,190 @@ describe("BlockStorage", () => {
...
+            });
+        });
+
+        describe("Different blocks with same hash but different coordinates", () => {
+            it("should return hash when storing block with same hash but different coordinates", () => {
+                // Store first block
+                const hash1 = storage.storeBlockConfirmation(
+                    mockBlockConfirmation
+                );
+                expect(hash1).to.equal(mockBlockHash);
```
daiagi:
```suggestion
```

test/storage/BlockStorage.test.ts
```diff
@@ -420,4 +420,190 @@ describe("BlockStorage", () => {
...
+                    height: 999
+                };
+                const result = storage.storeBlockConfirmation(
+                    mockBlockConfirmation,
+                    {
+                        coordinates: differentCoordinates
+                    }
+                );
+
+                expect(result).to.equal(mockBlockHash);
```
daiagi, line 491: it mean that we replaced the coordintates of an already stored block, doesn't it?

test/storage/BlockStorage.test.ts
```diff
@@ -420,4 +420,190 @@ describe("BlockStorage", () => {
...
+                // Should be retrievable by coordinates
+                const byCoords = storage.getBlockEntry(
+                    block.coordinates.forkId,
+                    block.coordinates.height
+                );
+                expect(byCoords?.blockConfirmation).to.equal(
+                    newBlockConfirmation
+                );
+            });
+        });
```
daiagi: repeating tests, delete
```suggestion

```

test/storage/BlockStorage.test.ts
```diff
@@ -420,4 +420,190 @@ describe("BlockStorage", () => {
...
+
+        describe("Reference equality", () => {
+            it("should maintain reference equality between hash and coordinates maps", () => {
+                // Store a block
+                const hash = storage.storeBlockConfirmation(
+                    mockBlockConfirmation
+                );
+                expect(hash).to.equal(mockBlockHash);
```
daiagi:
```suggestion
```

test/storage/BlockStorage.test.ts
```diff
@@ -420,4 +420,190 @@ describe("BlockStorage", () => {
...
+                expect(blockByCoords?.blockConfirmation).to.equal(
+                    mockBlockConfirmation
+                );
+
+                // Change the block by coordinates (add signature)
+                const newSignature = sig();
+                storage.insertSignature(newSignature, mockForkId, mockHeight);
+
+                // Get that block by hash
+                const blockByHash = storage.getBlockEntry(hash);
```
daiagi:
```suggestion
                const blockByHash = storage.getBlockEntry(hash!);
```

src/storage/BlockStorage.ts
```diff
@@ -271,29 +274,42 @@ export class BlockStorage {
...
-            // Merge signatures
-            const signaturesSet = new Set(
-                existingEntry.blockConfirmation.signatures
-            );
-            for (const newSignature of blockEntry.blockConfirmation
-                .signatures) {
-                signaturesSet.add(newSignature);
+        const existingEntry = this.coordinatesToBlockMap.get(coordinateKey);
+
+        if (existingEntry) {
```
daiagi: nitpick on code style:

it is simpler to read:

```ts
if (!existingEntry) {
    // short path
    return hashOfNewEntry
}
...
if (incomingEncodedBlock !== existingEncodedBlock) return undefeined

// now comes the "long" logic
```

daiagi (changes requested):
requesting changes due to:
1. the test `should return hash when storing block with same hash but different coordinates` reveals a logic error 
2. 4 redundant tests, these have been covered before

src/storage/BlockStorage.ts
cyony95: not a nitpick, it's valuable advice

src/storage/BlockStorage.ts
daiagi: yes, i think it would be better if:
- if `existingEntry.onChainTimestamp === undefined` => replace unconditionally (even if `blockEntry.onChainTimestamp` is undefined)
- else, replace only if `blockEntry.onChainTimestamp !== undefined && blockEntry.onChainTimestamp > existingEntry.onChainTimestamp`

test/storage/BlockStorage.test.ts
daiagi: not sure about this one
the trello card says to handle the case where confidantes are same but block is diff - which you did
and higher logic (aka a module that uses storage, such as state manger) would already check about "same block but not same coordinates" and reject  / handle it

not sure if block storage should consider this. probably not.
@lukastanisic99  wdyt?

daiagi (approved):

test/storage/BlockStorage.test.ts
lukastanisic99: In this case it means that a block will be stored and retrievable with 2 different `coordinates` -> ok for test but this shouldn't be the case anywhere in the code - we'll assume for now that our internal modules `are not lying` so we don't actually have to check in storage are the coordinates good for the given block since in that case why do the optimization of providing them in the first place

lukastanisic99 (approved):
looks ok - didn't think too much about test are some missing - will think more about them later both unit, integration and system testing

PR #117, opened 2025-07-17
DI Container

daiagi:
This PR offers an idea how to untangle the mess of passing class instances around WHILE maintaining an easy enough interface to isolate and test classes

Basically it offers an almost singleton feel to production code
while allowing test to mock dependencies

I am not promising the code here is without errors or that is perfect.
See it as a Experimental PR, opening the room for discussion.

If you like the direction, then I'll continue and fine tune it, otherwise,let's discuss.
Worst case we ditch this PR

see the AgreementManager test to get a feel for it


lukastanisic99:
Took a quick look at lunch - if I understood correctly, your concern with the normal singleton is that you can't easily inject a `mock` and test it in isolation?

My bigger concern with singletons in testing is that tests run async and singletons have a single global reference (that's shared across tests since it's the same process/memory) -> tests with singletons/global variables are `not isolated` from one another unless:
`1)` they don't use async code (are `sync`) and you cleanup after each test / reset the singleton
`2)` you use a `mutex` which turns async code into sync by not allowing another context to start execution until the mutex is released. -> same as `1)` 
I don't want to say `concurrent/parallel` since nodejs is single threaded, but since it's async while 1 context is `waiting` another context is executing (context switch). Mental model - think of starting multiple threads on a single core CPU. The `container` is still a singleton and since it instantiates other instances and checks are the instances already instantiated/mocked acting as a `cache` - it doesn't solve the problem above.

You would need to isolate the memory - e.g. run every test in a `WebWorker` which is a separate process (separate memory) and let the OS handle the "parallel" execution by performing the context switch on the OS level - this would work but is memory heavy since you're creating N processes where N is the number of tests that are `running` in parallel.

Without singletons every test creates its own instance which is not shared in other tests -> they can't touch that memory -> `isolated`

daiagi:
Yes, Singletons and testing do not mix well with together 

My proposal is not to use singleton in testing.
Tests create mock and put it into the DI container. The tested component reads it at runtime via the "inject" function


It is as simple as mocking in a "before" or "beforEach" hook and cleanup  the mock in the after / afterEach hook

test/agreementManager/AgreementManager.test.ts
```diff
@@ -0,0 +1,109 @@
...
+                getBlocksByForkId: sinon
+                    .stub()
+                    .returns([{ blockConfirmation: mockBlockConfirmation }])
+            }
+        };
+
+        // Get the real storage from container before mocking
+        const realStorage = inject(ServiceNames.STORAGE);
+        // put mock storage in container
+        container.mock(ServiceNames.STORAGE, mockStorage);
```
lukastanisic99, line 51: `container` is a shared global object, `ServiceNames.STORAGE` is constant -> if 2 tests that run in "parallel" define different `storageMocks` (e.g. `mock1` and `mock2`) and both call `container.mock(ServiceNames.STORAGE, storageMocks)` , with their respective mocks (`mock1` and `mock2`), they will override the object in the map (`cache`). E.g. `test 1` when it tries to `resolve` it may end up using `mock2` and vice versa, do to "parallel" execution modifying a `shared/global object`

-> same problem as with singletons -> they are singletons

lukastanisic99: Imo this is all equivalent to everything being a singleton and having a `singleton/global mockMap` and every `getInstance()` returning the mock if the `specificKey/name` is set otherwise just acting as a normal singleton and returning the instance 

In that case better to just implement `getInstance(obj?)` where when called for the first time it sets the `instance` to the `obj` if provided, else runs the normal instance creation (what you call factory) -> you'd have the same behavior as now 

And both have the same problem of `shared objects` being used across tests

daiagi:
image

https://mochajs.org/#behavior-driven-development-interface
___

from Hardat [docs](https://hardhat.org/hardhat-runner/docs/guides/test-contracts#running-tests-in-parallel):

"[Running tests in parallel](https://hardhat.org/hardhat-runner/docs/guides/test-contracts#running-tests-in-parallel)

You can run your tests in parallel by using the --parallel flag:

npx hardhat test --parallel

Alternatively, use parallel: true in the mocha section of your Hardhat config."

----

**So unless explicitly activated, test are ran serially**

So no, tests are not ran in parallel, and if they can be, then we simply should not
because the benefit of shaving of few seconds for test run time is much less then the benefit of removing manual dependency injection while being able to test neatly

daiagi:
do you mean something like this?
``` ts
class Storage {
    private static instance: Storage | null = null;
    private static mockInstance: Storage | null = null;

    static getInstance(mock?: Storage): Storage {
        // If mock provided, use it
        if (mock) {
            this.mockInstance = mock;
            return mock;
        }
        
        // If we have a mock set, return it
        if (this.mockInstance) {
            return this.mockInstance;
        }
        
        // Otherwise, create/return real instance
        if (!this.instance) {
            this.instance = new Storage();
        }
        return this.instance;
    }
}
```


but then the `AgreementManager` will obviously call `Storage.getInstance()`
and how are you going to test `AgreementManager` with mock storage?


you're going to need to instantiave agrreement manger with a mock storage instance => back to manual wiring => no benefit

lukastanisic99:
> image
> https://mochajs.org/#behavior-driven-development-interface
> 
> so no, tests are not ran in parallel, and if they can be, then we simply should not because the benefit of shaving of few seconds for test run time is much less then the benefit of removing manual dependency injection while being able to test neatly

Can you test this with `2 async tests` - `test1` that logs immediately something like "test 1 start" and goes to sleep for some time (yields its execution context) and after logs "test 1 end" and `test2` can just log "test 2 start". I remember when we were running compound test and there were A LOT and they we're literally executing for minutes and they all ran `async` ("in parallel") - I see mocha says they're async, so not sure what it means by `serially` - does it apply a `mutex` to every test ?!

Best to test with a simple test

lukastanisic99:
> do you mean something like this?
> 
> ```ts
> class Storage {
>     private static instance: Storage | null = null;
>     private static mockInstance: Storage | null = null;
> 
>     static getInstance(mock?: Storage): Storage {
>         // If mock provided, use it
>         if (mock) {
>             this.mockInstance = mock;
>             return mock;
>         }
>         
>         // If we have a mock set, return it
>         if (this.mockInstance) {
>             return this.mockInstance;
>         }
>         
>         // Otherwise, create/return real instance
>         if (!this.instance) {
>             this.instance = new Storage();
>         }
>         return this.instance;
>     }
> }
> ```
> 
> but then the `AgreementManager` will obviously call `Storage.getInstance()` and how are you going to test `AgreementManager` with mock storage?
> 
> you're going to need to instantiave agrreement manger with a mock storage instance => back to manual wiring => no benefit

Yes and since you have to write your `storage mock` before interacting with `AgreementManager` you also run `Storage.getInstance(mock)` to set it 

What I don't like about both of these approaches is that sdk code has `mocks` within it - mocks should be something that can be injected during testing without the `src` code accounting for it 

Plus this is all reliant on `IF` tests are actually sequential and if we end up having more test we'll want to make them `parallel` eventually

daiagi:
you are overthinking it
just tests run one after the other. simple as that
`async` means you can write `async/await` tests => aka the function you pass to `it` can return a promise

```bash
yarn test test/parallel_test*.ts
yarn run v1.22.22
warning package.json: "dependencies" has dependency "@nomicfoundation/hardhat-toolbox" with range "^4.0.0" that collides with a dependency in "devDependencies" of the same name with version "^5.0.0"
$ yarn hardhat test test/parallel_test2.ts test/parallel_test.ts
warning package.json: "dependencies" has dependency "@nomicfoundation/hardhat-toolbox" with range "^4.0.0" that collides with a dependency in "devDependencies" of the same name with version "^5.0.0"
$ /home/luke/Documents/peer3/state-channels-plus/node_modules/.bin/hardhat test test/parallel_test2.ts test/parallel_test.ts


  Second Parallel Test File - Suite A
file2 test 1 start - 2025-07-18T05:55:44.460Z
file2 test 1 end - 2025-07-18T05:55:46.963Z
    ✔ file2 test 1 - should sleep for 2.5 seconds (2505ms)
file2 test 2 start - 2025-07-18T05:55:46.967Z
file2 test 2 end - 2025-07-18T05:55:46.967Z
    ✔ file2 test 2 - should execute immediately
file2 test 3 start - 2025-07-18T05:55:46.968Z
file2 test 3 end - 2025-07-18T05:55:48.470Z
    ✔ file2 test 3 - should sleep for 1.5 seconds (1503ms)

  Second Parallel Test File - Suite B
file2 suite B test 1 start - 2025-07-18T05:55:48.473Z
file2 suite B test 1 end - 2025-07-18T05:55:48.973Z
    ✔ file2 suite B test 1 - should sleep for 0.5 seconds (501ms)
file2 suite B test 2 start - 2025-07-18T05:55:48.974Z
file2 suite B test 2 end - 2025-07-18T05:55:48.974Z
    ✔ file2 suite B test 2 - should execute immediately

  Parallel vs Serial Test Execution
test 1 start - 2025-07-18T05:55:48.975Z
test 1 end - 2025-07-18T05:55:51.980Z
    ✔ test 1 - should sleep for 3 seconds (3005ms)
test 2 start - 2025-07-18T05:55:51.981Z
test 2 end - 2025-07-18T05:55:51.981Z
    ✔ test 2 - should execute immediately
test 3 start - 2025-07-18T05:55:51.982Z
test 3 end - 2025-07-18T05:55:53.983Z
    ✔ test 3 - should sleep for 2 seconds (2002ms)
test 4 start - 2025-07-18T05:55:53.985Z
test 4 end - 2025-07-18T05:55:53.985Z
    ✔ test 4 - should execute immediately

  Additional Parallel Test Suite
parallel suite test 1 start - 2025-07-18T05:55:53.986Z
parallel suite test 1 end - 2025-07-18T05:55:54.988Z
    ✔ parallel suite test 1 - should sleep for 1 second (1002ms)
parallel suite test 2 start - 2025-07-18T05:55:54.989Z
parallel suite test 2 end - 2025-07-18T05:55:54.989Z
    ✔ parallel suite test 2 - should execute immediately


  11 passing (11s)
  ```
  
  >mocks within it - mocks should be something that can be injected during testing without the src code accounting for it
  
this is exactly what I'm suggesting
truly don't see the issue

the mocks are defined by test code and utilized there only
the SDK code just knows that it pulls a dependency via `inject`

daiagi:
closing

PR #118, opened 2025-07-18
Update-tests

daiagi:
- Closes https://trello.com/c/kUVIqaG3
- Follow up on #115

Add / update tests
Fix discovered bugs


cyony95 (approved):

daiagi:
@lukastanisic99 
Essence of the changes here:

- Assuming the the genesis snapshot of a fork corresponds to block height 0 => first block on a fork has height = 1
aka if requesting a state snapshot of block at height 0 => get the genesis snapshot 

this makes sense but i don't think we have clear defined rule for it across the code -> room for bugs
- technical: exclude generators / iterators from the deep copy proxy, naturally , a generator cannot be deep copied


these are pretty trivial stuff so i just merge, don't think it is worth your time reviewing

lukastanisic99:
> @lukastanisic99 Essence of the changes here:
> 
> * Assuming the the genesis snapshot of a fork corresponds to block height 0 => first block on a fork has height = 1
>   aka if requesting a state snapshot of block at height 0 => get the genesis snapshot
> 
> this makes sense but i don't think we have clear defined rule for it across the code -> room for bugs
> 
> * technical: exclude generators / iterators from the deep copy proxy, naturally , a generator cannot be deep copied
> 
> these are pretty trivial stuff so i just merge, don't think it is worth your time reviewing

I'd keep separate as blocks[ ] only referencing block logic and if you want the genesisSnapshot you explicitly need ask for it - I know you did `-1` which I was thinking should I comment, but left it like that since it's isolated and is something that's unusuall so it's like an 'explicit action' + doesn't affect the logic of the rest of the system so imo it's better to have it like that - [0...N] block, -1 genesis

PR #119, opened 2025-07-21
SDK Diagrams - Initial

lukastanisic99:
Initial Diagrams with Draw.io extension for VS Code. 

Diagrams:
`- User Interaction` - how the user/consumer uses the SDK vs what they would normally do. 
`- Architecture Overview` -  How does the system look like from a high level
`- Setup/Vm` - Explaining the flow what happens when the user calls `setup` to use the SDK and EvmStateMachine / VM or whatever we end up calling this - see notes. 
`- Dependency Graph` - Typescript components dependency graph (not including internal components like `ValidationService` which is part of `StateManager`). Also something to consider as an alternative to singletons that feels like singletons, but is not and is simple to reason and implement - explained there. 

In general read the collapsable notes. I'll continue in the morning and on the plane. I have my own ideas how to describe these systems, but feel free to add your suggestions/wishes what you'd like to see @daiagi @cyony95

### DIFF
--- a/diagrams/README.md
+++ b/diagrams/README.md
@@ -1,0 +1,3 @@
+# SDK Diagrams 
+
+Created with the [Draw.io](https://marketplace.visualstudio.com/items?itemName=hediet.vscode-drawio) extension for Visual Studio Code. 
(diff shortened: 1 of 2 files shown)


cyony95:
@lukastanisic99 just to make sure we're on the same page, this si how we want it to work, not necessarily how it is working 100% right now, correct?

lukastanisic99:
> @lukastanisic99 just to make sure we're on the same page, this is how we want it to work, not necessarily how it is working 100% right now, correct?

Yes, but most of the things there are already like that. In cases where that's not the case I say something 'for now ...' and it's mostly just naming. I didn't get to the low level stuff that is yet to be done and that diverges. If you have questions/doubts about any specific part, feel free to ping me

cyony95 (approved):

daiagi (approved):

PR #121, opened 2025-07-21
decode custom solidity errors

daiagi:
- Closes  https://trello.com/c/YpyemO4X


src/utils/ContractErrors.ts
```diff
@@ -0,0 +1,83 @@
+// All errors from contracts/V1/StateChannelDiamondProxy/Errors.sol
+
+export enum ContractErrors {
```
daiagi: I know, not too great to list it manually here instead of generating it from solidity
couldn't find a way to do so, open to ideas

lukastanisic99: About to takeoff so will look later - take a look at the generated `artifacts` and/or generated `typechain` types (it often has more than just types) - The artifacts should have the full interface so it should be listed there and since we're going to deploy the `diamond` universally locally we can take those generated artifacts and include them in the build since we're going to use them anyway

daiagi: done

src/utils/ContractErrors.ts
```diff
@@ -0,0 +1,29 @@
...
+    "DisputeFraudProofFacet",
+    "FraudProofFacet",
+    "JoinChannelFacet",
+    "StateChannelCommon",
+    "StateSnapshotFacet"
+];
+export const artifacts = facets.map((facet) => {
+    return require(`${artifacts_base_path}/${facet}.sol/${facet}.json`);
+});
+
```
lukastanisic99: We should include the artifacts (their json abi) in the build so it's available, since `require` would fail if installed through npm

src/stateManager/StateManager.ts
```diff
@@ -482,11 +484,34 @@ class StateManager {
...
             console.log("Posting calldata on chain!");
             this.p2pEventHooks.onPostingCalldata?.();
-            this.stateChannelManagerContract
-                .postBlockCalldata(signedBlock, Clock.getTimeInSeconds())
-                .then((txResponse) => txResponse.wait())
+
+            await call(() =>
+                this.stateChannelManagerContract.postBlockCalldata(
+                    signedBlock,
+                    Clock.getTimeInSeconds()
```
lukastanisic99: `maxTimestamp` that's passed to `postBlockCalldata` is `previousBlock.getRelevantTimestamp() + p2pTime + agreementTime + chainFallbacktime`

src/utils/contractCaller.ts
```diff
@@ -0,0 +1,56 @@
...
+        return await contractCall();
+    } catch (error: any) {
+        const customError = error.data ? decodeCustomError(error.data) : null;
+
+        if (customError) {
+            throw new CustomContractError(customError, error);
+        }
+
+        throw error;
+    }
```
lukastanisic99: I don't like this call function - Imo we shouldn't change the normal usage or flow of interacting with the contract and if we do better to do with `proxies` - maybe I'm biased towards proxies :D 

I like the idea of having a `CustomEvmError` class that we use internally within the SDK and throw that instead of what ethers throws - so we can inject that transformation with proxies and/or just transform `error.data` into our `CustomEvmError` when interacting with `EVM.call` directly -> and then we just interpret our own `CustomEvmError` that's already decoded

So we don't break any logical flow - we use everything the same internally. If we want to catch the error we catch it and do something (only difference it's now always decoded), else let it throw and log its description 

Imo all of this code here and the code from `ContractErrors` should be in one place `CustomEvmError` and we don't need the ContractError.enum - we'll just interpret our decoded custom error

src/utils/contractCaller.ts
```diff
@@ -0,0 +1,56 @@
+import { ErrorDescription, ethers } from "ethers";
+import { errorAbis } from "./ContractErrors";
+import { Bytes } from "@/types/types";
+
+// interface for parsing errors
+const errorInterface = new ethers.Interface(errorAbis);
+
+export interface CustomError {
```
lukastanisic99: also not sure if this is used anywhere

lukastanisic99 (changes requested):
Previous comments

src/utils/contractCaller.ts
daiagi: not a very good ideas i think
with proxy we will lose the type safety and autocomplete when interacting with the proxied contract
this will drag us down a hole of rebuilding the method surface of the contract OR not having autocomplete

i think you are indeed a bit biased towards proxies

src/stateManager/StateManager.ts
daiagi: ok... that's kinda besides the point
i changes this code as a usage example

the point of this PR is not to fix State Manger  pre-existing errors

src/utils/contractCaller.ts
lukastanisic99: Why do you think we lose type safety or autocomplete? Everything stays the same - take a look at debug proxy 

The only thing that the proxy does is catch the ethers error -> decode it and throw our error 

Everything else stays the same 

Only if we do a try catch we expect our error instead of ethers serialized error

src/utils/ContractErrors.ts
daiagi: done

src/utils/contractCaller.ts
daiagi: ill give it a shot. see how it goes

src/utils/contractCaller.ts
daiagi: removed

lukastanisic99 (changes requested):
Everything looks good, but missing throwing our `CustomEvmError` if the local `evm.runCall` errors/fails - we're mostly interested in this so we can know should we try and fetch missing data from the chain or not

daiagi:
>missing throwing our CustomEvmError

should be good to go

lukastanisic99 (approved):

PR #124, opened 2025-07-25
Universal deployment

daiagi:
@lukastanisic99  this is here i am at the moment

i got a failing test due to code size too large
please advice

```bash
Universal Deployment
  ▼ Local Diamond
    ✖ deploys successfully
Error: Failed to deploy in EVM code size to deposit exceeds maximum code size
    at deployLocal (scripts/V1/deploy.ts:37:15)
    at async deployFacet (scripts/V1/deploy.ts:88:16)
    at async Promise.all (index 0)
    at async deployAllFacets (scripts/V1/deploy.ts:124:9)
    at async deployLocalDiamond (scripts/V1/deploy.ts:185:20)
    at async Context.<anonymous> (test/V1/UniversalDeployment.test.ts:38:36)
```

### DIFF
--- a/contracts/V1/StateChannelDiamondProxy/AStateChannelManagerProxy.sol
+++ b/contracts/V1/StateChannelDiamondProxy/AStateChannelManagerProxy.sol
@@ -3,7 +3,7 @@ pragma solidity ^0.8.8;
 import "./StateChannelCommon.sol";
 import "../StateChannelManagerInterface.sol";
 import "./StateChannelUtilLibrary.sol";
-import "./ConsumerFacet.sol";
+import "./AConsumerFacet.sol";
 
 import "./DisputeManagerFacet.sol";
 import "./FraudProofFacet.sol";
@@ -17,7 +17,7 @@ contract AStateChannelManagerProxy is StateChannelManagerInterface, StateChannel
     DisputeFraudProofFacet disputeFraudProofFacet;
     StateSnapshotFacet stateSnapshotFacet;
     JoinChannelFacet joinChannelFacet;
-    ConsumerFacet consumerFacet;
+    AConsumerFacet consumerFacet;
 
     constructor(
         address _stateMachineImplementation,
@@ -34,7 +34,7 @@ contract AStateChannelManagerProxy is StateChannelManagerInterface, StateChannel
         disputeFraudProofFacet = DisputeFraudProofFacet(_disputeFraudProofFacet);
         stateSnapshotFacet = StateSnapshotFacet(_stateSnapshotFacet);
         joinChannelFacet = JoinChannelFacet(_joinChannelFacet);
-        consumerFacet = ConsumerFacet(_consumerFacet);
+        consumerFacet = AConsumerFacet(_consumerFacet);
         p2pTime = 15;
         agreementTime = 5;
         chainFallbackTime = 30;
--- a/contracts/V1/StateChannelDiamondProxy/AStateChannelManagerProxy.sol
+++ b/contracts/V1/StateChannelDiamondProxy/AStateChannelManagerProxy.sol
@@ -3,6 +3,7 @@ pragma solidity ^0.8.8;
 import "./StateChannelCommon.sol";
 import "../StateChannelManagerInterface.sol";
 import "./StateChannelUtilLibrary.sol";
+import "./ConsumerFacet.sol";
 
 import "./DisputeManagerFacet.sol";
 import "./FraudProofFacet.sol";
@@ -10,12 +11,13 @@ import "./DisputeFraudProofFacet.sol";
 import "./StateSnapshotFacet.sol";
 import "./JoinChannelFacet.sol";
 
-abstract contract AStateChannelManagerProxy is StateChannelManagerInterface, StateChannelCommon {
+contract AStateChannelManagerProxy is StateChannelManagerInterface, StateChannelCommon {
     DisputeManagerFacet disputeManagerFacet;
     FraudProofFacet fraudProofFacet;
     DisputeFraudProofFacet disputeFraudProofFacet;
     StateSnapshotFacet stateSnapshotFacet;
     JoinChannelFacet joinChannelFacet;
+    ConsumerFacet consumerFacet;
 
     constructor(
         address _stateMachineImplementation,
@@ -23,7 +25,8 @@ abstract contract AStateChannelManagerProxy is StateChannelManagerInterface, Sta
         address _fraudProofFacet,
         address _disputeFraudProofFacet,
         address _stateSnapshotFacet,
-        address _joinChannelFacet
+        address _joinChannelFacet,
+        address _consumerFacet
     ) {
         stateMachineImplementation = AStateMachine(_stateMachineImplementation);
         disputeManagerFacet = DisputeManagerFacet(_disputeManagerFacet);
@@ -31,6 +34,7 @@ abstract contract AStateChannelManagerProxy is StateChannelManagerInterface, Sta
         disputeFraudProofFacet = DisputeFraudProofFacet(_disputeFraudProofFacet);
         stateSnapshotFacet = StateSnapshotFacet(_stateSnapshotFacet);
         joinChannelFacet = JoinChannelFacet(_joinChannelFacet);
+        consumerFacet = ConsumerFacet(_consumerFacet);
         p2pTime = 15;
         agreementTime = 5;
         chainFallbackTime = 30;
@@ -68,6 +72,41 @@ abstract contract AStateChannelManagerProxy is StateChannelManagerInterface, Sta
         blockCalldataCommitments[channelId][msg.sender][forkId][transactionCnt] = commitment;
 
         emit BlockCalldataPosted(_block.transaction.header.channelId, msg.sender, signedBlock, block.timestamp);
+    }
+
+    // ********** Consumer Facet Delegation Functions **********
+
+    function openChannel(bytes32 channelId, bytes[] calldata openChannelData, bytes[] calldata signatures)
+        public
+        virtual
+        override
+    {
+        require(!isChannelOpen(channelId), "AStateChannelManagerProxy: openChannel - channel already open");
+        consumerFacet.openChannel(channelId, openChannelData, signatures);
+    }
+
+    function closeChannel(bytes32 channelId, bytes[] calldata closeChannelData, bytes[] calldata signatures)
+        public
+        virtual
+        override
+    {
+        consumerFacet.closeChannel(channelId, closeChannelData, signatures);
+    }
+
+    function removeParticipant(bytes32 channelId, bytes[] calldata removeParticipantData, bytes[] calldata signatures)
+        public
+        virtual
+        override
+    {
+        consumerFacet.removeParticipant(channelId, removeParticipantData, signatures);
+    }
+
+    function addParticipant(bytes32 channelId, bytes[] calldata addParticipantData, bytes[] calldata signatures)
+        public
+        virtual
+        override
+    {
+        consumerFacet.addParticipant(channelId, addParticipantData, signatures);
     }
 
     function uploadDispute(DisputeConfirmation memory disputeConfirmation) public override {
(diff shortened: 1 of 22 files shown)


scripts/V1/deploy.ts
```diff
@@ -0,0 +1,211 @@
...
+};
+
+function linkBytecode(bytecode: string, libraryAddress: string): string {
+    const placeholder = /__\$[a-f0-9]{34}\$__/g;
+    return bytecode.replace(placeholder, libraryAddress.slice(2)); // Remove 0x prefix
+}
+
+async function deployLocal(bytecode: string, evm: EVM): Promise<string> {
+    const result = await evm.runCall({
+        data: ethers.getBytes(bytecode)
```
lukastanisic99: not sure this is the correct way to do it - you can't just plug the bytecode. Take a look at this - https://github.com/peer3to/state-channels-plus/blob/9972328fcd0f3c0efdeb7cb46fba98b46ca5cc45/examples/TicTacToe/tic-tac-toe-vite/src/stateChannel/TicTacToeStateChannel.ts#L53

This is how we deploy the state machine - take a look at this and understand the lifecycle of the deployment tx - you don't need any of this `linkBytecode` stuff everything is in the `artifacts` and you just cast it to the typechain generated type so typescript helps with constructor arguments if any. 

To do a universal script, you'd need to take a signer as an argument (which you already do) but for local deployment you need to inject the signer that will redirect to the local evm and know how to process the tx receipt/response that you normally get from an RPC node - which `EVM.call` doesn't have the same format 

So best to have 2 scripts `deployLocal` which takes a `signer` but doesn't do `deploy` but `getDeployTransaction` and you plug it into to the EVM and get an address out of it. 

The normal `deploy` which we expose to the client can interact normally since it's interacting with RPC nodes and getting a transactionReposne/Receipt  from the RPC nodes.

You can still make it work as a universal script, but probably less work to have 2 separate scripts and expose only 1 to the client for them to use.

daiagi:
i pushed a NON WORKING version
>you don't need any of this linkBytecode stuff 

you say that and you should definitely be right, because regexp replacement is BS
BUT...  it simply doesn't work
everywhere we have the syntax:
```ts
    let disputeManagerFacetFactory = await ethers.getContractFactory(
        "DisputeManagerFacet",
        { libraries: { StateChannelUtilLibrary: libraryAddress } }
    );
```


we use `hardhat`


i have dug up docs and tries with this or that AI
and looked at `ethers.ContractFactory` code as well. 
did not find any util does does it nicely

as far as i see, manual replacement of lib code is the only way, and it DID work

looking forward for you to prove me wrong, i much prefer not to have to do manual byte replacements


except for this issue, i believe it is now correct


https://github.com/ethers-io/ethers.js/issues/195#issuecomment-396350174
https://github.com/ethers-io/ethers.js/issues/195#issuecomment-833432293

test/V1/UniversalDeployment.test.ts
```diff
@@ -0,0 +1,137 @@
...
+        // Get the deployment transaction data
+        const deployTx = await MathStateMachine.getDeployTransaction(5000000);
+        mathStateMachineBytecode = deployTx.data || "0x";
+    });
+
+    describe("Local Diamond", () => {
+        it("deploys successfully", async () => {
+            const diamondAddress = await deployLocalDiamond(
+                mathStateMachineBytecode,
+                evm
```
lukastanisic99: Am I missing something or you're using the `evm` as a `signer` ?!

scripts/V1/deploy.ts
```diff
@@ -0,0 +1,124 @@
...
+): Promise<{ address: string; contract: LocalDiamond }> {
+    const { address: libAddress } = await deployArtifact(
+        StateChannelUtilLibraryArtifact,
+        signer
+    );
+    const facetAddresses = await deployFacets(signer, {
+        StateChannelUtilLibrary: libAddress
+    });
+
+    const { address: stateMachineAddress } = await deployLocalFromTx(
```
lukastanisic99: Imo I think the problem is this is only used for the `StateMachine` deployment and other local deployments use the normal `factory.deploy()` connected with `signer` which won't route the deployment TX to the local EVM unless you've implemented a new signer - you can't substitute the signer with the evm :D so all local deployments should be done with `deployLocalFromTx` where the signer just signs and doesn't forward the tx to the RPC node, so you can manually forward the tx to the evm

contracts/V1/StateChannelDiamondProxy/ConsumerFacet.sol
```diff
@@ -0,0 +1,26 @@
+pragma solidity ^0.8.8;
+
+import "../types/DataTypes.sol";
+import "../StateChannelManagerInterface.sol";
+
+abstract contract ConsumerFacet {
```
lukastanisic99: `AConsumerFacet`

contracts/V1/StateChannelDiamondProxy/AStateChannelManagerProxy.sol
```diff
@@ -3,34 +3,38 @@ pragma solidity ^0.8.8;
...
+import "./ConsumerFacet.sol";
 
 import "./DisputeManagerFacet.sol";
 import "./FraudProofFacet.sol";
 import "./DisputeFraudProofFacet.sol";
 import "./StateSnapshotFacet.sol";
 import "./JoinChannelFacet.sol";
 
-abstract contract AStateChannelManagerProxy is StateChannelManagerInterface, StateChannelCommon {
+contract AStateChannelManagerProxy is StateChannelManagerInterface, StateChannelCommon {
```
lukastanisic99: `StateChannelManagerProxy`

test/V1/UniversalDeployment.test.ts
daiagi: It is outdated test code. I said the code is not working. This is not the reason why

scripts/V1/deploy.ts
daiagi: I was not trying to replace signer with evm.
I gave clear comment about the issue.
The issue Is library linking.
The facets deploy fail.you said no to manual linking but I have no other solution.
For examples. The into Al deeply of the util library runs. Because... No linking

Ethers documentation has nothing about linking either

daiagi: to prove my point i pushed test file that does not touch the local deploy
so let's put that aside for a moment

```bash
DisputeManagerFacet
▼ Universal Deployment
  ▼ consumer facet Deployment
    ✖ deploys with consumer facet
TypeError: invalid BytesLike value (argument="value", value="0x6080806040523460155761646f908161001b8239f35b600080fdfe6101408060405260043610156...
...
 at deployArtifact (scripts/V1/deploy.ts:49:21)
    at deployFacets (scripts/V1/deploy.ts:79:35)
    at deploy (scripts/V1/deploy.ts:97:34)
```

lukastanisic99: Yeah not sure why ethers didn't account for libraries if they accounted for contract factories - their docs really don't mention it and it seems this is a hardhat specific feature. 

4 options:
- take a look what hardhat does under the hood and use that (maybe it uses also the regex injection), so using the regex injection is fine for now IF IT WORKS - if you decide on this approach,  create a trello card for us to investigate this deeper when other higher priority tasks are completed
- instead of using libraries we can use a utility facet, like our other facets (it's the same stuff, but the contract address is not hardcoded into the bytecode, but it's in storage - it uses delegate call either way under the hood) - this is probably the ugliest and most amount of work since since all interaction has to happen through the diamond
- use `solidity free functions` since all the stuff defined in the library is stateless, so you can just make it part of the bytecode of every contract that's using it - the disadvantage is the bytecode of other contracts are larger 
- Use hardhat for the deployment scripts (probably you'll need to include the entire `artifacts` directory in the build for it to work when installed over npm)

lukastanisic99: Imo long term we shouldn't use libraries regardless since they're not upgradable and facets are - for now doesn't matter much

daiagi: https://github.com/peer3to/state-channels-plus/pull/124/commits/66a635dcaa057c065fdccc4eb46bba368f7b51de

Issue is done as far as I'm considered
no need for a Trello card

it works, it good interface and it does like hardhat does
https://github.com/NomicFoundation/hardhat/blob/294ef52103c850e3d3a2ca072522272daee1150a/packages/hardhat-ethers/src/internal/helpers.ts#L435

contracts/V1/StateChannelDiamondProxy/LocalDiamond.sol
```diff
@@ -0,0 +1,85 @@
...
+            bytes32 slot = slots[i];
+            bytes32 value;
+            assembly {
+                value := sload(slot)
+            }
+            values[i] = value;
+        }
+
+        return values;
+    }
```
lukastanisic99, line 83: This is fine for now, but we'll use explicit setters and getters (e.g. `setBlockCalldataCommitment`) but no need to do it now, as we implement `syncing` we'll populate and use the appropriate interface

scripts/V1/deploy.ts
```diff
@@ -0,0 +1,158 @@
...
+    AStateChannelManagerProxy
+} from "@typechain-types/index";
+import { Artifact } from "hardhat/types";
+
+export async function deployLocalFromTx<T extends Contract>(
+    tx: ContractDeployTransaction,
+    signer: Signer
+): Promise<{ address: string; contract?: T }> {
+    if (!signer) throw new Error("Signer required for deployment");
+    const sentTx = await signer.sendTransaction(tx);
```
lukastanisic99: do we have a special signer that sends this to our local evm (like `p2pSigner`)? This will always deploy to the RPC node

lukastanisic99: Would be great to reuse `p2pSigner` and then have a single deployment script, but probably more work for now than having 2 separate

contracts/V1/StateChannelDiamondProxy/AStateChannelManagerProxy.sol
daiagi: this is a big change that causes circular dependency issue
ill put it in separate pr to not muddy the water #126

scripts/V1/deploy.ts
daiagi: https://trello.com/c/3hHOkeva

daiagi:
>you'd need to take a signer as an argument (which you already do) but for local deployment you need to inject the signer that will redirect to the local evm

actually, why?

i changed it to be signer based due to your previous comments
why no do it like `createStandalone` does it?
```ts
        // Deploy the state machine contract
        const deploymentResult = await evm.runCall({
            data: ethers.getBytes(deployStateMachineTx.data)
        });

```

lukastanisic99:
>> To do a universal script, you'd need to take a signer as an argument (which you already do) but for local deployment you need to inject the signer that will redirect to the local evm and know how to process the tx receipt/response that you normally get from an RPC node - which `EVM.call` doesn't have the same format
> >
>> So best to have 2 scripts `deployLocal` which takes a `signer` but doesn't do `deploy` but `getDeployTransaction` and you plug it into to the EVM and get an address out of it.
>> 
> 
> actually, why?
> 
> i changed it to be signer based due to your previous comments why no do it like `createStandalone` does it?
> 
> ```ts
>         // Deploy the state machine contract
>         const deploymentResult = await evm.runCall({
>             data: ethers.getBytes(deployStateMachineTx.data)
>         });
> ```
yes you have to do it like that - if you wanted to do 1 script only that works for any EVM chain you'd take a general signer (the signer includes a provider - RPC endpoint) and the signer signs and sends it to the RPC node. 

In our case it's easier to do 2 scripts - one for the client (the normal one with the signer that also has the provider) and the other one that has just the signer but doesn't send to the provider. It just signs and then you just forward it the same way as in `createStandalone`. You still need some signer to create the `deploymentTX`

As for the Trello card - I didn't go over it yet in depth, but a good idea if it abstracts away the entire RPC node so we can `JUST` use ethers and get a txReceipt/response from the abstracted RPC node - will take a look later

What we're doing already with the `P2PSinger` is similar, it feeds it directly to the EVM, but doesn't return a txReceipt/Response which is the drawback. You can inject a signer like that still and not implement the Trello card and still have just 1 deployment script (that's how our `p2p contract instance` is routed locally).

lukastanisic99:
> As for the Trello card - I didn't go over it yet in depth, but a good idea if it abstracts away the entire RPC node so we can `JUST` use ethers and get a txReceipt/response from the abstracted RPC node - will take a look later
> 
> What we're doing already with the `P2PSinger` is similar, it feeds it directly to the EVM, but doesn't return a txReceipt/Response which is the drawback. You can inject a signer like that still and not implement the Trello card and still have just 1 deployment script (that's how our `p2p contract instance` is routed locally).

We'd still need our `P2PSigner` implementation regardless since it first needs to go through `StateManager` (the pipeline) and then to the EVM and then it needs to go over our RPC so it's gossiped to others

daiagi:
https://trello.com/c/8472PdJp

lukastanisic99 (approved):

PR #127, opened 2025-07-29
Added new logic for off chain exits

cyony95:
closes https://trello.com/c/wncLEYJD/145-agreementmanager-getfinalizedandlatest-getstateproof


cyony95:
still needs the interface. should the agreementmanager classs implement the stateproof interface? @daiagi @lukastanisic99

daiagi:
what is `dispute1` branch?
why is this PR to that branch?

cyony95:
> what is `dispute1` branch? why is this PR to that branch?

my mistake, I made it to make sure there are no conflicts and forgot to change it back

src/agreementManager/AgreementManager.ts
```diff
@@ -104,37 +112,293 @@ class AgreementManager {
...
+        if (!genesisSnapshot) {
             throw new Error("Fork not found");
         }
 
+        // Get all exit points for this fork - these are the milestones
+        const exitPoints = this.storage.exitPoints.getExitPointsInRange(forkId);
+
+        // Get all blocks sorted by height descending
+        const blockEntries = Array.from(
+            this.storage.blocks.getBlocksByForkId(forkId, SortOrder.DESC)
```
lukastanisic99: Is this some olds stuff before iterators? The code defines blockEntires twice

src/agreementManager/AgreementManager.ts
```diff
@@ -104,37 +112,293 @@ class AgreementManager {
...
-        let requiredSignatures = new Set<Address>(thresholdAddresses);
+        let milestoneProofs: MilestoneProofStruct[] = [];
+        let milestoneSnapshots: StateSnapshot[] = [];
+
+        // Start with genesis snapshot for the first milestone
+        let currentSnapshot = genesisSnapshot;
+
+        // Process exit points in chronological order - each exit point is a milestone
+        const sortedExitPoints = [...exitPoints].sort(
+            (a, b) => Number(a) - Number(b)
```
lukastanisic99: `getExitPointsInRange` returns `sorted BlockHeight[ ]` - no need to sort it again

src/agreementManager/AgreementManager.ts
```diff
@@ -104,37 +112,293 @@ class AgreementManager {
...
+        );
+
+        for (const exitPointHeight of sortedExitPoints) {
+            const milestoneData = this.processMilestone(
+                exitPointHeight,
+                blockEntries,
```
lukastanisic99: This shouldn't be a range of blocks, but an iterator starting from the first blocks that should be analyzed

src/agreementManager/AgreementManager.ts
```diff
@@ -104,37 +112,293 @@ class AgreementManager {
...
+            milestoneProofs,
+            milestoneSnapshots
+        };
+    }
+
+    /**
+     * Build a milestone proof for a given set of blocks and participant set
+     */
+    private buildMilestoneProof(
+        blocks: any[],
```
lukastanisic99: why `any`

src/agreementManager/AgreementManager.ts
```diff
@@ -104,37 +112,293 @@ class AgreementManager {
...
+        participantSet: Set<Address>
+    ): MilestoneProofStruct | null {
+        if (blocks.length === 0) {
+            return null;
+        }
+
+        const blockConfirmations: BlockConfirmationStruct[] = [];
+        let requiredSignatures = new Set<Address>(participantSet);
```
lukastanisic99: these are not signatures, these are addresses/participants -> so better rename to `requiredParticipants` or similar

src/agreementManager/AgreementManager.ts
```diff
@@ -104,37 +112,293 @@ class AgreementManager {
...
+
+        return null; // No milestone achieved
+    }
+
+    /**
+     * Process a single milestone at the given exit point height
+     */
+    private processMilestone(
+        exitPointHeight: BlockHeight,
+        blockEntries: any[],
```
lukastanisic99: again why `any` :D and this should be an iterator

src/agreementManager/AgreementManager.ts
```diff
@@ -104,37 +112,293 @@ class AgreementManager {
...
+        milestoneProof: MilestoneProofStruct;
+        snapshot: StateSnapshot;
+    } | null {
+        // Get blocks up to this exit point (milestone)
+        const blocksUpToExit = blockEntries
+            .filter((entry: any) => {
+                const block = Block.decode(
+                    entry.blockConfirmation.signedBlock.encodedBlock
+                );
+                return block.coordinates.height <= exitPointHeight;
```
lukastanisic99: you iterate forward from the `exitPointBlock` not backwards :D 

The idea is to prove `that block` is final and past blocks don't prove future blocks. 

So you start from that block, check signatures and threshold set remaining -> if not N/N (final) you iterate forward and if you don't have anymore forward to iterate and a threshold is not reached -> return null/undefined

src/agreementManager/AgreementManager.ts
```diff
@@ -104,37 +112,293 @@ class AgreementManager {
...
+            .sort((a: any, b: any) => {
+                const blockA = Block.decode(
+                    a.blockConfirmation.signedBlock.encodedBlock
+                );
+                const blockB = Block.decode(
+                    b.blockConfirmation.signedBlock.encodedBlock
+                );
+                return (
+                    Number(blockA.coordinates.height) -
+                    Number(blockB.coordinates.height)
```
lukastanisic99: not needed with iterators

src/agreementManager/AgreementManager.ts
```diff
@@ -104,37 +112,293 @@ class AgreementManager {
...
+        const blockConfirmations: BlockConfirmationStruct[] = [];
+        let requiredSignatures = new Set<Address>(participantSet);
+
+        for (const blockEntry of blocks) {
+            const block = Block.decode(
+                blockEntry.blockConfirmation.signedBlock.encodedBlock
+            );
+            const signersAddresses = block.getSignerAddresses(
+                blockEntry.blockConfirmation.signatures as Signature[]
+            );
```
lukastanisic99: you also have to take into account the signature from the `block author` - `blockConfirmation.SignedBlock` this is actually the most important part since we're here because cooperation failed and at some point the peer that is not cooperating (signing blocks) is forced to be a block author and issuing a valid block with a valid signature or be timedout

src/agreementManager/AgreementManager.ts
```diff
@@ -104,37 +112,293 @@ class AgreementManager {
...
+
+            // Add to block confirmations
+            blockConfirmations.push(blockEntry.blockConfirmation);
```
lukastanisic99: technically this is correct, but you don't want to repeat signatures from participants that you already accounted for in the set so you want to filter for signatures of new participants - so at most any milestone has N signatures to verify on-chain (worse case: N blocks with 1 signature and best case 1 block with N signatures) - like this we'd have to verify almost N*N signatures on-chain which is expensive

src/agreementManager/AgreementManager.ts
```diff
@@ -104,37 +112,293 @@ class AgreementManager {
...
+                currentSnapshot = milestoneData.snapshot;
+            }
+        }
+
+        // Find latest finalized and correct states using the final participant set
+        const finalParticipantSet = new Set<Address>(
+            currentSnapshot.snapshotData.participants as Address[]
+        );
+
+        const virtualVotingResult = this.performVirtualVoting(
```
lukastanisic99: This was already done in `buildMilestoneProof`

src/agreementManager/AgreementManager.ts
```diff
@@ -104,37 +112,293 @@ class AgreementManager {
...
+        return {
+            milestoneProof,
+            snapshot: exitPointSnapshot
+        };
+    }
+
+    /**
+     * Perform virtual voting to find latest finalized and correct states
+     */
+    private performVirtualVoting(
```
lukastanisic99: `buildMilestoneProof` is already doing virtual voting

src/agreementManager/AgreementManager.ts
```diff
@@ -104,37 +112,293 @@ class AgreementManager {
...
+        forkId: ForkId,
+        blockHeight: BlockHeight,
+        signerAddress: Address
+    ): Promise<{
+        encodedLatestFinalizedState: Bytes;
+        encodedLatestCorrectState: Bytes;
+        virtualVotingBlocks: BlockConfirmationStruct[];
+        milestoneProofs: MilestoneProofStruct[];
+        milestoneSnapshots: StateSnapshot[];
+    }> {
```
lukastanisic99: This function should return a `StateProof struct` - you don't need to worry about these other stuff - they're easy to fetch from the `StateProof`. 

Let's break down the problem - let's forget about `getFinalizedAndLatestWithMilestones` and only implement `StateProof` and from a state proof we can easily get everything we need 

Mostly you have all the components here already, but adjust for my comments. Mostly implement it with iterators and you'll see how simple and efficient it is :D 

add an optional argument to getIterator so it looks like `getIterator(forkId, SortOrder, BlockHeight?)` so you can tell it at which block to start and then just construct the stateProof using `exitPoints` and everything else is simple. 

For every `exit point` you `iterate forward`, and after all exit points you `iterate backwards once` for the `latest finalized block` (potential last milestone) and `latest block` (`signedBlock[ ]` of `stateProof`) - you'll have to think here a little what does everything mean - ping me if something is unclear

lukastanisic99 (changes requested):
previous comments

src/agreementManager/AgreementManager.ts
cyony95: I messed up the merge conflict, that s why :(

src/agreementManager/AgreementManager.ts
cyony95: you re absolutely right! I m getting O(N*N) flashbacks on this now that I look at it :)))

src/agreementManager/AgreementManager.ts
cyony95: i definitely missed that part but it makes sense, I'm so damn far away from understanding all of this.

src/agreementManager/AgreementManager.ts
cyony95: including the block author now as well

src/agreementManager/AgreementManager.ts
cyony95: including the block author now as well

src/agreementManager/AgreementManager.ts
cyony95: removed it.

src/agreementManager/AgreementManager.ts
cyony95: removed.

src/agreementManager/AgreementManager.ts
```diff
@@ -105,53 +117,149 @@ class AgreementManager {
...
 
     /**
-     * Get the latest finalized state and latest signed state (by signer) with virtual voting blocks
+     * Get state proof for a specific block height
      */
-    public getFinalizedAndLatestWithVotes(
+    public async getStateProof(
         forkId: ForkId,
+        blockHeight: BlockHeight,
         signerAddress: Address
```
lukastanisic99: `blockHeight` not used - will explain bellow where to use it and `signerAddress` not needed - we're creating the latest possible `StateProof`, not the latest signed by a signer

src/agreementManager/AgreementManager.ts
```diff
@@ -105,53 +117,149 @@ class AgreementManager {
...
-        let requiredSignatures = new Set<Address>(thresholdAddresses);
+        // Get all exit points for this fork
+        const exitPoints = this.storage.exitPoints.getExitPointsInRange(forkId);
+
+        const milestones: MilestoneProofStruct[] = [];
+        let currentSnapshot = genesisSnapshot;
+
+        // For each exit point, iterate forward to prove it's final
+        for (const exitPointHeight of exitPoints) {
+            const milestoneData = this.processMilestone(
```
lukastanisic99: `MilestoneData` is 

```typescript
{
        milestoneProof: MilestoneProofStruct;
        snapshot: StateSnapshot;
        encodedLatestFinalizedState: Bytes | undefined;
        encodedLatestCorrectState: Bytes | undefined;
        virtualVotingBlocks: BlockConfirmationStruct[];
    }
```
all of this is not needed the only thing that should be returned is a single `milestone` everything else can be fetched easily O(1) from the milestone 

as for the input arguments to the function the only arguments you need are (blockIterator,currentSnapshot) - everything else is not needed 

so the function should be `tryBuildMilestone(blockIterator,currentSnapshot) -> milestoneStruct | undefined` - in general we use `undefined` instead of `null` even though they are the same when used in something like `if( )` with both being equivalent to false

so the code should be here:
```typescript
for (const exitPointHeight of exitPoints) {
let blockInterator = this.storage.blocks.getIterator(
                forkId,
                SortOrder.ASC,
                exitPointHeight 
            );
const milestone = tryBuildMilestone(blockIterator,currentSnapshot);

if(milestone){
milestones.push(milestone)
currentSnapshot = getSnapshot(milestone) // this is a helper function that takes the first blockConfirmation in the milestone and fetches its snapshot from storage
}
}
```

src/agreementManager/AgreementManager.ts
```diff
@@ -105,53 +117,149 @@ class AgreementManager {
...
+                lastExitPointHeight
+            );
+            const firstBlock = blockEntries.next();
+
+            if (!firstBlock.done && firstBlock.value) {
+                signedBlocks.push(
+                    firstBlock.value.blockConfirmation.signedBlock
+                );
+            }
+        }
```
lukastanisic99: so at this point we've processed all `ExitPoint` milestones (if any) and maybe just need to build the `latest milestone` for the latest possible state or `signedBlocks` to form the `StateProof`

This is also easy and is basically this:

```typescript
let blockIterator = this.storage.blocks.getIterator(
    forkId,
    SortOrder.DESC,
    blockHeight
);

const milestone = tryBuildMilestone(blockIterator, currentSnapshot);

if (milestone) {
    milestones.push(milestone);
    currentSnapshot = getSnapshot(milestone);

    return {
        milestones,
        // signedBlocks are empty since the milestone already accounted the latest state
        signedBlocks: [] 
    };
} else {
    // If we're here we can't prove anything final so at most we have N-1 blocks
    // linked to currentSnapshot -> so we iterate backwards from blockHeight and
    // add all blockConfirmations into an array. We stop iteration when there's 
    // nothing else to iterate over (block 0 is reached) or we reach the block 
    // that commits to the currentSnapshot. We sort the array in ascending order 
    // (just reverse the array) and that's our signedBlocks.

    return {
        milestones,
        signedBlocks
    };
}

```

src/agreementManager/AgreementManager.ts
```diff
@@ -191,26 +297,83 @@ class AgreementManager {
...
+            encodedLatestFinalizedState,
+            encodedLatestCorrectState,
+            virtualVotingBlocks
+        };
+    }
+
+    /**
+     * Process a single milestone at the given exit point height
+     */
+    private processMilestone(
```
lukastanisic99: so this is the function I call `tryBuildMilestone(blockIterator, currentSnapshot) -> milestone | undefined`

its implementation should look like:

pseudo code:
```typescript
thresholdSet = new Set(currentSnapshot.participants)
filteredBlockConfirmations = [ ];
while(blockIterator.hasNext()){
    currentBlockConfirmation = blockIterator.value // current block or whatever the typescript interrator interface for this is
    allSignatures = getAllSignatures(currentBlockConfirmation)
    fillteredBlockConfirmation = objectCopy(currentBlockConfirmation);
    fillteredBlockConfirmation.signatures = [ ] //strip the signatures but keep the authors signature (that's always needed so the block is valid, so in the worse case we have 2N signatures)  
    for(signature in allSignatures ){
        participantAdr = block.getSignerAddress(signature)
        if(thresholdSet.has(participantAdr))thresholdSet.subtract(participant) //whatever the Set operation for that is - not looking at the code now to check
        if(participant!=blockAuthor) fillteredBlockConfirmation.signatures.push(signature);
    }
filteredBlockConfirmations.push(filteredBlockConfirmation)
if(thresholdSet.size==0) return filteredBlockConfirmations.sort(ASC) //so tryBuildMilestone works REGARDLESS of iterator direction
}
return undefined;
```

lukastanisic99: the above implementation/comment covers also `buildMilestoneProof` so that is not needed

src/agreementManager/AgreementManager.ts
```diff
@@ -105,113 +117,200 @@ class AgreementManager {
...
+                currentSnapshot
+            );
+
+            if (milestone) {
+                milestones.push(milestone);
+                const newSnapshot = this.getSnapshot(milestone);
+                if (newSnapshot) {
+                    currentSnapshot = newSnapshot;
+                }
+            }
```
lukastanisic99, line 158: if we can't form the milestone we should `break` the `for loop` early. 
`Explanation:` since we can't prove finality -> we would fallback on the logic of trying to prove the `latest state` final against the `currentSnapshot` (the current snapshot is always proven final) by iterating backwards up to that snapshot and if unsuccesfull populate signedBlocks instead. 

In the case that `getSnapshot(milestone)` deosn't find a snapshot it should `throw an error` since it shouldn't be possible that we've reached finality on a block and don't have the snapshot that coresponds to it.

src/agreementManager/AgreementManager.ts
```diff
@@ -105,113 +117,200 @@ class AgreementManager {
...
+                );
+
+                if (
+                    block.height === 0 ||
+                    block.stateSnapshotHash === currentSnapshot.hash
+                ) {
+                    break;
                 }
+
+                signedBlocks.push(blockEntry.blockConfirmation.signedBlock);
```
lukastanisic99: this should before the `if(block 0 ... )` so the first block is also included otherwise if you had blocks `A` and `B` you'd only include `B` in signed blocks and `B` is not linked to the genesis/currentSnapshot

src/agreementManager/AgreementManager.ts
```diff
@@ -105,113 +117,200 @@ class AgreementManager {
...
+    public tryBuildMilestone(
+        blockIterator: Generator<BlockEntry, void, unknown>,
+        currentSnapshot: StateSnapshot
+    ): MilestoneProofStruct | undefined {
+        const thresholdSet = new Set<Address>(
+            currentSnapshot.snapshotData.participants as Address[]
+        );
+        const filteredBlockConfirmations: BlockConfirmationStruct[] = [];
+
+        let result = blockIterator.next();
```
lukastanisic99, line 259: Will this skip the first block? E.g. `iterator = getIterator(height=100)` and you call `iterator.getNext()` you never processed block 100 and you went to `99` or `101` - in both cases you need to process `block 100` too.

src/agreementManager/AgreementManager.ts
```diff
@@ -105,113 +117,200 @@ class AgreementManager {
...
-                const stateSnapshot =
-                    this.storage.stateSnapshots.getStateSnapshotByHash(
-                        block.stateSnapshotHash
+            const filteredBlockConfirmation: BlockConfirmationStruct = {
+                signedBlock: currentBlockConfirmation.signedBlock,
+                signatures: [] // Strip signatures but keep author's signature in signedBlock
+            };
+
+            for (const signature of allSignatures) {
+                const participantAddress = Block.decode(
```
lukastanisic99: decode the block once at the top so you don't decode it all the time

src/agreementManager/AgreementManager.ts
```diff
@@ -105,113 +117,200 @@ class AgreementManager {
...
+                    currentBlockConfirmation.signedBlock.encodedBlock
+                )
+                    .getSignerAddresses([signature])
+                    .values()
+                    .next().value;
```
lukastanisic99: https://github.com/peer3to/state-channels-plus/blob/39c61df52c2f48d0c658c348009dae60f68a5dd2/src/models/Block.ts#L118

src/agreementManager/AgreementManager.ts
```diff
@@ -105,113 +117,200 @@ class AgreementManager {
...
+                        }
+                    )
+                };
+            }
 
-        // If no correct state found, use genesis
-        if (!encodedLatestCorrectState) {
-            encodedLatestCorrectState =
-                this.storage.getGenesisStateMachineState(forkId) as Bytes;
+            result = blockIterator.next();
```
lukastanisic99, line 318: you can add a termination condition here while you have the old block before you transition it `if(block commits to currentSnapshot)break` that means we didn't find finality -> will return `undefined` - later when Snapshot has a `height field` this can also be written in the while above as `while(!result.done && currentSnapshot.height>=block.height)`

src/agreementManager/AgreementManager.ts
cyony95: sorry, I missed block 0 here.

src/agreementManager/AgreementManager.ts
cyony95: This will not skip the next block, because when you call getIterator you get a generator actually 🗡️ . the first value will be yielded only when you call next() on it.
https://www.geeksforgeeks.org/javascript/javascript-generator-next-method/

src/agreementManager/AgreementManager.ts
cyony95: you re 100% right, I will be more careful next time.

lukastanisic99 (approved):

src/agreementManager/AgreementManager.ts
```diff
@@ -105,113 +117,208 @@ class AgreementManager {
...
+                const participantAddress = block.getSignerAddress(signature);
+
+                if (
+                    participantAddress &&
+                    thresholdSet.has(participantAddress)
+                ) {
+                    thresholdSet.delete(participantAddress); // Subtract participant from threshold set
+
+                    // Add signature to filtered block confirmation if it's not the author's signature
+                    if (participantAddress !== block.author) {
```
lukastanisic99, line 287: in regards to #143 I'll comment here @daiagi @cyony95 - the code here was my original idea to save on signature verification on-chain. The author signature is always taken into account for the threshold as used above `allSignatures` and is not part of `filtering` only `confirmation signatures` are filtered (the original `signedBlock` is always included), mostly to not have `N*N` signatures to verify on-chain. The reason why the author signature is always included (not filtered) is because the block is useless without the authors signature. So at most we do `2*N` signatures verifications on-chain, which is better than `N*N`.

I'm not sure why #143 which is different code than this doesn't show the difference as changes even though it's targeting the dispute branch which merged this.

PR #130, opened 2025-08-01
Unify-pipeline

daiagi:
- Closes https://trello.com/c/XxziBREi


lukastanisic99:
@daiagi should I review this or still `wip`

src/stateManager/StateManager.ts
```diff
@@ -235,21 +234,17 @@ class StateManager {
         }
     }
     private async tryConfirmFromQueue(): Promise<void> {
-        //TODO! race condition and skipping a txCount
```
lukastanisic99, line 238: in a uniformed pipeline there should only be `1` queue that calls `onBlockConfirmation`

src/stateManager/StateManager.ts
```diff
@@ -278,122 +273,80 @@ class StateManager {
...
+        });
     }
 
     // Passes the block confirmation through a verification pipeline and returns an execution flag
     public async onBlockConfirmation(
-        signedBlock: SignedBlockStruct,
-        confirmationSignature: Signature,
-        block?: Block
+        blockConfirmation: BlockConfirmationStruct
     ): Promise<ExecutionFlags> {
```
lukastanisic99: Take a look at the right part of the diagram where it explains the execution flags meaning and the conclusion - we don't have to return the execution flag just a flag `shouldDisconnect` and the flags can be removed all together and substituted with functions `success(), dispute(), notReady() ...` that do the common operation of those functions respectively - if for some reason you still need `ExecutionFlags` internally within `StateManager` sure, but in general they're not needed

src/stateManager/StateManager.ts
```diff
@@ -278,122 +273,80 @@ class StateManager {
...
-        let finalExecutionFlag: ExecutionFlags = ExecutionFlags.SUCCESS; // Default to SUCCESS
-        const decodedBlock = block ?? Block.decode(signedBlock.encodedBlock);
+        const validationFlag =
```
lukastanisic99: missing `mutex` here

src/stateManager/StateManager.ts
```diff
@@ -278,122 +273,80 @@ class StateManager {
...
-                    confirmationSignature
+            if (!success) {
+                const previousEntity = getPreviousBlockOrSnapshot(
+                    block.coordinates,
+                    this.storage
+                );
+                throw new InvalidStateTransitionException(
+                    previousEntity,
+                    blockConfirmation.signedBlock
                 );
```
lukastanisic99: Still think it's better to not throw an error that kills the logical flow of the pipeline, but explicity create the specific fraud proof and persist it in storage and call `dispute()` - this way we don't have to polute our error handling with checking for fraud proofs

src/stateManager/StateManager.ts
```diff
@@ -278,122 +273,80 @@ class StateManager {
...
-                    decodedBlock
-                );
-            finalExecutionFlag = result.flag;
+            const {
+                success,
+                encodedState,
+                previousStateHash,
+                successCallback,
+                exitChannels
+            } = await this.applyTransaction(block.transaction);
```
lukastanisic99: in general since `BlockConfirmation pipeline` is the largest part of the `StateManager` you extract and implement that pipeline in a seperate file as part of `StateManager` e.g. `BlockConfirmationPipeline.ts` instead of `ValidationService.ts` - stateTransition stuff can also be a seperate file so you can re-use it there and for `playTransaction`

That will also help with not decoding the block twice as on line 298 (a few above)

btw `successCallback` and `exitChannels` are not used - they should be part of `success()`

src/stateManager/StateManager.ts
```diff
@@ -278,122 +273,80 @@ class StateManager {
...
+                const fraudProof =
+                    this.fraudProofService.createInvalidStateTransitionProof(
+                        error.previousEntity,
+                        error.invalidBlock
+                    );
+                // TODO: Persist fraud proof to storage
+                return ExecutionFlags.DISPUTE;
             }
+            throw error; // Re-throw non-fraud exceptions
         }
```
lukastanisic99, line 288: `finally` to release the `mutex`

src/stateManager/StateManager.ts
lukastanisic99: or `BlockConfirmationService` if we're using that convention

src/stateManager/utils/FraudProofService.ts
```diff
@@ -0,0 +1,178 @@
...
+/**
+ * Service class for handling fraud proof creation and validation
+ */
+export class FraudProofService {
+    constructor(private readonly storage: Storage) {}
+
+    /**
+     * Create invalid state transition proof
+     */
+    createInvalidStateTransitionProof(
```
lukastanisic99, line 27: so instead of throwing the errors which have the same arguments as these functions, just call these functions on the spot and all they do is `create` the `specific fraud proof`, `encode` it and `wrap it` into a `general fraud proof struct` as defined in solidity and perisit it in storage in a way where it's easy to retrive fraud proofs based for participants (e.g. `getFraudproof(participant)`) - so everything is in storage and when we eventually call `dispute()` later - we don't have to pass fraud proofs to it. It can fetch it from storage in O(1) for every participant.

src/stateManager/utils/FraudProofService.ts
```diff
@@ -0,0 +1,178 @@
...
+            invalidBlock: signedBlock,
+            previousBlock: prevSignedBlock,
+            previousStateSnapshot: prevStateSnapshot.toStruct()
+        };
+    }
+
+    /**
+     * Fraud proof creators mapped by exception name
+     */
+    private readonly fraudProofCreators = {
```
lukastanisic99: when taken into account with my previous comment above this entire section with fraud proof errors is over engineering

src/stateManager/utils/channelValidation.ts
```diff
@@ -0,0 +1,36 @@
...
+export type PreviousEntity = {
+    blockConfirmation?: BlockConfirmationStruct;
+    stateSnapshot?: StateSnapshot;
+};
+
+export function isChannelOpen(forkId: ForkId): boolean {
+    return forkId !== ethers.ZeroHash && forkId !== NULL;
+}
+
+export function getPreviousBlockOrSnapshot(
```
lukastanisic99: This can be a `general storage` helper function

src/stateManager/utils/channelValidation.ts
```diff
@@ -0,0 +1,36 @@
+import { ethers } from "ethers";
+import Storage from "@/storage";
+import { BlockCoordinates, StateSnapshot } from "@/models";
+import { ForkId } from "@/types/types";
+import { BlockConfirmationStruct } from "@typechain-types/contracts/V1/types/DataTypes";
+
+const NULL = "0x00";
+
+export type PreviousEntity = {
```
lukastanisic99: rename to `BlockOrSnapshot` and you can add this to `types` and since this is uses in just a few places better to name the variables where used `previousBlockOrSnapshot` eventhough it's more text - `previousEntity` doesn't mean anything

src/stateManager/utils/channelValidation.ts
```diff
@@ -0,0 +1,36 @@
...
+    blockConfirmation?: BlockConfirmationStruct;
+    stateSnapshot?: StateSnapshot;
+};
+
+export function isChannelOpen(forkId: ForkId): boolean {
```
lukastanisic99: add this somewhere else so it's not the only thing that remains in this file

src/stateManager/ValidationService.ts
```diff
@@ -1,384 +1,488 @@
...
+        // 1. Check if block is in queue
+        if (
+            this.storage.queues.isBlockQueued(blockConfirmation, {
+                hash: block.hash
+            })
+        ) {
+            const signerAddresses = block.getSignerAddresses(
+                blockConfirmation.signatures as Signature[]
+            );
+            const areAllParticipants = isSubset(signerAddresses, participants);
```
lukastanisic99, line 215: this should be a general check regardless if queue or block storage

src/stateManager/ValidationService.ts
```diff
@@ -1,384 +1,488 @@
...
+                block.forkId,
+                block.height
+            );
+            const maybePreExistingBlockConfirmation =
+                blockEntry?.blockConfirmation;
 
-        // Validate past block in current fork
-        if (blk.height < this.getNextHeight()) {
-            const agreementFlag = this.checkBlock(signedBlock);
+            const { conflict, fraudType } = this.checkConflictingBlock(
```
lukastanisic99: so instead of fetching and naming a potential conflicting block `maybePreExistingBlockConfirmation` here, just pass the current block to `checkConflictingBlock` and it will fetch a `conflictingBlock` and if the conflictingBlock is defined, proceed with checking what type of fraud proof it is (if any) and instead of returing the `fraudType` call teh function to create and persist the fraudProof in storage while you have the context what fraud proof it is.

src/stateManager/ValidationService.ts
```diff
@@ -1,384 +1,488 @@
...
-            // Return NOT_READY to allow retry
-            return notReady();
-        }
+    private async isDisputedFork(
+        forkId: ForkId,
+        channelId: ChannelId
+    ): Promise<boolean> {
+        return (
+            this.storage.disputes.getDisputedFork(forkId) ||
+            (await this.stateChannelManagerContract.isForkDisputed(
```
lukastanisic99: this should be localEvm.diamond.isForkDisputed - we don't want to spam the RPC node and should do as much of the computation locaaly as possible

src/stateManager/ValidationService.ts
daiagi: it is done for both
when block is in storage the validation may reject earlier  if now new signatures and will not check if `areAllParticipants`

daiagi:
all done
still using "flags" although newly named subset of them
they are used only internally in the pipeline, they are not reflected outwards

there are still TODOS left, to be handled AFTER this PR

fraud proofs ARE saved into (newly created) FraudProofStorage
this storage is is a quick sketch. also meant to be refined in follow ups

but the creation of dispute / gossiping it etc is not included in this PR
this PR also does not handle  the success/disconnect functions / actions. this is also left as a TODO

src/stateManager/ValidationService.ts
```diff
@@ -1,384 +1,494 @@
...
+                return {
+                    shouldDisconnect: true,
+                    action: BlockValidationAction.DISPUTE
+                };
+            }
+            // If valid, continue with rest of validation
+
+            if (
+                block.onChainTimestamp === undefined ||
+                block.onChainTimestamp <= block.timestamp
```
lukastanisic99, line 424: this `if` should be on the outside of the initial/first `if(!isValidTimestamp)` - take a look at the diagram in the diagram branch under the blockConfirmation pipeline - it has TimeLogic

src/stateManager/ValidationService.ts
```diff
@@ -1,384 +1,494 @@
...
-        if (
-            blk.onChainTimestamp === undefined ||
-            blk.onChainTimestamp <= blk.timestamp
-        ) {
-            if (
-                Math.abs(Clock.getTimeInSeconds() - blk.timestamp) >=
-                this.timeCfg.agreementTime
-            )
-                return timestampTooOld();
+            return genesisSnapshot.hash === block.previousBlockHash;
```
lukastanisic99: I think we went for the convention `(block 0).previousBlockHash = 0x00`, and `block.forkId = hash(genesisStateSnapshot.snapshotData)` which is always true - so block 0 `can not be not linked <=> is always linked`, since it already commited to the genesis by defining the `forkId`

src/stateManager/ValidationService.ts
```diff
@@ -1,384 +1,494 @@
...
+        return { shouldDisconnect: true };
+    }
+
+    private async isDisputedFork(
+        forkId: ForkId,
+        channelId: ChannelId
+    ): Promise<boolean> {
+        return (
+            this.storage.disputes.getDisputedFork(forkId) ||
+            (await this.localDiamond.isForkDisputed(channelId, forkId))
```
lukastanisic99: EvmDiamondStateMachine.localDiamond (or however you expose the localDiamond interface)

src/storage/DisputeStorage.ts
```diff
@@ -39,6 +42,10 @@ export class DisputeStorage {
         );
     }
 
+    storeDisputedFork(forkId: ForkId): void {
```
lukastanisic99: should be able to unset also in the case that the dispute tx failed - maybe `setDidIDisputeFork(forkId,bool)`

src/storage/DisputeStorage.ts
```diff
@@ -62,6 +69,10 @@ export class DisputeStorage {
         return this.disputes.get(disputeHash);
     }
 
+    getDisputedFork(forkId: ForkId): DidIDispute {
```
lukastanisic99: `didIDispute(forkId)` - so it's clear that only that is taken into account and not the existance of a `disputeWindow` + if you name it `getDisputedFork` it implies that the return value is a fork

lukastanisic99 (changes requested):
Overall good just small quick fixes/adjustments and add the todos as a trello card

src/stateManager/ValidationService.ts
daiagi: this is enough for this PR
will be done in follow

https://trello.com/c/gI7XhT3M

lukastanisic99 (approved):

src/stateManager/StateManager.ts
```diff
@@ -875,10 +761,92 @@ class StateManager {
...
+        if (!success) {
+            this.fraudProofService.createInvalidStateTransitionProof(
+                block.coordinates,
+                blockConfirmation.signedBlock
+            );
+            await this.dispute(blockConfirmation);
+            return;
+        }
+
+        const stateTransitionFlag = this.isValidStateTransition(
```
lukastanisic99, line 802: Just realized while looking at Adrians PR #140 that here we don't check `isBlockCommitmentCorrect` - don't know how I missed it while reviewing (it's in the diagram) 

We have to construct the `stateSnapshot` and check does the block commit to it 

@daiagi

PR #134, opened 2025-08-09
clear solidity warnings

daiagi:
## 📋 Description

clear `yarn compile` warnings

### DIFF
--- a/contracts/V1/AStateMachine.sol
+++ b/contracts/V1/AStateMachine.sol
@@ -82,16 +82,16 @@ abstract contract AStateMachine {
         // emit SetStateA(encodedState);
     }
 
-    function joinChannel(JoinChannel memory joinChannel) external _nonReentrant returns (bool) {
-        return _joinChannel(joinChannel);
+    function joinChannel(JoinChannel memory joinChannelData) external _nonReentrant returns (bool) {
+        return _joinChannel(joinChannelData);
     }
 
     function slashParticipant(address adr) external _nonReentrant returns (bool, ExitChannel memory exitChannel) {
-        (bool success, ExitChannel memory exitChannel) = _slashParticipant(adr);
+        (bool success, ExitChannel memory exitChannelData) = _slashParticipant(adr);
         if (success) {
-            _addExitChannel(exitChannel);
+            _addExitChannel(exitChannelData);
         }
-        return (success, exitChannel);
+        return (success, exitChannelData);
     }
 
     function removeParticipant(address adr)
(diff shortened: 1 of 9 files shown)


contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol
```diff
@@ -8,10 +8,8 @@ import "../types/DisputeFraudProofTypes.sol";
 import "./utils/DisputeUtils.sol";
 
 contract DisputeFraudProofFacet is StateChannelCommon {
-    mapping(
-        DisputeFraudProofType
-            => function(bytes memory encodedFraudProof, Dispute memory dispute) internal returns (address)
-    ) private proofHandlers;
+    mapping(DisputeFraudProofType => function(bytes memory, Dispute memory) internal returns (address)) private
```
lukastanisic99, line 11: isn't it better to have `bytest memmory encodedFraudProof` than `bytes memory` which doesn't mean anything

daiagi: Naming function type parameters is deprecated.solidity(6162)

image


---


The actual functions DO have named arguments, but the mapping declaration does not 


```solidity
    function _handleTimeoutThreshold(bytes memory encodedFraudProof, Dispute memory dispute)
        internal
        pure
        returns (address)
```

contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol
```diff
@@ -122,9 +120,9 @@ contract DisputeFraudProofFacet is StateChannelCommon {
         return dispute.disputer;
     }
 
-    function _handleTimeoutParticipantNotNext(bytes memory encodedFraudProof, Dispute memory dispute)
+    function _handleTimeoutParticipantNotNext(bytes memory, /* encodedFraudProof */ Dispute memory dispute)
```
lukastanisic99, line 123: Not sure what's the need for this - of course the compiler is giving a warning that the variable is not used, but it will be used eventually when this gets populated - so better to have the warning as a reminder that this needs to be implemented - we need the variable either way for access

contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol
lukastanisic99: Yeah I saw that but it still compiled - I guess you can put a comment above so people know what the `bytes` are

contracts/V1/AStateMachine.sol
```diff
@@ -82,16 +82,16 @@ abstract contract AStateMachine {
         // emit SetStateA(encodedState);
     }
 
-    function joinChannel(JoinChannel memory joinChannel) external _nonReentrant returns (bool) {
-        return _joinChannel(joinChannel);
+    function joinChannel(JoinChannel memory joinChannelData) external _nonReentrant returns (bool) {
```
lukastanisic99, line 85: what warning was given here

PR #136, opened 2025-08-10
Block model wraps BlockConfirmationStruct

daiagi:
Move signatures in `Block` Model
aka `Block` wraps around`BlockSignatures`

this simplifies the code all over and is nicer to work with

Tests are not updated

when PR is approved, i'll update them

### DIFF
--- a/src/agreementManager/AgreementManager.ts
+++ b/src/agreementManager/AgreementManager.ts
@@ -5,15 +5,7 @@ import {
     StateProofStruct
 } from "@typechain-types/contracts/V1/types/ProofTypes";
 import Storage, { BlockEntry, SortOrder } from "@/storage";
-import {
-    Address,
-    BlockHeight,
-    Bytes,
-    ForkId,
-    Signature,
-    Hash
-} from "@/types/types";
-import { BytesLike } from "ethers";
+import { Address, BlockHeight, ForkId, Signature } from "@/types/types";
 import { Block, StateSnapshot } from "@/models";
 import { Codec, Type } from "@/utils";
 import { ethers } from "ethers";
@@ -145,14 +137,10 @@ class AgreementManager {
                 blockHeight
             );
 
-            let result = blockIterator.next();
-            while (!result.done) {
-                const blockEntry = result.value;
-                const block = Block.decode(
-                    blockEntry.blockConfirmation.signedBlock.encodedBlock
-                );
+            for (const blockEntry of blockIterator) {
+                const block = blockEntry.block;
 
-                signedBlocks.push(blockEntry.blockConfirmation.signedBlock);
+                signedBlocks.push(block.signedBlock);
 
                 if (
                     block.height === 0 ||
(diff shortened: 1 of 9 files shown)


src/storage/BlockStorage.ts
```diff
@@ -1,16 +1,8 @@
 import {
-    BlockConfirmationStruct,
-    SignedBlockStruct
+    SignedBlockStruct,
```
cyony95, line 2: why do we still have these? we dont use them anymore, right?

src/storage/QueueStorage.ts
```diff
@@ -1,77 +1,56 @@
-import {
-    SignedBlockStruct,
-    BlockConfirmationStruct
-} from "@typechain-types/contracts/V1/types/DataTypes";
 import { Block } from "@/models";
 import { Hash, ForkId, BlockHeight } from "@/types/types";
-import { hash } from "@/utils";
+import { BlockConfirmationStruct } from "@typechain-types/contracts/V1/types/DataTypes";
```
cyony95, line 3: same question here, we just use Block now, right?

cyony95: nvm, I saw where you're using it here.

src/storage/BlockStorage.ts
```diff
@@ -52,7 +44,7 @@ export class BlockStorage {
     /*────────────────────────────────────────────────────────────────────────────
       STORE  BLOCK - IMPLEMENTATION
     ────────────────────────────────────────────────────────────────────────────*/
-    storeBlock(
+    storeSignedBlock(
```
lukastanisic99: Imo we can remove this - currently not used. Everytime we get a `signedBlock` or `blockConfirmation` from the p2p network or DLT we can immediately transform it into a `block` by calling `Block.fromSignedBlock`/`Block.fromBlockConfirmation` and just use the `block` across the typescript code

src/storage/QueueStorage.ts
```diff
@@ -1,77 +1,56 @@
...
         // Store the new block confirmation
-        this.queuedBlocks.set(block.hash, blockConfirmation);
+        this.queuedBlocks.set(block.hash, block);
         this.addHashToCoordinateIndex(block.hash, block.forkId, block.height);
 
         return block.hash;
     }
 
+    /** Queue a block confirmation for future processing */
+    queueConfirmation(blockConfirmation: BlockConfirmationStruct): Hash {
```
lukastanisic99, line 33: Imo we can remove this also

src/agreementManager/AgreementManager.ts
```diff
@@ -135,14 +90,8 @@ class AgreementManager {
...
-        for (const blockEntry of blockEntries) {
-            const block = Block.decode(
-                blockEntry.blockConfirmation.signedBlock.encodedBlock
-            );
-
-            const signersAddresses = block.getSignerAddresses(
-                blockEntry.blockConfirmation.signatures as Signature[]
-            );
+        for (const { block } of blockEntries) {
+            const signersAddresses = block.confirmationSignerAddresses;
```
lukastanisic99: this should take into account the original signature also

lukastanisic99 (approved):
Overall I think this is a good and usefull

PR #138, opened 2025-08-14
Fix killdispute

lukastanisic99:
Quick fix so `killDispute` uses `killTime` to check is the `DisputeWindow expired` instead of `evidenceTime` - I'll create a PR for the dispute part of the diagrams these days when it's all done, but there's enough to take a look already (diagrams branch).


cyony95 (approved):

cyony95:
> Quick fix so `killDispute` uses `killTime` to check is the `DisputeWindow expired` instead of `evidenceTime` - I'll create a PR for the dispute part of the diagrams these days when it's all done, but there's enough to take a look already (diagrams branch).

this basically moves up the window available for the dispute to be killed. before, it was killing the dispute if the dispute window expired, and not it kills it before being finalized, right?

PR #143, opened 2025-08-20
Update-tests

daiagi:
- update a bunch of test to sync with changes in the code
- fix and improve API of some code where the test revealed that 
a. a bug
b. non convenebt code to use


daiagi:
@cyony95 
in this commit https://github.com/peer3to/state-channels-plus/pull/136/commits/0320eeaf03a2ff3ff6d09aea9363165f2ceeca87

i have made changes to your milestones code

I'm pretty sure i only changed style and did not alter function AT ALL, but...

quite a few tests in `AgreementManager` test suite are not passing
so i probably did f** up something

could you please take a look and make sure that that all of them pass?
some of them however, seems redundant or testing similar stuff, it would be good if you could also trim it down

cyony95:
> @cyony95 in this commit [0320eea](https://github.com/peer3to/state-channels-plus/commit/0320eeaf03a2ff3ff6d09aea9363165f2ceeca87)
> 
> i have made changes to your milestones code
> 
> I'm pretty sure i only changed style and did not alter function AT ALL, but...
> 
> quite a few tests in `AgreementManager` test suite are not passing so i probably did f** up something
> 
> could you please take a look and make sure that that all of them pass? some of them however, seems redundant or testing similar stuff, it would be good if you could also trim it down

one of the changes is changing confirmationSignatures to allSignatures. That means it takes the author into consideration. The tests were written for confirmationSignatures, that s why they fail.

daiagi:
>one of the changes is changing confirmationSignatures to allSignatures

not sure why you say that, before, `tryBuildMilestone` was using 
```ts
const allSignatures = this.getAllSignatures(currentBlockConfirmation);
```
 
 in any case, you tell me, which is correct? only confirmations or include author
 i think to include author is correct.
 reason:
 we want to see that all the current channel participant have directly or indirectly signed a block, if we dont include the author event the most straightforward block will not pass
 since the author is a participant and their signatures is not supposed to be in confirmation signatures, if we dont count it we will be short 1 vote EVEN if all are cooperating and signiing

cyony95:
> > one of the changes is changing confirmationSignatures to allSignatures
> 
> not sure why you say that, before, `tryBuildMilestone` was using
> 
> ```ts
> const allSignatures = this.getAllSignatures(currentBlockConfirmation);
> ```
> 
> in any case, you tell me, which is correct? only confirmations or include author i think to include author is correct. reason: we want to see that all the current channel participant have directly or indirectly signed a block, if we dont include the author event the most straightforward block will not pass since the author is a participant and their signatures is not supposed to be in confirmation signatures, if we dont count it we will be short 1 vote EVEN if all are cooperating and signiing

you are correct. i was just confused by this PR's differences, i checked my branch and you re right. I initially missed the author, then added it as Luka pointed out.

cyony95 (approved):

src/agreementManager/AgreementManager.ts
```diff
@@ -203,7 +204,7 @@ class AgreementManager {
                 currentBlock.signedBlock
             );
 
-            for (const signature of currentBlock.confirmationSignatures) {
+            for (const signature of currentBlock.allSignatures) {
```
lukastanisic99, line 207: I'm not sure which PR changed this code after #127 - I commented there on the logic - the logic there is good and here I see stuff that is incorrect and missing stuff like `if (participantAddress !== block.author)` which does not show as difference in code

lukastanisic99: #136

lukastanisic99 (approved):

PR #145, opened 2025-08-21
Sync-storage

daiagi:
- Closes https://trello.com/c/4Yy32C5n
- Closes https://trello.com/c/dAGKfnT7


daiagi:
@lukastanisic99  updated according to our call yesterday
- event consolidations
- use StateChannelEventListener
- remove "auto" mirroring
- use `LocalDiamondSigner`

contracts/V1/StateChannelDiamondProxy/JoinChannelFacet.sol
```diff
@@ -73,18 +73,13 @@ contract JoinChannelFacet is StateChannelCommon {
         channelBalance.latestJoinChannelBlockHash = blockHash;
 
         // Update pending participants
-        _updatePendingParticipants(jc);
+        disputeData[channelId].pendingParticipants.push(jc.participant);
```
lukastanisic99, line 76: isn't it cleaner to do it with a function and if the logic changes you just update it in one place

src/evm/LocalDiamondSigner.ts
```diff
@@ -0,0 +1,307 @@
...
+            throw new Error("Invalid transaction data");
+        }
+
+        try {
+            const decoded = this.interface.parseTransaction({ data });
+            if (!decoded) {
+                throw new Error("Failed to decode transaction data");
+            }
+
+            await this.localDiamond[decoded.name](...decoded.args);
```
lukastanisic99: This doesn't go to the local evm - it goes to the destination where the `signer.provider/runner` is pointing to and the signer that's passed to the SDK has a provider/runner to the on-chain RPC node - not local.

Most of this file is over engineering 

The signer should not know about the contract, the contract connects to a signer and uses the signer `sendTransaction` for state mutating transactions and `call` for read/view. 

The interaction is always `EthersContractInstance` -> which is connected to a `signer` which knows where to send the `payload`, not `signer` -> `contract` which would end up being `signer`->`contract`->`signer` either way, since ethers would call the signer internally anyway

 
So this signer class should be really simple - same as p2pSigner - it just overrides `call` and `sendTransaction` to route the payload to `EvmStateMachine.diamondContractExecutor` - it doesn't do any of the encoding decoding etc. Basically it's simple as `p2pSigner.runView` (take a look) 

The only reason `p2pSigner.sendTransaction` 'looks' more complex is because it's playing the transaction and wrapping it into a format used for our SDK (our block format) since it's replicated and gossiped - this is internal and in can be simple as `runView` - nothing else is needed 

So then you create a localDiamond ethers contract and connect it to an instance of this signer and use that ethers contract to interact with the diamond it will just work :D

lukastanisic99 (changes requested):
Didn't look at everything yet, but the `LocalDiamondSigner` is over engineering and It doesn't route to the local EVM

contracts/V1/StateChannelDiamondProxy/LocalDiamond.sol
```diff
@@ -41,46 +39,139 @@ contract LocalDiamond is StateChannelManagerProxy {
...
+
+    // Called by StateSnapshotUpdated event
+    function onStateSnapshotUpdated(bytes32 channelId, StateSnapshot calldata stateSnapshot, uint256 timestamp)
+        external
+    {
+        stateSnapshots[channelId] = stateSnapshot;
+    }
+
+    // Called by JoinChannelProcessed event
+    function onJoinChannelProcessed(
```
lukastanisic99, line 52: this function is not called in the eventListener

contracts/V1/StateChannelDiamondProxy/LocalDiamond.sol
```diff
@@ -41,46 +39,139 @@ contract LocalDiamond is StateChannelManagerProxy {
...
         }
-        emit StorageSet(slot, value);
     }
 
-    function getStorageSlot(bytes32 slot) external view returns (bytes32) {
-        bytes32 value;
-        assembly {
-            value := sload(slot)
+    // Called by BlockCalldataPosted event
+    function onBlockCalldataPosted(
```
lukastanisic99, line 74: this one also

scripts/generate-artifacts.js
```diff
@@ -19,7 +19,8 @@ const facets = [
     "FraudProofFacet",
     "JoinChannelFacet",
     "StateChannelCommon",
-    "StateSnapshotFacet"
+    "StateSnapshotFacet",
+    "LocalDiamond"
```
lukastanisic99, line 23: We have too many places in the code that we hardcode stuff either in some records mapping or like this or even ethers types and it's becoming too much to remember to update/manage - for instance even here we forgot to add `DisputeVerificationFacet` that was the result of splitting `DisputeManagerFacet` maybe other stuff is missing too 

Not sure what's the best way of addressing this - we could write a script for this that does some regex on the generated artifacts to extract all `*Facet` - that would automate this

Add some trello card for this and generally thinking how to NOT have hard coded stuff in our code that we have to remember to update

src/StateChannelEventListener.ts
```diff
@@ -85,20 +87,7 @@ class StateChannelEventListener {
...
-            handler: (logObj: any) => {
-                const { outputStateSnapshot, disputeCommitment } = logObj.args;
-                console.log("OutputStateSnapshotVerified EVENT ");
-                this.stateManager.onOutputStateSnapshotVerified(
-                    outputStateSnapshot,
-                    disputeCommitment
-                );
-            }
-        },
+
```
lukastanisic99, line 105: The above (old) events are missing updating the state in the localDiamond

lukastanisic99 (changes requested):
Overall fine - just need to address the few missing things

lukastanisic99 (changes requested):
Overall fine - just need to address the few missing things

lukastanisic99 (approved):

PR #146, opened 2025-08-22
Update snapshot for different forks

cyony95:
closes https://trello.com/c/ldQeKwn8/151-update-snapshot-fork-dependent-on-storage-statemanager-agreements

Algo:
1. Fetch current on-chain snapshot
2. Check if the fork is disputed
3. Traverse if a reduced result already exists
4. Reduce if no reduced result present(but don't commit)
5. Update on-chain snapshot


src/stateManager/StateManager.ts
```diff
@@ -640,6 +640,180 @@ class StateManager {
...
+                    await this.stateChannelManagerContract.isReduceChallengePeriodExpired(
+                        this.channelId,
+                        currentForkId
+                    );
+
+                if (!isChallengePeriodExpired) {
+                    console.log(
+                        `Challenge period not expired for fork ${currentForkId}`
+                    );
+                    break;
```
cyony95, line 717: @lukastanisic99  @daiagi  break or wait?

src/stateManager/StateManager.ts
```diff
@@ -640,6 +640,180 @@ class StateManager {
...
+                }
+
+                // Get the dispute window creation timestamp from the on-chain contract
+                const creationTimestamp =
+                    await this.stateChannelManagerContract.getDisputeWindowCreationTimestamp(
+                        this.channelId,
+                        currentForkId
+                    );
+
+                // Call reduce on-chain to get the reduced result
```
cyony95: @lukastanisic99  @daiagi  maybe check if the result already has been reduced?

src/stateManager/StateManager.ts
```diff
@@ -640,6 +640,180 @@ class StateManager {
...
+                const reducedOutput =
+                    await this.stateChannelManagerContract.reduce(
+                        [dispute],
+                        creationTimestamp
+                    );
+
+                // TODO, this is not implemented yet
+                // Also run reduce locally in the local EVM for validation
```
cyony95: @lukastanisic99 @daiagi not sure i did this right.

src/stateManager/StateManager.ts
```diff
@@ -640,6 +640,180 @@ class StateManager {
...
+                // Check if the next fork has a dispute window
+                isDisputed =
+                    await this.stateChannelManagerContract.isForkDisputed(
+                        this.channelId,
+                        currentForkId
+                    );
+            }
+
+            // Get the state snapshot for the final fork
+            const latestBlockHeight =
```
cyony95: @lukastanisic99 this is all recycled code from the updatesnapshotsamefork

contracts/V1/StateChannelDiamondProxy/StateChannelManagerProxy.sol
```diff
@@ -384,6 +385,41 @@ contract StateChannelManagerProxy is StateChannelManagerInterface, StateChannelC
...
+        DisputeWindow storage disputeWindow = _disputeData.disputeWindowMap[forkId];
+        return disputeWindow.evidence.creationTimestamp;
+    }
+
+    function reduce(Dispute[] memory disputes, uint256 disputeWindowCreationTimestamp)
+        public
+        view
+        returns (ReduceOutput memory reducedOutput)
+    {
+        return disputeVerificationFacet.reduce(disputes, disputeWindowCreationTimestamp);
```
lukastanisic99: this needs to be a `delegatecall` so it executes in the storage context of the `StateChannelManagerProxy` like all other operations

contracts/V1/StateChannelDiamondProxy/StateChannelManagerProxy.sol
```diff
@@ -384,6 +385,41 @@ contract StateChannelManagerProxy is StateChannelManagerInterface, StateChannelC
...
+    }
+
+    function commitToReducedResult(
+        bytes32 channelId,
+        bytes32 disputedForkId,
+        bytes32 reducedForkId,
+        uint256 reducedForkGenesisTimestamps
+    ) public {
+        disputeManagerFacet.commitToReducedResult(
+            channelId, disputedForkId, reducedForkId, reducedForkGenesisTimestamps
```
lukastanisic99: delegatecall

src/stateManager/StateManager.ts
```diff
@@ -640,6 +640,180 @@ class StateManager {
...
+
+                if (!isChallengePeriodExpired) {
+                    console.log(
+                        `Challenge period not expired for fork ${currentForkId}`
+                    );
+                    break;
+                }
+
+                // Get the dispute data from the DisputeHandler for this fork
+                const dispute = this.disputeHandler.disputes.get(currentForkId);
```
lukastanisic99: `disputeHandler` is outdated code (not sure we made any changes to it after the w3f grant) - in general this code here should be `disputeCommitments = evmStateMachine.localDiamond.getDisputeCommitments(forkId)` or in your case for now `disputeCommitments = stateChannelManagerContract.getDisputeCommitments(forkId)`
then we need a storage module that persists all disputes that were emited as calldata when `uploading` the dispute on-chain. So you'll have something like:

```typescript
disputes = []
for(commitment in disputeCommitments){
    dispute = storage.getDispute(commitment);
    if(!dispute){
        // querry on-chain logs for the event that commited to `commitment` (maybe the commitment in the event needs to be `indexed`)
        dispute = querriedEvent.dispute
        if(!dispute) throw `Missing Data Availability for dispute commitment ${commitment}`
        storage.insertDispute(commitment,dispute)
    }
    disputes.push(dispute);
}
```

src/stateManager/StateManager.ts
```diff
@@ -640,6 +640,180 @@ class StateManager {
...
+                if (!dispute) {
+                    console.log(
+                        `Reached final undisputed fork: ${currentForkId}`
```
lukastanisic99: you can delete this entire `if block` - You don't care for the single dispute to deduct have you reached the end, you just care does the DisputeWindow exist -> `isForkDisputed` above is enough

src/stateManager/StateManager.ts
lukastanisic99: Yes - Ultimately we'll do automatic reduction when the `killPeriod` expires and here we'll just do 
`reducedResult = DisputeWindow.reducedResult`

src/stateManager/StateManager.ts
lukastanisic99: We'll move all of this reduce logic to trigger automatically when the killPeriod expires for a disputeWindow - maybe best to remove it from here and create a trello card

src/stateManager/StateManager.ts
```diff
@@ -640,6 +640,180 @@ class StateManager {
...
+                        this.channelId,
+                        currentForkId,
+                        reducedOutput.latestBlock.stateSnapshotHash,
+                        reducedOutput.forkGenesisTimestamp
+                    );
+
+                await commitTx.wait();
+                console.log(
+                    `Committed reduced result for fork ${currentForkId}`
+                );
```
lukastanisic99: up to here everything will be part of `auto-commit` when `killPeriod expires`

lukastanisic99: We'll probably just use reduce on-chain (not commit) and do the computation on-chain so we don't have to wait for the challenge period and later we can consider optimizing on gas (eventhough solidity code accounts for it already) - add a trello card

src/stateManager/StateManager.ts
lukastanisic99: break - it's fine to update snapshot in steps - e.g. Updating directly `A` to `D` is equivalent as upodating in steps `A` to `B` to `C` to `D` or any combination that get's to `D` eventually

src/stateManager/StateManager.ts
lukastanisic99: if too much for current PR create a trello card

src/stateManager/StateManager.ts
lukastanisic99: So here you're not updated to the 'latestSnapshot' of a fork -> you're updating to the `genesisSnapshot`. To update to the `latestSnapshot` you'd first update to the genesis and then to the latest using a `multicall` and calling the snapshot update with the milestone stuff.
The genesis snapshot should be available in storage as O(1) lookup for a forkId - esentially the forkId = hash(StateSnapshot.snapshotData)

lukastanisic99: So all of the code bellow is not needed - there are no milestones or blocks in this snapshot logic :D

lukastanisic99:
Before I forget 

https://github.com/peer3to/state-channels-plus/blob/4a596f919065dfe035ff8faf921153ca5ff7b497/contracts/V1/StateChannelDiamondProxy/StateSnapshotFacet.sol#L18

Above this line we also need to add a `require(newSnapshot.height == 0)` 
Technically we never generate the `genesis StateSnapshot` on-chain, just `snapshotData`, but some things are implicit like `height = 0`, `snapshot.forkId = hash(snapshotData)`, `snapshot.timestamp=disputeWindow.genesisTimestamp`

src/stateManager/StateManager.ts
cyony95: added this check. it considers that the disputes are in storage, will add them in https://trello.com/c/Nv7AGVyR/166-update-snapshot-fork-reduce-off-chain-commit-on-chain-wait-it-out-mechanism

src/stateManager/StateManager.ts
cyony95: https://trello.com/c/Nv7AGVyR/166-update-snapshot-fork-reduce-off-chain-commit-on-chain-wait-it-out-mechanism

src/stateManager/StateManager.ts
cyony95: https://trello.com/c/Nv7AGVyR/166-update-snapshot-fork-reduce-off-chain-commit-on-chain-wait-it-out-mechanism

src/stateManager/StateManager.ts
cyony95: https://trello.com/c/Nv7AGVyR/166-update-snapshot-fork-reduce-off-chain-commit-on-chain-wait-it-out-mechanism

src/stateManager/StateManager.ts
```diff
@@ -650,6 +651,129 @@ class StateManager {
...
 
+    /**
+     * Updates the state snapshot when the fork is different
+     */
+    public async updateSnapshotFork(): Promise<void> {
+        try {
+            // Get the current on-chain snapshot first
+            const currentOnChainSnapshot = StateSnapshot.from(
+                await this.stateChannelManagerContract.getStateSnapshot(
+                    this.channelId
```
lukastanisic99, line 667: add a trelo card `[optimization] - use cached/local 'currentOnChainSnapshot' instead of alwways querying the RPC node` - this is low priority so but it at the bottom of the TODO list

src/stateManager/StateManager.ts
```diff
@@ -700,17 +700,22 @@ class StateManager {
...
 
-                // Build Dispute[] from local storage;
-                // TODO, implement getDispute(commitment)
+                // Build Dispute[] from local storage confirmations
                 const disputes: DisputeStruct[] = [];
                 for (const commitment of disputeCommitments) {
-                    const dispute =
-                        this.storage.disputes.getDispute(commitment);
-                    if (!dispute) {
+                    const confirmation =
```
lukastanisic99, line 723: add a trello card to generally to transform data structures like we did with `Block` so we just use our structure which includes the confirmation and everything and the actual `DisputeConfirmation` is just a data structure that gets passed through distributed networks (p2p/dlt) that way we have a clean interface like for `Block` and always have everything

src/stateManager/StateManager.ts
```diff
@@ -650,6 +651,129 @@ class StateManager {
...
+                await this.stateChannelManagerContract.getStateSnapshot(
+                    this.channelId
+                )
+            );
+
+            let currentForkId = currentOnChainSnapshot.forkId;
+
+            // Traverse through dispute windows until we reach a fork with no disputes
+            let isDisputed =
+                await this.stateChannelManagerContract.isForkDisputed(
```
lukastanisic99, line 675: this should also be an optimization - to check locally - add a trello card

src/stateManager/StateManager.ts
```diff
@@ -650,6 +651,129 @@ class StateManager {
...
+                );
+
+            while (isDisputed) {
+                // If reduced result already exists on-chain, traverse to it
+                const existingReducedResult =
+                    await this.stateChannelManagerContract.getReducedResult(
+                        this.channelId,
+                        currentForkId
+                    );
+                if (existingReducedResult[0]) {
```
lukastanisic99, line 699: add a comment above `// if reduceResult exists and is final` since the logic for this would change if we we're doing commitments - so we have it in mind later

src/stateManager/StateManager.ts
```diff
@@ -650,6 +651,129 @@ class StateManager {
...
+                    const dispute = Codec.decode(
+                        confirmation.signedDispute.encodedDispute,
+                        Type.Dispute
+                    ) as DisputeStruct;
+                    disputes.push(dispute);
+                }
+
+                // Get the dispute window creation timestamp from the on-chain contract
+                const creationTimestamp =
+                    await this.stateChannelManagerContract.getDisputeWindowCreationTimestamp(
```
lukastanisic99, line 741: optimization - trello card

src/stateManager/StateManager.ts
```diff
@@ -650,6 +651,129 @@ class StateManager {
...
+                        this.channelId,
+                        currentForkId
+                    );
+
+                // Reduce on-chain to obtain the reduced fork id
+                await this.stateChannelManagerContract.reduce(
```
lukastanisic99: The reduce funtion in `DisputeVerificationFacet` is view only -> doesn't set the state

Take a look at `challengeDisputeReduction` to understand how a `final reduceResult is applied` it has the pipeline. So you have to create and expose a new function `reduceAndFinalize` that does the pipeline

src/stateManager/StateManager.ts
```diff
@@ -650,6 +651,129 @@ class StateManager {
...
+                    `No genesis snapshot found for fork ${currentForkId}`
+                );
+            }
+
+            // Update snapshot
+            const txResponse =
+                await this.stateChannelManagerContract.updateStateSnapshotFork(
+                    this.channelId,
+                    genesisSnapshot.toStruct(),
+                    []
```
lukastanisic99: for the exit blocks portion the `[]` what you do is 

```typescript
let latestExitBlockHash = genesisSnapshot.latesExitBlockHash;
let currentOnChainExitBlockHash = currentOnChainSnapshot.latestExitBlockHash;
let exitBlocks = [];
let currentExitBlock = storage.getExitBlock(latestExitBlockHash);
let currentExitBlockHash = hash(currentExitBlock);
while(currentExitBlockHahs != currentOnChainExitBlockHash){
    exitBlocks.pushAtBeginig(currentExitBlock); //unshift or something
    currentExitBlock = storage.getExitBlock(currentExitBlock.previousBlockHash);
    currentExitBlockHash = hash(currentExitBlock);
}
...updateStateSnapshotFork(...,exitBlocks);

```

lukastanisic99 (changes requested):
Most comments are optimization, except the last 2

src/stateManager/StateManager.ts
cyony95: you re right, that s why I called the commit function in the previous implementation but you said not to commit and i got confused. i had it exposed even. however, if you want it this way, I will do it in a reduceandFinalize one.

cyony95: function commitToReducedResult(
        bytes32 channelId,
        bytes32 disputedForkId,
        bytes32 reducedForkId,
        uint256 reducedForkGenesisTimestamps
    ) public {
        _delegatecall(
            address(disputeManagerFacet),
            abi.encodeCall(
                disputeManagerFacet.commitToReducedResult,
                (channelId, disputedForkId, reducedForkId, reducedForkGenesisTimestamps)
            )
        );
    } I'll remove this since everything is being done now in reduceAndFinalize

lukastanisic99: Yeah the commit one is exposed so you can commit, but for this logic we're expecting that it's finalized, otherwise the logic is a bit more complicated and we can do that later as an optimization - first lets get stuff working

src/stateManager/StateManager.ts
```diff
@@ -650,6 +652,216 @@ class StateManager {
...
+                    throw new Error(
+                        `No genesis snapshot found for fork ${currentForkId}`
+                    );
+                }
+
+                // Get the encoded state machine state for this fork
+                const encodedStateMachineState =
+                    this.storage.stateMachineStates.getStateMachineState(
+                        forkSnapshot.snapshotData.stateMachineStateHash
+                    );
```
lukastanisic99, line 770: You're now kinda in a tricky situation:
`encodedStateMachineState` is not the state of the genesisSnapshot of the fork but the `latestState` of all disputes reduced (e.g. DisputeA.latesState=block50, DisputeB.latestState=block55 -> reduce(DisputeA,DisputeB) -> latestState=block55 -> encodedStateMachineState = storage.getState(storage.getSnapshot(storage.getBlock(latestBlockHash // this was reduced to block 55).snapshotHash).stateMachineStateHash) )

the reason you're in a tricky situation is that 'reduce' eventhough it's fundamentally a `view` function which would return a `ReducedOutput` from the RPC node when called, is exposed to you through the `StateChannelManagerProxy` which means the `view` function is hidded behind a `delegatecall` which makes the `StateChannelManagerProxy.reduce` function not a `view` function, so if it can mutate state, regardless if it returns `ReducedOutput` it only returns that within the composable domain of the EVM, but to you when you call `StateChannelManagerProxy.reduce(...)` you'll get a `TransactionResponse` object since from the perspective of the RPC node it's a transaction not a call 

If we ran this in our localEvm, we can easily get the `ReducedOutpu` and not the `TransactionResponse` which is useless to us + we don't want to do a transaction, we want to do a fast and cheap querry of the RPC node or our localEvm. 

If it was a `pure` function we could simply call `DisputeVerificationFacet.reduce()`, but now we can't do that since the `view` reads `storage` and the correct storage is in the `StateChannelManagerProxy` which would provide its storage under the `delegatecall` to the `DisputeVerificationFacet` that would interpret that storage and compute reduce on it 

We could emit an event from `StateChannelManagerProxy.reduce()` and collect that event, but I don't want to populate more events + I don't want a transaction that someone has to pay TX fees and wait for finality when it's a view operation on the state. 

The simplest thing is to move reduce into `StateChannelCommon` but that increaces the bytecode of all facets and logically it belongs in disputes and not something that's used elsewhere 

Think a bit and we'll discuss tomorrow in the call some options

cyony95: I spent more than a few hours yesterday to try to get past this issue before I submitted the last commit. That's why I tried to go around with the whole "forksnapshot" implementation but that s still not right" I thought I was doing something wrong but thanks for clearing it up. Let's talk today about it.

contracts/V1/StateChannelDiamondProxy/StateChannelManagerProxy.sol
```diff
@@ -384,6 +385,87 @@ contract StateChannelManagerProxy is StateChannelManagerInterface, StateChannelC
...
+
+    function reduceProxyView(Dispute[] memory disputes, uint256 disputeWindowCreationTimestamp)
+        external
+        view
+        returns (ReduceOutput memory reducedOutput)
+    {
+        // This should trick the compiler, but still use delegatecall under the hood. This doesn't magically allow us to mutate the state in view functions that use delegate call.
+        // Ethers and other tools won't issue a transaction, but a call that runs on a single node so even if it did modify the state it wouldn't be reflected on-chain/persisted.
+        return DisputeVerificationFacet(address(this)).reduce(disputes, disputeWindowCreationTimestamp);
+    }
```
lukastanisic99, line 418: @cyony95 take a look and try and test this - long term we'll remove it the moment we use the localEvm, but this should allow you to use reduce from the proxy

under the hood, it thinks it's calling `DisputeVerificationFacet.reduce` which is a `view`, but it's given the address of the proxy (`this`) which also has the same method signature `reduce(...)` which should map to the same function identifier and actually call `StateChannelManagerProxy.reduce(...)` which will use `delegatecall` under the hood, but Ethers and other tools don't know that, so they won't/shouldn't issue a transaction, but a call and the RPC node will use `staticcall` which can NOT modify state, eventhough technically delegatecall can do that. In our case we really do call a `view` function and don't modify the state, but if we tried to do it, we'd realize that the state doesn't get modified and replicated on other nodes and it's discarded after the call on the single RPC node that performs the call locally

I didn't test this, but I belive this works like this - see if you can test it

src/stateManager/StateManager.ts
```diff
@@ -650,6 +652,225 @@ class StateManager {
...
+                        );
+                    if (!currentJoinChannelBlock) {
+                        break;
+                    }
+                    latestJoinChannelBlockHash = ethers.keccak256(
+                        Codec.encode(
+                            currentJoinChannelBlock.block,
+                            Type.JoinChannelBlock
+                        )
+                    );
```
lukastanisic99: You almost got lucky by accident :D 
If you swtiched what you call `latestJoinChannelBlockHash` and `targetJoinChannelBlockHash` it would work :P 

The logic of iterating should be `from(reduced.latestJoinChannelBlockHash iterate backwards to genesisSnapshot.latestJoinChannelBlockHash)` you used `latestSnapshot (notGenesis)`, but since joining doesn't happen off-chain (in the same fork) the `.latestJoinChannelHash` doesn't change and all snapshots in the fork have the same value which is `genesisSnapshot.latestJoinChannelBlockHash`, but you started iterating from that hash instead of stoping when that has is reached so you would fetch all joinChannelBlocks excpet the ones we're interested in :D 

A simple fix would be to just rotate how you assign `latestJoinChannelBlockHash` and `targetJoinChannelBlockHash` but I'd still be more explicit for easy reading and use the `genesisSnapshot` here for this (plus you already fetched it above)

src/stateManager/StateManager.ts
```diff
@@ -650,6 +652,225 @@ class StateManager {
...
+
+                // Reduce and finalize on-chain to obtain the reduced fork id
+                await this.stateChannelManagerContract.reduceAndFinalize(
+                    disputes,
+                    creationTimestamp,
+                    latestSnapshot.toStruct(),
+                    encodedStateForReduce,
+                    joinChannelBlocks
+                );
+
```
lukastanisic99, line 820: here you'd have to wait for the tx to finalize first, since this `await` yields when the RPC node gives you your `TransactionReceipt` (the transaction isn't included in the block and it didn't modify state yet) - so you'd have to do:

```typescript
let txReceipt = await reduceAndFinalize(...);
let txResponse = await txReceipt.wait() // I think this is the interface - check
checkForError(txResponse) // not sure will it throw implicitly in that case you need to catch it and cast it into our customError so you can easily interpret it 

// only if success continue with the rest
```

`reduceAndFinalize` will triger automatically when the `killPeriod` expires so it won't be part of this code and it will be moved there and this code will assume that it's `final` on-chain when it runs - let's first get this right and then we'll move the logic to `setTimeout` - create a Trello card

cyony95:
```
if (customError?.name === "ErrorDisputeAlreadyReduced") {
     break; // The reduced result was already computed, continue the pipeline
}
throw new Error(`reduceAndFinalize failed: ${customError?.name || "unknown error"}`);
```
the only way we want to move forward is if the dispute was already reduced, otherwise we throw an error, checked all the other error messages and they re (in my opinion) all breaking errors
```
ErrorNoDisputesProvided()
ErrorCantParticipateInDispute()
ErrorDisputeCommitmentNotAvailable()
ErrorInvalidStateSnapshot()
ErrorInvalidLatestState()
ErrorDisputeJoinChannelBlocksInvalid()
ErrorDisputeKillPeriodNotExpired()
ErrorDisputeAlreadyReduced()
ErrorDisputeStateMachineJoiningFailed()
```

src/stateManager/StateManager.ts
cyony95: you're totally right, I guess it's from all the drilling in the walls today

src/stateManager/StateManager.ts
```diff
@@ -650,6 +653,240 @@ class StateManager {
...
+                const txReceipt = await txResponse.wait();
+                if (txReceipt?.status === 0) {
+                    const errorData = txReceipt.logs?.[0]?.data;
+                    const customError = errorData
+                        ? new ethers.Interface(errorAbis).parseError(errorData)
+                        : null;
+
+                    if (customError?.name === "ErrorDisputeAlreadyReduced") {
+                        break; // The reduced result was already computed, continue the pipeline
+                    }
```
lukastanisic99, line 818: This `break` will break the `while` and it won't execute the code bellow - not sure this is the intended behavior since `AlreadyReduced` should be the same as no error 

@daiagi can you take a look at the way that these custom errors are interpreted - I belive we had a proxy and that we're just interacting we our `custom error` without having to decode this stuff again with ethers

src/stateManager/StateManager.ts
```diff
@@ -650,6 +653,240 @@ class StateManager {
...
+                        );
+                    if (!currentJoinChannelBlock) {
+                        break;
+                    }
+                    currentJoinChannelBlockHash = ethers.keccak256(
+                        Codec.encode(
+                            currentJoinChannelBlock.block,
+                            Type.JoinChannelBlock
+                        )
+                    );
```
lukastanisic99:
```suggestion
                    currentJoinChannelBlockHash = currentJoinChannelBlock.block.previousBlockHash;
                    currentJoinChannelBlock =
                        this.storage.joinChannelBlocks.getJoinChannelBlockEntry(
                            currentJoinChannelBlock.block.previousBlockHash
                        );
```

src/stateManager/StateManager.ts
```diff
@@ -650,6 +653,240 @@ class StateManager {
...
+                currentExitBlock =
+                    this.storage.exitChannelBlocks.getExitChannelBlockEntry(
+                        currentExitBlock.block.previousBlockHash
+                    );
+                if (!currentExitBlock) {
+                    break;
+                }
+                latestExitBlockHash = ethers.keccak256(
+                    Codec.encode(currentExitBlock.block, Type.ExitChannelBlock)
+                );
```
lukastanisic99:
```suggestion
                latestExitBlockHash = currentExitBlock.block.previousBlockHash;
                currentExitBlock =
                    this.storage.exitChannelBlocks.getExitChannelBlockEntry(
                        currentExitBlock.block.previousBlockHash
                    );
```

src/stateManager/StateManager.ts
cyony95: good catch with the break, I forgot i can only break out of loops and switches in TS :D I redid the check to throw an error only if the error is not ErrorDisputeAlreadyReduced, otherwise it's business as usual

cyony95: @lukastanisic99 you are right, I checked and we have a wrapper called decodeErrorProxy. I added the error and an additional try catch block to re-throw the error if it s not ErrorDisputeAlreadyReduced

src/stateManager/StateManager.ts
cyony95: @lukastanisic99 in here, I had to cast currentJoinChannelBlock.block.previousBlockHash to hex string to satisfy the compiler.
```
                    currentJoinChannelBlockHash = ethers.hexlify(
                        currentJoinChannelBlock.block.previousBlockHash
                    ); 
```

lukastanisic99 (approved):
Made a small change 

Overall fine for now - some future PR will introduce more changes

PR #147, opened 2025-08-28
Check-state-snapshot-commitment

daiagi:
- Closes https://trello.com/c/Bh1Hnb4H

### DIFF
--- a/src/stateManager/StateManager.ts
+++ b/src/stateManager/StateManager.ts
@@ -283,7 +283,9 @@ class StateManager {
         });
     }
 
-    // Passes the block confirmation through a verification pipeline and returns shouldDisconnect flag
+    // Passes the block confirmation through a verification pipeline
+    // returns true if the block is valid and the state transition is successful
+    // returns false -> the calling context should disconnect from the peer
     public async onBlockConfirmation(
         blockConfirmation: BlockConfirmationStruct
     ): Promise<boolean> {
@@ -355,8 +357,8 @@ class StateManager {
                 exitChannelBlock
             );
 
-            // sucess - no disconnect
-            return false;
+            // success - no disconnect
+            return true;
         } catch (error) {
             throw error;
         } finally {
--- a/src/stateManager/StateManager.ts
+++ b/src/stateManager/StateManager.ts
@@ -23,7 +23,6 @@ import {
 // TypeChain types - Dispute types
 import {
     DisputeStruct,
-    ReduceOutputStruct,
     SignedDisputeStruct
 } from "@typechain-types/contracts/V1/types/DisputeTypes";
 
@@ -44,7 +43,7 @@ import Storage from "@/storage";
 import P2pEventHooks from "@/P2pEventHooks";
 
 // Models
-import { Block, BlockCoordinates, StateSnapshot } from "@/models";
+import { Block, StateSnapshot } from "@/models";
 
 // Utils
 import {
@@ -57,6 +56,7 @@ import {
     isCustomEvmError,
     getActiveParticipants,
     SignatureUtils,
+    difference,
     decodeErrorProxy
 } from "@/utils";
 // Types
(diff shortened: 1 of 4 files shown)


src/stateManager/StateManager.ts
```diff
@@ -846,40 +852,75 @@ class StateManager {
...
+        )!;
 
         const latestJoinChannelBlockHash =
-            this.storage.joinChannelBlocks.getLatestJoinChannelBlockHash();
-        const latestExitChannelBlockHash =
-            this.storage.exitChannelBlocks.getLatestExitChannelBlockHash();
-        const totalDeposits = this.storage.joinChannelBlocks.getTotalDeposits();
-        const totalWithdrawals =
-            this.storage.exitChannelBlocks.getTotalWithdrawals();
+            (await this.localDiamondContract.getLatestJoinChannelBlockHash(
```
lukastanisic99: this should be - `genesisSnapshot.latestJoinChannelBlockHash` - you can't join off-chain in the same fork - only on-chain which creates another fork

src/stateManager/StateManager.ts
```diff
@@ -846,40 +852,75 @@ class StateManager {
...
+                this.channelId
+            )) as Hash;
+        const totalDeposits = await this.localDiamondContract.getTotalDeposits(
```
lukastanisic99: this should also be `genesisSnapshot.totalDeposits`

src/stateManager/StateManager.ts
```diff
@@ -846,40 +852,75 @@ class StateManager {
...
+
+        let exitChannelBlock: ExitChannelBlockStruct | undefined;
+
+        if (exitChannels && exitChannels.length > 0) {
+            participants = Array.from(
+                difference(
+                    new Set(participants),
+                    new Set(
+                        exitChannels.map(
+                            (exitChannel) => exitChannel.participant
```
lukastanisic99: We want to account for partial exits (when someone withdraws less then their full balance and remains in the channel) - so this fields is fetched from the stateMachine as `participants = stateMachine.getParticipants(posteriorState)` the stateMachine will do the accounting will it decide to premove the participant (full exit) or keep (partial exit)

So also the trigger for including an `exit point` should be `if(oldParticipants != posteriorParticipants) -> exit point` 

You can add a trello card for the exit points if it's not accounted for already

src/stateManager/StateManager.ts
```diff
@@ -846,40 +852,75 @@ class StateManager {
...
+            );
+        }
 
         const stateSnapshot: StateSnapshotStruct = {
-            forkId,
-            blockHeight: blockHeight,
-            timestamp: Clock.getTimeInSeconds(),
+            forkId: block.coordinates.forkId,
+            blockHeight: BigInt(block.coordinates.height),
+            timestamp: BigInt(Clock.getTimeInSeconds()),
```
lukastanisic99: Block.timestamp - in reality this field is only usefull for the genesisSnapshot for now

src/stateManager/StateManager.ts
```diff
@@ -99,8 +99,7 @@ class StateManager {
     private dispatcher = new Map([
         [BlockValidationAction.DISPUTE, this.dispute],
         [BlockValidationAction.BROADCAST, this.broadcast],
-        [BlockValidationAction.NOT_ENOUGH_TIME, this.notEnoughTime],
-        [BlockValidationAction.SUCCESS, this.success]
+        [BlockValidationAction.NOT_ENOUGH_TIME, this.notEnoughTime]
```
lukastanisic99: @daiagi missing merge with `dispute` to account for new changes - what I never understood, why doesn't github display updated files in a PR when the destination branch is updated (e.g. today's PR got merged into `dispute` and it's not reflected here)

daiagi: image


because i didn't rebase yet

lukastanisic99:
@daiagi there's still some conflict - probably because Adrians PR got merged

lukastanisic99 (approved):

lukastanisic99 (approved):
Overall good - just resolve the merge conflict

PR #149, opened 2025-08-28
update time validation

daiagi:
update time logic check to follow the diagram

## 💪 Related Issues

- Closes https://trello.com/c/8cLJYFfm
- Closes https://trello.com/c/gI7XhT3M

### DIFF
--- a/src/stateManager/ValidationService.ts
+++ b/src/stateManager/ValidationService.ts
@@ -302,6 +302,7 @@ export default class ValidationService {
         previousblockOrSnapshot: BlockOrSnapshot,
         channelId: ChannelId
     ): Promise<ValidationResult | undefined> {
+        // Calculate previousTimestamp
         let previousTimestamp: Timestamp;
         let previousBlock: Block | undefined;
         let previousStateSnapshot: StateSnapshot | undefined;
@@ -316,18 +317,19 @@ export default class ValidationService {
             previousTimestamp = previousStateSnapshot!.timestamp;
         }
 
+        // OBJECTIVE: isValidTimestamp check
         const isValidTimestamp =
             block.timestamp >= previousTimestamp &&
             block.timestamp <= previousTimestamp + this.timeConfig.p2pTime;
 
         if (!isValidTimestamp) {
             if (
-                block.height <= 0 || // genesis block
-                previousBlock === undefined || // no previous block == genesis block
+                // first block
+                previousBlock === undefined ||
+                //  has on-chain timestamp
                 previousBlock.onChainTimestamp !== undefined
             ) {
-                // already has best timestamp
-                // no point in trying to update the timestamp => fraud proof
+                // Already has best timestamp - persist InvalidTimestamp fraud proof
                 this.fraudProofService.createInvalidTimestampProof(block);
                 return {
                     shouldDisconnect: true,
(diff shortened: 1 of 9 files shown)


lukastanisic99:
@daiagi I pushed a commit - take a look 

I've also added a Trello card that STF and other parts of the pipeline need to be outside of success() since success() should only trigger when the final flag is SUCCESS not as an intermediary thing 

Most of the 'logical' changes here are to timelogic and there are not much - other are structural/simplification - was easier for me to just change and commit than to write a text 

Let me know your thoughts 

Btw I didn't run/update tests

daiagi:
>STF and other parts of the pipeline need to be outside of success() 

Already moved it out (other PR)
https://github.com/peer3to/state-channels-plus/pull/147/files#diff-f06ee01cce3e8caa18790180dfd8f6ee12ff7eb9df4afae8eff20dab3c7d5fda

src/stateManager/ValidationService.ts
```diff
@@ -40,20 +35,20 @@ export default class ValidationService {
...
+    ): Promise<BlockValidationResult> {
         const forkId = this.getForkId();
         const channelId = this.channelId;
 
         // 1. Check if channel is open
         if (!this.isChannelOpen(forkId)) {
             // not ready
             this.queueForLater(blockConfirmation);
-            return { shouldDisconnect: false };
+            return BlockValidationResult.DISCONNECT;
```
daiagi, line 40: `{ shouldDisconnect: false }` converted to DISCONNECT

mistake or on purpose?

src/stateManager/ValidationService.ts
```diff
@@ -420,54 +397,90 @@ export default class ValidationService {
...
         }
     }
+
+    private async isPostedOnChainTooLate(
+        previousTimestamp: Timestamp,
+        block: Block
+    ): Promise<boolean> {
+        // if doesn't have on-chain timestamp try and fetch it
+        if (!block.onChainTimestamp) {
+            let onChainTimestamp = await this.fetchOnChainTimestamp(block);
```
daiagi, line 462:
```suggestion
            const onChainTimestamp = await this.fetchOnChainTimestamp(block);
```

src/stateManager/ValidationService.ts
```diff
@@ -264,22 +247,23 @@ export default class ValidationService {
...
-                shouldDisconnect: true,
-                action: BlockValidationAction.DISPUTE
-            };
+            return BlockValidationResult.DISPUTE;
         }
 
-        return { shouldDisconnect: true };
+        // if first block -> wrong genesis
+        if (conflictingBlock.height === 0) {
+            //TODO
```
daiagi, line 255: TODO what?
fraud proof?

daiagi:
The changes you made to time logic are fine
but i don't like what you did with the return values

the flags were meant to represent the final outcome of the entire pipe, and to be actionable 
you have turned them to a mix of actionable and non actionable items

AFTER your change
- SUCCESS became an  intermediate result, not the final one,  i don't like it
-  DUPLICATE is a non actionable flag, it will just be passed around to be converted finaly into false (representing do not disconnect)
- NOT_READY - same as  DUPLICATE, not actionable


you are stepping back in the direction you explicitly asked to move away from, too many flags which are not actionable

lukastanisic99:
Yes you're right, they might not imply the final decision and you don't need them at all you could just call a function success() instead of returning a flag and calling it in the end which may be better 

The flags, unlike before are not exposed to other parts of the system, they stay visible only inside the pipe and to the outside world (p2pManager) a single boolean is returned

I know that some flags like `duplicate` may not have any side-effects (for now) and you can omit them, but they're there for reading clarity if you're looking at the diagrams

I know you returned `undefined` as an intermediary state to chain the next validation check, while I did it more explicitly with SUCCESS to indicate this check is correct, which doesn't guarantee that all checks are correct which is like an intermediary state

Imo it doesn't make much of a difference, if you want `undefined` feel free to change back, imo it's just less clear when a check returns undefined you don't really associate undefined with success.

Again you can remove all the flags and just use functions as the final decision and some intermediary boolean for every check `isValid` if you want to know should you continue the pipe

We can discuss tomorrow

PR #150, opened 2025-09-01
state manager clean up

daiagi:
## 💪 Related Issues

- Closes https://trello.com/c/3MqMZIg8

### DIFF
--- a/src/stateManager/StateManager.ts
+++ b/src/stateManager/StateManager.ts
@@ -927,40 +927,106 @@ class StateManager {
     private async tryTimeoutParticipant(
         forkId: ForkId,
         blockHeight: BlockHeight,
-        participantAdr: Address
-    ) {
-        if (participantAdr == this.signerAddress) return;
+        participantAddress: Address
+    ): Promise<void> {
+        if (participantAddress === this.signerAddress) {
+            return;
+        }
 
+        // Get the block entry - if it doesn't exist (can happen ONLY from setState), skip timeout
         const blockEntry = this.storage.blocks.getBlockEntry(
             forkId,
             blockHeight
         );
-        if (blockEntry) {
-            const block = blockEntry.block;
-            const participants = this.storage.getParticipants(
-                block.coordinates
-            );
-
-            if (block.didEveryoneSign(participants)) return;
-
-            if (blockEntry.onChainTimestamp !== undefined) return;
+        if (!blockEntry) {
+            return;
         }
 
-        const response =
+        const block = blockEntry.block;
+        const participants = this.storage.getParticipants(block.coordinates);
+
+        // If everyone already signed or block has already onChainTimestamp, no timeout needed
+        if (
+            block.didEveryoneSign(participants) ||
+            blockEntry.onChainTimestamp !== undefined
+        ) {
+            return;
+        }
+
+        // Check if participant posted a commitment on-chain
+        const commitmentResponse =
             await this.stateChannelManagerContract.getBlockCallDataCommitment(
                 this.channelId,
                 forkId,
                 blockHeight,
-                participantAdr
+                participantAddress
             );
-        if (response.found) return;
 
-        // we have a "block" hash, but we need to make sure it is a commitment to a real block
-        // can re-use fetchOnChainTimestamp to get the real block and time stamp (underlying commitment)
-        // if the block commited is junk the we force timeout (see bool isForced on Timeout struct)
+        if (!commitmentResponse.found) {
+            // No commitment posted - proceed with timeout
+            await this.createTimeOutDispute(
+                forkId,
+                blockHeight,
+                participantAddress
+            );
+            return;
+        }
 
-        // after getting the block we feed it to the event handler that was suppose to fire when the block data was posted
+        // Validate the on-chain commitment is legitimate
+        const isValidCommitment = await this.validateBlockCommitment(
+            blockEntry.block,
+            commitmentResponse.blockCalldataCommitment,
+            participantAddress
+        );
 
+        if (!isValidCommitment) {
+            // Invalid commitment - force timeout
+            // TODO
+        }
+    }
+
+    private async validateBlockCommitment(
+        block: Block,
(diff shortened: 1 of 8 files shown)


cyony95 (approved):

daiagi:
@lukastanisic99  updated the `tryTimeoutParticipant` with the comments we wrote down during the meet
please tell me once again what it means a forced timeout?
just to set the flag on the contract storage and the regular timeout dispute i assume.
anything more?

waiting on your review

Thanks

src/stateManager/StateManager.ts
```diff
@@ -292,7 +264,8 @@ class StateManager {
 
             const validationResult =
                 await this.validationService.validateBlockConfirmation(
-                    blockConfirmation
+                    blockConfirmation,
+                    onChainTimestamp
```
lukastanisic99, line 265: We need to meld this (`onChainTimestamp`) into the block model as soon as possible and try to use the block model as soon as possible - probably the first thing we should do is authenticate the block, so we just use our block model for the pipeline 

Currently if the channel is not opened, the block is queued without the onChainTimestamp, so the onChainTimestamp is lost

src/stateManager/StateManager.ts
```diff
@@ -733,70 +708,158 @@ class StateManager {
...
         if (
-            Clock.getTimeInSeconds() <
-            this.agreementManager.getChainLatestBlockTimestamp(
-                forkId,
-                transactionCnt
-            ) +
-                this.getTimeoutWaitTimeSeconds()
-        )
+            block.didEveryoneSign(participants) ||
+            blockEntry.onChainTimestamp !== undefined
```
lukastanisic99: or `didISign` this is a subset of `didEveryoneSign`. This esentially redenrs `didEveryoneSign` redundant - but we can keep it for mental model/clarity even though it's inefficiency in the code 

As for background on this design for people interested:
The person invoking timeout on-chain is esentially `vouching` that the timedout participant can NOT challenge the timeout, so if they're wrong they're fully slashed - One of the ways to challenge it is to shown an N/N threshold. A peer can't rely on `didEveryoneSign`, since peers can be coluding -> witholding signatures, so an honest peer may see a subset of sigantures and invoke timeout on-chain, just to realize that they've received a full slash since the attacker presented the witheled signatures on-chain. So the only relevant signature to them is theirown, which they know the attacker can not forge. So if their signature is missing on the block they would invoke timeout and know that the attacker would not be able to present an N/N threshold.

They still may observe that a threshold is missing, but that doesn't matter the state will eventually be finalized through `virtual voting` or `disputes`. 

We can remove `didEveryoneSign` here if it's confusing, eventhough it was part of the original design

src/stateManager/StateManager.ts
```diff
@@ -733,70 +708,158 @@ class StateManager {
...
+        if (!onChainResult) {
+            return false;
+        }
+
+        const { signedBlock, timestamp } = onChainResult;
+        const expectedCommitment = hash(
+            Codec.encode({ signedBlock, timestamp }, Type.BlockCommitment)
+        );
+
+        const isValid = expectedCommitment === blockCalldataCommitment;
```
lukastanisic99, line 999: this is of course 1 of the logical and intuitive checks, but it's not usefull since the way you fetched this data (through `fetchBlockCommitmentCalldata` and `getBlockCallDataCommitment`) guarantees that this relation is always `true`, assuming the blockchain is honest - the hash is computed on-chain and the data that was used to copute the has is emited and you just do the same computation off-chain and compare it.

When I said we need to validate the block against the commitment it was to validate that the original data posted on-chain is a `valid block` that can correctly be applied to the state at that block height, so it's trying to run the blockConfirmation pipeline  - see how to get that information in a practical way - I'll let you think about it and if needed I'll suggest something later

src/stateManager/StateManager.ts
```diff
@@ -733,70 +708,158 @@ class StateManager {
...
+                participantAddress,
+                block.signedBlock,
+                timestamp
+            );
+        }
+
+        return isValid;
+    }
+
+    private async createTimeOutDispute(
```
lukastanisic99, line 1018: Feel free to leave a `todo` for this (crating a dispute) and a trello card so we don't forget to full this todo, but a trello card exists for creating a dispute that takes into account this and it's just a simple call after it's implemented

lukastanisic99 (commented):
> @lukastanisic99  updated the `tryTimeoutParticipant` with the comments we wrote down during the meet
> please tell me once again what it means a forced timeout?
> just to set the flag on the contract storage and the regular timeout dispute i assume.
> anything more?
> 
> waiting on your review
> 
> Thanks
>  

Left some comments above

As for what a `force timeout` is - It's when someone posted `junk` / not a valid block on-chain, so you just set `isForced` to true as 1 of the arguments when you call `createDispute` ( https://trello.com/c/qwpYPLj8/127-disputes-creation-management-submission-dependent-on-storage-statemanager ) which will create the `Timeout Struct` (part of the `Dispute Struct`) with isForced set to true, which tells the on-chain contract "I'm aware that a `participant` commited calldata for `frokId`, `height` and I vouche that it's junk and not applicable to the state, so ignore the race condition check that is there to protect me in the case that I was not aware that calldata was posted and proceed to timeout`
If the block is not junk, the challenger will be able produce a valid proof on-chain will will slash the disputer and in the process cancel the timeout.

src/stateManager/StateManager.ts
daiagi: well the name maybe is not accurate
this function does not necessarily create the dispute
it may call internally th (outdated) `disputeHandler.createDispute`, which is the call that will be replaced later

daiagi:
>As for what a force timeout

i see, thank you for the clarification.
beyond scope for this PR

>this relation is always true

:face_exhaling:  of course, that was silly

i fixed it by using the block that was fetched from storage for the commitment validation 
and passing `onChainResult.signedBlock` to `stateChannelEventListener.handleBlockCalldataPosted`

lukastanisic99 (approved):

PR #151, opened 2025-09-04
Spectate RPC

cyony95:
closes https://trello.com/c/iKqCQbb5/125-sync-spectate-rpc-dependent-on-stateproof

RPC Methods (Following InitHandshakeService pattern):
- spectateSync(channelId): Local initiation method
- onSpectateRequest(channelId, time): Runs on participant - generates proof data
- onSpectateResponse(channelId, snapshotPayload, time): Runs on spectator - verifies proof data

onSpectateRequest:

1. Validate time - Ensure request is recent
2. Get current state
3. Get fork genesis snapshot
4. Check disputes - Query blockchain for dispute status
5. Build dispute windows - Decode disputes from local storage
6. Package payload - Combine all proof data
7. Send response - Return payload to spectator                

onSpectateResponse:

1. Validate RTT - Ensure response is recent
2. Fetch on-chain snapshot
3. Verify fork proof - Traverse disputes to compute latest fork ID
4. Verify fork ID - Compare computed vs participant's claimed fork ID
5. Verify state proof - Compare encodedState computed stateproof with participant provided stateProof
6. Handle verification failure - Exit if verification fails and disconnect from all peers
7. Sync to proven state - Store proven state and update fork ID - Not sure I did it right
8. Enable block processing - Ready to process blocks like a participant  TODO, need info


src/rpc/services/SpectateService.ts
```diff
@@ -0,0 +1,418 @@
...
+        } catch (error) {
+            console.error("Failed to sync to proven state:", error);
+            throw error;
+        }
+    }
+
+    /**
+     * TODO Apply blocks from queue in blockConfirmationPipeline
+     */
+    private async applyQueuedBlocksFromPipeline(channelId: ChannelId) {
```
cyony95, line 560: placeholder implementation here

src/rpc/services/SpectateService.ts
```diff
@@ -0,0 +1,418 @@
...
+                `onSpectateRequest - time difference too big - time:${time} localTime:${localTime} diff:${
+                    time - localTime
+                } aggreeTime:${
+                    this.mainRpcService.p2pManager.stateManager.timeConfig
+                        .agreementTime
+                }`
+            );
+            return;
+        }
+        console.log(`onSpectateRequest - localTime:${localTime} time:${time}`);
```
lukastanisic99: You don't have to check time here, since by default `InitHandhsahe` will always run first and it's a requirement to continue 

If someone wanted to be byzantine they can lie about the timestamp regardless. Spectating is fully subjective and not byzantine tolerant, you're not inheriting security from a blockchain and it has no cost - it's a soft check to see if people are cooperating and if they're not, you just abort before commiting on-chain (induring cost). 

RTT (Round trip time) would be usefull to a spectator to know is the other person 'responding in time' and if not disconnect/abort

src/rpc/services/SpectateService.ts
```diff
@@ -0,0 +1,418 @@
...
+    }
+
+    public async onSpectateResponse(
+        channelId: ChannelId,
+        snapshotPayload: SnapshotPayload,
+        responseTime: Timestamp
+    ) {
+        console.log(`onSpectateResponse - start`);
+        let localTime = Clock.getTimeInSeconds();
+        let rtt = localTime - responseTime;
```
lukastanisic99: this is not RTT - RTT is the difference bewtween when you sent the request and when you got the response - it's the difference between `your` local clock at those 2 points - check in `initHanskahe` how it's caluclated

src/rpc/services/SpectateService.ts
```diff
@@ -0,0 +1,418 @@
...
+        if (
+            rtt >
+            this.mainRpcService.p2pManager.stateManager.timeConfig.agreementTime
+        ) {
+            console.log("onSpectateResponse - RTT too high, ignoring");
+            return;
```
lukastanisic99, line 92: you want to abort this channel all together - disconnect from everyone if you detect at least 1 dishonest peer

src/rpc/services/SpectateService.ts
```diff
@@ -0,0 +1,418 @@
...
+            if (!confirmation) {
+                throw new Error(
+                    `Missing Data Availability for dispute commitment ${commitment}`
+                );
+            }
+            const dispute = Codec.decode(
+                confirmation.signedDispute.encodedDispute,
+                Type.Dispute
+            ) as DisputeStruct;
+            disputeWindows.push(dispute);
```
lukastanisic99: A `disputeWindow` is not a `dispute` so not sure why you're pushing `disputes` into it. A single `disputeWindow` can contain multiple `disputes` and there can be multiple windows - the verifier/spectator can get all `disputeWindows` on-chain, but you need to provide `disputes` for every `disputeWindow` so they can verify each window following the fork choice rule as in `diagrams` as in what you kinda already did for updating the snapshot on different forks - so you have to iterate from the `onChainSnapshot` to the `latestFork` and package all the `disputes` for every `disputeWindow` along the way (maybe some other data too, probably not)

The spectator can iterate from `onChainSnapshot` to `latestFork` following the fork choice rule and ignore if `disputeWindows` don't have finalized `reduced outputs` on-chain since they can run reduce on the data you provided to them - they only care that for all `disputeWidnows`, `killPeriods` have expired, so they know that data `can NOT` change and once they verify the data you provided them, they know with certanty that they have all the data available (DA) to sync and start verifying the state and later join the channel and participate. 

The stateProof part is only relevant once they've proven/verified the `latestFork, latestGenesisSnapshot`

If anything fails - they abort and disconnect from everyone and go interact with another channel or start a new one on their own

src/rpc/services/SpectateService.ts
```diff
@@ -0,0 +1,418 @@
...
+    StateProofStruct
+} from "@typechain-types/contracts/V1/types/ProofTypes";
+import { DisputeStruct } from "@typechain-types/contracts/V1/types/DisputeTypes";
+import { Codec, Type } from "@/utils";
+import { isEqual } from "lodash";
+
+export interface SnapshotPayload {
+    disputeWindows: DisputeStruct[];
+    stateProof: StateProofStruct;
+    latestForkGenesisSnapshot: StateSnapshot;
```
lukastanisic99, line 13: You for sure have to share the `encodedState` that stateProof proves in the end, so they can start replicating the state from that state

lukastanisic99: And think does this design cover cases when `StateProof` is not needed (no blocks on the latestGenesisSnapshot - prove just the genesis - e.g. analogy to calling `updateSnapshotFork`) and when disputeWindows are not needed (e.g. analogy to calling `updateSnapshotSameFork`)

cyony95: you're right, we should keep StateProof(with encodedState) and disputeWindows optional.

src/rpc/services/SpectateService.ts
```diff
@@ -0,0 +1,531 @@
...
+} from "@typechain-types/contracts/V1/types/ProofTypes";
+import { DisputeStruct } from "@typechain-types/contracts/V1/types/DisputeTypes";
+import { Codec, Type } from "@/utils";
+import { isEqual } from "lodash";
+
+export interface SnapshotPayload {
+    latestForkGenesisSnapshot: StateSnapshot;
+    disputeWindows?: DisputeStruct[][];
+    stateProof?: StateProofStruct;
+    encodedState?: Hash;
```
lukastanisic99: this should be `bytes` not `Hash`

src/rpc/services/SpectateService.ts
```diff
@@ -0,0 +1,531 @@
...
+}
+
+class SpectateService extends ARpcService {
+    private spectateInitTime: number | null = null;
```
lukastanisic99: doing it like this is fine if you expect only 1 RPC call and having 1 per channel is fine, but we'll run multiple channels in parralel so if some fail you connect to at least 1 successfully with some degree of high probability - in that case this single number is overwritten - this should be a map like in `InitHandshakeService`

src/rpc/services/SpectateService.ts
```diff
@@ -0,0 +1,531 @@
...
+        }
+
+        const latestBlockHeight =
+            stateManager.storage.blocks.getNextBlockHeight(forkId) - 1;
+
+        // Get the latest state proof data (only if there are blocks for same-fork update)
+        let latestStateProof: StateProofStruct | undefined;
+        let encodedState: Hash | undefined;
+
+        if (latestBlockHeight >= 0) {
```
lukastanisic99: I've accounted for no blocks in `stateProof` verification is solitiy (in my latest PRs) - it will enforce `genesisSnapshot == latestSnapshot` in those cases. 

`getStateProof` should return an empty `stateProof` in those cases and solidity know how to verify it, so this check is not needed

src/rpc/services/SpectateService.ts
```diff
@@ -0,0 +1,531 @@
...
+            if (existingReducedResult[0]) {
+                // Traverse to the reduced fork
+                currentForkId = existingReducedResult[0];
+                isDisputed =
+                    await stateManager.stateChannelManagerContract.isForkDisputed(
+                        channelId,
+                        currentForkId
+                    );
+                continue;
+            }
```
lukastanisic99: On a first though this seems like the optimal way to do it "If it's reduced I don't need to provide the data to them to reduce it, they can just check on-chain and if it's not reduced I'll provide the data, so they can verify for themselves - keep in mind when you do it like that there could be race condition on-chain where somone reduced a disputeWindow on-chain `after you packaged the payload` and `before the spectator verifies it`, so you have to account for it in the verification code

lukastanisic99: plus I think this check doesn't take into account is the reduced result on-chain final, only with final ones you can have this optimization

lukastanisic99: I'd keep it simple and not optimized (to not think about these race conditions and edge cases for now) and just provide all disputeWindows and they will run reduce on everything regardless to derive the final forId - if there's inconsistency they would just abort

src/rpc/services/SpectateService.ts
```diff
@@ -0,0 +1,531 @@
...
+                }
+
+                // No existing reduced result, use the disputes provided by participant
+                if (
+                    snapshotPayload.disputeWindows &&
+                    snapshotPayload.disputeWindows.length > 0
+                ) {
+                    // Pop the next dispute window
+                    const currentWindowDisputes =
+                        snapshotPayload.disputeWindows.shift()!;
```
lukastanisic99: yep doesn't account for race condtion / the posibility that this disputeWindow could be reduced on-chain and skiped above

src/rpc/services/SpectateService.ts
```diff
@@ -0,0 +1,531 @@
...
+                            currentWindowDisputes,
+                            creationTimestamp
+                        );
+                    console.log(
+                        `Computed reduced output: forkGenesisTimestamp=${reducedOutput.forkGenesisTimestamp}`
+                    );
+
+                    // Move to next fork
+                    const nextForkId =
+                        reducedOutput.latestBlock.stateSnapshotHash;
```
lukastanisic99: so the forkId is not the hash of the `StateSnapshot` but the hash of `SnapshotData`, but that doesn't change much here - you have to run the entire pipeline `reduceOutputToSnapshotData` which you can see happens in `reduceAndFinalize` - the `winningForkId` is your next fork

src/rpc/services/SpectateService.ts
```diff
@@ -0,0 +1,531 @@
...
+    StateProofStruct
+} from "@typechain-types/contracts/V1/types/ProofTypes";
+import { DisputeStruct } from "@typechain-types/contracts/V1/types/DisputeTypes";
+import { Codec, Type } from "@/utils";
+import { isEqual } from "lodash";
+
+export interface SnapshotPayload {
+    latestForkGenesisSnapshot: StateSnapshot;
+    disputeWindows?: DisputeStruct[][];
+    stateProof?: StateProofStruct;
```
lukastanisic99, line 19: you also need milestoneSnapshots as used and required in the stateProof verification logic

src/rpc/services/SpectateService.ts
```diff
@@ -0,0 +1,531 @@
...
+            // Verify state proof to prove the latest state from the proven fork
+            if (
+                snapshotPayload.stateProof &&
+                snapshotPayload.stateProof.milestones.length > 0
+            ) {
+                console.log(`Verifying state proof from proven fork`);
+
+                // Get the latest state proof from the proven fork to compare with participant's state proof
+                const latestBlockHeight =
+                    stateManager.storage.blocks.getNextBlockHeight(
```
lukastanisic99: the verifyier/spectator doesn't have any blocks in storage - to verify the stateProof specificaly the milestones which we're interested in, only the you would reuse the solidity code already defined - see `updateStateSnapshotSameFork` in `StateSnapshotFacet` - the only difference is that `updateStateSnapshotSameFork` reads the `currentStateSnapshot` from storage and you'd pas in `genesisStateSnapshot`, so you'd slighly have to refactor the code to have a function that returns true/false and takes in an additional snapshot as an argument and use that one (that you can pass the genesisSnapshot) here and `updateStateSnapshotSameFork` uses that internally and passes the snapshot read from storage

src/rpc/services/SpectateService.ts
```diff
@@ -0,0 +1,531 @@
...
+                        error: "State proof does not match computed state proof"
+                    };
+                }
+
+                console.log(
+                    `Verified state proof matches computed state proof`
+                );
+
+                // Verify the encoded state by generating a state proof from it and comparing
+                if (snapshotPayload.encodedState) {
```
lukastanisic99, line 477: don't understand the comment above this line of code - to verify the encoded state you just hash it and compare it to the latestVerifiedSnpashot from stateProof after stateProof is verified - if stateProof is empty it should be the genesisSnapshot and than the snapshot has a field `snapshotData.stateMachineStateHash` to which you compare the hash

src/rpc/services/SpectateService.ts
```diff
@@ -0,0 +1,531 @@
...
+            ) {
+                console.log(`Verifying state proof from proven fork`);
+
+                // Get the latest state proof from the proven fork to compare with participant's state proof
+                const latestBlockHeight =
+                    stateManager.storage.blocks.getNextBlockHeight(
+                        currentForkId
+                    ) - 1;
+                const computedStateProof =
+                    await stateManager.agreementManager.getStateProof(
```
lukastanisic99: why is the verifier calling this? Their storage is empty - the verification should be exlusivley a function of (the data provided and the data available on-chain)

src/rpc/services/SpectateService.ts
cyony95: not going to happen anymore since we just send all of the disputes now, according to the suggestion above.

contracts/V1/StateChannelDiamondProxy/StateChannelManagerProxy.sol
```diff
@@ -363,6 +363,19 @@ contract StateChannelManagerProxy is StateChannelManagerInterface, StateChannelC
...
     }
 
+    function verifyMilestonesProxyView(
+        MilestoneProof[] memory milestoneProofs,
+        StateSnapshot[] memory milestoneSnapshots,
+        StateSnapshot memory genesisSnapshot
+    ) external view returns (bool) {
+        // This should trick the compiler, but still use delegatecall under the hood. This doesn't magically allow us to mutate the state in view functions that use delegate call.
+        // Ethers and other tools won't issue a transaction, but a call that runs on a single node so even if it did modify the state it wouldn't be reflected on-chain/persisted.
+        (bool isValid,) = DisputeVerificationFacet(address(this)).verifyMilestones(
```
lukastanisic99: `DisputeVerificationFacet.verifyMilestones` is a `pure` function, which means it does not read or modify storage/state - it's stateless (does all the computation on the provided input), so you don't have to call it with a `delegatecall` which makes your life simple 


So after merging/rebasing with the `dispute` branch - you can change `StateChannelManagerProxy.verifyMilestone` to `view` - (why not `pure`? - because it reads the address of the `DisputeVerificationFacet` from its own storage before calling it) - view is also fine since ethers returns an object not a tx response 

So after rebase/merge:

```solidity
function verifyMilestones(
        MilestoneProof[] memory milestoneProofs,
        StateSnapshot[] memory milestoneSnapshots,
        SnapshotData memory genesisSnapshotData
    ) public view returns (bool isValid, bytes memory lastBlockEncoded) {
        return DisputeVerificationFacet(disputeVerificationFacetAddress).verifyMilestones(milestoneProofs, milestoneSnapshots, genesisSnapshotData);
    }
```

contracts/V1/StateChannelDiamondProxy/DisputeVerificationFacet.sol
```diff
@@ -60,6 +60,61 @@ contract DisputeVerificationFacet is StateChannelCommon {
...
+        outputSnapshotData = SnapshotData({
+            originForkId: latestStateSnapshot.forkId,
+            stateMachineStateHash: keccak256(outputState.encodedModifiedState),
+            participants: getStatemachineParticipants(outputState.encodedModifiedState),
+            latestJoinChannelBlockHash: reducedOutput.latestJoinChannelBlockHash,
+            latestExitChannelBlockHash: keccak256(abi.encode(outputState.exitBlock)),
+            totalDeposits: outputState.totalDeposits,
+            totalWithdrawals: outputState.totalWithdrawals
+        });
+    }
```
lukastanisic99, line 116: Am I missing something or this fuction is a duplicate of `reduceOutputToSnapshotData` defined bellow at line 304? 
Plus this implementation doesn't account for the genesisState being the latest state as the implementation at 304 does - apart from that everything else is the same??

contracts/V1/StateChannelDiamondProxy/LocalDiamond.sol
```diff
@@ -200,6 +200,25 @@ contract LocalDiamond is StateChannelManagerProxy {
...
+        ReduceOutput memory reducedOutput,
+        StateSnapshot memory latestStateSnapshot,
+        bytes memory latestStateMachineState,
+        JoinChannelBlock[] memory joinChannelBlocks
+    ) public view returns (SnapshotData memory outputSnapshotData) {
+        bytes memory data = abi.encodeCall(
+            DisputeVerificationFacet.computeReducedOutputSnapshotData,
+            (reducedOutput, latestStateSnapshot, latestStateMachineState, joinChannelBlocks)
+        );
+        (bool success, bytes memory returnData) = address(this).staticcall(data);
```
lukastanisic99, line 213: this will revert when called since semantically what's called with the staticcall is state mutating which will cause a revert at runtime. The compiler doesn't know that since the compiler can not make the logical deduction what's behind the data even though it's defined above. 

You'll have to do this with a `delegatecall` (if state is actually required) and if its's logically stateless you can do it with a `call` - in either cases this can not bew a view tranaction. Further you can't do `address(this)` and think you'll call `DisputeVerificationFacet.computeReducedOutputSnapshotData` - address(this) corresponds to the address of the `LocalDiamond` so you'll endup calling this function recursivly in a loop until you run out of gas - this is not the same trick that I used to trick the compiler - for this you can't trick it and what I did is yet to be tested at runtime 

The correct way to do this is: `disputeVerificationFacetAddress.delegatecall/call(data)`

It doesn't matter since this is not a view transaction since locally you'll get the object directly from the EVM - the only difference is that you can't reuse `ethers` but have to expose the function yourself as you did

lukastanisic99 (commented):
I'll push a PR in a few minutes on top of this one

PR #153, opened 2025-09-05
Post state snapshot

cyony95:
closes https://trello.com/c/x9c32hDb/177-poststatesnapshot-call-both-adrians-call-as-a-single-mulyi-tx

### DIFF
--- a/src/stateManager/StateManager.ts
+++ b/src/stateManager/StateManager.ts
@@ -481,7 +481,7 @@ class StateManager {
             const txResponse =
                 await this.stateChannelManagerContract.multicall(callData);
             await txResponse.wait();
-            console.log("Successfully posted state snapshot using multicall");
+            console.log("Successfully posted state snapshot");
         } else {
             console.log("No state snapshot updates needed");
         }
--- a/src/stateManager/StateManager.ts
+++ b/src/stateManager/StateManager.ts
@@ -446,109 +446,58 @@ class StateManager {
         milestoneSnapshots: StateSnapshot[],
         exitChannelBlocks: ExitChannelBlockStruct[] = []
     ) {
-        // Get on-chain state
-        const onChainforkId = await this.stateChannelManagerContract.getforkId(
-            this.channelId
-        );
-        const onChainDisputeLength =
-            await this.stateChannelManagerContract.getDisputeLength(
+        // Get current on-chain state
+        const currentOnChainSnapshot = StateSnapshot.from(
+            await this.stateChannelManagerContract.getStateSnapshot(
                 this.channelId
-            );
-
-        if (onChainDisputeLength == onChainforkId) {
-            // Call contract without dispute
-            return this.stateChannelManagerContract.updateStateSnapshotWithoutDispute(
-                this.channelId,
-                milestoneProofs,
-                milestoneSnapshots,
-                exitChannelBlocks
-            );
-        }
-
-        // Need to include a dispute
-        const disputeData = this.agreementManager.forks.getLatestDispute();
-        if (!disputeData) {
-            throw new Error(
-                "No dispute data available but dispute length > fork count"
-            );
-        }
-
-        // Get output state snapshot data
-        const encodedDispute = Codec.encode(disputeData.dispute, Type.Dispute);
-        const commitment = ethers.keccak256(
-            ethers.AbiCoder.defaultAbiCoder().encode(
-                ["bytes", "uint256"],
-                [encodedDispute, disputeData.timestamp]
             )
         );
 
-        const outputStateSnapshot =
-            this.outputStateSnapshotData.get(commitment);
-        if (!outputStateSnapshot) {
-            throw new Error("No output state snapshot data available");
+        // Get the target fork ID from the latest milestone snapshot
+        const targetForkId =
+            milestoneSnapshots[milestoneSnapshots.length - 1].forkId;
+
+        // Update to the correct fork if needed
+        if (currentOnChainSnapshot.forkId !== targetForkId) {
+            console.log(
+                `Fork mismatch: on-chain=${currentOnChainSnapshot.forkId}, target=${targetForkId}`
+            );
+            console.log("Updating to correct fork first...");
+
+            // Call updateSnapshotFork to transition to the correct fork
+            await this.updateSnapshotFork();
+
+            // Get the updated on-chain snapshot after fork transition
+            const updatedOnChainSnapshot = StateSnapshot.from(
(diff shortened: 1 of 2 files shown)


src/stateManager/StateManager.ts
```diff
@@ -696,6 +662,10 @@ class StateManager {
                     currentForkId
                 );
 
+            if (!isDisputed) {
```
cyony95, line 704: @lukastanisic99 @daiagi I realized I missed this opportunity for an early return, not really related to the current ticket. Let me know if it's fine.

src/stateManager/StateManager.ts
```diff
@@ -441,120 +441,63 @@ class StateManager {
...
-                exitChannelBlocks
-            );
+        // Execute the final snapshot updates in a single multicall transaction
+        if (callData.length > 0) {
+            const txResponse =
+                await this.stateChannelManagerContract.multicall(callData);
+            await txResponse.wait();
+            console.log("Successfully posted state snapshot");
+        } else {
+            console.log("No state snapshot updates needed");
```
lukastanisic99, line 547: The multicall bubbles up the error so you should catch it here and interpret it to know what to do

src/stateManager/StateManager.ts
```diff
@@ -656,29 +599,52 @@ class StateManager {
...
+                error
+            );
+            throw error;
+        }
+    }
+
+    /**
+     * Updates the state snapshot when the fork is the same
+     */
+    public async updateSnapshotSameFork(forkId: ForkId): Promise<void> {
```
lukastanisic99: is this used? Probably the single function call that takes into account to both update the fork and the state in the fork is enough - you'd need to inerpret the custom error here also

daiagi:
some comments:
- `prepareUpdateStateSnapshotFork` is very long function (~210 lines), can you break it up?
Example idea of possible sub-functions:
 `findUndisputedFork(startingForkId: Hash): Promise<Hash>`
 `processDisputesAndReduce(forkId: Hash): Promise<Hash>`
`buildDisputesFromCommitments(commitments: Hash[]): DisputeStruct[]`
possibly more...

Same goes for `prepareUpdateSnapshotSameFork`

-  `postStateSnapshot` - if the new stateSnsaphot is on the same fork as the on-chain one
a. no need to call prepareUpdateStateSnapshotFork
b. direct call rather the multi call is cleaner and clearer

src/stateManager/StateManager.ts
```diff
@@ -441,120 +441,63 @@ class StateManager {
...
-                this.channelId,
-                milestoneProofs,
-                milestoneSnapshots,
-                disputeProof,
-                exitChannelBlocks
-            );
+    public async postStateSnapshot(forkId: ForkId): Promise<void> {
+        // Prepare data for the final two snapshot update calls
+        const forkData = await this.prepareUpdateStateSnapshotFork();
+        const sameForkData = await this.prepareUpdateSnapshotSameFork(forkId);
```
lukastanisic99, line 488: you have no gurantee that `prepareUpdateStateSnapshotFork` will update to `forkId` even if forkId is correct - maybe it stops half way - e.g. `A->B->C` it only updates to `B`, so you can't chain `C,height=X`

You have to extract the forkId to which `prepareUpdateStateSnapshotFork` can update through `forkData` and if there's a missmatch throw - I'll add a trello card for better checks and fallback logic if there's a missmatch

src/stateManager/StateManager.ts
cyony95: @lukastanisic99  it's not used anywhere else, I left it so we can just call it if we need it, but if we don't, I'll just remove it and updateSnapshotFork as well

cyony95: removed

cyony95:
@daiagi I implemented your check to call just updateSnapshotSameFork if the fork is the same. I haven't broken up the functions yet as I wanted to keep the PR clear and concise. I'll break them up in a separate PR. Thanks for the hints.

src/stateManager/StateManager.ts
```diff
@@ -441,120 +441,116 @@ class StateManager {
...
+                    ]
+                );
+            callData.push(forkCalldata);
         }
-
-        // Check if we have threshold signatures on the dispute
-        const fork = this.agreementManager.forks.latestFork();
-        if (!fork) {
-            throw new Error("No latest fork found");
+        if (sameForkData) {
```
lukastanisic99, line 512: This still doesn't check if `forkData` will update the fork to `sameForkData`

src/stateManager/StateManager.ts
```diff
@@ -441,120 +441,116 @@ class StateManager {
...
+        if (callData.length > 0) {
+            try {
+                const txResponse =
+                    await this.stateChannelManagerContract.multicall(callData);
+                await txResponse.wait();
+                console.log("Successfully posted state snapshot");
+            } catch (error) {
+                if (isCustomEvmError(error)) {
+                    console.error(
+                        "Error posting state snapshot:",
```
lukastanisic99, line 538: you have to interpret the error to see if you have to do something - that's why we've crated these custom errors

cyony95: Possible Errors from updateStateSnapshotFork:
- ErrorStateSnapshotNotValid
- ErrorInvalidStateSnapshot
- ErrorFirstExitChannelBlockInvalid
- ErrorExitChannelBlocksNotLinked
- ErrorLastSnapshotInvalid
- ErrorLastSnapshotDoesNotMatchGenesis

Possible Errors from updateStateSnapshotSameFork:

- ErrorSnapshotsNotProvided
- ErrorSnapshotForkMismatch - this one shouldn't even happen since I added a check before calling the function
- ErrorBlockHeightTooOld
- ErrorInvalidStateProof
- ErrorWithdrawalFailed:
- CantWithdrawMoreThanDeposits

I am interpreting the errors, that s why I'm using isCustomEvmError. But for which of these can I take an action? Except ErrorDisputeAlreadyReduced, which I already handle in the updateStateSnapshotFork by itself, I'm not sure if there's an action to take for any of these errors here.
```
                } catch (error) {
                    if (
                        isCustomEvmError(error) &&
                        error.errorDescription.name !==
                            "ErrorDisputeAlreadyReduced"
                    ) {
                        throw error; // Re-throw other errors
                    }
```

src/stateManager/StateManager.ts
cyony95: added check for the resulting forkid to be compared with the argument forkid

lukastanisic99 (approved):

PR #154, opened 2025-09-08
post block confirmation actions

daiagi:
## 💪 Related Issues

- Closes https://trello.com/c/zQWUSwXc
- Closes https://trello.com/c/l8RxDJ8O

@lukastanisic99 
regarding #152 , `setState` and `playTransaction` should do what? call `success()`?

PR #152, opened 2025-09-05
remove onSuccessCommon()

cyony95:
closes https://trello.com/c/YbPfLemN/185-onsuccesscommon-kill-it-this-logic-should-go-into-success-function

PR #157, opened 2025-09-08
use `success()`

daiagi:
## 📋 Description

Correct `success` actions:

_"regarding #152 , setState and playTransaction should do what? call success()?_

_playTransaction - yes
setState - called after a dispute to set the genesis of a new fork - a block is not produced so only a subset of success() can be called - onTurn() startTryTimeoutNext() tryDequeueNext - everything else assumes a block is produced"_

## 💪 Related Issues

https://github.com/peer3to/state-channels-plus/pull/154#pullrequestreview-3196983739

### DIFF
--- a/src/stateManager/StateManager.ts
+++ b/src/stateManager/StateManager.ts
@@ -236,9 +236,6 @@ class StateManager {
         const nextToWrite = await this.diamondStateMachine.getNextToWrite();
 
         this.p2pEventHooks.onTurn?.(nextToWrite);
-
-        scheduleTask(this.tryExecuteFromQueue, 0, "queueProcessing");
-
         const nextTransactionCnt =
             this.storage.blocks.getNextBlockHeight(_forkId);
         scheduleTask(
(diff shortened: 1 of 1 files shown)


src/stateManager/StateManager.ts
```diff
@@ -229,9 +230,27 @@ class StateManager {
...
+        scheduleTask(
+            () =>
+                this.tryTimeoutParticipant(
+                    _forkId,
+                    nextTransactionCnt,
+                    nextToWrite
+                ),
+            this.getTimeoutWaitTimeSeconds() * 1000,
+            "participantTimeout"
+        );
```
lukastanisic99:
```suggestion
        const nextTransactionCnt =
            this.storage.blocks.getNextBlockHeight(_forkId);
        scheduleTask(
            () =>
                this.tryTimeoutParticipant(
                    _forkId,
                    nextTransactionCnt,
                    nextToWrite
                ),
            this.getTimeoutWaitTimeSeconds() * 1000,
            "participantTimeout"
        );
        
         scheduleTask(this.tryExecuteFromQueue, 0, "queueProcessing");
```

lukastanisic99: just putted dequeue at the bottom so even if storage becomes async later it preserves the order of execution

lukastanisic99 (approved):

PR #159, opened 2025-09-09
Diagrams - Event Handlers

lukastanisic99:
Most event handlers are covered and explained. Only JoinChannel one is left which I'll do later when spectating is done. Take a look, think and ask questions


lukastanisic99:
Merging so I can look at it in working branches - in general I'll auto merge diagrams going forward

PR #160, opened 2025-09-10
InvalidTimestamp fraud proof

daiagi:
- Closes https://trello.com/c/TfL7XEP8

### DIFF
--- a/contracts/V1/StateChannelDiamondProxy/FraudProofFacet.sol
+++ b/contracts/V1/StateChannelDiamondProxy/FraudProofFacet.sol
@@ -219,7 +219,7 @@ contract FraudProofFacet is StateChannelCommon {
             // Validate previous block hash matches previous block
             Block memory previousBlock = abi.decode(invalidTimestampProof.previousBlock.encodedBlock, (Block));
             require(
-                invalidBlock.previousBlockHash == keccak256(abi.encode(previousBlock)), ErrorInvalidPreviousBlockHash()
+                invalidBlock.previousBlockHash == keccak256(invalidTimestampProof.previousBlock.encodedBlock), ErrorInvalidPreviousBlockHash()
             );
 
             // Check if calldata is posted - previous block
--- a/contracts/V1/StateChannelDiamondProxy/FraudProofFacet.sol
+++ b/contracts/V1/StateChannelDiamondProxy/FraudProofFacet.sol
@@ -7,6 +7,22 @@ import "./Errors.sol";
 import "../types/FraudProofTypes.sol";
 
 contract FraudProofFacet is StateChannelCommon {
+    mapping(
+        FraudProofType
+            => function(bytes memory encodedFraudProof, FraudProofVerificationContext memory fraudProofVerificationContext) internal returns (address)
+    ) private proofHandlers;
+
+    constructor() {
+        //If we endup having too many fraud proofs, we'll refactor them into a seperate 'facet' (ERC-2535)
+        proofHandlers[FraudProofType.BlockDoubleSign] = _handleBlockDoubleSign;
+        proofHandlers[FraudProofType.BlockEmptyBlock] = _handleBlockEmptyBlock;
+        proofHandlers[FraudProofType.BlockInvalidStateTransition] = _handleBlockInvalidStateTransition;
+        proofHandlers[FraudProofType.InvalidTimestamp] = _handleInvalidTimestamp;
+        // proofHandlers[FraudProofType.TimeoutThreshold] = _handleTimeoutThreshold;
+        // proofHandlers[FraudProofType.TimeoutPriorInvalid] = _handleTimeoutPriorInvalid;
+        // proofHandlers[FraudProofType.DisputeInvalidPreviousRecursive] = _handleDisputeInvalidPreviousRecursive;
+    }
+
     //This is a bit inefficient, since public/external functions always do a deep copy unlike internal/private that pas by reference, but this shares the context
     function applyFraudProofs(
         FraudProof[] memory fraudProofs,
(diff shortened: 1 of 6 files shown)


contracts/V1/StateChannelDiamondProxy/FraudProofFacet.sol
```diff
@@ -172,4 +173,108 @@ contract FraudProofFacet is StateChannelCommon {
...
+            require(
+                invalidBlock.previousBlockHash == keccak256(abi.encode(previousStateSnapshot)),
+                ErrorInvalidPreviousSnapshotHash()
+            );
+            previousTimestamp = previousStateSnapshot.timestamp;
+        } else {
+            // Validate previous block hash matches previous block
+            Block memory previousBlock = abi.decode(invalidTimestampProof.previousBlock.encodedBlock, (Block));
+            require(
+                invalidBlock.previousBlockHash == keccak256(abi.encode(previousBlock)), ErrorInvalidPreviousBlockHash()
```
lukastanisic99:
```suggestion
                invalidBlock.previousBlockHash == keccak256(invalidTimestampProof.previousBlock.encodedBlock), ErrorInvalidPreviousBlockHash()
```

contracts/V1/StateChannelDiamondProxy/FraudProofFacet.sol
```diff
@@ -172,4 +173,108 @@ contract FraudProofFacet is StateChannelCommon {
 
         return signer;
     }
+
+    function _handleInvalidTimestamp(
+        bytes memory encodedProof,
+        FraudProofVerificationContext memory fraudProofVerificationContext
+    ) internal view returns (address) {
+        InvalidTimestampProof memory invalidTimestampProof = abi.decode(encodedProof, (InvalidTimestampProof));
+        Block memory invalidBlock = abi.decode(invalidTimestampProof.invalidBlock.encodedBlock, (Block));
```
lukastanisic99, line 198: We'd have to authenticate the siganture of the `invalidBlock` so we know it's authored by the correct author that we're trying to slash - I'll add this to the pseudo code in the diagrams

contracts/V1/StateChannelDiamondProxy/FraudProofFacet.sol
```diff
@@ -172,4 +173,108 @@ contract FraudProofFacet is StateChannelCommon {
...
+        (bool currentBlockFound, bytes32 currentBlockCommitment) = getBlockCallDataCommitment(
+            channelId, forkId, invalidBlock.transaction.header.transactionCnt, invalidBlockAuthor
+        );
+
+        uint256 currentBlockTimestamp;
+        if (currentBlockFound) {
+            bytes32 expectedCurrentCommitment = keccak256(
+                abi.encode(invalidTimestampProof.invalidBlock, invalidTimestampProof.invalidBlockOnChainTimestamp)
+            );
+            require(currentBlockCommitment == expectedCurrentCommitment, ErrorInvalidCurrentBlockCalldataCommitment());
```
lukastanisic99, line 272: I've been thinking about this line of code for some time and my thoughts have changed a few times (even as I was writing the comment bellow), so I'll just comment on the thought process:

`Short:`
I need to restructure things - I didn't account for race condition as I did in disputes, so `revert/require` have to revert the tx, not slash the disputer - returing `address(0)` will mean that the tx was successfull, but the fraud proof was invalid which will have the consequence of slashing the disputer - reverts will be bubbled up, so we can catch the error message and interpret the error message and maybe try again. 

`Long:`
I like this one - in the pseudo code I wrote this is an `if` to deduct the timestamp, while thinking why you used `require` I realized the logic makes sense, but not to punish the `disputer` if the require doesn't hold, but to punish `invalidBlockAuthor` since they commited to something else (junk or another block different than invalidBlock), but also produced `invalidBlock`, but I also realized I made a mistake unlike for the `previousBlock` I didn't include `invalidBlockCalldata` to validate the input data against the commitment so this part should be 

```typescript
if(currentBlockCommitment){
    require(currentBlockCommitment = hash(currentBlockCalldata, invalidBlockOnChainTimestamp)) // if this doesn't hold punish the disputer since they didn't supply correct data
    if(currentBlockCalldata != invalidBlock)return invalidAuthorAddress // the author commited to some junk or another block, but still produced and signed invalidBlock
    currentBlockTimestamp = invalidBlockOnChainTimestamp
}
```

In the version I did in the diagrams (with the if, not require) if the disputer 'lied' about the invalidBlockOnChainTimestamp, the if would use `invalidBlock.timestamp` which is less than `invalidOnChainTimestamp`, which could result in the last check `if (currentBlockTimestamp > previousTimestamp + getP2pTime() + getAgreementTime() + getChainFallbackTime())` being evaluated with a smaller `currentBlockTimestamp` which would be less beneficial to the disputer, that wants to prove that the `currentBlockTimestamp` is over the equation on the right - this on its own is not the problem, since you can assume that the disputer wouldn't put themselves in a less favorable position.

The actual problem is race condition. 

I realized that both commitments (`previousBlockCommitment` and `currentBlockCommitment`) are susceptible to race condtion on-chain even thoug the acceptable window for submitting calldata for those commitments should be expired when this code runs, but this code on its own does NOT know that and can NOT deduct it, if data is missing (calldata not uploaded), which would be the case since from the perspective of the disputer, when they were creating this transaction, commitments wouldn't be there -> not uploaded. 

So like with timeout in disputes, that have race condition guards to revert/abort the transaction this has to too have such guards. 


What we can do - merge this PR and create a trello card and I'll modify what's needed - I'll just finish my review bellow to see if anything else is missing

lukastanisic99: The underlying problem is that this fraud proof is `statefull` (requires reading on-chain state) and it's probably the only `fraud proof` that's not stateless - the stateless ones are non susceptible to race condition. Luckily `dispute fraud proofs` are all statefull and I had that in mind when designing them, so this pipeline can change to that - not a big change on solidity - I'd argue that the solidity code even becomes simpler since now fraud proofs would be decoupled from disputes and run in isolation and disputes would just rely on what's in `onChainSlashes`, which both `fraud proofs` and `dispute fraud proofs` append to, but it becomes a bit `annoying on typescript`. Instead of detecting a fraud proof and creating a dispute and appending that fraud proof to the dispute and knowing that the fraud proof is always true, now it would have to `run on-chain` and for good UX it would have to run as a `multicall`, where you first run the fraud proof on-chain to append to `onChainSlashes` and chain the dispute on success, but it doesn't stop there - every peer would try and invoke this fraud proof on-chain when they detect it and only the first run would actually append to onChainSlashes, so we'd have to make the others `gracefully fail without reverting`, so they can chain the dispute they want to chain on success (not a problem, just don't throw and return if the disputed peer is already in on-chain slashes) - further the fraud proof can fail due to race condition, where we'd have to `bubble up the error` and interpret it in typescript to understand should we try again with different argument (to account for the race condition that happened) or what the fuck should we due.

I'll do a blockchain like PR on my recent PR to introduce these changes and we'll come back to this PR later

This `custom error handing` is getting anoying since apparently we have it everywhere - maybe have it in a single place as an universal pipeline (e.g. when this error is thrown always do this action and than we proxy that and from the calling context were we're interaction with the on-chain contract we just return the end result of the call instead of interpreting errors there) - What are your thoughts? Think about this

daiagi:
Closing

PR #161, opened 2025-09-10
remove BlockEntry from storage

daiagi:
- Closes https://trello.com/c/HBQTxyfI

### DIFF
--- a/src/stateManager/ValidationService.ts
+++ b/src/stateManager/ValidationService.ts
@@ -123,14 +123,11 @@ export default class ValidationService {
             return genesisSnapshot?.hash === block.previousBlockHash;
         }
 
-        const prevBlockEntry = this.storage.blocks.getBlockEntry(
-            forkId,
-            height - 1
-        );
-        if (!prevBlockEntry) {
+        const prevBlock = this.storage.blocks.getBlock(forkId, height - 1);
+        if (!prevBlock) {
             return false;
         }
-        return prevBlockEntry.block.hash === block.previousBlockHash;
+        return prevBlock.hash === block.previousBlockHash;
     }
 
     authenticateBlock(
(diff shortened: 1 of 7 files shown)


cyony95 (approved):

src/storage/BlockStorage.ts
```diff
@@ -79,55 +70,9 @@ export class BlockStorage {
...
-    ): boolean;
-
-    /*────────────────────────────────────────────────────────────────────────────
-      IMPLEMENTATION
-    ────────────────────────────────────────────────────────────────────────────*/
-    setOnChainTimestamp(
-        hashOrForkId: Hash | ForkId,
-        timestampOrHeight: Timestamp | BlockHeight,
-        timestamp?: Timestamp
-    ): boolean {
```
lukastanisic99, line 102: How is this accounted for now? Updating by reference outside of storage doesn't persist the changes here - there's a `deepCopy`

test/storage/BlockStorage.test.ts
```diff
@@ -242,83 +237,26 @@ describe("BlockStorage", () => {
...
 
             expect(
-                storage.getBlockEntry(mockBlockHash)?.block
-                    .confirmationSignatures.size
+                storage.getBlock(mockBlockHash)?.confirmationSignatures.size
             ).to.equal(expectedNumSignatures);
         });
     });
 
-    describe("UPDATE - setOnChainTimestamp()", () => {
```
lukastanisic99, line 251: Why is this not relevant anymore

daiagi: ha?
this is the point of the trello card / this PR
it is put on the Block model itself

src/stateManager/ValidationService.ts
```diff
@@ -328,11 +321,6 @@ export default class ValidationService {
 
             // True - Update the previous block with the on-chain timestamp
             previousBlock.onChainTimestamp = previousBlockOnChainTimestamp;
```
daiagi, line 323: @lukastanisic99 
right here

before: duplicate accounting, on the block and in storage
after: only on storage

lukastanisic99: I'm on my phone now will check when I get home

To me it seems that the accounting is done only on the block model and it will be persisted in storage only if it's writing the block to storage the first time - if it's an update to an already existing block it won't be accounted for 

It's not much of a difference in logic than signature expansion - a block can exist in storage and signatures are added later

lukastanisic99 (approved):

PR #163, opened 2025-09-10
disputed refactor - seperate fraud proof from disputes

lukastanisic99:
Mostly all the logic to create a dispute. Some solidity code is refactored to assist in re-using it locally


contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol
```diff
@@ -151,18 +154,24 @@ contract DisputeFraudProofFacet is StateChannelCommon {
...
 
-        // Check is block newer
-        if (_getBlockHeight(newerBlock) <= _getBlockHeight(latestBlock)) revert();
+            // Check latestBlock and newerBlock same forkId
+            if (!_areBlocksSameFork(newerBlock, latestBlock)) revert();
+
+            // Check is block newer
+            if (_getBlockHeight(newerBlock) <= _getBlockHeight(latestBlock)) revert();
+        }
+        // if !hasBlock -> latestState should be genesis state -> if the disputer signed any block this proof is valid
```
lukastanisic99: Changes in this file are mostly `_getLatestBlock(stateProof)` now accounting for stateProof being empty (can create a dispute with the `latestState == genesisState`)

contracts/V1/StateChannelDiamondProxy/DisputeVerificationFacet.sol
```diff
@@ -18,26 +18,53 @@ contract DisputeVerificationFacet is StateChannelCommon {
...
+        bytes32 channelId,
+        FraudProof[] memory fraudProofs,
+        bool selfRemoval,
+        address[] memory onChainSlashes,
+        address disputer,
+        Timeout memory timeout,
+        StateSnapshot memory latestStateSnapshot,
+        bytes memory latestStateMachineState,
+        bytes32 latestJoinChannelBlockHash
+    ) public returns (SnapshotData memory outputSnapshotData, address[] memory slashes) {
```
lukastanisic99, line 42: This function just extract the pipeline from auditing in a single function so it's easy to re-use in typescript - I was lazy to refactor `DisputeStruct` into `disputeInput` and `disputeOutput` - most function run on disputeInput and you use that to calculate disputeOutput, but currently everything is jammed in - the problem is when you're creating a `disputeStruct` you don't know the `output` - you need to generate it so to reuse these functions they can't expect a disputeStruct so I just extracted what I needed in a bit ugly way

lukastanisic99: I've added a trello card to refactor this

contracts/V1/StateChannelDiamondProxy/DisputeVerificationFacet.sol
```diff
@@ -272,16 +297,24 @@ contract DisputeVerificationFacet is StateChannelCommon {
...
     ) public returns (SnapshotData memory outputSnapshotData) {
         //verify snapshot linked to reducedOutput.latestBlock
         Block memory latestBlock = reducedOutput.latestBlock;
-        require(
-            latestBlock.stateSnapshotHash == keccak256(abi.encode(latestStateSnapshot)), ErrorInvalidStateSnapshot()
-        );
+        if (latestBlock.transaction.header.forkId == bytes32(0)) {
+            //no blocks in reducedOutput - must be genesis
+            // TODO - think - this ensures that snapshotData is correct, not the snapshot, which means someone can lie about the time, but it shouldn't matter here - we just care to perform the correct State Transition
+            require(keccak256(abi.encode(latestStateSnapshot.snapshotData)) == forkId, ErrorInvalidStateSnapshot());
```
lukastanisic99, line 261: this is the case where `StateProof` is empty (doesn't have any blocks) - need to see what's the best way to approach this

contracts/V1/StateChannelDiamondProxy/DisputeVerificationFacet.sol
cyony95: that should be more clean but then there wouldn't be a link between disputeInput and disputeOutput. How do we know which output belongs to which input? we must have a way to link them, I think

lukastanisic99: I missed this comment :D - The link is the `generateDisputeOutputState` function - logically everything stays the same. The current `DisputeStruct` has all `input data` except `outputSnapshotDataHash` - this is the only `output` data of the `DisputeStruct`. The change that should happen is a structural change so instead of passing 10 variables 1 by 1 since we can't always pass the entire dispute, since we may not know the output data (we want to reuse this code to generate the outputData), we structure everything in a way where DisputeStruct has a field called `input` and all the input fields go there, so these functions take an argument `DisputeInput` instead of 10 separate variables

src/disputeManager/DisputeManager.ts
```diff
@@ -0,0 +1,311 @@
...
+        // this can be a subset of on-chain slashes, so we don't need to run any race condition checks
+        let onChainSlashes = new Set<Address>(
+            await this.diamondStateMachine.localDiamondContract.getOnChainSlashedParticipants(
+                this.channelId
+            )
+        );
+        let participants = new Set<Address>(
+            await this.diamondStateMachine.getParticipants()
+        );
+
```
daiagi: why different interface?
one is `diamondStateMachine.localDiamondContract.<function>` and the other `diamondStateMachine.<function>`


I would expect both to be `diamondStateMachine.<function>`


is there a reason or just quick code?

src/disputeManager/DisputeManager.ts
```diff
@@ -0,0 +1,311 @@
...
+        // to make sure we're trying to slash only participants - even though onChainSlashes should always be a subset of participants
+        onChainSlashes = intersection(onChainSlashes, participants);
+
+        // fraudProofs
+        let fraudProofs: FraudProofStruct[] = [];
```
daiagi: FYI:
_"Many style guides (including [MDN's](https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Code_style_guide/JavaScript#variable_declarations)) recommend using const over [let](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let) whenever a variable is not reassigned in its scope. This makes the intent clear that a variable's type (or value, in the case of a primitive) can never change. Others may prefer let for non-primitives that are mutated."_


#166

src/storage/JoinChannelBlockStorage.ts
```diff
@@ -66,4 +66,32 @@ export class JoinChannelBlockStorage {
...
+        fromBlockHash: Hash,
+        toBlockHash?: Hash
+    ): Generator<JoinChannelBlockEntry, void, unknown> {
+        if (fromBlockHash == ethers.ZeroHash) return;
+        let currentHash = fromBlockHash;
+        while (
+            currentHash != ethers.ZeroHash &&
+            (!toBlockHash || currentHash != toBlockHash)
+        ) {
+            const entry = this.blockMap.get(currentHash);
```
daiagi:
```suggestion
        while (currentHash != ethers.ZeroHash) {
            if (toBlockHash && currentHash === toBlockHash) break;
            const entry = this.blockMap.get(currentHash);
```

src/storage/JoinChannelBlockStorage.ts
```diff
@@ -66,4 +66,32 @@ export class JoinChannelBlockStorage {
...
+            if (!entry) return;
+            yield entry;
+            currentHash = entry.block.previousBlockHash;
+        }
+    }
+
+    getBlocksInRange(
```
daiagi, line 85: kind of not-needed
you could instead use Array.from in `DisputeManager`

```ts
const exitChannelBlocks = Array.from(
    this.storage.exitChannelBlocks.getIterator(fromBlockHash, toBlockHash)
).map(entry => entry.block);
```

src/storage/ExitChannelBlockStorage.ts
```diff
@@ -66,4 +66,32 @@ export class ExitChannelBlockStorage {
     ): ExitChannelBlockEntry | undefined {
         return this.blockMap.get(blockHash);
     }
+
```
daiagi, line 69: same comment as on `JoinChannelBlockStorage`

src/disputeManager/DisputeManager.ts
```diff
@@ -0,0 +1,311 @@
...
+
+    public async createDispute(
+        forkId: ForkId,
+        selfRemoval: boolean = false,
+        timeoutOptions?: TimeoutOptions
+    ): Promise<void> {
+        const latestBlockHeight =
+            this.storage.blocks.getNextBlockHeight(forkId) - 1;
+        // StateProof
+        const stateProof = await this.agreementManager.getStateProof(
```
daiagi: performance:
consider parallelizing

```ts
    const [stateProof, latestStateSnapshot, latestStateMachineState, onChainSlashes, participants] = await Promise.all([
        this.agreementManager.getStateProof(forkId, latestBlockHeight),
        this.storage.getStateSnapshot({ forkId, height: latestBlockHeight }),
        this.diamondStateMachine.getState(),
        this.diamondStateMachine.localDiamondContract.getOnChainSlashedParticipants(this.channelId),
        this.diamondStateMachine.getParticipants()
    ]);

    // valitate, sanity checks
```

daiagi (approved):

daiagi:
great
now we can review the ACTUAL changes

@lukastanisic99  @cyony95 

Edit:

merging, @cyony95  if you have any objections please open trello card for them.
let's move on

PR #164, opened 2025-09-13
Fraud proofs are stateful and decoupled from disputes - 1

lukastanisic99:
My series of blockchain like PRs to introduce digestible and isolated changes - any merge conflict with the `dispute branch` will be handled in the last PR in the chain, up to that point all changes are relative to the previous PR in the chain 

This PR removes/decouples fraud proofs from disputes - they run in isolation and directly append to `onChainSlashes` - disputes may slash only participant present in onChainSlashes and don't know anymore what fraud proofs are nor can they run them


contracts/V1/StateChannelDiamondProxy/FraudProofFacet.sol
```diff
@@ -23,23 +23,25 @@ contract FraudProofFacet is StateChannelCommon {
...
-                revert ErrorInvalidFraudProof();
+            if (!isParticipantSlashedOnChain(fraudProofVerificationContext.channelId, proofs[i].participant)) {
+                address slashedParticipant =
+                    proofHandlers[proofs[i].proofType](proofs[i].encodedProof, fraudProofVerificationContext);
+                if (slashedParticipant == address(0) || slashedParticipant != proofs[i].participant) {
+                    // slash the disputer
+                    slashedParticipant = msg.sender;
+                }
+                // if in (participants || pendingParticipants) && !slashedOnChain
+                if (_canParticipateInDisputes(fraudProofVerificationContext.channelId, slashedParticipant)) {
```
daiagi, line 25: Why does the slashed participant need to particiapte ins diapute?
Shouldn't we confirm that the reporter (the fraud submitter), if the proof is valid, should be able to particiapte in disputes?

```
if (valid proof) {
    if (canParticipateInDispute(msg.sender) && slashedParticipant != msg.sender) {
        addOnChainSlashedParticipant
    }
}
```

PR #168, opened 2025-09-16
Auditing removal initial - 3

lukastanisic99:
Mostly all the logic to substitute `auditing` - the dispute confirmation pipeline (the auditing substitution part) - for the full pipeline what's in diagrams under events needs to be implemented prior. 

Dispute Fraud Proofs are yet to be implemented, but they're all explained and left in the code and all except 1 are simple - the 1 I need to think how to reuse the `blockConfirmation pipeline` since it needs to cover the case when `new/unfinalized blocks` can commit to the `latestState` to which the dispute is applied - since it's not finalized, it may have not went through the blockConfirmation pipeline and we `CAN NOT just plug it` since if it's invalid we need a `dispute fraud proof` (not a normal fraud proof) to actually kill the invalid dispute. 

Other stuff in the PR are mostly cleanup and some suggestions from older PRs - the linter stuff wasn't merged here but in some future PR in this chain it will be so it will cascade eventually


src/stateManager/StateManager.ts
```diff
@@ -507,109 +512,7 @@ class StateManager {
...
-                disputeProof,
-                exitChannelBlocks
-            );
-        }
-
-        // Dispute is not finalized
-        console.log(
-            "Dispute is not finalized, state snapshot was not submitted"
-        );
+        throw new Error("TODO - Not implemented");
```
daiagi: why you deleted this?

src/stateManager/utils/DisputeFraudProofService.ts
```diff
@@ -0,0 +1,10 @@
+import Storage from "@/storage";
+
+// ────────────────────── FRAUD PROOF SERVICE ─────────────────────
+
+/**
+ * Service class for handling fraud proof creation and validation
+ */
+export default class DisputeFraudProofService {
```
daiagi, line 8: why not use FraudProofService?

daiagi:
see PR
* #169

PR #170, opened 2025-09-18
reduce hard code

daiagi:


lukastanisic99:
I'll leave this one also hanging for now since I need to think about it - some stuff is still hardcoded and we have to remember to change that in scripts - Getting everything implemented is of higher priority at the moment and don't want to think too much about this until then - may even get good ideas subconsciously in the meantime

daiagi:
Closing

PR #171, opened 2025-09-20
Dispute fraud proofs - 4

lukastanisic99:
Most dispute fraud proofs are implemented - missing 2 that use at least some parts of the `BlockConfirmation` pipeline - so I'll think what's the best way to reuse it especially since the proofs need to run in solidity - maybe I'll rewrite parts of the `BlockConfirmation` pipeline in solidity so it can be reused in `typescript` and can `run on-chain` when needed - in those cases the fraud proofs are simple they just call some `boolean function` as demonstrated with this PR for most dispute fraud proofs

Apart from dispute fraud proofs this PR also removed explicit `genesisTimestamp` from the `disputeWindow` and adds `dynamic killTime` - mostly needed a simple way to lookup the timestamp which doesn't rely on `reduce` and logically the timestamp shouldn't rely on it, just when things are posted on-chain.


lukastanisic99:
I'll just continue from here but in general my PRs don't have to be blockchain like [1 <- 2 <- 3...] so as thigs get merged I'll just PR to the latest thing available (disputes or some of my PRs that are left in review)

daiagi:
> I'll just continue from here but in general my PRs don't have to be blockchain like [1 <- 2 <- 3...] so as thigs get merged I'll just PR to the latest thing available (disputes or some of my PRs that are left in review)

let's merge them all
i prefer to be working on the latest version rather then waiting for change

PR #172, opened 2025-09-20
Events

daiagi:
start of Event handling following the diagrams.

### DIFF
--- a/contracts/V1/StateChannelDiamondProxy/LocalDiamond.sol
+++ b/contracts/V1/StateChannelDiamondProxy/LocalDiamond.sol
@@ -291,32 +291,4 @@ contract LocalDiamond is StateChannelManagerProxy {
             dispute, disputeAuditingData, auditingDataIntegrityVerified
         );
     }
-
-    function isDisputeWindowRelevant(bytes32 channelId, bytes32 disputedForkId) external view returns (bool) {
-        StateSnapshot storage currentStateSnapshot = stateSnapshots[channelId];
-        DisputeData storage disputeData = disputeData[channelId];
-
-        bytes32 currentForkId = currentStateSnapshot.forkId;
-
-        if (currentForkId == disputedForkId) {
-            return true;
-        }
-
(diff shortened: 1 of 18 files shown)


contracts/V1/StateChannelDiamondProxy/LocalDiamond.sol
```diff
@@ -287,4 +291,32 @@ contract LocalDiamond is StateChannelManagerProxy {
...
+        ) {
+            currentForkId = disputeWindow.reducedResult.forkId;
+
+            if (currentForkId == disputedForkId) {
+                return true;
+            }
+
+            disputeWindow = disputeData.disputeWindowMap[currentForkId];
+        }
+
```
lukastanisic99: A dispute is relevant if it can change our fork choice by altering a disputeWindow so it reduces to something else - already reduced and finalized dispute windows can't change - so we have to find the ones along the path that are not finalized and observe them

daiagi: removed

contracts/V1/StateChannelManagerEvents.sol
```diff
@@ -4,36 +4,40 @@ import "./types/DisputeTypes.sol";
...
-        bytes32 indexed channelId,
-        bytes32 indexed commitmentHash,
-        address sender,
-        SignedBlock signedBlock,
-        uint256 timestamp
-    );
-    event SetState(bytes32 indexed channelId, bytes encodedState, bytes32 forkId, uint256 timestamp);
+    event StateSnapshotUpdated(bytes32 indexed channelId, StateSnapshot stateSnapshot, uint256 timestamp);
+
+    event BlockCalldataPosted(bytes32 indexed channelId, address sender, SignedBlock signedBlock, uint256 timestamp);
```
lukastanisic99: why remove `bytes32 indexed commitmentHash` you need it https://github.com/peer3to/state-channels-plus/blob/4d778e5905c7505ff3a57753ae99b72b51c7215b/src/stateManager/ValidationService.ts#L397

PR #174, opened 2025-09-23
Validation strategy

lukastanisic99:
Multiple components will reuse our `BlockConfirmation pipeline` (blockConfimation, disputes when they need to catchup to the state, syncing which is similar to blockConfirmation but peers don't initiate disputes) with different 'side effects' - this implements the `strategy design pattern` where all failures are handles by a specific strategy - e.g disputes can have their own, syncing/spectating their own etc etc - the pipeline stays unchanged

The pipeline can be refactored a bit since it's a bit weir like this - start in `StateManger` go to `ValidationService` (do things there in sub functions) and eventually come back to `StateManager` to continue doing checks - not high priority to refactor this at the moment - still should be more readable like this, since now in a single file you have all the failing function handlers


cyony95 (approved):

src/stateManager/validationStrategy/BlockValidationStrategy.ts
```diff
@@ -0,0 +1,154 @@
...
+    BlockConfirmationStruct,
+    ExitChannelBlockStruct
+} from "@typechain-types/contracts/V1/StateChannelManagerEvents";
+import AValidationStrategy from "./AValidationStrategy";
+import FraudProofService from "../utils/FraudProofService";
+import ADiamondStateMachine from "@/ADiamondStateMachine";
+import { StateChannelManagerProxy } from "@typechain-types/index";
+import Storage from "@/storage";
+import P2PManager from "@/P2PManager";
+import DisputeManager from "@/disputeManager";
```
daiagi: many unused imports
not sure how eslint let's you get away with this

src/stateManager/validationStrategy/BlockValidationStrategy.ts
```diff
@@ -0,0 +1,154 @@
...
+        this.fraudProofService.createInvalidStateTransitionProof(block);
+        // TODO this.disputeManager.createDispute()
+        return BlockValidationResult.DISPUTE;
+    }
+    public async wrongGenesisDetected(
+        block: Block
+    ): Promise<BlockValidationResult> {
+        //TODO
+        throw new Error("Not implemented");
+        // TODO this.disputeManager.createDispute()
```
daiagi, line 108: not sure why you removed the previous code
```ts
this.fraudProofService.createWrongGenesisProof(block);
```

just losing work that's been done before
also, isn't `createDispute()` a bit too general and vague
at what point are the fraud proofs being created and stored / immediatly uploaded?

src/stateManager/validationStrategy/BlockValidationStrategy.ts
```diff
@@ -0,0 +1,154 @@
...
+        this.storage.queues.queueBlock(block);
+        return BlockValidationResult.DISCONNECT;
+    }
+    public async notAllSingersAreParticipants(
+        block: Block
+    ): Promise<BlockValidationResult> {
+        return BlockValidationResult.DISCONNECT;
+    }
+    public async noNewSignaturesOnExistingBlock(
+        block: Block
```
daiagi, line 68:
```suggestion
        _block: Block
```

src/stateManager/validationStrategy/BlockValidationStrategy.ts
```diff
@@ -0,0 +1,154 @@
...
+            case BlockValidationResult.DISCONNECT:
+                // disconnect
+                return false;
+            case BlockValidationResult.BROADCAST:
+                return true;
+            case BlockValidationResult.DISPUTE:
+                return false;
+            default:
+                return true;
+        }
```
daiagi, line 45: might as well just remove the flags and return boolean
or, for a bit of clarity `{disconnet: boolean}

src/stateManager/validationStrategy/AValidationStrategy.ts
```diff
@@ -0,0 +1,77 @@
+import { Block, StateSnapshot } from "@/models";
+import { BlockValidationResult } from "@/types";
+import { Address, Bytes } from "@/types/types";
```
daiagi: unused imports

src/stateManager/validationStrategy/BlockValidationStrategy.ts
lukastanisic99: Not sure - I never liked eslint because it always errors out
Now I'm manually running `yarn lint:fix-only` and it corrected only some files (didn't push changes), but for instance this file wasn't fixed

It's forcing me to add `type:module` in `package.json` which I'm not sure how it works for you without and even after adding it fixes 2 files and errors with some internal error `Object.hasOwn is not a function` at `....node_modules/espree/dist/espree.cjs:917:20` - pure shit this eslint - maybe I have to use it with a specific version of node or? I installed all dependencies again btw

src/stateManager/validationStrategy/BlockValidationStrategy.ts
lukastanisic99: Why do you prefer this? it's a small isolated function argument - to me it doesn't make a big difference I can change all of them to `_block` just curious why

src/stateManager/validationStrategy/BlockValidationStrategy.ts
lukastanisic99: actually you can't since the pipeline is structured a bit weird where you have in some cases `StateManager` -> `validateBlockConfirmation` -> `another internal call`

so for example if `another internal call` returns any flag that resolves to `true` (e.g. `DUPLICATE`) the higher call (e.g.  `validateBlockConfirmation` will interpret that as `success` and continue the pipeline, when it should stop) - so this function is called at the very end in `StateManager` to just deduct what bool is returned to the calling context that invoked the pipeline - it will be different for disputes and if a bool is not enough in the end the context will expand, but I think the bool will be enough

src/stateManager/validationStrategy/BlockValidationStrategy.ts
lukastanisic99: My mistake - didn't see that line of code - restored it - check if I missed anything else 

As for create dispute - it won't be like that it has more arguments as you see - I just commented it out today so it's there as a reminder - I'll populate those fields probably today

src/stateManager/validationStrategy/AValidationStrategy.ts
lukastanisic99: eslint :D

daiagi (approved):

src/stateManager/validationStrategy/BlockValidationStrategy.ts
daiagi: just to not have unused var warning

PR #176, opened 2025-09-25
Sender transport - Idea

cyony95:
- P2PManager receives RPC with optional senderTransport parameter
- executeWithTransport() sets the transport in context before RPC execution
- RPC methods use getCurrentSenderTransport() to access the transport
- Context is automatically cleaned up after RPC execution

### DIFF
--- a/src/rpc/MainRpcService.ts
+++ b/src/rpc/MainRpcService.ts
@@ -31,8 +31,8 @@ class MainRpcService {
     p2pManager: P2PManager;
     rpcProxy = RpcProxy.createProxy(this);
 
-    //execution context
-    senderTransport: ATransport | undefined; //TODO! set this
+    //execution context - using context approach instead of shared variable
+    private currentTransport: ATransport | undefined;
     self = DEBUG_RPC ? DebugProxy.createProxy(this) : this;
 
     //RPC Services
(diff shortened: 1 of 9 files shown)


src/rpc/MainRpcService.ts
```diff
@@ -48,6 +48,26 @@ class MainRpcService {
...
+        } finally {
+            this.currentTransport = previousTransport;
+        }
+    }
+
+    /**
+     * Get the current sender transport from context
+     */
+    public getCurrentSenderTransport(): ATransport | undefined {
+        return this.currentTransport;
```
lukastanisic99, line 68: this is still a shared variable - restoring the old value after the function finishes doesn't help since another function can start executing while the first one `awaits` - think of it as multithreading on a single core cpu - there is only 1 thing happening at any 1 point in time, but a context switch can happen whenever, so another thread my start executing while the first one isn't finished 

with a shared variable we would have to have a `mutex` that controls access to a shared variable or for the calling function to not be async which is the same as the mutex logic - blocking. 

The async context was a really nice idea - too bad it doesn't work in browsers 

The underlying problem is we want to combine `exclusive` and `shared` variables in the same class/object/context - E.g. we still want the `service` to be able to define encapsulated `properties/fields` that it uses internally as a `shared` context across the different RPC methods of the service (e.g. how we calculate `rtt`), but we still want the `exclusive` variable (`msg.sender`) to be available without populating every RPC method with it. 

Closures are a nice way to capture the context, and while we can use them, the calling methods wouldn't be aware of the surrounding closure and couldn't access it

If we use shared variables with an async function we need to restore the value when it resumes and this is only true for single treaded execution pipelines

What we'll have to do in the end if we can't find a way to reuse the idea of async local storage / async local context is to create objects for every RPC call that will encapsulate the exclusive context and pass the global context by reference

cyony95: you're right, as usual. I'll continue with this, keeping this PR open and changing the implementation with a new idea if I can find one :D

PR #178, opened 2025-09-26
Events <- Events 2 

lukastanisic99:
Some changes - wanted to also implemented setTimeout -> reduce - started and realized I need stuff I added in another PR, so I left a todo and will implement post merge

### DIFF
--- a/src/stateManager/StateManager.ts
+++ b/src/stateManager/StateManager.ts
@@ -61,6 +61,7 @@ import {
     ChannelId,
     ForkId,
     Hash,
+    ReductionTimeoutHandle,
     Timestamp
 } from "@/types/types";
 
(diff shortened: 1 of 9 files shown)


src/stateManager/StateManager.ts
```diff
@@ -94,6 +94,7 @@ class StateManager {
     fraudProofService: FraudProofService;
     latestForkId: ForkId = NULL;
     defaultValidationStrategy: AValidationStrategy;
+    reductionTriggerMap: Map<ForkId, ReductionTimeoutHandle> = new Map();
```
daiagi, line 97: Maybe move this to storage

daiagi (approved):

src/stateManager/StateManager.ts
lukastanisic99: in browsers it could be in storage since the handle for `setTimeout` is an number id - in node it's a reference so we'd lose that in storage + it's only used in the StateManager

PR #182, opened 2025-09-29
Spectate feature 3

lukastanisic99:
@cyony95 take a look - it builds on your and my version, but it goes to disputes directly so the other 2 stay intact when we want to come back to them later 

This one uses local compute in most cases to try and detect failure early, buy has to double check in the end with a tx simulation over an RPC node `staticcall`

Apart from that there are other stuff implemented to accommodate these changes 

In general - the `staticcall` is a nice feature to simulate stuff on the RPC node or even in our local EVM

General problem with our local EVM is that we do `NOT` know is our local state sync - we never compare state roots as light clients do, but at the same time light clients don't actually sync state just headers and they're used to prove state by giving a merkle proof which the verify against the state root in the header - so they have to query the RPC node for that proof which is more than just asking the RPC node for the data, but with the proof they verify the data + know that they're synced (it adds up to the state root commitment in the block header) - this is more communication with the RPC nodes + it's lazy - the light client only request the data once it needs it, so we don't save any time nor compute - SOOOO the conclusion for now - using local EVM is fine for stateless compute (not reading EVM state) - it's deterministic and we have a nice interface to use it, but for stateful operations we need to run the simulation on an RPC node that's synced to the actual on-chain state - `staticcall` is a nice way to do it. 

You'll notice it also cleans up the interface in our local EVM, by not having to expose stuff that would normally return a `TransactionResponse` and we don't need to do any 'compiler tricks` - not sure if I actually removed that code that does the tricks

BTW - eslint not working again so I just ignored it :D + is there a plugin for grammar highlighting - I write these comments and some times even variables with spelling mistakes and I realize later (or never) that they're misspelled


contracts/V1/StateChannelDiamondProxy/utils/DisputeUtils.sol
```diff
@@ -88,3 +89,11 @@ function areDisputesCommitted(DisputeWindow storage disputeWindow, Dispute[] mem
     }
     return true;
 }
+
+function _hadParticipantPostedEvidence(DisputeWindow storage disputeWindow, address participant) view returns (bool) {
```
daiagi, line 93: typo:
`_hadParticipantPostedEvidence`

src/evm/ContractExecuter.ts
```diff
@@ -19,10 +21,20 @@ export default class ContractExecuter {
         return this.contractAddress;
     }
 
-    async executeCall(data: Bytes): Promise<ExecResult> {
+    async executeCall(
+        data: Bytes,
+        caller?: Address,
```
daiagi, line 26: 'caller' is declared but its value is never read.ts(6133)

src/rpc/MainRpcService.ts
```diff
@@ -124,5 +127,22 @@ class MainRpcService {
...
+
+    public async onSpectateResponse(
+        channelId: ChannelId,
+        snapshotPayload: SyncPayload,
+        responseTime: Timestamp
+    ) {
+        this.spectateService.onSpectateResponse(
+            channelId,
+            snapshotPayload,
+            responseTime
```
daiagi: Expected 2 arguments, but got 3.

src/evm/ContractExecuter.ts
lukastanisic99: ah forgot to pass it to the EVM - will fix

src/rpc/MainRpcService.ts
lukastanisic99: good catch

daiagi (approved):

PR #187, opened 2025-10-03
New RPC design - senderTransport and RPC entry points isolated

lukastanisic99:
Everything builds - typescript types are correct. Didn't test. 

This new design should be really simple and straight forward - the complexity is 'slightly' in the proxies, but that's implemented once and that too isn't that complex (needs to be tested)


src/rpc/RemoteRpcProxy.ts
```diff
@@ -0,0 +1,51 @@
...
+                const val = Reflect.get(target, prop, receiver);
+
+                if (typeof val != "object" || !(val instanceof ARpcService)) {
+                    throw new Error("RemoteRpcProxy can only access services");
+                }
+                // val is a service
+                const serviceName = prop.toString();
+                ctx.serviceName = serviceName;
+                // @ts-ignore
+                ctx.service = val;
```
lukastanisic99, line 45: so the ctx is an object that's passed to `RpcMethodsProxy` above and it's by reference, so we can change its properties before the above proxy is invoked so we can package the correct service name and have the correct service during the execution of the above proxy. 

If someone has a better way of doing this - let me know.
However - this is isolated. 

Needs to be tested!!!

daiagi (approved):

PR #188, opened 2025-10-05
E2e-tests-infra

daiagi:
- added actual assertions to `EvmStateMachine.test.ts`
- fixed all sorts of bugs that popped up
- created `PeerTestHarness`
- as demo, created `EvmStateMachine-harness.test.ts`, which tests exactly same as  `EvmStateMachine.test.ts`, but is much shorter, this is to showcase that the harness allows us to write e2e tests much easier and much more focus on business logic

- Until opening channel and setting genesis state snapshot are resolved, it is hard to write tests, since it is hard to "initiate", so more e2e tests to follow later

---
## Edit:

**For clarity, these are the bugs that were fixed:**
1. `src/evm/LocalDiamondSigner.ts\resolveName`  [doc](https://docs.ethers.org/v5/api/signer/#Signer-resolveName)
hardhat does not support ENS name resolution out of the box, nor do we care about it, but the Signer interface requires this method to be implemented.
Naive simple fix by wrapping in try/catch, this is not something we care about.

2. `src/utils/evmErrorHandler.ts`  aka `decodeErrorProxy`:  was wrapping everything with a promise.
resulting in `localDiamondContract.getDiamondAddress` returning a promise, instead of direct value.
solved by making the proxy aware of the original return value : 
` const isAsync = originalProperty.constructor.name === "AsyncFunction";`

3. removed 
```ts
        if (hash(encodedState) === stateSnapshot.stateMachineStateHash) {
                validationResult =
                    await strategy.invalidStateTransitionDetected(block);
                return await strategy.interpretFinalValidationResult(
                    validationResult
                );
```

From StateManager

this check was nonsense, the function above it that creates the state snapshot takes in as an argument `hash(encodedState) ` and put it as `stateSnapshot.stateMachineStateHash`
simply put, this condition is always true and leads to dispute for no reason


src/stateManager/StateManager.ts
```diff
@@ -308,104 +306,102 @@ class StateManager {
...
+                    stateSnapshot,
+                    encodedState,
+                    successCallback,
+                    totalWithdrawals,
+                    leftParticipants,
+                    exitChannelBlock
                 );
-            }
 
-            if (hash(encodedState) === stateSnapshot.stateMachineStateHash) {
```
lukastanisic99, line 385: the mistake here was that it was `===` it should have been `!=`, but since we check `if(createdSnapshot.hash != block.snapshotHash)` the encodedState state check is implicit and not needed here

daiagi:
@lukastanisic99  let me know if this commit is ok with you
5b88006


I've added it because some test was failing

lukastanisic99:
> @lukastanisic99 let me know if this commit is ok with you [5b88006](https://github.com/peer3to/state-channels-plus/commit/5b880063ded6e6dcf2c2b9a9e25d1a600ae3fb30)
> 
> I've added it because some test was failing

It needs to be the proxy, since that proxy is what captures the `method` and `args` part - so the method and args don't really exist as a callable function on that target, but the proxy never calls them - it just serializes them and builds the rpc

.github/workflows/ci.yml
```diff
@@ -7,27 +7,30 @@ on:
...
 
-      # Step 4: Compile and run tests
-      - name: Compile and run tests
-        run: yarn testc
\ No newline at end of file
+      - name: Compile
+        run: yarn compile
+
+      - name: Run tests
+        run: yarn testc
```
lukastanisic99: `yarn testc` compiles it again - it was short for `compile and test`

daiagi: yep

daiagi:
@cyony95 @lukastanisic99 
the GH workflow works, all tests are passing

image

From this moment on, in order to be accepted, a PR will need to pass all the tests (aka not break anything)
PR's that include tests for the new functionality, even better

src/stateManager/StateManager.ts
```diff
@@ -308,104 +312,102 @@ class StateManager {
...
+            // no error is actually expected to happen, and the catch block just re-throws the error
+            const strategy =
+                validationStrategy || this.getStrategyByStatus(this.status);
+            try {
+                await this.mutex.lock();
+                let validationResult: BlockValidationResult =
+                    BlockValidationResult.SUCCESS;
+                const isAuthentic =
+                    await this.diamondStateMachine.localDiamondContract.isBlockAuthentic(
+                        blockConfirmation.signedBlock
```
lukastanisic99, line 326: this didn't change don't know why github is highlighting it

src/stateManager/StateManager.ts
```diff
@@ -308,104 +312,102 @@ class StateManager {
...
+                    );
 
-            const block = Block.fromBlockConfirmation(
-                blockConfirmation,
-                onChainTimestamp
-            );
+                if (!isAuthentic) {
+                    validationResult =
+                        await strategy.authenticateBlockFailed(
+                            blockConfirmation
```
lukastanisic99, line 332: this just has different formatting - not sure is it from eslint or prettier - prettier works for me, so I guess it's from eslint, otherwise it would have the same format?

src/stateManager/StateManager.ts
daiagi: Indentation I think
It didn't change you are right

src/utils/config.ts
```diff
@@ -1,55 +1,65 @@
...
+};
+
+// Create configuration: if node -> test config, else -> regular config, then apply overrides
+function createConfig(overrides: Partial<Config> = {}): Config {
+    let baseConfig: Partial<Config> = {};
+
+    // If running in Node.js, use test config, otherwise use regular config
+    if (isNode()) {
+        try {
+            const testConfigModule = require("../../test/peer3.test.config");
```
lukastanisic99, line 36: why did we get rid of `loadConfigFromFile` and why do we have `test stuff` in `src` 

if you need a config for testing it should be possible to inject/load it

daiagi: That stuff never was triggered, it relied on json file that never existed, and in order to load a json you need to be in node anyway, so it will end up exactly at the same point 

The assumption is that running in node means running tests, so it loads test config
This assumptions can be made more specific if needs


Whats the issue? Why the word test bothers you so much?

The destinction between the two envs must be somewhere

lukastanisic99: This was more to be used by the consumer of the sdk - someone that installs it through `npm` - so the `config` is outside. Regardless if it runs in node/browsers etc - they'd build it with node (or we assume) 

How do other projects do config files?

lukastanisic99 (approved):

PR #189, opened 2025-10-05
Add rate limiter

cyony95:
closes https://trello.com/c/EMbLjLrj/222-ratelimit-spam

Initially, I wanted to configure each of the transports natively to be rate limited. Holepunch doesn't support rate limiting, but webRTC does. The universal solution was a rate limiter implemented on the SDK level, independent of transport type.

Right now, the rateLimiter is very simple: 
- it works with a token bucket algorithm and it refills tokens over time. 
- it supports bursting (with maxTokens). 
However:
- it only refills tokens when checkAndConsume is called, so lazy evaluation. 
- it drops messages when rate limited
- it works for both outbound and inbound (with inbound having a token bucket per connection)

Let me know if you want it to queue the messages or some other functions.


src/transport/WebRTCTransport.ts
```diff
@@ -23,8 +25,18 @@ class WebRTCTransport extends ATransport {
...
         console.log("WebRTC - SendingRPC", serializedRPC);
+
+        if (globalRateLimiter) {
+            const dataSizeBytes = Buffer.byteLength(serializedRPC, "utf8");
+            if (!globalRateLimiter.checkAndConsume(dataSizeBytes)) {
+                console.warn("WebRTC rate limit exceeded, dropping message");
+                return; // Drop the message
+            }
+        }
+
```
lukastanisic99: We're more concerned about incoming data not being spam/abusive, than us spamming others - for honest peers this should never trigger - but nice to have as a warning especially if we have things like chat 

The rate limiting should be per connection/PeerProfile on incoming traffic - if we're being spammed too much we should disconnect from the peer

src/utils/RateLimiter.ts
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ * Implements a token bucket algorithm for rate limiting
+ */
+export class RateLimiter {
+    private tokens: number;
+    private lastRefillTime: number;
+    private readonly maxTokens: number;
+    private readonly refillRate: number; // tokens per second
+    private readonly tokenSize: number; // bytes per token
```
lukastanisic99: is there are benefit/reason to thinking in `tokens*tokenSize` vs just thinking in `bytes`

cyony95: the reason is that this way we are sure the messages are complete. if we implement it in bytes, we might send incomplete messages, so this way we technically maintain message atomicity.

src/transport/WebRTCTransport.ts
cyony95: added  functionality for inbound rate limiting.

cyony95: @lukastanisic99 question is, should we rate limit outbound per connection as well? right now it's global

src/utils/RateLimiter.ts
lukastanisic99: not sure I follow how tokens make a difference - we're checking whole bytes of a constructed msg (serializedRPC) in both cases - the only difference here we're dividing them with `tokenSize` so the checks is:

tokensRemaining > tokensNeeded <=> bytesRemaining/tokenSize > bytesNeeded/tokenSize and we just have an unnecessary division on both sides 

Maybe I don't see something

src/transport/HolepunchTransport.ts
```diff
@@ -31,6 +33,15 @@ class HolepunchTransport extends ATransport {
     }
     send(serializedRPC: string): void {
         console.log("SENDING RPC", serializedRPC);
+
+        if (outboundRateLimiter) {
+            const dataSizeBytes = Buffer.byteLength(serializedRPC, "utf8");
```
lukastanisic99: Would consider extracting this and writing it only once in the base class for all transports and then the actual concrete transports be maybe private methods that are called by the base class if they can write, so we don't end up copy pasting this in every transport we implement 

```typescript
ATransport{
  abstract _send()
  send(){
    rateLimit
    // more shared code here
    this._send()
  }
}
```

src/utils/RateLimiter.ts
cyony95: changed it to bytes

src/transport/HolepunchTransport.ts
cyony95: great suggestion!

lukastanisic99:
@cyony95 without a clean solution to address gossip in bandwidth, no point in integrating this at this stage - I have other higher priority stuff to think about, so you can either let this hang if you want to do something else or think about this problem more in depth.
My 2min thinking: probably we'll have to add a message origin to every RPC and authentication (signing) so we know who's bandwidth to deduct, but this is also slightly annoying since very rarely do we actually `just gossip` the message (at least in regards to blocks) - usually we extend the message with our signature and gossip that, but this logic is specific to the RPC regarding that - if we implemented a chat we would of course gossip the unmodified message (or maybe not even gossip it, since chat is not a `critical` feature and if messages go missing/aren't delivered nothing breaks, there are no security assumptions) - can't give you a general solution without spending some time to think

src/rpc/Rpc.ts
```diff
@@ -2,18 +2,36 @@ type Rpc = {
...
+export function createMessageContent(
+    method: string,
+    params: any[],
+    timestamp: number
+): string {
+    const messageContent = {
+        method: method,
+        params: params,
+        timestamp: timestamp
+    };
```
lukastanisic99, line 25: why not include `service` in the msgContent - that's all part of the msg

Better to follow our typical convention:
```typescript
type Rpc = {
    service: string;
    method: string;
    params: any[];
    timestamp: number;
}

type SignedRpc = {
    serializedRpc: string;
    signature: string;
}
```

src/rpc/RpcHandleProxy.ts
```diff
@@ -35,12 +35,32 @@ class RpcMethodsProxy {
                         return Reflect.get(target, prop, receiver);
                     }
                     if (typeof prop === "symbol") return;
-                    return (...args: any) => {
+                    return async (...args: any) => {
+                        const timestamp = Date.now();
```
lukastanisic99: this should be `Clock.getTo,eInSeconds()` we're always using our logical clock

src/rpc/RpcHandleProxy.ts
```diff
@@ -35,12 +35,32 @@ class RpcMethodsProxy {
                         return Reflect.get(target, prop, receiver);
                     }
                     if (typeof prop === "symbol") return;
-                    return (...args: any) => {
+                    return async (...args: any) => {
```
lukastanisic99: this makes the usage ugly - it's really nice to be able to chain e.g. (`remote.service.method(args).broadcast()`)

Better to make the end functions async if you need to sign -> in `RpcHandler` make `broadcast, sendOne, sendMultiple` async and sign there - they all seralize the RPC and those functions are so small and don't change and then we can keep the nice `chaining` and awaiting them is not relevant - they will eventually just send out the message to the network which will regardless travel with some network delay which is aysnc anyway

src/utils/RateLimiter.ts
```diff
@@ -0,0 +1,288 @@
...
+                return false; // Invalid signature
+            }
+
+            // Check for message deduplication
+            const messageHash = this.createMessageHash(rpc);
+            const cachedMessage = this.messageCache.get(messageHash);
+
+            if (cachedMessage) {
+                // Message already processed, don't charge bandwidth again
+                return true;
```
lukastanisic99: why is message validation and caching the responsability of the rate limiter

src/rpc/RpcHandleProxy.ts
cyony95: moved signing to rpchandler and made broadcast, sendone and sendmultiple async.

src/utils/RateLimiter.ts
cyony95: moved the validation and the caching to separate classes

src/rpc/RpcHandler.ts
```diff
@@ -1,24 +1,58 @@
...
+    constructor(rpcPayload: UnsignedRpc, p2pManager: P2PManager) {
+        this.rpcPayload = rpcPayload;
         this.p2pManager = p2pManager;
     }
 
-    public broadcast() {
-        this.p2pManager.broadcastRpc(serializeRpc(this.rpc));
+    private async buildSignedRpc(): Promise<Rpc> {
+        if (!this.signedRpcPromise) {
+            this.signedRpcPromise = (async () => {
```
lukastanisic99: Imo better to `await` here and have the `signedRpc` instead of `signedRpcPromise`

src/transport/ATransport.ts
```diff
@@ -1,10 +1,35 @@
...
 
-    abstract send(serializedRPC: string): void;
+    /**
+     * Public send method that handles rate limiting and other shared logic
+     */
+    send(serializedRPC: string): void {
+        // Check outbound rate limiting
+        if (outboundRateLimiter) {
+            const dataSizeBytes = Buffer.byteLength(serializedRPC, "utf8");
+            if (!outboundRateLimiter.checkAndConsume(dataSizeBytes)) {
```
lukastanisic99: I'm boarding now, so will continue later - I pushed some changes already

Here - this rate limmits all outgoing RPCs in the `SAME` bucket - we don't want that

on the outgoing side we should care to just limits `OUR` messages so we potentially become aware (get a warning) that we should `cool off` - we should NOT count other peers messages towards this limit as they (should) have their own bucket that is deduced onRPC by analyzing the message origin of the RPC with the TTL (didn't review that part yet)

lukastanisic99:
So I've implemented most changes today, which I'll push soon to this PR to have for later, but we'll close for now and come back later - why?

I don't have a clean solution for general gossip - an attacker can create an infinite amount of addresses that we're not even connected to and gossip their messages, which we'd either drop or create rate buckets for. Using `isParticipant` has race conditions which we can't address easily here (depending on the state, the opinion `isParticipant` may very, unlike with onBlockConfirmation where we're doing block by block as a function of state, so the outcome is deterministic for everyone) 

So what we'll do:
I'll create a new PR for rate limit without general gossip - that will strip half of this away - you vouch for the messages sent and they strictly originate from you and it's your responsibility to send them to relevant peers - the things that has `gossip like behavior` is the `BlockConfirmation pipeline` which has it implemented already in a way that the message comes from you - this is the only relevant pipeline for now - other RPC are either direct communication or we don't care if a message is dropped

src/utils/RpcFilter.ts
```diff
@@ -0,0 +1,118 @@
...
+                }
+                const issuerAddress =
+                    await this.validationService.recoverAddressFromRpc(rpc);
+                if (!issuerAddress) {
+                    this.p2pManager.disconnectAndBlacklistPeer(transport);
+                    return undefined; // Invalid signature
+                }
+                // TODO
+                /*
+                    Solution for general gossip needed here
```
lukastanisic99, line 80: So what's missing for general gossip - is a solution here - for now this is not important and the gossip we use is not general (the RPC that uses it, implements the logic)

lukastanisic99:
@cyony95 take a look at the changes if you want - I'll close the PR and do a simple PR for non gossip rate limiting

PR #196, opened 2025-10-12
time and broadcast fixes

daiagi:
the test EvmStateMachine.test.ts was flaky, sometime passing sometime not.
I identified three issues:

1. Double broadcast, p2pSigner was broadcasting again after playTransaction already does so
2. Too strict time validation, i added 1 second leeway, that solved the issue on my machine
3. init handshake was using `settimeout` value in seconds instead of ms => connections were closed too soon


src/stateManager/ValidationService.ts
```diff
@@ -268,8 +268,10 @@ export default class ValidationService {
         }
 
         // OBJECTIVE: isValidTimestamp check
+        // Allow small clock drift tolerance (1 second backwards)
+        const CLOCK_DRIFT_TOLERANCE = 1; // seconds
         const isValidTimestamp =
-            block.timestamp >= previousTimestamp &&
+            block.timestamp + CLOCK_DRIFT_TOLERANCE >= previousTimestamp &&
```
lukastanisic99: This is an objective check - that shouldn't have anything to do with your local clock or any `clock_drift_tolerance` - it's just comparing timestamps that are recorded measures - the previous timestamp is also an objective measurement 

Drift/Skewness is relevant when using your local clock to either record a timestamp or compare a timestamp and we already tolerate skewness/drift in `agreementTime` and when recording the timestamp in `playTransaction` so if the recorded timestamp is not correct than `adjustTimestampIfNeeded` has a bug and by just looking at it - it seems that way that it doesn't take into account the genesisTimestamp 

further I think I made a mistake here  `block.timestamp >= previousTimestamp && block.timestamp <= previousTimestamp + this.timeConfig.p2pTime` which is a mistake in this trello card https://trello.com/c/8cLJYFfm

previousTimestamp is block.relevantTimestamp which forces the block author to produce the block after the previous one has been posted on-chain, which we don't want and may even cause a slash to an honest peer that produced the next block before the previous one was posted on-chain 

what it should be is:

`previousOriginalTimestamp = genesis.timestap || previousBlock.timestamp`
`previousTimestamp = same....` 

and the check should be 
block.timestamp >= previousOriginalTimestamp  && block.timestamp <= previousAdjustedTimestamp + this.timeConfig.p2pTime

probably need to account for it in solidity too - will add a trello card and edit the description of the trello card above so it has everything and edits are visible

daiagi: removed the drift, updated `adjustTimestampIfNeeded` to adjust timestamp of first block on a fork
that solved the issue for me, and i think it makes sense.
i am not sure there is a mistake
see here, https://github.com/peer3to/state-channels-plus/blob/872def6e1a129b2f080c4a90dacedd26677e3b95/src/stateManager/ValidationService.ts#L262
we are using the `getRelevantTimestamp`

anyway i did not adjust that, i was only concerned with an annoying race condition in test

src/stateManager/StateManager.ts
```diff
@@ -1264,15 +1264,24 @@ class StateManager {
...
+        } else {
+            previousTimestamp = latestBlock.timestamp;
         }
 
-        const latestBlockTimestamp = latestBlock.timestamp;
-
-        if (Number(tx.header.timestamp) < latestBlockTimestamp) {
-            tx.header.timestamp = latestBlockTimestamp + 1;
+        if (Number(tx.header.timestamp) <= previousTimestamp) {
+            tx.header.timestamp = BigInt(previousTimestamp + 1);
```
lukastanisic99: even if we didn't do +1 it should be fine - added that more as safety but we can remove it it might be better if blocks are being produced too fast since the subjective check will fail thinking they're too far in the future in that case

src/stateManager/ValidationService.ts
```diff
@@ -268,10 +268,8 @@ export default class ValidationService {
         }
 
         // OBJECTIVE: isValidTimestamp check
-        // Allow small clock drift tolerance (1 second backwards)
-        const CLOCK_DRIFT_TOLERANCE = 1; // seconds
         const isValidTimestamp =
-            block.timestamp + CLOCK_DRIFT_TOLERANCE >= previousTimestamp &&
+            block.timestamp >= previousTimestamp &&
```
lukastanisic99: `block.timestamp >= previousTimestamp` -> `block.timestamp >= previousOriginalTimestamp` 
https://trello.com/c/8cLJYFfm

src/stateManager/StateManager.ts
daiagi: Removed the +1

src/stateManager/ValidationService.ts
daiagi: This PR is only about stabilizing the race condition
changing the time validation logic will be done separately

daiagi: https://github.com/peer3to/state-channels-plus/pull/198

lukastanisic99 (approved):

PR #197, opened 2025-10-13
skeleton for e2e testing

daiagi:
This PR includes list of test cases to be divided up in trello cards and implemenetd

There is 0 implemetned code here, just a blueprint for us to agree on and start implementing

The idea is to use `peerTestHarness` to facilitate writing these tests

**Edit**
---

started implementing already, @lukastanisic99  note d76c48c324e72a8b26eff6b65e09036d4fd12c66


lukastanisic99:
@daiagi we need an easy way to 'teleport' time in tests that will trigger things (without actually waiting for time to elapse) - we need that for testing in the evm also - I'll let you think about it

daiagi:
@lukastanisic99 
Although I agree with you that teleporting time is better, I disagree that it priority

By allowing the PR to go through, Adrian can start writing tests as, and i get the approval from you to continue writing them as well

Ideally, by the end of the month the test suite is done, the bugs are fixed,  and you have confidence to think about shipping

Even if the tests take long time to run (can ball park 90s, see calc below), this is just a minor dev inconvenience and can be improved on later, not something we should delay for.

### Time estimate:

In the skeleton we have  35 tests
right now 5 are implemented and they take 8 seconds (this can proabbly be improved but nvm )

so: `35 X (8s/5) ~ 88 seconds

src/stateManager/StateManager.ts
```diff
@@ -657,8 +657,14 @@ class StateManager {
         if (!block.didEveryoneSign(participants)) {
             this.p2pEventHooks.onPostingCalldata?.();
 
+            const maxTimestamp =
+                Clock.getTimeInSeconds() +
```
lukastanisic99: maxTimestamp = previousRelevantTimestamp+p2p+agreement+fallback

lukastanisic99:
> @lukastanisic99 Although I agree with you that teleporting time is better, I disagree that it priority
> 
> By allowing the PR to go through, Adrian can start writing tests as, and i get the approval from you to continue writing them as well
> 
> Ideally, by the end of the month the test suite is done, the bugs are fixed, and you have confidence to think about shipping
> 
> Even if the tests take long time to run (can ball park 90s, see calc below), this is just a minor dev inconvenience and can be improved on later, not something we should delay for.
> 
> ### Time estimate:
> In the skeleton we have 35 tests right now 5 are implemented and they take 8 seconds (this can proabbly be improved but nvm )
> 
> so: `35 X (8s/5) ~ 88 seconds

If it makes a difference to you sure - do it this way, but I think you'll get annoyed waiting also realize that it will be better to spend some time (a day or 2) to implement teleportation especially since I believe there will be a lot more tests than 35 - I just didn't spend time thinking about them yet

src/stateManager/StateManager.ts
daiagi: That doesn't work
I tried

lukastanisic99: Then there's a bug somewhere

lukastanisic99: The only reason tweaking this would cause it to fail on-chain is if the on-chain timestamp (block.tinestamp) is not correct

daiagi: fixed, it was due to multiple setTimouts polluting each other between tests

src/evm/EvmDiamondStateMachine.ts
```diff
@@ -376,13 +378,21 @@ class EvmDiamondStateMachine extends ADiamondStateMachine {
...
             chainFallbackTime: Number(configTimes[2]),
             evidenceTime: Number(configTimes[3])
         };
 
+        // Apply time configuration override if provided (for testing)
+        if (timeConfigOverride) {
+            timeConfig = {
+                ...timeConfig,
+                ...timeConfigOverride
+            };
```
lukastanisic99, line 393: btw you know that you can just define these values in the test contract (MathStateMachine) and that you don't need to override anything - the values in the contract will be fetched

lukastanisic99 (approved):

PR #198, opened 2025-10-13
fix: time validation

daiagi:
https://github.com/peer3to/state-channels-plus/pull/196#discussion_r2425290104


src/stateManager/ValidationService.ts
```diff
@@ -262,15 +263,34 @@ export default class ValidationService {
...
             previousStateSnapshot = previousBlockOrSnapshot.stateSnapshot;
             previousTimestamp = previousStateSnapshot!.timestamp;
+            previousOriginalTimestamp = previousStateSnapshot!.timestamp;
         }
 
         // OBJECTIVE: isValidTimestamp check
+
+        // Check if block timestamp is not in the past
+        const isTimestampNotInPast =
+            block.timestamp - previousOriginalTimestamp >= 0;
```
lukastanisic99, line 277: to me this is less readable `isTimestampNOTinPast..` <=> `isTimestampInTheFuture `- still I'd keep the original comparison I though it's simple

src/stateManager/ValidationService.ts
```diff
@@ -262,15 +263,34 @@ export default class ValidationService {
...
+            this.timeConfig.p2pTime +
+                this.timeConfig.agreementTime +
+                this.timeConfig.chainFallbackTime;
+
         const isValidTimestamp =
-            block.timestamp >= previousTimestamp &&
-            block.timestamp <= previousTimestamp + this.timeConfig.p2pTime;
+            isTimestampNotInPast &&
+            isWithinP2PTimeWindow &&
+            isWithinMaxAllowedTime;
```
lukastanisic99: if we do this here `isWithinMaxAllowedTime ` than bellow we have to remove `isPostedOnChainTooLate` - didn't think too much about this is it logically equivalent in regard to race condition on-chain, but I guess it is

daiagi: Isn't it what you asked for?

image

lukastanisic99: > Isn't it what you asked for?
> 


yes but isn't it already done bellow in `isPostedOnChainTooLate` - I think here is better, but in that case we have to remove `isPostedOnChainTooLate` - 
The edits in Trello show all Edits - I made that one before in the comments - the latest one from yesterday was just `previousOriginalTimestamp`

src/stateManager/ValidationService.ts
daiagi: done

src/stateManager/ValidationService.ts
daiagi: removed 'isWithinMaxAllowedTime'

lukastanisic99 (approved):

PR #199, opened 2025-10-14
added stateProof RPC && isForkDisputed RPC

cyony95:
closes https://trello.com/c/KkkgtrYx/189-rpc-to-prove-forkid-reuse-sync-when-peers-detect-theyre-not-on-same-forks
closes https://trello.com/c/tJRQR6hq/232-think-about-honest-peer-on-wrong-fork-related-to-the-rpcto-prove-forkid

### stateProof
- Fork mismatch detected → Request state proof for specific fork + block height
- Peer generates proof for that exact state
- Peer sends proof → Success! (If they can send it, they can prove it)
- Timeout → Disconnect if no response

### isForkDisputed
- Peer builds on disputed fork → Strategy triggers handshake
- Check if already acknowledged → If yes, disconnect immediately
- Handshake exchange → Both peers check and share dispute status
- Agreement check → If disagree, disconnect; if agree on disputed, mark acknowledged
- Block processing → Queue block for later processing
- Future attempts → If building on acknowledged disputed fork again, disconnect


src/rpc/services/stateProof/StateProofRpcMethods.ts
```diff
@@ -0,0 +1,91 @@
...
+            return this.p2pManager.disconnectAndBlacklistPeer(senderTransport);
+        }
+
+        console.log(`Peer's fork ${provedForkId} is the canonical fork!`);
+        console.log(
+            `I was on fork ${myForkId}, syncing to canonical fork ${provedForkId}`
+        );
+
+        // Trigger a spectate sync to actually sync to their state
+        this.service.spectateService.spectateSync(senderTransport, channelId);
```
cyony95: @lukastanisic99  should we even do this? I'm not sure, since we will queue the block I think this is unnecessary.

src/rpc/services/stateProof/StateProofService.ts
```diff
@@ -0,0 +1,172 @@
...
+        // Timeout if they don't respond
+        setTimeout(() => {
+            if (this.forkProofInitTimes.has(transport)) {
+                console.log(
+                    `Peer failed to prove fork ${peerForkId}, disconnecting`
+                );
+                this.p2pManager.disconnectAndBlacklistPeer(transport);
+                this.forkProofInitTimes.delete(transport);
+            }
+        }, this.p2pManager.stateManager.timeConfig.agreementTime * 1000);
```
cyony95: @lukastanisic99 setTimeout expects milliseconds. is agreementTime set in milliseconds as well?

src/stateManager/ValidationService.ts
```diff
@@ -46,6 +48,24 @@ export default class ValidationService {
...
+
+            // Challenge peer to prove their fork
+            this.stateManager.p2pManager.localRpc.forkProofService.challengePeerFork(
+                senderTransport,
+                channelId,
+                block.forkId
+            );
+
+            // Queue the block - will process after sync if peer proves canonical fork
+            this.storage.queues.queueBlock(block);
```
cyony95: @lukastanisic99 not sure if I should queue the block here or not

src/rpc/services/forkProof/ForkProofRpcMethods.ts
```diff
@@ -0,0 +1,91 @@
...
+
+    constructor(transport: ATransport, service: ForkProofService) {
+        super(transport, service.p2pManager);
+        this.service = service;
+    }
+
+    /**
+     * Peer receives challenge: "Prove your fork"
+     */
+    public async onProveForkRequest(
```
lukastanisic99: this should be prove state which is {`fork` & `blockHeight`}

src/stateManager/ValidationService.ts
lukastanisic99: you don't need an explicit check for this - other parts of the code will deduct are we on the correct fork and apply it or will try and queue it - where we queue it, we send this request to the other peer to prove the `state` - this is done in the specific `strategy` and think for every strategy what should be the strategy for this

This + working rateLimit is enough to protect vs spam 

This covers all non-canonical forks & all non-provable blockHeights on canonical forks 

What it doesn't cover when someone is building on a fork that's disputed since they may not be aware that the fork is disputed - this is also simple to add 

expose an RPC that just does a handshake between peers acknowledging that the fork is disputed (this causes the peers to explicitly check is it - if they don't agree on the status isDisputed they disconnect from one another - for every forkId this request can be sent only once - so if a peer `acknowledges` that the fork is disputed the next time they send something on that fork the above RPC will trigger a disconnect)

so the only thing you need to do in that case in the `strategy.blockForkIsDisputed` is to trigger the RPC handshake

daiagi:
@cyony95  i resolved the merge conflicts for you

src/stateManager/ValidationService.ts
cyony95: Added isForkDisputedService

daiagi:
the failing test will pass after merge of #203

src/rpc/services/isForkDisputedService/IsForkDisputedService.ts
```diff
@@ -0,0 +1,68 @@
...
+    public acknowledgeDisputedFork(transport: ATransport, forkId: ForkId) {
+        let acknowledgedForks = this.acknowledgedDisputedForks.get(transport);
+        if (!acknowledgedForks) {
+            acknowledgedForks = new Set();
+            this.acknowledgedDisputedForks.set(transport, acknowledgedForks);
+        }
+        acknowledgedForks.add(forkId);
+    }
+}
+
```
lukastanisic99, line 158: I'll comment here for the entire service 

Imo best to do this handshake the moment the `disputeWindow` is created on-chain, not prior, so everyone can check in O(1) in the worst case on the RPC node without much delay (propagation time / transaction inclusion time) - anything prior is premature and not needed

this handshake will be substituted later for a disputeConfirmation for early finality of the disputeWindow, for now it's fine to do it like this:

1) event : onDisputeCommitted -> if window opened (if first dispute in window) -> RPC.isDisputed.request(forkId).sendAll() -> if response doesn't come within agreementTime assume dishonest and disconnect 
2) request(forkId){
if(transport already sent request for same forkId) disconnect
if(localDiamond.isDisputed(forkId)) return respond(forkId)
if(stateChannelManagerContract.isDisputed(forkId)) return respond(forkId)
else return disconnect // no need to respond 
}
3) respond(forkId){
if(transport already responded for forkId) disconnect
map<transport,set<forkId>>.set(transport,forkId)
}

so when you get a block of a disputed fork you just check `localRpc.isDisputed.didTransportAckDispute(transport,forkId)` if true -> disconnect, else queue block but don't disconnect

lukastanisic99:
Tried Copilot to see how he reviews things - useless

src/rpc/services/isForkDisputedService/IsForkDisputedService.ts
cyony95: great idea to trigger it when a disputeWindow is created on chain!

cyony95:
@daiagi added also e2e tests descriptions. I added more but you can thin them out if you see anything too "extra".

src/rpc/services/isForkDisputedService/IsForkDisputedService.ts
```diff
@@ -0,0 +1,67 @@
...
+     * This should be called when a dispute window is created on-chain
+     */
+    public requestDisputeAcknowledgment(channelId: ChannelId, forkId: ForkId) {
+        console.log(
+            `Requesting all peers to acknowledge disputed fork ${forkId}`
+        );
+
+        this.remoteRpc.isForkDisputedService
+            .onDisputeAcknowledgmentRequest(channelId, forkId)
+            .sendMultiple(this.p2pManager.openConnections);
```
lukastanisic99: this is same as just calling broadcast()

lukastanisic99: Also here you should trigger setTimeout() for 2*agreementTime to observe have the open connections `acknowledged` and disonnect from the ones that havent 

make a copy of the refferences to openConnections or peer addresses to whom you've sent these requests, so you don't disconnect from peers that may connect to you within 2*agreementTime.. So the copy is like a snapshot which will be caputured in the closure that runs when setTimeout expires

src/stateManager/StateManager.ts
```diff
@@ -462,7 +463,8 @@ class StateManager {
     public async onBlockConfirmation(
         blockConfirmation: BlockConfirmationStruct,
         onChainTimestamp?: Timestamp,
-        validationStrategy?: AValidationStrategy
+        validationStrategy?: AValidationStrategy,
+        senderTransport?: ATransport
```
lukastanisic99: senderTransport is never set - these 3 optional args have to be extracted into `options` since like this it's not possible to set senderTransport without setting `onChainTimestamp` and `validationStrategy` and other combinations that don't include prior args, which in some cases is not possible (e.g. if the block comes from the p2p network you can't set onChainTimestamp, since that is reserved when the block is posted as calldata and you get it from the chain etc..)

cyony95: good observation!

src/rpc/services/isForkDisputedService/IsForkDisputedRpcMethods.ts
```diff
@@ -0,0 +1,123 @@
...
+
+        // Fork is not disputed - disconnect
+        console.log(`Fork ${forkId} is not disputed, disconnecting`);
+        return this.p2pManager.disconnectAndBlacklistPeer(this.senderTransport);
+    }
+
+    /**
+     * Respond to dispute acknowledgment
+     */
+    private async respondToDisputeAcknowledgment(
```
lukastanisic99: these helper functions should be in the service and only entry points should be exposed as public functions in `RpcMethods` - during runtime `typescript` is striped and a byzantine actor can invoke these private functions as rpc{serviceName,privateMethodName,args}

src/rpc/services/isForkDisputedService/IsForkDisputedRpcMethods.ts
```diff
@@ -0,0 +1,123 @@
...
+        channelId: ChannelId,
+        forkId: ForkId
+    ) {
+        // Check if we already responded for this forkId
+        if (
+            this.service.hasAcknowledgedDisputedFork(
+                this.senderTransport,
+                forkId
+            )
+        ) {
```
lukastanisic99: why are we performing the same check again

src/rpc/services/stateProof/StateProofService.ts
```diff
@@ -0,0 +1,332 @@
...
+    public requestStateProof(
+        transport: ATransport,
+        channelId: ChannelId,
+        peerForkId: ForkId,
+        blockHeight: BlockHeight
+    ) {
+        console.log(
+            `Requesting peer to prove their state at block height ${blockHeight}`
+        );
+        const time = Date.now();
```
lukastanisic99: Clock..

src/rpc/services/stateProof/StateProofService.ts
```diff
@@ -0,0 +1,332 @@
...
+import { StateProofStruct } from "@typechain-types/contracts/V1/types/ProofTypes";
+
+export interface DisputeWindowVerification {
+    disputeConfirmations: DisputeConfirmationStruct[];
+    forkId: Hash;
+    latestStateSnapshot: StateSnapshotStruct;
+    latestEncodedStateMachineState: Bytes;
+    joinChannelBlocksAppliedInReduce: JoinChannelBlockStruct[];
+    reducedForkId: Hash;
+}
```
lukastanisic99: This entire service feels like a poor copy-paste of spectate

I'd expect to see stuff resued, so I don't have to go over line by line and think `was it copy-pasted correctly` or `if some logic changes in spectate, don't forget to copy-paste this here too`

src/rpc/services/isForkDisputedService/IsForkDisputedService.ts
```diff
@@ -0,0 +1,98 @@
...
+                        transportsToDisconnect.push(transport);
+                    }
+                }
+
+                // Disconnect from peers that haven't acknowledged
+                for (const transport of transportsToDisconnect) {
+                    console.log(
+                        `Peer did not acknowledge disputed fork ${forkId}, disconnecting`
+                    );
+                    this.p2pManager.disconnectConnection(transport);
```
lukastanisic99: disconnect and blacklist so they can't connect again

src/rpc/services/isForkDisputedService/IsForkDisputedService.ts
```diff
@@ -0,0 +1,98 @@
...
+
+    /**
+     * Check if a transport has acknowledged a specific disputed fork
+     */
+    public didTransportAckDispute(
+        transport: ATransport,
+        forkId: ForkId
+    ): boolean {
+        return this.hasAcknowledgedDisputedFork(transport, forkId);
+    }
```
lukastanisic99, line 102: what's the point of this function just calling the function bellow

src/stateManager/StateManager.ts
```diff
@@ -1578,7 +1584,15 @@ class StateManager {
         dispute: DisputeStruct,
         timestamp: Timestamp
     ) {
-        throw new Error("TODO - Not implemented");
+        console.log(
+            `Dispute committed for channel ${dispute.input.channelId}, requesting acknowledgment from all peers`
+        );
+
+        // Request all peers to acknowledge the disputed fork
+        this.p2pManager.localRpc.isForkDisputedService.requestDisputeAcknowledgment(
```
lukastanisic99: this `onDisputeCommitted` is not called anywhere, so it's some outdated stuff - the event is collected in `EventHandler`

cyony95: in this case, I moved the trigger to the EventHandler

src/rpc/services/stateProof/StateProofService.ts
cyony95: that s what I did initially (https://github.com/peer3to/state-channels-plus/pull/199/commits/6f8fbdba1ade17b5205b1897aead16481be61e74) , but after one of the conversations we had in the meetings I understood that I should just copy the functionality (I understood wrong). I will reference spectate again, to make it nice and tidy :D

src/rpc/services/isForkDisputedService/IsForkDisputedService.ts
```diff
@@ -0,0 +1,106 @@
...
+        return new IsForkDisputedRpcMethods(transport, this);
+    }
+
+    /**
+     * Request all peers to acknowledge a disputed fork
+     * This should be called when a dispute window is created on-chain
+     */
+    public requestDisputeAcknowledgment(channelId: ChannelId, forkId: ForkId) {
+        console.log(
+            `Requesting all peers to acknowledge disputed fork ${forkId}`
```
lukastanisic99, line 42: Note for later - Luke implemented a logger we're using for debugging - when testing change this to that

src/rpc/services/isForkDisputedService/IsForkDisputedService.ts
```diff
@@ -0,0 +1,106 @@
...
+        // This captures the peers we sent the request to, so we don't disconnect
+        // from peers that connect after we sent the request
+        const snapshotTransports = [...this.p2pManager.openConnections];
+
+        // Broadcast the request
+        this.remoteRpc.isForkDisputedService
+            .onDisputeAcknowledgmentRequest(channelId, forkId)
+            .broadcast();
+
+        setTimeout(
```
lukastanisic99: we have a timeoutManager now - best we use that for all setTimeouts, since it takes into account disposal and potentially canceling tasks

src/rpc/services/isForkDisputedService/IsForkDisputedRpcMethods.ts
```diff
@@ -0,0 +1,94 @@
...
+            console.log(`Fork ${forkId} is disputed on-chain, responding`);
+            const senderTransport = this.senderTransport;
+            if (!senderTransport) {
+                return;
+            }
+            return this.service.respondToDisputeAcknowledgment(
+                senderTransport,
+                channelId,
+                forkId
+            );
```
lukastanisic99, line 52: cleaner to do this:
```typescript
isDisputed = local.isForkDisputed...
if(!isDisputed)isDisputed = onChain.isForkDisputed...
if(isDisputed){
    respond()
    return
}
//else
disconnectAndBlacklist
```
That way you don't have to keep track of 2 reponse paths, just 1

src/rpc/services/isForkDisputedService/IsForkDisputedService.ts
```diff
@@ -0,0 +1,106 @@
...
+        );
+    }
+
+    public respondToDisputeAcknowledgment(
+        transport: ATransport,
+        channelId: ChannelId,
+        forkId: ForkId
+    ): Promise<void> | void {
+        if (this.hasAcknowledgedDisputedFork(transport, forkId)) {
+            console.log(`Already responded for fork ${forkId}, disconnecting`);
```
lukastanisic99: The console.log message is correct, but that's not what would happen :D 
`hasAcknowledgedDisputedFork` does not track `your responses` it tracks responses from `connected peers`

Peers initiate this handshake `bidirectionally` (A->B & B->A) so, e.g. `A` responds to `B`, `B` will mark `A has acknowledged` and when `B` wants to responds to `A` instead of asking `have I responed to A already?`, it checks if `A responed` which would cause this to fail on `B's` end and this handhsake would never be succesfull (unless they both respond at the same time and due to network delay these checks don't get triggered)

TLDR:
This should be `haveIAcknowledged(transport,forkId)`

src/rpc/services/isForkDisputedService/IsForkDisputedService.ts
```diff
@@ -0,0 +1,106 @@
...
+            return this.p2pManager.disconnectAndBlacklistPeer(transport);
+        }
+
+        this.acknowledgeDisputedFork(transport, forkId);
```
lukastanisic99: there should be 2 maps:
- `didPeersAcknowledge` that tracks `N-1` responses from other peers
- `didIAcknowledge` that tracks `N-1` reponses `this` peer sends out to others 

here you record that you've responed and don't modify anything related to other peers responses

src/rpc/services/isForkDisputedService/IsForkDisputedRpcMethods.ts
```diff
@@ -0,0 +1,94 @@
...
+            `Received dispute acknowledgment response for fork ${forkId}`
+        );
+
+        const senderTransport = this.senderTransport;
+        if (!senderTransport) {
+            return;
+        }
+
+        // Mark that this peer has acknowledged (from our perspective)
+        this.service.acknowledgeDisputedFork(senderTransport, forkId);
```
lukastanisic99: since `onDisputeAcknowledgmentResponse` can be freely called like any other RPC method, here we'd have to protect the endpoint and if the fork was already acknowledged disconnect - in theory an attacker can also acknowledge a fork that is not disputed, but this doesn't trigger anything else and it only prevents them early from sending blocks on disputed forks

in general if you don't want to think too much where to do these checks and disconnect, you can do them in only 2 places - where you're keeping track of responses 

```typescript
iAcknowledge(transport,fork){
    if(didIRespond(transport,fork)) disconnect
    myMap.set(transport,fork)...
}
peerAcknoledges(transport,fork){
    if(didPeerAcknowledge(transport,fork)) disconnect
    peerMap.set(transport,fork)....
}
```
This way you can just do the `happy case` without checking in multiple places

src/rpc/services/stateProof/StateProofService.ts
```diff
@@ -0,0 +1,255 @@
...
+        const stateManager = this.p2pManager.stateManager;
+        const agreementManager = stateManager.agreementManager;
+        const { disputeWindows, currentForkId } = await collectDisputeWindows(
+            stateManager,
+            channelId
+        );
+
+        if (currentForkId != forkId)
+            throw new Error(
+                "Reduce and iterate didn't derive the requested fork"
```
lukastanisic99: we don't have to check this - we don't care if it's latest or some in the past - we try and prove it, but the proof is just `stateProof` so we forward this to `getStateProof` - what has to change now is that `getStateProof` should NOT throw since a byzantine peer can request any `{forkId,blockHeight}` so we don't want our application crashing from the peer requesting junk

Imo a quick and easy fix is to just expose 1 more function in `agreementManager` `tryGetStateProof`

```typescript
// in AgreementManager
tryGetStateProof(same args): StateProof | undefined {
    try{
    return this.getStateProof(...)
    }catch(e){
        return undefined;
    }
}
```

and use this when other peers can ask us to generate stateProofs, otherwise use the normal one internally since internally it should never throw and if it throws it's a bug that we should fix

in general it shoud not be possible for a byzantine peer to cause us to throw and kill our application regardless of their input


What should be the protection here against spam that a peer can only ask us to prouce a stateProof every K seconds - where probably it makes sense to have K as the TTL of RPC messages

src/rpc/services/stateProof/StateProofService.ts
```diff
@@ -0,0 +1,255 @@
...
+import {
+    collectDisputeWindows,
+    DisputeWindowVerification
+} from "../spectate/SpectateService";
+
+export interface StateProofPayload {
+    disputeWindows: DisputeWindowVerification[];
+    forkGenesisSnapshot: StateSnapshotStruct;
+    stateProof: StateProofStruct;
+}
```
lukastanisic99: How is this payload a subset of `SyncPayload` - how can the recepient prove and validate stateProof if they're missing the data, and an honest peer invoking this is missing the data

lukastanisic99:
@daiagi again mocks failing when e2e tests work :D 
image

image

I'll think about testing in more detail next week

cyony95:
@lukastanisic99 normal tests fail but I will check those and adapt them to the new code.

lukastanisic99:
> @lukastanisic99 normal tests also fail but I will check those and adapt the to the new code.

On my machine and github CI - only 4 tests and they're related to mocks 

I pushed 1 more commit directly to this branch (merge with dispute)

daiagi:
we can remove the `stateManager` unit tests entirely, they have lost much of their value with better e2e testing now falling into place.
i still think my other PR is better version of unit tests for state manager then the current code
but, being JUST unit tests,  there are going to be mocks
if nothing is mocked, it is not a unit test anymore.

cyony95:
> we can remove the `stateManager` unit tests entirely, they have lost much of their value with better e2e testing now falling into place. i still think my other PR is better version of unit tests for state manager then the current code but, being JUST unit tests, there are going to be mocks if nothing is mocked, it is not a unit test anymore.

regarding the unit tests and mocked stuff, I agree with @daiagi. that s basically what a unit test is.

lukastanisic99:
> > we can remove the `stateManager` unit tests entirely, they have lost much of their value with better e2e testing now falling into place. i still think my other PR is better version of unit tests for state manager then the current code but, being JUST unit tests, there are going to be mocks if nothing is mocked, it is not a unit test anymore.
> 
> regarding the unit tests and mocked stuff, I agree with @daiagi. that s basically what a unit test is.

We can do the mocks for now, but I'll build some stuff later to ditch the mocks, since I think they're useless - I'll however take a look more deeply at Luke's new PR 

Will want to see how you guys mock more complex stuff that humans can't easily compute like `reduce`, the simple stuff like 'store blocks' `get blocks` is so simple that even if we didn't have tests for that it wouldn't be a big deal. Plus our mocks are working with junk data

cyony95:
> > > we can remove the `stateManager` unit tests entirely, they have lost much of their value with better e2e testing now falling into place. i still think my other PR is better version of unit tests for state manager then the current code but, being JUST unit tests, there are going to be mocks if nothing is mocked, it is not a unit test anymore.
> > 
> > 
> > regarding the unit tests and mocked stuff, I agree with @daiagi. that s basically what a unit test is.
> 
> We can do the mocks for now, but I'll build some stuff later to ditch the mocks, since I think they're useless - I'll however take a look more deeply at Luke's new PR
> 
> Will want to see how you guys mock more complex stuff that humans can't easily compute like `reduce`, the simple stuff like 'store blocks' `get blocks` is so simple that even if we didn't have tests for that it wouldn't be a big deal. Plus our mocks are working with junk data

so, according to @daiagi , there are better statemanager tests in his other PR, so shall I just remove these ones?

lukastanisic99:
> so, according to @daiagi , there are better statemanager tests in his other PR, so shall I just remove these ones?

Sure - I didn't look at his PR in detail yet (I'll do it now), but as far as I'm concerned these tests with mocks are shit

PR #200, opened 2025-10-17
close channel check && reduceOutput.latestBlock undefined

cyony95:
closes https://trello.com/c/TNnHsjgO/227-update-statesnapshot-0-participants-remaining-close-channel
closes https://trello.com/c/IcfQWybk/223-take-into-account-that-reduceoutputlatestblock-may-not-exist-in-the-case-of-a-genesis-state-relevant-for-fetching-data

### DIFF
--- a/contracts/V1/StateChannelDiamondProxy/StateSnapshotFacet.sol
+++ b/contracts/V1/StateChannelDiamondProxy/StateSnapshotFacet.sol
@@ -79,7 +79,6 @@ contract StateSnapshotFacet is StateChannelCommon {
 
         // Check if channel should be closed (0 participants remaining)
         if (newSnapshot.snapshotData.participants.length == 0) {
-            emit ChannelClosed(channelId);
             // Clear storage when channel is closed (0 participants)
             _clearStorage(channelId, newSnapshot.snapshotData.latestJoinChannelBlockHash);
             // Clear the state snapshot
--- a/contracts/V1/StateChannelDiamondProxy/StateSnapshotFacet.sol
+++ b/contracts/V1/StateChannelDiamondProxy/StateSnapshotFacet.sol
@@ -82,6 +82,9 @@ contract StateSnapshotFacet is StateChannelCommon {
             emit ChannelClosed(channelId);
             // Clear storage when channel is closed (0 participants)
             _clearStorage(channelId, newSnapshot.snapshotData.latestJoinChannelBlockHash);
+            // Clear the state snapshot
+            delete stateSnapshots[channelId];
+            // TODO! send all remaining funds to the treasury
         }
 
         //check if last fork -> clearStorage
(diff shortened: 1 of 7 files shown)


lukastanisic99:
Solidity implementation also needed

lukastanisic99:
Solidity

src/agreementManager/AgreementManager.ts
```diff
@@ -368,14 +368,29 @@ class AgreementManager {
...
         // reducedOutput latestStateSnapshot
-        const reducedLatestStateSnapshot =
-            this.storage.stateSnapshots.getStateSnapshotByHash(
-                reducedOutput.latestBlock.stateSnapshotHash // TODO - latestBlock may not exist -> genesis
-            );
-        if (!reducedLatestStateSnapshot)
-            throw new Error(
-                "Missing latestStateSnapshot for reducedOutput in storage for syncing"
+        let reducedLatestStateSnapshot: StateSnapshot;
+        if (!reducedOutput.latestBlock) {
```
lukastanisic99: I think ethers will return an object with all `0s` so you'll probably have to add ` || reducedOutput.latestBlock.forkId == bytes32(0)` (think ethers has a macro defined for hashZero or something zero

contracts/V1/StateChannelManagerEvents.sol
```diff
@@ -43,6 +43,8 @@ interface StateChannelManagerEvents {
 
     event ChannelStorageCleared(bytes32 indexed channelId, bytes32 latestJoinChannelBlockHash);
 
+    event ChannelClosed(bytes32 indexed channelId);
```
lukastanisic99: This event is being emitted, but never collected on the client. I think we don't even need it, since people will leave when they observe they left or as you did when a new snapshot updates and they register 0 participants (they left) - so we can delete this explicit event emission unless we find a good reason to have it later

src/agreementManager/AgreementManager.ts
cyony95: @lukastanisic99 you are entirely correct, the object will be interpreted as truthy in this case, I will add the check. Nice catch!

daiagi (changes requested):
compilation error:

https://github.com/peer3to/state-channels-plus/actions/runs/18650582138/job/53167437877?pr=200#step:6:153

contracts/V1/StateChannelDiamondProxy/StateSnapshotFacet.sol
```diff
@@ -77,6 +77,16 @@ contract StateSnapshotFacet is StateChannelCommon {
         // Update the state snapshot
         stateSnapshots[channelId] = newSnapshot;
 
+        // Check if channel should be closed (0 participants remaining)
+        if (newSnapshot.snapshotData.participants.length == 0) {
+            emit ChannelClosed(channelId);
```
lukastanisic99: This shouldn't compile since this is still here while the definition was removed :D

cyony95: vs code search played me :)))

lukastanisic99:
rebase/merge dispute - that's why tests are failing probably

daiagi:
the failing test will pass after merge of #203

lukastanisic99 (approved):

PR #202, opened 2025-10-19
partial fix of Timeout dispute creation

daiagi:
I wrote a test to assert hat timeout dispute is being created
the test revealed quite a few bug in the code path

with this PR, the test does not pass yet, but quite a few of the bugs fixed

the current bug is from
`computeDisputeOutputSnapshotDatastaticCall`
in `DisputeManager`
```
Error: LocalDiamond call failed: Error: EVM execution failed: revert
    at LocalDiamondSigner.call (/home/luke/Documents/peer3/state-channels-plus/src/evm/LocalDiamondSigner.ts:56:19)
    at async staticCallResult (/home/luke/Documents/peer3/state-channels-plus/node_modules/ethers/src.ts/contract/contract.ts:337:22)
    at async AsyncFunction.staticCall (/home/luke/Documents/peer3/state-channels-plus/node_modules/ethers/src.ts/contract/contract.ts:303:24)
    at async DisputeManager.createDispute (/home/luke/Documents/peer3/state-channels-plus/src/disputeManager/DisputeManager.ts:185:13)
    at async StateManager.createTimeOutDispute (/home/luke/Documents/peer3/state-channels-plus/src/stateManager/StateManager.ts:1251:13)
    at async StateManager.tryTimeoutParticipant (/home/luke/Documents/peer3/state-channels-plus/src/stateManager/StateManager.ts:1150:13)
    at async Timeout._onTimeout (/home/luke/Documents/peer3/state-channels-plus/src/utils/TimeoutManager.ts:27:21)
```

@lukastanisic99  something is wrong with `computeDisputeOutputSnapshotData`
i put debug console logs, and even the first line of it is not reaced
the revert is happening before, the function code is nevr entered into
I'm guessing it is something about the static call

### DIFF
--- a/src/disputeManager/DisputeManager.ts
+++ b/src/disputeManager/DisputeManager.ts
@@ -147,7 +147,7 @@ class DisputeManager {
                     timeoutOptions.previousBlockProducerPostedCalldata || false,
                 participantSignatureOnPreviousBlock:
                     (timeoutOptions.participantSignatureOnPreviousBlock as Bytes) ||
-                    ""
+                    "0x"
             };
         } else {
             timeoutStruct = this.getEmptyTimeoutStruct();
@@ -215,7 +215,9 @@ class DisputeManager {
             signatures: []
         };
 
-        this.stateChannelManagerContract.uploadDispute(disputeConfirmation);
+        await this.stateChannelManagerContract.uploadDispute(
+            disputeConfirmation
+        );
 
         this.p2pEventHooks.onInitiatingDispute?.();
 
@@ -254,7 +256,7 @@ class DisputeManager {
             latestStateSnapshot = genesisStateSnapshot;
         } else {
             const snapshot = this.storage.stateSnapshots.getStateSnapshotByHash(
-                latestBlock.hash
+                latestBlock.stateSnapshotHash
             );
             if (!snapshot) {
                 isPartial = true;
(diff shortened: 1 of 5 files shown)


lukastanisic99:
> @lukastanisic99 something is wrong with `computeDisputeOutputSnapshotData` i put debug console logs, and even the first line of it is not reaced the revert is happening before, the function code is nevr entered into I'm guessing it is something about the static call

Yeah maybe - I never tested staticcall especially in the local diamond - maybe there's something we need to do in `ethereumjs/evm` - I just set `isStatic` to true and assumed that is equivalent to `eth_call` on an RPC node which means simulate a transaction without persisting changes, but maybe it doesn't mean that - maybe it means assume only view/pure functions can run (like the low level `staticcall` in the evm) - didn't read docs 

As for `console.log` if you're referring to `hardhat console.log` in solidity code you have that only on the hardhat node that knows how to interpret that - not if you're running in the localEVM 

Quick fix you can substitute `localDiamond` with `StateChannelmanagerContract` and run the same staticcall with ethers which will run it on the hardhat node (if you're using that) which will have the console log and everything and when that's working in the end change it to localDiamond and try it like that - if it's still not working just keep the staticCall on the StateChannelManagerContract and fuck the localDiamond (it probably means some state is not synced) and fuck the optimization and we'll run it on the RPC node

lukastanisic99 (approved):

PR #203, opened 2025-10-20
Timeout-test

daiagi:
1. Static call doesn't work, not on local diamond nor on `StateChannelManager`.
`staticCall` is not a magic bullet, it required the called function to be `view`, otherwise it does not work, if we want to use it, `computeDisputeOutputSnapshotData` need to be refactored to become `view`

2. For this reason, i have replaced the call with `encode` -> `provider.call` -> `decode` "trick"
however, even that does not work on `LocalDiamond`, I don't know why

3. Other notable change is change to `minTimeStamp` on `disputeInput.timeout` to avoid failing:
```solidity
         if (block.timestamp > dispute.input.timeout.minTimeStamp) {
                revert ErrorDisputeTimeoutNotMinTimestamp();
            }
```

the test (`"should handle timeout when author peer disconnects"`) passes now, but the contract call is not pleasing


lukastanisic99:
@daiagi Ethers V5 clearly explains what's the intended behvior - https://docs.ethers.org/v5/api/contract/contract/#contract-callStatic

V6 doesn't have so verbose docs for this so I assumed that this is the equivalent - https://docs.ethers.org/v6/api/contract/#BaseContractMethod-staticCall

daiagi:
you are somewhat right, but not in the way you would like
I was wrong that this would work with view function (i tried with `getP2pTime`)

it doesn't work in any way what so ever
not as `stateChannelManagerContract.callStatic. METHOD_NAME`
and not as `stateChannelManagerContract.METHOD_NAME.staticCall(...args)`


to be clear, by "doesn't work" i mean:

`TypeError: this.stateChannelManagerContract.computeDisputeOutputSnapshotData.staticCall is not a function`

lukastanisic99:
> you are somewhat right, but not in the way you would like I was wrong that this would work with view function (i tried with `getP2pTime`)
> 
> it doesn't work in any way what so ever not as `stateChannelManagerContract.callStatic. METHOD_NAME` and not as `stateChannelManagerContract.METHOD_NAME.staticCall(...args)`
> 
> to be clear, by "doesn't work" i mean:
> 
> `TypeError: this.stateChannelManagerContract.computeDisputeOutputSnapshotData.staticCall is not a function`

try updating ethers 

`callStatic` is v5 syntax and I'm not sure it's smart to downgrade 
`staticCall` is v6, but maybe the version we're using has a bug - try on an empty project with a simple contract - I was looking at their github of issues related to statiCall and no one is complaining `staticCall is not a function`

lukastanisic99:
@daiagi I setup a sandbox project to test `staticCall in ethers v6` take a look at tests - https://github.com/peer3to/state-channels-plus/tree/staticcall

daiagi:
> @daiagi I setup a sandbox project to test `staticCall in ethers v6` take a look at tests - https://github.com/peer3to/state-channels-plus/tree/staticcall


thank you for this, I found the issue
it was the damn decodeErrorProxy

Edit:

@lukastanisic99  it works now, with static call on stateChannelManagerContract

src/disputeManager/DisputeManager.ts
```diff
@@ -136,10 +137,22 @@ class DisputeManager {
...
+                    this.diamondStateMachine.localDiamondContract.getChainFallbackTime()
+                ])
+            ).map(Number);
+
             timeoutStruct = {
                 participant: participantToTimeout,
                 blockHeight: timeoutOptions.blockHeightToTimeout,
-                minTimeStamp: Clock.getTimeInSeconds(), // we're here since some higher level logic decided it want to timeout based on our own subjective clock
+                minTimeStamp:
+                    Clock.getTimeInSeconds() +
```
lukastanisic99, line 151: previousRelevantTimestamp + p2p + agreement + fallback 

since these checks were done in `stataManager.tryTimeoutParticipant` which calls this code, by the time this runs `Clock.getTimestamp` on its own is already >=  `previousRelevantTimestamp + p2p + agreement + fallback`

Either way I'll make this more clear and this timeout stuff will be read from storage and most things will be read from storage/evm when creating the dispute

daiagi: same as `maybePostCalldata`?
```ts
        const previousBlockOrGenesis = this.storage.getPreviousBlockOrSnapshot(
            block.coordinates
        );
        const previousRelevantTimestamp = previousBlockOrGenesis.block
            ? previousBlockOrGenesis.block.getRelevantTimestamp(block.author)
            : previousBlockOrGenesis.stateSnapshot!.timestamp;
```

lukastanisic99: yes and I just realized tryTimeout doesn't account for that (it does in `createTimeoutDispute`, but I'll try to make it more clean) - I'll see to do a PR before the call today

daiagi:
@lukastanisic99 
update:
onChannelOpened.staticCall + `isStatic = false`
works

lukastanisic99:
> @lukastanisic99 
> update:
> onChannelOpened.staticCall + `isStatic = false`
> works

So does that mean that `isStatic` in ethereumjs/evm doesn't mean "simulate a tx and discard any state changes", but "only view/pure functions can run"

daiagi:
@lukastanisic99 
We have some progress
I have found and fixed the deployment issue
and got it working **IF** using the very reduced version of `computeDisputeOutputSnapshotData`
point is, the `delgatecall` in `localDiamond` works

Now, there are other bugs, all seems to be around the `stateMachineImplemention`

#### Failures:
- getStateMachineParticipants, quite sure it failes becuse `tateMachineImplementation.setState(encodedState)` fails
- `_applySlashesToStateMachine`, I didn't debug but i would bet it fails  in `stateMachineImplementation.slashParticipant`




So, progress has been made, but the bugs keep on cropping up

lukastanisic99:
> @lukastanisic99 
> We have some progress
> I have found and fixed the deployment issue
> and got it working **IF** using the very reduced version of `computeDisputeOutputSnapshotData`
> point is, the `delgatecall` in `localDiamond` works
> 
> Now, there are other bugs, all seems to be around the `stateMachineImplemention`
> 
> #### Failures:
> - getStateMachineParticipants, quite sure it failes becuse `tateMachineImplementation.setState(encodedState)` fails
> - `_applySlashesToStateMachine`, I didn't debug but i would bet it fails  in `stateMachineImplementation.slashParticipant`
> 
> 
> 
> 
> So, progress has been made, but the bugs keep on cropping up
> 

Most likely a bug in deploying `stateMachineImplementstion`

daiagi:
>Most likely a bug in deploying stateMachineImplementstion

as far as i can see, no. it seems to be deployed correctly
the revert happens in `stateMachineImplementation.setState`
and i am so far unable to get any glimpse why

i have confirmed that the StateMachine address passed to diamond is the same adddress the deploy script deployed the state machine into
i have confirmes that that address indeed hold mathStateMachien code
and still, the call `stateMachineImplementation.setState` reverts
i even did `stateMachineImplementation.setState(stateMachineImplementation.getState())
so this is not about undecodable state, fails

clueless...

daiagi:
@lukastanisic99  I think it's time to merge this
because you new PR will have merge conflicts that we better resolve early and anyway this PR already fixes a few bugs
the remaining bugs will wait for next PR

daiagi:
@lukastanisic99  good to go? mergeable?

lukastanisic99 (approved):

src/utils/LocalDiscoveryServer.ts
```diff
@@ -68,23 +91,14 @@ export class LocalDiscoveryServer {
             const [peerPort, peerChannelId] = JSON.parse(message.toString());
             if (duplicateSet.has(peerPort)) return;
             duplicateSet.add(peerPort);
+
             if (
-                peerPort > myPort &&
+                peerPort !== myPort && // Don't connect to self
```
lukastanisic99, line 96: @daiagi shouldn't this connect double connections? A<=>B and B<=>A 

with portA>portB only from 1 peers perspective is that relation true and a single connection is maintained - or I'm not seeing something

daiagi: fixed in #204

PR #205, opened 2025-10-24
Timeout-test-2 | fix bugs

lukastanisic99:
Fix bugs that caused evm to revert 

@daiagi I kept the test case as `only` since I had a hard time running all of them, tests either timeout (go over 40s limit) or peers couldn't connect in time - didn't look into that


types/ethers-contracts/factories/Counter__factory.ts
```diff
@@ -0,0 +1,118 @@
...
+import type { Addressable } from "ethers";
+import {
+    Contract,
+    ContractFactory,
+    ContractTransactionResponse,
+    Interface
+} from "ethers";
+import type { Signer, ContractDeployTransaction, ContractRunner } from "ethers";
+import type { NonPayableOverrides } from "../common.js";
+import type { Counter, CounterInterface } from "../Counter.js";
```
lukastanisic99: @daiagi are we generating these files? Not sure where this is used and where it's coming from

daiagi: no idea where they came from, removed

PR #206, opened 2025-10-29
imporoved logging, color by peer

daiagi:
## 📋 Description
use winston for better logging, 

logs now look like 
```sh
[11:39:10][INFO][Peer 0][0xf39Fd6...][EventHandler] <message>
```

image


daiagi:
@lukastanisic99  I'm merging
there is a follow up Pr on top of this already, so you might as well review all in one place
I don't expect you will have objections to this one

PR #208, opened 2025-11-04
test simple timeout, postCallData bug fix

daiagi:
## 📋 Description

another timeout test
another bug fixed
improvements to test harness assertion capabilities

### DIFF
--- a/src/stateManager/StateManager.ts
+++ b/src/stateManager/StateManager.ts
@@ -679,17 +679,12 @@ class StateManager {
         }
     }
 
-    private async maybePostBlockOnChainByHash(blockHash: Hash): Promise<void> {
+    private async maybePostBlockOnChain(blockHash: Hash): Promise<void> {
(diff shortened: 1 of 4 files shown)


src/stateManager/StateManager.ts
```diff
@@ -679,6 +679,16 @@ class StateManager {
         }
     }
 
+    private async maybePostBlockOnChainByHash(blockHash: Hash): Promise<void> {
+        // Retrieve the latest version of the block from storage (with all collected signatures)
+        const block = this.storage.blocks.getBlock(blockHash);
+        if (!block) {
+            return;
+        }
+
```
lukastanisic99: yes this is a nice catch - imo best to remove `maybePostBlockOnChain(block)` and just use `maybePostBlockOnChain(block.hash)` so people can't accidentally use the wrong one

lukastanisic99 (approved):

PR #204, opened 2025-10-23
Timeout dispute

lukastanisic99:
Timeout detection that accounts for race condition & dispute creation (multicall when needed) - hopefully simple - everything is `f(storage)`

PR #209, opened 2025-11-05
Test: forced timeout dispute

daiagi:
- Test for forced timeout dispute
- Minor bug fix: avoid scheduling tasks when state manager is disposed
- Bug fix: `onChainTimestamp`  was not being set on the storage `Block`
    * removed the "band-aid" code that was superficially fixing this bug ([link](https://github.com/peer3to/state-channels-plus/commit/3cbe64ea1a1cd1583ccdd9b26c697c02609a5c27#r169464660))
- This and that cleaner API's on `PeerTestHarness` and of-course, added the ability to submit "junk" as block calldata


src/stateManager/StateManager.ts
```diff
@@ -1308,22 +1313,6 @@ class StateManager {
...
-                previousBlockProducerPostedCalldata = true;
-            } else {
-                previousBlockProducerPostedCalldata = (
-                    await this.diamondStateMachine.localDiamondContract.getBlockCallDataCommitment(
-                        this.channelId,
-                        forkId,
-                        previousBlock.height,
-                        previousBlock.author
-                    )
-                ).found;
```
lukastanisic99, line 1323: why is this removed

daiagi: see issue description:
> removed the "band-aid" code that was superficially fixing this bug ([link](https://github.com/peer3to/state-channels-plus/commit/3cbe64ea1a1cd1583ccdd9b26c697c02609a5c27#r169464660))

src/stateManager/ValidationService.ts
```diff
@@ -164,6 +164,14 @@ export default class ValidationService {
                 existingSignatures
             );
 
+            if (block.onChainTimestamp && !existingBlock.onChainTimestamp) {
```
lukastanisic99: Does this change fix anything? There could only be 1 on-chain timestamp and currently we don't allow overwriting existing calldata on-chain, so another one wouldn't even be emitted, but in the case that that logic change on-chain we'd expect that the off-chain client also updates the on-chain timestamp (even if set, since the one recored and persisted on-chain is the on relevant)

test/utils/SyncCoordinator.ts
```diff
@@ -1,6 +1,11 @@
 import { ForkId, Address } from "@/types/types";
 import { Logger } from "@/utils";
 
+export type WaitForPeersInSyncOptions = {
```
lukastanisic99, line 4: Is this entire file related to tests? If so, why not put it in `test/utils`

src/stateManager/ValidationService.ts
daiagi: yes, it does change
it fixes a bug: the block that comes in via the "onBlockCalldataPosted" event handler => the onchaintimstamp was not being persisted on the storage block

test/utils/SyncCoordinator.ts
daiagi: moved

src/stateManager/StateManager.ts
lukastanisic99: `tryTimeoutParticipant` should take into account the `previousBlock.relevantTimestamp` - that means that pipeline has run the block and if everything is ok it persisted the on-chain timestamp that will be taken into account for the calculation and if the timestamp is not persisted what's posted on-chain is mostly junk, `but it's still posted and persisted on-chain`, so regardless we'd have to set `previousBlockProducerPostedCalldata = true` (even though we might timeout early since it's junk), but it has to be set as `true` so we signal on-chain `I'm aware that calldata for this block exists` when making the on-chain transaction - this is a race condition guard that protects the sender of the transaction from potential race condition - so.... if junk is persisted and this is set to false the transaction will revert since there's a mismatch between the expectation and reality 

This code here just takes into account if the on-chain timestamp is persisted in storage, which means it's accepted - it doesn't cover the junk case (not in storage, but available on-chain)

src/stateManager/ValidationService.ts
lukastanisic99: I assumed that, but don't understand why it's like this - regardless of the presence or absence of an existing block it would be fine to unconditionally update the timestamp in storage if the block (block.hash) is already in storage or am I not getting something?

src/stateManager/StateManager.ts
daiagi: honestly i am not sure what you are talking about
this is a new code that i added in the commit i asked you to review in the other PR
https://github.com/peer3to/state-channels-plus/pull/204#issuecomment-3472306867
it was just masking the issue that onChainTimestamp was undefined

since i have fixed it from the root cause, i have removed this code.
not see what the issue is.
original code, that you have written is

```ts
           previousBlockProducerPostedCalldata: previousBlock
                ? Boolean(previousBlock.onChainTimestamp)
                : false,
```

src/stateManager/ValidationService.ts
daiagi: ok so basically you are saying this is sufficient:
`if (block.onChainTimestamp)`

correct?

src/stateManager/StateManager.ts
lukastanisic99: Yes I know and when I saw your `aid implementation` and made the [comment](https://github.com/peer3to/state-channels-plus/commit/3cbe64ea1a1cd1583ccdd9b26c697c02609a5c27#r169464660) I realized I forgot to do the check above. 

Test for this case:
- Produce and gossip a correct block N and post some junk as calldata for block N on-chain, 
- The next peer doesn't produce block N+1
- A timeout will be initiated for block N+1 after some time

What should happen:
- previousBlock.onChainTimestamp will be undefined (pipeline didn't accept junk) -> relevantTimestamp == block.timestamp; Temeout will initiate after p2p+agreement+fallback time (which is good). `previousBlockProducerPostedCalldata` should be set to `true` matching the expectation that the storage slot that corresponds to the previousBlock on-chain storage slot is set. The timeout goes through and is accepted. 

What will happen under: 
```tyepscript
previousBlockProducerPostedCalldata: previousBlock
                ? Boolean(previousBlock.onChainTimestamp)
                : false,
```
`previousBlockProducerPostedCalldata` will be `false` even though the storage slot on-chain is set, which will cause the transaction to revert - https://github.com/peer3to/state-channels-plus/blob/dac95097d2a321488f187540b1d36607b219eaa1/contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol#L124

src/stateManager/ValidationService.ts
lukastanisic99: yes as long as the same block is in storage already e.g. updating by hash `setOnChainTimestamp(block.hash, onChainTimestamp)`

daiagi: fixed

src/stateManager/StateManager.ts
daiagi: restored and added a test
however, this triggered a new , yet untested, buggy behavior:
- Produce and gossip a correct block N and post some junk as calldata for block N on-chain, 
    * The posting of junk calldata triggers the other peers to create a dispute
    * The uploading of this dispute fails for unknown reason
- The next peer doesn't produce block N+1
- A timeout will be initiated for block N+1 after some time
    * The failure of the dispute above, corrupts something (idk what yet) that causes the timeout dispute to not be created

daiagi: see comments in the `"should create forced timeout when peer posts junk calldata that is rejected"` test
and the error in the gh workflow [log](https://github.com/peer3to/state-channels-plus/actions/runs/19128698933/job/54664061954#step:8:59)

lukastanisic99: when I ran all tests I did see that `revert` (probably out of gas), but when I ran the single test, no error - 


image


I'll debug more

lukastanisic99: image


So from the call stack the pipeline has detected a double sign block 

image


so probably the `junk` posted is not `junk` but a conflicting block - I'll check

lukastanisic99: @daiagi `postJunkCalldataOnChain` is creating a conflicting block that meets all the requirements for a double sign (same block author, same fork, same height, different block, signed) - that's why the dispute is initiated for double sign (why it fails, that's another story - probably out of gas) - so this is expected behavior 

if you want junk data, you can just change the signature so it's invalid (not signed by the block author) and try with that 

I'll debug to see why the transaction reverted - in general this doubleSign dispute should also be its own test

lukastanisic99: Debugged typescript - everything working as designed :D will take a look at solidity - this is doing a `multicall` running a fraud proof and chaining a dispute - probably the fraud proof fails :D

daiagi:
@lukastanisic99 
- fixed `postJunkCalldataOnChain`
- fixed `retrieveSignerAddress` in `UtilityFacet` to not revert

good to go now, i think. tests are all passsing

src/stateManager/StateManager.ts
```diff
@@ -1551,6 +1555,11 @@ class StateManager {
         }
 
         // step 11 - schedule a timeout check for the next participant
+        // Don't schedule timeout tasks if StateManager is disposed
+        if (this.isDisposed) {
+            return;
```
lukastanisic99: Imo not needed, since `timeoutManager.scheduleTask` considers disposal when scheduling

src/stateManager/StateManager.ts
```diff
@@ -450,6 +450,12 @@ class StateManager {
             this.storage.blocks.getNextBlockHeight(forkId);
         let timeLost = Clock.getTimeInSeconds() - genesisTimestamp;
         timeLost = timeLost < 0 ? 0 : timeLost; // if genesisTimestamp is in the future - no time is lost
+
+        // Don't schedule timeout tasks if StateManager is disposed
+        if (this.isDisposed) {
+            return;
```
lukastanisic99: Yeah better to have this logic in `timeoutManager` so we don't have to think to add this to every place

contracts/V1/StateChannelDiamondProxy/UtilityFacet.sol
```diff
@@ -62,7 +62,12 @@ contract UtilityFacet {
         // EIP-191 - This is what actually gets signed
         bytes32 signedHash = keccak256(abi.encodePacked("\x19Ethereum Signed Message:\n32", _hash));
 
-        return ECDSA.recover(signedHash, signature);
+        // Use tryRecover to handle invalid signatures (recover reverts if signature is invalid)
+        (address recovered, ECDSA.RecoverError error,) = ECDSA.tryRecover(signedHash, signature);
+        if (error != ECDSA.RecoverError.NoError) {
```
lukastanisic99, line 71: so here we just have to do 

```solidity
if (error != ECDSA.RecoverError.NoError) return(recovered, false);
return(recovered, true);
````

For threshold verification we don't need to do this, since the caller should know before calling do they have N correct signatures or not and it's fine for threshold to revert (`recover` calls `tryRecover` and if it's not `NoError` it reverts) - here we want to continue execution in the case of failure so we can potentially slash someone

lukastanisic99 (changes requested):
Overall fine - just the few things

lukastanisic99 (approved):

lukastanisic99 (changes requested):
Seems there are some compiler errors (at least in tests) - let me check

lukastanisic99:
yep, compilation fails locally too

daiagi:
oops
fixed

lukastanisic99 (approved):

PR #211, opened 2025-11-08
RPC - prove state & isForkDisputed - 2

lukastanisic99:
@cyony95 take a look


cyony95:
> @cyony95 take a look

@lukastanisic99 I see you added the logger and used the timeoutManager, that makes sense.  But why don't we need stateProof RPC functionality anymore?

lukastanisic99:
> > @cyony95 take a look
> 
> @lukastanisic99 I see you added the logger and used the timeoutManager, that makes sense.  But why don't we need stateProof RPC functionality anymore? 

Spectate is fully reused for that - you see the optional arguments it takes? 95% of the code is the same and just a few lines vary depending how you call the service

cyony95:
> > > @cyony95 take a look
> > 
> > 
> > @lukastanisic99 I see you added the logger and used the timeoutManager, that makes sense.  But why don't we need stateProof RPC functionality anymore?
> 
> Spectate is fully reused for that - you see the optional arguments it takes? 95% of the code is the same and just a few lines vary depending how you call the service

ahh yes, you just get the state at whatever height you want now, not just the latest. that's clean as hell. why didn't I think of that :(

cyony95 (approved):

PR #212, opened 2025-11-09
Cleanup-tests

daiagi:
- remove dead test code
- partial migration of `StateManager` unit tests to a cleaner, more managable form (hope you like it @lukastanisic99 )
- bug fix in `StateManager.prepareUpdateStateSnapshotFork`


lukastanisic99:
> * remove dead test code
> * partial migration of `StateManager` unit tests to a cleaner, more manageable form (hope you like it @lukastanisic99 )
> * bug fix in `StateManager.prepareUpdateStateSnapshotFork`

I didn't go over it too deep, but again feels `mock-ish` with more implementations that need to be managed 

I was thinking for unit tests:
- We go bottom up, with components that have the least amount of dependencies and once tested use them as 'facts'/'unconditional truth' in higher components (like we use other libraries/modules e.g. ethers, evm..) till we get to e2e/system testing to prove the system as a whole
- When things break, we dig deeper to understand did we miss a test case in the lower level component or the current logic and adjust
- I don't like mocks since they're implementations that need to be maintained and I don't think that substituting internal components with their mocked versions is even correct (the mocked versions take in data that the normal component would error - e.g. hashes that are not 256bit representation of hashes, where ethers would be the first to panic on encoding) 
-  I like the harness framework for e2e testing since it uses all correct system components and just injects `spies` to observe internal state as things are executing/progressing. I like the TRAP style (proxies) debugging which doesn't modify the system, but allows you to observe what's going on and assert if needed. 
- So I'd much rather we write tests like that where we observe what has happened, what was triggered, what state was derived and persisted by observing and asserting on correct components. 

- I'd much rather see we extend harness or another framework that allows us to inject `different inputs` into our `correct system components` so we can observe how they act under different cases - the logic what we're doing for e2e but for unit tests too. 

- We should have factories for different versions of the system state (deterministic), load that state easily to observe what the component does on that state given inputs that we change. 

E.g. we're testing state manager: we load a state with 3 blocks already processed and on that state we run different tests and observe:
1) correct block - does it trigger success and other stuff 
2) incorrect blocks - does it trigger disconnect (without thinking what will P2pManager do) 
3) malicious blocks - does it trigger a dispute (without thinking what does DisputeManager do)

So later when we test DisputeManager we don't think about stateManager we just test different inputs in isolation - all using correct system components 

The only tricky part is the state factory where we have to load correct state at the beginning - for stateless components that's not the case 

Most of this is logical and nothing new, but I think this reduces the decision tree to a finite number of definite steps, since there's 1 happy case or a few and we just have to deviate from those happy cases in N ways

E.g. 
We need correct system snapshot for edge cases 0,1,2,3 participants/blocks... that's in most cases all since anything that applies to 3 would apply to general N

Then when we're testing different inputs we run those inputs on different starting states and observe what's going on and assert - e.g. (load state 0 + run input), (load state 1 + run input)... 

So we just need a nice framework for injecting starting state and make it modular like lego blocks so we can re use it

TLDR - I like spies,proxies, trap style testing that allows all correct components to actually run and don't like mocks since they don't depict the `actual` implementation and need to be maintained 

Let's discuss this on our next call @daiagi

daiagi:
I'll answer in short:
These here are unit tests, testing state manager in isolation, so complicated dependencies like scm proxy contract needs to be mocked.
That's really all there is to it

And yes, since it's just unit tests they have limited value (although I did find one bug)
Harness is for e2e

daiagi:
@lukastanisic99  ok to merge then?

lukastanisic99 (approved):

PR #213, opened 2025-11-12
Test-double sign

daiagi:
## 📋 Description

- Fix the revert with no reason bug
- Fix other various bugs
- test for double sign dispute - passing

### DIFF
--- a/src/stateManager/StateManager.ts
+++ b/src/stateManager/StateManager.ts
@@ -444,6 +444,9 @@ class StateManager {
         await this.diamondStateMachine.setState(encodedState);
         // Update the forkId to the new fork
         this.forkId = forkId;
+
+        this.setStatus(Status.PARTICIPATING);
+
         const nextToWrite = await this.diamondStateMachine.getNextToWrite();
         this.p2pEventHooks.onTurn?.(nextToWrite);
         const nextTransactionCnt =
--- a/src/stateManager/StateManager.ts
+++ b/src/stateManager/StateManager.ts
@@ -445,7 +445,11 @@ class StateManager {
         // Update the forkId to the new fork
         this.forkId = forkId;
 
-        this.setStatus(Status.PARTICIPATING);
+        const participants = await this.diamondStateMachine.getParticipants();
+        const isParticipant = participants.includes(this.signerAddress);
+        if (isParticipant) {
+            this.setStatus(Status.PARTICIPATING);
+        }
 
         const nextToWrite = await this.diamondStateMachine.getNextToWrite();
         this.p2pEventHooks.onTurn?.(nextToWrite);
--- a/src/types/disputes.ts
+++ b/src/types/disputes.ts
@@ -14,11 +14,11 @@ import { BytesLike } from "ethers";
 
 export enum FraudProofType {
     // Block related fraud proofs
-    BlockDoubleSign = 100,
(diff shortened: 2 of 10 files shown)


src/stateManager/StateManager.ts
```diff
@@ -444,6 +444,9 @@ class StateManager {
         await this.diamondStateMachine.setState(encodedState);
         // Update the forkId to the new fork
         this.forkId = forkId;
+
+        this.setStatus(Status.PARTICIPATING);
```
lukastanisic99: should check are we actually a participant

src/types/disputes.ts
```diff
@@ -14,11 +14,11 @@ import { BytesLike } from "ethers";
 
 export enum FraudProofType {
     // Block related fraud proofs
-    BlockDoubleSign = 100,
```
lukastanisic99, line 17: Where did this 100 come from? What was the error? Solidity doing the accounting from 0 while typescript doing from 100?

src/utils/Codec.ts
```diff
@@ -67,19 +67,19 @@ type StructType =
...
+    OpenChannel = "OpenChannel",
+    BlockConfirmation = "BlockConfirmation",
+    Transaction = "Transaction",
+    Dispute = "Dispute",
+    StateSnapshot = "StateSnapshot",
+    SnapshotData = "SnapshotData",
+    JoinChannelBlock = "JoinChannelBlock",
+    ExitChannelBlock = "ExitChannelBlock",
+    ExitChannel = "ExitChannel",
+    DisputeAuditingData = "DisputeAuditingData"
```
lukastanisic99, line 82: I assume the `strings` are added for debugging purposases and are not the cause of the error / don't make a logical difference?

lukastanisic99 (approved):

src/utils/Codec.ts
daiagi: Let's have a call. The 100 was the source of the bug and the strings are part of the solution

src/stateManager/StateManager.ts

src/types/disputes.ts
daiagi: >Solidity doing the accounting from 0 while typescript doing from 100?

yes, essentially
https://github.com/peer3to/state-channels-plus/blob/fb8020d66f6784a7e17f90e8f1cf84378817873b/contracts/V1/types/ProofTypes.sol#L30-L34

And in ts, we have FraudProofService, which created a proof with enum based in 100, which solidity did not recognize.

Now, why was there a 100 there to begin with?
i did that when i made a type union map for codec https://github.com/peer3to/state-channels-plus/blob/fb8020d66f6784a7e17f90e8f1cf84378817873b/src/utils/Codec.ts#L87

there was an ambiguity between the two enums,  since TS is only strongly typed in compile time.
i solved by basing one of the enums from a 100.

so now instead, i made the `Type` enum be a string enum and left the enum that mirror the sol enum alone

 this is also the reason i explicitly gave numbers to `FraudProofType`, to make it less fragile
 
 
 I acknowledge this is also a band-aid crappy solution
 the sol enums values are not present not in the typegen nor in the artifacts
 
 
 ```json
           "components": [
            {
              "internalType": "enum FraudProofType",
              "name": "proofType",
              "type": "uint8"
            },
 ```
 I'll explore alternatives and propose a PR

PR #215, opened 2025-11-12
Parallel-e2e-tests

daiagi:
## 📋 Description

run E2E tests in parallel
it mostly works, but a bit more fragile then the sequential execution
consider it a first iteration

on my machine:
- `yarn test:e2e` => ~44 seconds
- `yarn test:e2e:parallel` => ~23 seconds


lukastanisic99 (approved):

package.json
```diff
@@ -66,7 +67,7 @@
     "yarn": ">=1.15"
   },
   "scripts": {
-    "compile": "yarn hardhat clean && yarn hardhat compile --force && yarn hardhat typechain && yarn generate-enums",
+    "compile": "yarn hardhat compile && yarn hardhat typechain && yarn generate-enums",
```
lukastanisic99: why remove this

daiagi: Really it is not needed. But I'm trying out to reduce time on gh workflow by caching compilation artifacts

PR #217, opened 2025-11-13
auto-gen ts counterparts for sol enums

daiagi:
- auto generation of TS counterparts for solidy enums

- prevents stupid bugs
https://github.com/peer3to/state-channels-plus/pull/213#discussion_r2521336634

### DIFF
--- a/src/types/sol-enums.ts
+++ b/src/types/sol-enums.ts
@@ -1,14 +1,14 @@
 // Auto-generated from Solidity contracts. Do not edit manually.
 
 export enum FraudProofType {
-    BlockDoubleSign,
+    BlockDoubleSign = 100,
     BlockInvalidStateTransition,
     WrongGenesis,
     InvalidTimestamp
 }
 
 export enum DisputeFraudProofType {
-    DisputeNotLatestState,
+    DisputeNotLatestState = 200,
     DisputeInvalidOutputState,
     DisputeInvalidStateProofWithoutAuditingDataIntegrityVerified,
     DisputeInvalidStateProofWithAuditingDataIntegrityVerified,
@@ -22,3 +22,8 @@ export enum DisputeFraudProofType {
     TimeoutParticipantNotNext,
     TimeoutTooEarly
 }
+
+export const toSolidityFraudProofType = (value: FraudProofType) => value % 100;
+
+export const toSolidityDisputeFraudProofType = (value: DisputeFraudProofType) =>
+    value % 200;
--- a/src/types/sol-enums.ts
+++ b/src/types/sol-enums.ts
@@ -1,0 +1,24 @@
+// Auto-generated from Solidity contracts. Do not edit manually.
+
+export enum FraudProofType {
+    BlockDoubleSign = 0,
+    BlockInvalidStateTransition = 1,
+    WrongGenesis = 2,
+    InvalidTimestamp = 3
+}
+
+export enum DisputeFraudProofType {
+    DisputeNotLatestState = 0,
+    DisputeInvalidOutputState = 1,
+    DisputeInvalidStateProofWithoutAuditingDataIntegrityVerified = 2,
+    DisputeInvalidStateProofWithAuditingDataIntegrityVerified = 3,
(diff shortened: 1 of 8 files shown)


src/types/sol-enums.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export enum FraudProofType {
+    BlockDoubleSign = 0,
+    BlockInvalidStateTransition = 1,
+    WrongGenesis = 2,
+    InvalidTimestamp = 3
+}
+
+export enum DisputeFraudProofType {
+    DisputeNotLatestState = 0,
```
lukastanisic99: How does this solve the problem if you're doing a type union? At runtime they both start at 0 and don't they implicitly start from 0, so why explicitly state 0,1,2..? 

This is however nice to not have to maintain double implementations sol/ts for these enums - not sure why typechain doesn't generate them

daiagi: 1.Because the  enum `Type` in `Codec` is string enum
but the `Type` enum is Ts internal, so it doesn't interfere with Sol

2. why explicit? because this can create unexpected bugs, so explicit is better then implicit. 

3. >not sure why typechain doesn't generate them

yes, that would be nice if it had

lukastanisic99: > 1.Because the enum `Type` in `Codec` is string enum but the `Type` enum is Ts internal, so it doesn't interfere with Sol

Yes, but it doesn't account for DisputeFraudProofs so eventually when I populate the `dispute confirmation pipeline` the union will expand to ` | DisputeFraudProofType`

> 2. why explicit? because this can create unexpected bugs, so explicit is better then implicit.

Sure, but the explicit is the same as the implicit - you'll have the same bugs

daiagi: and?
is the code not better with the pr then without it?
 yes, it doesn't solve the future problem of adding another sol enum to the union, this will cause a bug, so don't do it, another solution should be found
 yes, being explicit does not magically solve bugs, it is just visually showing the info that is hidden.  what is the point? you want it be implicit?

daiagi: > Sure, but the explicit is the same as the implicit - you'll have the same bugs

done, made it implicit

lukastanisic99: I don't care that it's explicit - I'm just saying this doesn't solve our problem. Yes it's better in a sense that we don't have to copy paste the enum into typescript, but if we know we'll have a problem more or less `now` (I should have populated that code already) we should account for it

daiagi: ok. i'll try for a better solution

daiagi: have another look

src/utils/Codec.ts
```diff
@@ -109,14 +114,66 @@ export class Codec {
...
+        }
+        throw new Error(`No ethers type mapping found for ${type}`);
+    }
+
+    private static isFraudProofStruct(struct: StructType): boolean {
+        return (
+            ("block1" in struct && "block2" in struct) || // BlockDoubleSignProof
+            ("invalidBlock" in struct &&
+                "previousBlockStateSnapshot" in struct) || // BlockInvalidStateTransitionProof
+            ("invalidBlock" in struct && "previousStateSnapshot" in struct) // InvalidTimestampProof or WrongGenesisProof
```
lukastanisic99: @daiagi  I don't like this - more hardcoded stuff that we need to remember to change - over engineering 

Let's do this: 
- generate the enums from solidity like now (good) - that way we have the same order and only need to maintain solidity
- Start each new enum family as you did with an offset e.g. `Types = 0,1..` `FraudProofTypes = 100,101..` `DisputeFraudProofTypes = 200, 201...`
- When assigning to a solidity type like in `FraudProofService` do a `modular devision` (e.g. `type = FraudProofType.SomeType % 100`) 

That way for solidity they will all start from 0, but in typescript remain different and our encoding/decoding can be simple as before without hardcoding and over engineering stuff

daiagi: Done.
your idea is indeed more elegant

lukastanisic99 (approved):

PR #224, opened 2025-11-17
Spectate extra features - wip

cyony95:
- added spectator logic to isSnapshotInPast(), so if the peer is spectating and has no blocks, the snapshot is allowed. before, spectate would call fetchAndPersistOnChainSnapshot() which in turn calls onStateSnapshotUpdated() which itself calls isSnapshotInPast(), even though the snapshot is from the chain.
- added getStatus() getter

### DIFF
--- a/src/eventHandlers/EventHandler.ts
+++ b/src/eventHandlers/EventHandler.ts
@@ -24,7 +24,6 @@ import { Codec, hash, Logger, Type } from "@/utils";
 import { isEqual } from "lodash";
 import { ZeroHash } from "ethers";
 import CalldataCommittedStrategy from "@/stateManager/validationStrategy/CalldataCommittedStrategy";
-import { Status } from "@/types";
 
 export class EventHandler {
     private logger: Logger;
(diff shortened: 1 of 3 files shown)


cyony95:
@lukastanisic99 @daiagi another solution to this would be to add a parameter to skip validation if the snapshot is on-chain? let me know what you think.

src/eventHandlers/EventHandler.ts
```diff
@@ -414,6 +415,14 @@ export class EventHandler {
         channelId: ChannelId,
         incomingSnapshot: StateSnapshotStruct
     ): Promise<boolean> {
+        // For spectators with no local state, allow the first snapshot
```
lukastanisic99: Imo we can probably remove this `isSnapshotInPast` - I had some ideas here, but in reality if the chain accepted a snapshot we have to accept it unconditionally. I'm looking at the implementation bellow and some things are mixed up anyway - what we're really interested in is having the latest snapshot, but I think we query for the latest snapshot regardless everywhere where we use it, so all of this is more of an optimization. 

On chain the chain logic should enforce correct snapshot submission in ascending order and if events are also emitted and collected in total order (the same order as they're emitted) all of this is useless.

TLDR: for now remove `isSnapshotInPast` entirely and add a `//TODO - make sure snapshots are in ascending order if events can be collected in random order - e.g we have the latest one always`

daiagi (approved):

PR #225, opened 2025-11-19
Dispute confirmation pipeline

lukastanisic99:
Hopefully this is all of the dispute stuff - some minor things remain (which I'll have to add at some point - some race condition stuff mostly which I left as a `todo` in the code and using `multicall` in some places) 

`This still needs to be tested!`

GPT5.1 codex in agent mode, generated the ethers types in the end, since I was lazy to do it on my own, at to me it looks good and the AI even added some missing fraud proof types which we left out before - overall we should utilize it more - it saved me time on tasks I don't want to do and it's observing what I'm doing and suggesting small fixes when I make logical mistakes, that would take us some time to find and debug - not saying there aren't any bugs here we still need to test.

### DIFF
--- a/src/stateManager/utils/DisputeFraudProofService.ts
+++ b/src/stateManager/utils/DisputeFraudProofService.ts
@@ -1,3 +1,4 @@
+import { Block } from "@/models";
 import Storage from "@/storage";
 import {
     DisputeFraudProofType,
--- a/src/stateManager/utils/DisputeFraudProofService.ts
+++ b/src/stateManager/utils/DisputeFraudProofService.ts
@@ -1,4 +1,3 @@
-import { Block } from "@/models";
 import Storage from "@/storage";
 import {
     DisputeFraudProofType,
--- a/src/disputeManager/DisputeManager.ts
+++ b/src/disputeManager/DisputeManager.ts
@@ -73,12 +73,8 @@ class DisputeManager {
         try {
             await this.mutex.lock();
             if (this.storage.disputes.didIDispute(forkId)) return;
-            const {
-                dispute,
-                disputeConfirmation,
-                auditingData,
-                fraudProofsToApply
-            } = await this.constructDispute(forkId);
+            const { disputeConfirmation, auditingData, fraudProofsToApply } =
+                await this.constructDispute(forkId);
 
             const pendingParticipants =
                 await this.stateChannelManagerContract.getPendingParticipants(
--- a/src/disputeManager/DisputeManager.ts
+++ b/src/disputeManager/DisputeManager.ts
@@ -73,8 +73,12 @@ class DisputeManager {
         try {
             await this.mutex.lock();
             if (this.storage.disputes.didIDispute(forkId)) return;
-            const { disputeConfirmation, auditingData, fraudProofsToApply } =
-                await this.constructDispute(forkId);
+            const {
+                dispute,
+                disputeConfirmation,
+                auditingData,
+                fraudProofsToApply
+            } = await this.constructDispute(forkId);
 
             const pendingParticipants =
                 await this.stateChannelManagerContract.getPendingParticipants(
--- a/src/stateManager/DisputeValidationService.ts
+++ b/src/stateManager/DisputeValidationService.ts
@@ -457,4 +457,17 @@ export default class DisputeValidationService {
         // if we're here - it's all good
         return true;
     }
+
+    private async isTimeoutTooEarly(
+        dispute: DisputeStruct,
+        disputeAuditingData: DisputeAuditingDataStruct
+    ): Promise<boolean> {
+        const [hasBlock, latestBlock] =
+            await this.diamondStateMachine.localDiamondContract.getLatestBlockFromStateProof(
+                dispute.input.stateProof
+            );
+        const previousTimestamp = 0;
+
+        return true;
+    }
 }
--- a/src/stateManager/DisputeValidationService.ts
+++ b/src/stateManager/DisputeValidationService.ts
@@ -457,17 +457,4 @@ export default class DisputeValidationService {
         // if we're here - it's all good
         return true;
     }
-
-    private async isTimeoutTooEarly(
-        dispute: DisputeStruct,
-        disputeAuditingData: DisputeAuditingDataStruct
-    ): Promise<boolean> {
-        const [hasBlock, latestBlock] =
-            await this.diamondStateMachine.localDiamondContract.getLatestBlockFromStateProof(
-                dispute.input.stateProof
-            );
-        const previousTimestamp = 0;
-
-        return true;
-    }
 }
--- a/src/stateManager/validationStrategy/DisputeValidationStrategy.ts
+++ b/src/stateManager/validationStrategy/DisputeValidationStrategy.ts
@@ -1,14 +1,13 @@
 import { Block } from "@/models";
 import { BlockValidationResult } from "@/types";
-import {
-    BlockConfirmationStruct,
-    DisputeStruct
-} from "@typechain-types/contracts/V1/types/DisputeTypes";
+import { DisputeStruct } from "@typechain-types/contracts/V1/types/DisputeTypes";
+
 import AValidationStrategy from "./AValidationStrategy";
 import FraudProofService from "../utils/FraudProofService";
 import Storage from "@/storage";
 import ATransport from "@/transport/ATransport";
 import DisputeFraudProofService from "../utils/DisputeFraudProofService";
+import { BlockConfirmationStruct } from "@typechain-types/contracts/V1/types/DataTypes";
 
 export default class DisputeValidationStrategy extends AValidationStrategy {
     readonly fraudProofService: FraudProofService;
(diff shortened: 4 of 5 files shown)


lukastanisic99:
@daiagi tests are still unstable - sometimes they work and sometimes still bugs like this ^ ` failed to synchronize in 8s`

src/utils/Codec.ts
evainfeld, line 1: I wonder what the advantage is of splitting dispute cases this much? Instead of all of these combinations of conditions having their own dispute cases, it may be easier to give them additional variables for data integrity / exit channel etc ?

lukastanisic99: Before it was all one large expensive function which we called `auditing` - most of this is partitioning that into smaller steps so we don't have to run everything on-chain, just run the failing step. 

If the `dispute` data structure on its own had `all the data`, many of these steps wouldn't be needed, but on-chain that would be very expensive in terms of gas, so we split it - the `dispute` to a large degree (apart from `stateProof`, which kinda has to be there) only has commitments and `auditingData` in most cases isn't uploaded to the chain, since any `participant` can reconstruct it in full or prove it's malicious - most of the dispute fraud proofs handle a variation of that

`AuditingData` mostly has everything you need to verify `stateProof` and it provides the `latestState` to which the STF is applied, but apart from that it has `ExitChannels`, which in the general case you can think of as `outbound messages` as discussed yesterday (messages from L2 to L1) - so that's all the data you need to make available to the chain so it can verify and transition the state (by force if needed) 

You can combine the `disputeFraudProofs` back up into larger execution steps, but that would be more costly to the end user - ideally we find a way to split it into even smaller steps or upload even less data, while guaranteeing security and liveness

evainfeld: That 's fair enough - it does seem like there is a lot of granularity here though and there may be some middle ground? DisputeInvalidStateProofWithoutAuditingDataIntegrityVerifiedProofEthersType vs DisputeInvalidStateProofWithAuditingDataIntegrityVerifiedProofEthersType could be a variable inside one function. 

I'm nitpicking because you asked for review though, the code looks fine to me. ;)

lukastanisic99: > That 's fair enough - it does seem like there is a lot of granularity here though and there may be some middle ground? DisputeInvalidStateProofWithoutAuditingDataIntegrityVerifiedProofEthersType vs DisputeInvalidStateProofWithAuditingDataIntegrityVerifiedProofEthersType could be a variable inside one function.
> 
> I'm nitpicking because you asked for review though, the code looks fine to me. ;)

Actually I'm glad you mentioned the 2 above since those 2 have a deeper meaning and are structured like that intentionally since:

1) The one `with` the `data integrity verified` means that we have provided the `full data (auditingData)` to which the disputer committed and any failure detected means that the disputer committed to something invalid -> objectively slash the disputer

2) The one `without` data integrity verified, means that we can `not` reconstruct `auditingData` in full, so we can't upload it in whole to be verified against the hash, so we take a more rigorous and restrictive verification path which doesn't resolve always in `true/false`, but could also `revert`. The premiss is this:
starting from some genesisState which is objective and available to everyone (it has the starting participant set), verify `milestone per milestone` in the `stateProof`, proving that the participant set unanimously reached a `threshold on the milestone` and forcing the auditor/challenger to upload a correct chunk for that milestone so that the verification can continue, otherwise the transaction reverts like it never happened. Meaning if the `stateProof` is valid and a threshold is reach, the data had to be available so it's the responsibility of the uploader to provide a correct chunk. Then we move to the next milestone and do that recursively until everything is verified. 
- If the disputer committed to junk (that's why we can't reconstruct auditing data in full), the threshold verification will fail before it's our turn to upload the underlying chunk (which we can't reconstruct), meaning the verification will result to `false` meaning that the proof is invalid and the dispute is invalid
- If the proof resolves to true, it means that all the thresholds are reached (we have that data), so the reason why we can't reconstruct it in full is that something in the `unfinalized` part of the stateProof is missing (e.g. we're on blockHeight 50, while someone created the dispute at blockHeight 51), since this is a finite number of blocks (N-1 with round robbing leader election) we plug in the unfinalized part of the stateProof (at most N-1 block) into the block confirmation pipeline for us to verify block by block and try to catchup to the state (e.g. blockHeight 51), or observe failure and point at which block the failure is detected - so the result is we either sync in full to 51 meaning we can construct the full auditing data, or we point at failure 
- the 3rd case `revert` is us trying to game the system, by not uploading a correct chunk, after finality of the chunk has been proven

The difference in 1) and 2) is that if we have the `full data` and can verify it against the hash - we remove the `revert` from 2) and treat that as `false/failure` to slash the disputer that committed to it

lukastanisic99:
@evainfeld though a little about the case `valid StateProof` but invalid `auditingData` - it should be covered, but not with `verifyStateProof`, but with this fraud proof - https://github.com/peer3to/state-channels-plus/blob/138e11f61ac6cc0379753e1ed44e7a533e3c0fdb/contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol#L154

The name is a bit verbose and at first I didn't like it since these ExitChannelBlocks that are in the name kinda don't feel right to be there, but during implementation I realized they had to be verified too, so I included the verbose name, but but but when we change that to general outbound messages (L2->L1) it makes more sense, at least to me :D

Sooo the flow is:
https://github.com/peer3to/state-channels-plus/blob/138e11f61ac6cc0379753e1ed44e7a533e3c0fdb/src/stateManager/DisputeValidationService.ts#L63
This function will try and construct `auditingData` based on `stateProof` (it doesn't verify the integrity it just constructs a blob of data that corresponds to the stateProof), so it may return `isPartial = true` meaning our local storage doesn't have the data -> we verify the stateProof (talked about this in the call) 

In the case it's `invalid` - this code runs and creates a fraud proof https://github.com/peer3to/state-channels-plus/blob/138e11f61ac6cc0379753e1ed44e7a533e3c0fdb/src/stateManager/DisputeValidationService.ts#L83

In the case the `proof is valid` - this code runs https://github.com/peer3to/state-channels-plus/blob/138e11f61ac6cc0379753e1ed44e7a533e3c0fdb/src/stateManager/DisputeValidationService.ts#L77
Which pushes the `unfinalized` blocks through the `block confirmation pipeline` for validation. The result of this is either a block is invalid and caught by the pipeline in which case fraud is detected https://github.com/peer3to/state-channels-plus/blob/138e11f61ac6cc0379753e1ed44e7a533e3c0fdb/src/stateManager/DisputeValidationService.ts#L203
In the case that the pipeline accepts all blocks, it means they're all valid and we synced to the latest state - our storage should reflect that -> when we run `again` https://github.com/peer3to/state-channels-plus/blob/138e11f61ac6cc0379753e1ed44e7a533e3c0fdb/src/stateManager/DisputeValidationService.ts#L63
It should `always` return the full `auditingData` with `isPartial = false`, so we for sure constructed `the correct data`, so in the case that participant committed to junk/invalid `auditingData` - this code should catch it https://github.com/peer3to/state-channels-plus/blob/138e11f61ac6cc0379753e1ed44e7a533e3c0fdb/src/stateManager/DisputeValidationService.ts#L124

src/stateManager/utils/DisputeFraudProofService.ts
```diff
@@ -7,4 +41,225 @@ import Storage from "@/storage";
...
+        encodedBlock: BytesLike,
+        signature: Signature
+    ): Hash {
+        const proof: DisputeNotLatestStateStruct = {
+            encodedBlock,
+            signature: signature as BytesLike
+        };
+
+        return this.storeFraudProof(dispute, {
+            type: DisputeFraudProofType.DisputeInvalidOutputState,
```
daiagi: mismatched FraudProofType
fixed in #233

src/stateManager/utils/DisputeFraudProofService.ts
```diff
@@ -7,4 +41,225 @@ import Storage from "@/storage";
...
+                auditingData
+            };
+
+        return this.storeFraudProof(dispute, {
+            type: DisputeFraudProofType.DisputeInvalidStateProofWithAuditingDataIntegrityVerified,
+            struct: proof
+        });
+    }
+
+    createDisputeIncorrectAuditingDataCommitmentWithValidStateProofAndValidExitChannelBlocks(
```
daiagi: https://en.wikipedia.org/wiki/Taumatawhakatangi%C2%ADhangakoauauotamatea%C2%ADturipukakapikimaunga%C2%ADhoronukupokaiwhen%C2%ADuakitanatahu

src/storage/BlockCalldataStorage.ts
```diff
@@ -0,0 +1,42 @@
+import { ForkId, BlockHeight, BlockCalldata, Address } from "@/types/types";
+import { Block } from "@/models";
+
+type CalldataCoordinateKey = string;
+
+export class BlockCalldataStorage {
+    // ====================================
+    // STORAGE MAPS
+    // ====================================
+    private coordinatesToBlockMap: Map<CalldataCoordinateKey, BlockCalldata>;
```
daiagi, line 10: why not hash as key?

src/types/disputes.ts
```diff
@@ -99,3 +100,66 @@ export type ReduceData = {
     encodedStateMachineState: BytesLike;
     joinChannelBlocks: JoinChannelBlockStruct[];
 };
+
```
daiagi, line 105: i see future bugs coming when something changes in the source structs

daiagi (approved):

daiagi:
this PR is now basically nothing
merging

PR #226, opened 2025-11-21
Dispute-test

daiagi:
## 📋 Description

got a basic dispute resolution test to work
quite a few bugs fixed along the way

- Use correct fork id from dispute:
`const forkId = dispute.input.genesisSnapshotDataHash;`

- Fix: No disputes were ever stored into storage => causing an error to be thrown
### Rabbit hole warning:
we had `DisputeConfirmation` storage, BUT, if i am not mistaken, this was under previous design where peers would p2p communicate disputes, like they do with blocks.
For now at least, we have moved away from this design and peers submit their disputes directly on chain, and, if the dispute is committed (chain approved) other peers observe it via the `onDisputeCommited` event
BUT, the payload from this event is `Dispute` **NOT** `DisputeConfirmaition`, nor does it make sense for it to be `DisputeConfirmation`.

So, I have removed the `DisputeConfirmation` storage, since it seems to me not relevant right now and put instead a simpler dispute storage
### END Rabbit Hole

- fix `ExitChannelEthersType`  :sleeping: 
- fix: Static calls return result object, very annoying. :angry: 
    Made proxy to "unwrap" them into our own data structures. so we don't need to fall into this bug over and over again

- simplified the `tryReduce` timeout logic in `StateManager`
a. it was too complicated i think, hard to read
b. it was not quite working, not sure what was the bug, due to problem _a_. I could not read and make sense of it
so I refactored and cleaned it up, Now it works, hope I didn't mess it up

- other "test realm" issues: warping hardhat time, fixing the test contract balance invariant

lot's of work to get the most basic dispute resolution test working :weary:


contracts/V1/examples/MathStateMachine/MathConsumerFacet.sol
```diff
@@ -18,13 +18,19 @@ contract MathConsumerFacet is AConsumerFacet {
...
+
+        // Initialize state.number to sum of all balances to satisfy balance invariant:
+        // totalDeposits == totalWithdrawals + getTotalStateBalance()
+        uint256 totalBalance = 0;
         for (uint256 i = 0; i < successfulJoinChannels.length; i++) {
             genesisState.participants[i] = successfulJoinChannels[i].participant;
             genesisState.balances[i] = successfulJoinChannels[i].balance.amount;
+            totalBalance += successfulJoinChannels[i].balance.amount;
         }
+        genesisState.number = totalBalance;
```
lukastanisic99, line 32: probably this doesn't make a difference, but the `number` field doesn't have anything to do with balances and it will change as the state transitions :D - the reason it was failing was probably because in `MathStateMachine` the function `getTotalStateBalance` was returning this number instead of summing up `balances`

What would be usefull for testing is to add 2 functions `infalte`/`steal` that will modify `balances` to break the invariant, so we can test also that 

Currently the MathStateMachine doesn't do anything with the balances, so the invariant will always hold 

The idea behind this invariant check was to prevent N colluding peers from infalting the balance within the channel, which would allow them to steal new deposits from new joining peers as they arrive

src/agreementManager/AgreementManager.ts
```diff
@@ -324,58 +323,25 @@ class AgreementManager {
...
-                ethers.getBytes(disputeHash),
-                sig as Signature
-            );
-            if (signer === participant) return true;
-        }
-
-        return false;
-    }
-
-    public async getForkDisputeConfirmations(
```
lukastanisic99, line 357: This code is not used currently and we're skipping it for now to ship faster, but I'd leave it here with DisputeConfriamtions since it was already implemented and a `DisputeConfirmation` is a super set of `Dispute` and you can easily go from `DisputeConfirmation` to `Dispute` and vice versa - solidity code accounted for finality on this and parts of the event pipeline also account for this - we just didn't finish the implementation which we'll eventually finish

src/disputeManager/DisputeManager.ts
```diff
@@ -127,6 +131,7 @@ class DisputeManager {
                 }
             }
 
+            this.storage.disputes.storeDispute(dispute);
```
lukastanisic99: This should be stored when the event is observed, since here you have no gurantee that the dispute would succeed and if you do it like this you have to pair it with `delete` in the case of an error as I did with `this.storage.disputes.storeDisputedFork(forkId, true)` and `this.storage.disputes.storeDisputedFork(forkId, false)` bellow, but I think it's simpler to just store it when you observe it since the pipeline is clean and simple

src/eventHandlers/EventHandler.ts
```diff
@@ -138,7 +137,7 @@ export class EventHandler {
         windowCreationTimestamp: Timestamp,
         disputeAuditingData?: DisputeAuditingDataStruct
     ): Promise<void> {
-        const forkId = dispute.input.disputeAuditingDataHash;
+        const forkId = dispute.input.genesisSnapshotDataHash;
```
lukastanisic99: ah yes - `gemini 3` was telling me this the other day, but aparently I didn't listen since I just read `DataHash` and though it was correct and that gemini didn't understand that the forkId is a hash 

Maybe better to refactor this into `dispute.input.forkId`?

src/eventHandlers/EventHandler.ts
```diff
@@ -196,6 +195,7 @@ export class EventHandler {
             );
 
         if (isValid) {
+            this.storage.disputes.storeDispute(dispute);
```
lukastanisic99: Yes this should be done here after validation

src/eventHandlers/EventHandler.ts
```diff
@@ -207,10 +207,11 @@ export class EventHandler {
                     forkId, // originForkId is this forkId
                     ZeroHash // resulting forkId is not relevant here
                 );
-            this.stateManager.setReductionTimeout(
-                forkId,
-                Number(potentialGenesisTimestamp)
+            const delayInSeconds = Number(potentialGenesisTimestamp);
```
lukastanisic99:
```typescript
const delayInSeconds = Number(potentialGenesisTimestamp) - Clock.getTimeInSeconds()
```

daiagi:
call notes:
- restore `DisputeConfirmation` storage
- emit `DisputeConfirmation` from the chain event
- local diamond has bug that makes the genesis timestamp Not a timestamp - something not syncing
- probably remove the teleport and create test contract that extends the SCM proxy

src/eventHandlers/EventHandler.ts
lukastanisic99: onDisputeCommitted doesn't set `evidence.lastEvidenceSubmissionTimestamp`

src/stateManager/StateManager.ts
```diff
@@ -357,15 +381,20 @@ class StateManager {
...
+        // Update local state to the reduced fork
+        this.logger.debug(
+            `Reduction complete: transitioning to fork ${reducedForkId}`
+        );
         this.setGenesisState(
             snapshotData,
             encodedStateMachineState,
             reducedForkId,
-            genesisTimestamp,
+            Clock.getTimeInSeconds(),
```
lukastanisic99: We can't do this - this has to be an objective timestamp, since it goes into the snapshot and the snapshot needs to be the same for everyone. That's why the timestamp is passed around since it originates from the chain

src/stateManager/StateManager.ts
```diff
@@ -217,116 +221,118 @@ class StateManager {
...
-                this.tryReduce(forkId, triggerTimestamp);
-            }, delayInMilliseconds);
-            this.reductionTriggerMap.set(forkId, {
-                handle: newHandle,
-                triggerTimestamp: triggerTimestamp
-            });
+            this.logger.debug(
+                `Existing timeout at ${existingHandle.triggerTimestamp} is earlier than new ${newTriggerTime}, keeping it`
+            );
+            return;
```
lukastanisic99: This isn't true - if the `newTriggerTime` > `currentTriggerTimestamp` we want to reschedule - what is happening someone submitted more evidence so the time is prolonged, so you do the reduction later

src/stateManager/StateManager.ts
```diff
@@ -217,116 +221,118 @@ class StateManager {
...
+
+        // Step 2: If local state says not ready, reschedule check
+        if (!canReduceLocally) {
+            if (timeRemaining > 0) {
+                this.logger.debug(
+                    `Rescheduling reduction check in ${timeRemaining}s`
                 );
+                return this.setReductionTimeout(forkId, timeRemaining);
             }
+            // timeRemaining is 0 but can't reduce -> local state not synced, fall through to on-chain check
```
lukastanisic99, line 289: `fall through` - I like this

src/utils/evmErrorHandler.ts
```diff
@@ -26,6 +26,41 @@ function decodeCustomError(errorData: Bytes): ErrorDescription | null {
...
+                args: [] as any,
+                signature: `${errorName}()`,
+                selector: "0x00000000", // placeholder
+                fragment: null as any // placeholder
+            };
+            return new CustomEvmError(mockErrorDescription, error);
+        }
+    }
+
+    return null;
```
lukastanisic99: what were your toughts behind this function? I see it's only used in 1 place and it does more stuff than the typical error handling we have for custom errors

lukastanisic99:
@daiagi 

So `staticCall` returns a `...StructOutput` not a `...Struct`
    
image

image

And when it's debugged it really is a proxy as we observed - not injected by us it seems
image


I've updated the staticCall tests for complex types - take a look at the `staticcall` branch for mini tests

lukastanisic99:
@daiagi tests are failing - can you take a look so I can debug later 

What's hard for me to understand is - A `...StructOutput` is a super set of  `...Struct` - it's a union - so why does it make a difference? It should be possible to use it interchangeably - my only guess it makes a different in `encoding` - so I'll take a look later in more depth

lukastanisic99:
Btw I'm not sure why I implemented `ethersResultToObjectRecursive` when this exists https://docs.ethers.org/v6/api/abi/#Result-toObject

cyony95:
> Btw I'm not sure why I implemented `ethersResultToObjectRecursive` when this exists https://docs.ethers.org/v6/api/abi/#Result-toObject

@lukastanisic99  I read that and changed in one of my branches your function to use these, and it fails even more tests than before. Looks like it's not the same thing in the end.

lukastanisic99:
> > Btw I'm not sure why I implemented `ethersResultToObjectRecursive` when this exists https://docs.ethers.org/v6/api/abi/#Result-toObject
> 
> @lukastanisic99  I read that and changed in one of my branches your function to use these, and it fails even more tests than before. Looks like it's not the same thing in the end.


Yeah I remember I was looking into this, I just don't remember did I try with the `deep` flag or not, but regardless for some reason I decided to implement this in the end..

src/utils/evmErrorHandler.ts
daiagi: it was nonsense, removed

daiagi:
@lukastanisic99 
>probably remove the teleport and create test contract that extends the SCM proxy


this is becoming a problem, since `createStandalone` deploys the local diamond internal, so in order to replace the local diamond with a test version
1. pass some bit of info to least `createStandalone` know it should deploy the test local dimaond
2. deploy in the test harness and pass an address to p2psetup


both these ideas are not great

daiagi:
> just don't remember did I try with the deep flag or not,

so the problem is that Result.toObject with deep flag  replaces arrays with objects
so 

```ts
            stateProof: {
                milestones: [],
                signedBlocks: []
            },
```


becomes 
```ts
            stateProof: {
                milestones: {},
                signedBlocks: {}
            },
```

and it's worse without the deep flag (inner structs stay Result)

no clean solution, might as well roll our own, as you already did

lukastanisic99:
> @lukastanisic99 
> >probably remove the teleport and create test contract that extends the SCM proxy
> 
> 
> this is becoming a problem, since `createStandalone` deploys the local diamond internal, so in order to replace the local diamond with a test version
> 1. pass some bit of info to least `createStandalone` know it should deploy the test local dimaond
> 2. deploy in the test harness and pass an address to p2psetup
> 
> 
> both these ideas are not great

What was the reason again for doing this? Changing the config time in the contract? Just change it in the constructor like other stuff

daiagi:
>What was the reason again for doing this? Changing the config time in the contract? Just change it in the constructor like other stuff

questions:
1. do we agree that the local diamond and the scm need to have matching time config? ( they currently don't)
2. as for the scm, yes, could pass time config in the constructor, the harness deploys it, easy peasy no issues
3. as for the local diamond, it is deployed internally by `createStandalone`, so some way of overriding the production default times must be created, that is unless, you disagree on question 1

lukastanisic99:
> >What was the reason again for doing this? Changing the config time in the contract? Just change it in the constructor like other stuff
> 
> questions:
> 1. do we agree that the local diamond and the scm need to have matching time config? ( they currently don't)
> 2. as for the scm, yes, could pass time config in the constructor, the harness deploys it, easy peasy no issues
> 3. as for the local diamond, it is deployed internally by `createStandalone`, so some way of overriding the production default times must be created, that is unless, you disagree on question 1
> 

1) yes
2) question?
3) question? Probably the answer you're looking for is, the local diamond fetches the time config from the deployed SCM when deploying

daiagi:
@lukastanisic99  ready

src/StateChannelEventListener.ts
```diff
@@ -105,15 +107,17 @@ class StateChannelEventListener {
...
                     isFinal,
                     windowCreationTimestamp
                 } = logObj.args;
 
                 this.eventHandler.onDisputeCommitted(
                     channelId,
-                    Codec.convertEthersResultToObject(dispute),
+                    (disputeConfirmation as Result).toObject(
+                        true
+                    ) as DisputeConfirmationStruct,
```
lukastanisic99: is this intentional or forgot to change back to our custom implementation?

daiagi: I changed to try, it worked. I didn't change back.

Edit:
Changed back

src/stateManager/StateManager.ts
```diff
@@ -217,116 +221,108 @@ class StateManager {
...
+    public setReductionTimeout(forkId: ForkId, triggerTimestamp: Timestamp) {
+        if (this.forkId !== forkId) return;
+
+        const existingHandle = this.reductionTriggerMap.get(forkId);
+        const now = Clock.getTimeInSeconds();
+
+        // Cancel existing timeout if present
+        if (existingHandle) {
+            this.timeoutManager.cancelTask(existingHandle.handle);
         }
```
lukastanisic99, line 236: this works if we know there can't be a race condition (e.g. timestamps always apear in ascending order), otherwise we're ineterested in keeping the largest timestamp

src/stateManager/StateManager.ts
```diff
@@ -217,116 +221,108 @@ class StateManager {
...
+            `Reduction check for fork ${forkId}: canReduce=${canReduceLocally}, timeRemaining=${timeRemaining}s`
+        );
+
+        // Step 2: If local state says not ready, reschedule check
+        if (!canReduceLocally) {
+            if (timeRemaining > 0) {
+                this.logger.debug(
+                    `Rescheduling reduction check in ${timeRemaining}s`
                 );
+                return this.setReductionTimeout(forkId, timeRemaining);
```
lukastanisic99: this should be the objective timestamp, not a delay

src/stateManager/StateManager.ts
```diff
@@ -217,116 +221,108 @@ class StateManager {
...
-        } catch (error) {
-            // this should never be the case since:
-            // 1) disputeWindows is expired - double checked on-chain
-            // 2) dispute commitments - collected on-chain -> we for sure have the correct data
-            // 3) even if someone else reduces on-chain -> they would have to reduce to the same output, so race condition is not a problem
-            this.logger.error("tryReduce failed", {
-                error: error instanceof Error ? error.message : String(error)
-            });
-            throw error;
+                return this.setReductionTimeout(forkId, remaining);
```
lukastanisic99: same

src/stateManager/StateManager.ts
```diff
@@ -217,116 +221,108 @@ class StateManager {
...
+                `Local state not synced, checking on-chain state`
+            );
         }
-        // reduce on-chain
-        const disputeConfirmations =
-            await this.agreementManager.getForkDisputeConfirmations(
+
+        // Step 3: Verify on-chain before committing to reduction
+        const [canReduceOnChain, onChainTimeRemaining] =
+            await this.stateChannelManagerContract.isKillPeriodExpired(
```
lukastanisic99, line 297: maybe best to change the interface for this to return `timestamp` not `timeRemaining` so we stay `objective` and locally when calculating timeoutDelay we can be subjective and do it against our local clock

lukastanisic99 (changes requested):
Overall looks good - just the small stuff mentioned

daiagi:
@lukastanisic99  Done

lukastanisic99 (approved):

PR #227, opened 2025-11-24
dispute confirmation pipeline fixes

daiagi:
### PR targets `dispute-confimation-pipeline` branch

1. fix / optimize: avoid submitting empty disputes
2. fix blockHeight comparison in `[check] isLinked to stateProof`

one side is a number, the other (was) a big Int 

3.fix `[check] N/N Threshold`, does not apply when timeout is forced

4. small modifications to per-exisitng e2e tests
a. raise number of participants from 2 to 3 => avoid yet unimplemented final disputes
b. slight increase p2p time => avoid undesired timeouts

### DIFF
--- a/src/disputeManager/DisputeManager.ts
+++ b/src/disputeManager/DisputeManager.ts
@@ -73,31 +73,8 @@ class DisputeManager {
         try {
             await this.mutex.lock();
             if (this.storage.disputes.didIDispute(forkId)) return;
-            const {
-                dispute,
-                disputeConfirmation,
-                auditingData,
-                fraudProofsToApply
-            } = await this.constructDispute(forkId);
-
-            // Check if there's anything meaningful to dispute
-            const hasTimeout =
-                dispute.input.timeout.participant !== ethers.ZeroAddress;
-            const hasFraudProofs = fraudProofsToApply.length > 0;
-            const hasSelfRemoval = dispute.input.selfRemoval;
-
-            if (!hasTimeout && !hasFraudProofs && !hasSelfRemoval) {
-                this.logger.verbose(
-                    "No timeout, fraud proofs, or self-removal to submit - skipping dispute",
-                    {
-                        forkId,
-                        hasTimeout,
-                        hasFraudProofs,
-                        hasSelfRemoval
-                    }
-                );
-                return;
-            }
+            const { disputeConfirmation, auditingData, fraudProofsToApply } =
+                await this.constructDispute(forkId);
 
             const pendingParticipants =
                 await this.stateChannelManagerContract.getPendingParticipants(
--- a/src/disputeManager/DisputeManager.ts
+++ b/src/disputeManager/DisputeManager.ts
@@ -80,6 +80,25 @@ class DisputeManager {
                 fraudProofsToApply
             } = await this.constructDispute(forkId);
 
+            // Check if there's anything meaningful to dispute
+            const hasTimeout =
+                dispute.input.timeout.participant !== ethers.ZeroAddress;
+            const hasFraudProofs = fraudProofsToApply.length > 0;
+            const hasSelfRemoval = dispute.input.selfRemoval;
+
+            if (!hasTimeout && !hasFraudProofs && !hasSelfRemoval) {
+                this.logger.verbose(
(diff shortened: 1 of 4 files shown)


src/disputeManager/DisputeManager.ts
```diff
@@ -80,6 +80,25 @@ class DisputeManager {
...
+                    {
+                        forkId,
+                        hasTimeout,
+                        hasFraudProofs,
+                        hasSelfRemoval
+                    }
+                );
+                return;
+            }
+
```
lukastanisic99, line 78: The question we need to ask is why and where is dispute being initiated if there are no slashes, no timeout, no self removal and (joining/force processing of an L1 message, but that TS code is yet to be added - hopefully this week just thinking about some stuff)

src/stateManager/DisputeValidationService.ts
```diff
@@ -402,8 +405,14 @@ export default class DisputeValidationService {
                 return false;
             }
 
-            // [check] N/N Threshold
-            if (block && block.didEveryoneSign(participants)) {
+            // [check] N/N Threshold - only for normal timeouts
+            // For forced timeouts: skip this check
+            //   the participant just posted junk calldata, so we timeout them anyway.
+            if (
+                !dispute.input.timeout.isForced &&
```
lukastanisic99: Well not really - first anyone can use `force` even if the blockCalldata is `not` posted, so this would allow anyone to timeout any block with an N/N threshold. `Force` just means `i'm aware of all the data and potential race condition checks that are there to protect me and I willingly choose to ignore them and proceed to vouch for a timeout, since I know that regardless of the the commitment that's stored on-chain I know that no one can successfully challenge the timeout`

The timeout can still be challenged in all cases, so if someone uses `force` that ignores the race condition guards that are there to protect them, they're doing it at their own risk

daiagi: ok. i hear you.
the reason for this was that , in the test `should create forced timeout when peer posts junk calldata that is rejected`

this check considered the timeout to be invalid

lukastanisic99: well if there exists a block at that blockHeight with a threshold - then of course it's invalid 

You should have 2 versions of that test: 
1) No block at blockHeight `i` was produced, but junk was posted on-chain -> we use timeout `force` and it works normally
2) Blocks are normally produced and reach a threshold and junk is posted -> timeout shouldn't be initiated and for now the junk should just be ignored and the channel should progress (we just have to make sure that posting junk doesn't grant extra time to others) - maybe the simples solution for this is to add a fraud proof since it's easy to prove `hash(validBlock) != hash(junk)`

cyony95:
> one side is a number, the other (was) a big Int

had the same issue while working on the spectate tests, but I just wrote a temp converter :)))

lukastanisic99:
> > one side is a number, the other (was) a big Int
> 
> had the same issue while working on the spectate tests, but I just wrote a temp converter :)))
> 

The conversion is Number(BigInt)

We use that everywhere, but we should be careful - it would be better to use BigInt everywhere since it's safer (can represent large numbers), but for now, I guess it's fine using numbers and if needed convert later

cyony95:
> > > one side is a number, the other (was) a big Int
> > 
> > 
> > had the same issue while working on the spectate tests, but I just wrote a temp converter :)))
> 
> The conversion is Number(BigInt)
> 
> We use that everywhere, but we should be careful - it would be better to use BigInt everywhere since it's safer (can represent large numbers), but for now, I guess it's fine using numbers and if needed convert later

there's also an issue using stringify with BigInt. stringify uses toJSON and toJSON can't serialize BigInt. when I meant a temp converter, I meant it in this context. So something like this in the PeerTestHarness:
```
if (typeof (BigInt.prototype as any).toJSON !== "function") {
    (BigInt.prototype as any).toJSON = function () {
        return Number(this);
    };
}
```

lukastanisic99:
@daiagi why is this closed? 
https://github.com/peer3to/state-channels-plus/pull/229 Builds upon this and I have a large PR coming soon upon my CRDT fix

daiagi:
merging for a simpler git tree
this PR is now trivial

PR #228, opened 2025-11-24
Add E2E RPC Tests for InitHandshake Service

cyony95:
- added BigInt serialization(not necessarily needed for these tests)
- moved some helpers to PeerTestHarness
- added initHandshake e2e test

### DIFF
--- a/test/e2e/E2E-RPC.test.ts
+++ b/test/e2e/E2E-RPC.test.ts
@@ -4,6 +4,8 @@ import { AStateMachine, MathStateMachine } from "@typechain-types/index";
 import { ForkId } from "@/types/types";
 import { hash } from "../factory";
 import { ATransport } from "@/transport";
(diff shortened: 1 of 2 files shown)


cyony95:
I might have overcomplicated the tests with the last commit, if you think it s unnecessary I will revert it but I think it's a little bit more accurate.

test/e2e/E2E-RPC.test.ts
```diff
@@ -528,6 +545,666 @@ describe("E2E: RPC Services", function () {
         });
     });
 
+    describe("InitHandshake RPC", function () {
+        let LocalDiscoveryServer: (typeof import("@/utils/LocalDiscoveryServer"))["LocalDiscoveryServer"];
```
lukastanisic99: why not just import it

PR #229, opened 2025-11-28
reduce CRDT fix - pick lowest hash block in case height is the same

lukastanisic99:
Reduce should output the same result regardless of the order of `disputes[ ]`.

Previously, when picking the `latestBlock` it selected the block with the highest height, but in the case the heights were equal it selected the one it `processed last`, so the order influenced the result. E.g. peerA reduces([X,Y]) and peerB reduces([Y,X]) and if X & Y are the same height peerA picks Y, while peerB pick X and they diverge. 

Simple solution, in that case pick the block with the lowest hash. 

@daiagi tests are still unstable - can we pls first fix and stabilize them so they always run deterministically - we have no threads and all the potential race conditions of a distributed system should be contained and controlled. Once and for all we should figure out why peers don't always establish a local connection within 5s.


evainfeld (approved):

daiagi:
how about?

```solidity
    function _isBetterBlock(Block memory a, Block memory b)
        internal
        view
        returns (bool)
    {
        uint256 aTx = a.transaction.header.transactionCnt;
        uint256 bTx = b.transaction.header.transactionCnt;

        if (aTx != bTx) return aTx > bTx;
        return _getBlockHash(a) < _getBlockHash(b);
    }
```


and 
```solidity
            if (hasBlock) {
                if (!latestBlockInitialized || _isBetterBlock(disputeLatestBlock, reducedOutput.latestBlock)) {
                    reducedOutput.latestBlock = disputeLatestBlock;
                    latestBlockInitialized = true;
                }
            }
```



>tests are still unstable

working on it

daiagi (approved):

PR #230, opened 2025-12-01
Fix  sporadic test fail

daiagi:
- fix: prevent Event handler from acting on chain events after state manger is disposed
- fix: allow test harness to wait for a peer's turn before submitting transaction
- fix: state manager awaits this.success
- refactor: make `LocalDisvcoeryServer` more functional, more readable; added retry on port selection for the main discovery server, "promisified" ws events so that can await on that promise until ws server is ready and listening

basically all of the above fixes various race and timing issues

### DIFF
--- a/src/P2PManager.ts
+++ b/src/P2PManager.ts
@@ -81,8 +81,8 @@ class P2PManager implements IOnMessage {
     }
     public async tryOpenConnectionToChannel(channelId: string) {
         if (DEBUG_LOCAL_TRANSPORT) {
-            LocalDiscoveryServer.tryStart();
-            LocalDiscoveryServer.connectToPeers(this.self, channelId);
+            await LocalDiscoveryServer.tryStart();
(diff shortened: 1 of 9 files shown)


cyony95 (approved):
I tested it on one of my branches with no new test cases and with some new test cases I'm working on and I can see tests that we know are correct and should always finish successfully do now always finish successfully. My intentionally failing tests keep failing so I dont see any problems here, the additional logging is always welcome.

daiagi:
@lukastanisic99 
Apparently, this https://github.com/peer3to/state-channels-plus/pull/227#discussion_r2554724671
does not happen in this PR, and was most likely a result of a chain events (onDisputeCommuted probably) "spilling" over to the next test. 
fixed by "dispose aware" Event manager

I'm merging this PR so i can focus or real bugs in the dispute-confirmation-pipeline PR, if any

PR #233, opened 2025-12-02
Dispute confirmation pipeline fix 2

daiagi:
another round of small fixes to make all tests pass in `dispute-confirmation-pipeline` PR


src/StateChannelEventListener.ts
```diff
@@ -224,14 +226,14 @@ class StateChannelEventListener {
...
-            handler: (logObj: any) => {
+            handler: async (logObj: any) => {
                 const {
                     channelId,
                     joinChannelBlock,
                     timestamp,
                     totalDeposits
                 } = logObj.args;
-                this.eventHandler.onJoinChannelProcessed(
+                await this.eventHandler.onJoinChannelProcessed(
```
lukastanisic99: Don't think these `awaits` make a difference - there's nothing bellow them (nothing is really waiting on them) 

These other things in `src` are human errors (me) and magic already resolved them on my end, except the extension of `decode` which isn't currently even used

There are some other bugs which I'll find out soon

test/e2e/E2E-RPC.test.ts
```diff
@@ -25,17 +25,25 @@ describe("E2E: RPC Services", function () {
...
+
+            // Wait for the peer to be ready for their turn
+            await harness!.waitForTurn(0);
+            await harness!.submitTransaction(harness.peers[0], (contract) =>
+                contract.add(1)
+            );
+            await harness!.waitForTurn(byzantinePeer.index);
+            await harness!.submitTransaction(byzantinePeer, (contract) =>
+                contract.add(1)
+            );
```
lukastanisic99: This is the thing introduced with the refactor of LocalDiscoveryServer - I don't like this verbose syntax - I understand the problem - better do a semaphore/barrier (take a look at `BarrierLocal.ts`) in `submitNextTransaction` and await it and release it like this:
```typescript
p2pEventHooks.onTurn(address){
    if(myAddress == address) barrier.allowOne()
}
```

So the syntax can be restored to `await submitNextTransaction(...)`

daiagi: It has nothing to do with local discovery server.
And absolutely nothing to do with the refcator, which changed very little functionality.
It has to do with the time it takes a peer action to propagate through the channel.
It is also simple common sense that a honest peer will submit a TX only when it is their turn.

lukastanisic99: Yeah I know I just said it was introduced in that PR first - I'm aware what it does

src/StateChannelEventListener.ts
daiagi: ok let's play a game:

change `DisputeCommitted` handler:

```ts
            handler: async (logObj: any) => {
                const {
                    channelId,
                    disputeConfirmation,
                    disputeCreationTimestamp,
                    isFinal,
                    windowCreationTimestamp
                } = logObj.args;
                await new Promise((resolve) => setTimeout(resolve, 10000)).then(
                    () =>
                        this.eventHandler.onDisputeCommitted(
                            channelId,
                            Codec.convertEthersResultToObject(
                                disputeConfirmation
                            ),
                            disputeCreationTimestamp,
                            isFinal,
                            windowCreationTimestamp
                        )
                );
            }
```


and run: `yarn test:e2e:debug --grep "should handle timeout when next peer to write, does not author a block"`


```sh
[08:40:44][INFO][Peer 1][0x709979...][P2pEventHooks] Initiating dispute
[08:40:44][INFO][Peer 0][0xf39Fd6...][P2pEventHooks] Initiating dispute
[08:40:54][DEBUG][Peer 0][0xf39Fd6...][EventHandler] Dispute committed {"channelId":"0xad3c1332b3af936661ef56ae83c46b567f216169d932eb04ad64453eba3f8c07","forkId":"0x703b528410d6ed12764ec959d58d9fe250edcb3124c60f679befbe6afa8910c7","isFinal":false,"disputeCreationTimestamp":"1764726054","isForced":false}
```


Notice the 10 seconds gap?


good, now lets remove the "sleep" promise:


```ts
            handler: async (logObj: any) => {
                const {
                    channelId,
                    disputeConfirmation,
                    disputeCreationTimestamp,
                    isFinal,
                    windowCreationTimestamp
                } = logObj.args;

                await this.eventHandler.onDisputeCommitted(
                    channelId,
                    Codec.convertEthersResultToObject(disputeConfirmation),
                    disputeCreationTimestamp,
                    isFinal,
                    windowCreationTimestamp
                );
            }
              
```


No gap:

```sh
[08:44:54][INFO][Peer 1][0x709979...][P2pEventHooks] Initiating dispute
[08:44:54][DEBUG][Peer 0][0xf39Fd6...][EventHandler] Dispute committed {"channelId":"0x369efef9d2d05a6eda3d6c8299cbc3092477b3950d8e466c2ca52d93f616b05b","forkId":"0x214cc0420fa678dd4928f2d531429b4aadd37613b78058d344639cdccee5ef80","isFinal":false,"disputeCreationTimestamp":"1764726304","isForced":false}
```



Regardless, I only did it to have uniform interface on the handler:
`handler: (logObj: any) => Promise<void>`


my point was about the cleanup logic, not the awaits
could of simply return the `eventHandler` method, since all of them return promises already
only reason handler was `Promise<void> | void` is because the handler were not uniform, some (like `StateSnapshotUpdated`) had `await` and others did not

lukastanisic99: Not sure I follow what point you want to prove - ofc there's a 10s gap since you sleep 10s - my point was that `await` in `await this.eventHandler.onDisputeCommitted(...)` doesn't make any difference 

Regardless I'm fine with requiring `Promise<void>` if you want a consistent interface, but performance wise and logic wise it doesn't make any difference

lukastanisic99: As for `StateSnapshotUpdated` in EventListener the await doesn't make a difference (not sure why it's there), but in SpectateService which uses `onStateSnapshotUpdated` the `await` does make a difference and it should be there and I made it async to account for that, but apparently I added the await in the wrong place - human error :D 
I'll correct it

PR #237, opened 2025-12-03
General message passing

lukastanisic99:
General L1->L2 and L2->L1 message passing

Missing a few things for feature complete 

All tests are passing

I've double checked the logic in `src`/`contracts` - not saying it doesn't have bugs, saying I vouch for it

@daiagi please check tests is the logic good


contracts/V1/StateChannelDiamondProxy/DisputeVerificationFacet.sol
```diff
@@ -310,12 +336,42 @@ contract DisputeVerificationFacet is StateChannelCommon {
...
+        outboundMessageBlock.previousBlockHash = latestStateSnapshot.snapshotData.latestOutboundMessageBlockHash;
+        uint256 outboundHeight = latestStateSnapshot.snapshotData.latestOutboundMessageBlockHeight;
+        if (outboundMessages.length > 0) {
+            outboundMessageBlock.blockHeight = outboundHeight + 1;
+        } else {
+            outboundMessageBlock.blockHeight = outboundHeight;
+        }
+        outboundMessageBlock.messages = outboundMessages;
+        outboundMessageBlock.totalBalance = outputState.totalWithdrawals;
+        outboundMessageBlock.timestamp = 0; // timestamp is not relevant, but more importantly this needs to be deterministic
```
lukastanisic99, line 358: There was a bug here - I had to find it manually 
This was `timestamp = block.timestamp` which caused non-determinism during `auditing` since `block.timestamp` changes

src/storage/Storage.ts
```diff
@@ -53,7 +58,7 @@ export class Storage {
     getStateSnapshot(coordinates: BlockCoordinates): StateSnapshot | undefined {
         const { forkId, height } = coordinates;
 
-        if (height <= 0) {
+        if (height < 0) {
```
lukastanisic99, line 61: magic

src/utils/EthersResultProxy.ts
```diff
@@ -0,0 +1,186 @@
...
+                    }
+                    return logs;
+                };
+            }
+
+            // Non-function properties return as-is
+            return original;
+        }
+    }) as T;
+}
```
lukastanisic99, line 192: @daiagi @cyony95 this may interest you

daiagi:
right off the bat:

```sh
Warning: Contract code size is 25924 bytes and exceeds 24576 bytes (a limit introduced in Spurious Dragon). This contract may not be deployable on Mainnet. Consider enabling the optimizer (with a low "runs" value!), turning off revert strings, or using libraries.
  --> contracts/V1/StateChannelDiamondProxy/DisputeVerificationFacet.sol:10:1:
  
  
  Warning: Contract code size is 25263 bytes and exceeds 24576 bytes (a limit introduced in Spurious Dragon). This contract may not be deployable on Mainnet. Consider enabling the optimizer (with a low "runs" value!), turning off revert strings, or using libraries.
  --> contracts/V1/StateChannelDiamondProxy/StateChannelManagerProxy.sol:18:1:
```

src/eventHandlers/EventHandler.ts
```diff
@@ -198,13 +206,13 @@ export class EventHandler {
...
-            throw new Error("Not implemented yet - final dispute");
-            // return this.stateManager.setGenesisState(
-            //     disputeAuditingData.latestStateStateMachineState,
-            //     dispute.outputSnapshotDataHash,
-            //     disputeCreationTimestamp
-            // );
+            await this.stateManager.setGenesisState(
+                disputeAuditingData.latestStateSnapshot.snapshotData,
+                disputeAuditingData.latestStateStateMachineState,
+                dispute.outputSnapshotDataHash as ForkId,
```
daiagi, line 212: isn't that `dispute.forkId`?

lukastanisic99:
> right off the bat:
> 
> ```shell
> Warning: Contract code size is 25924 bytes and exceeds 24576 bytes (a limit introduced in Spurious Dragon). This contract may not be deployable on Mainnet. Consider enabling the optimizer (with a low "runs" value!), turning off revert strings, or using libraries.
>   --> contracts/V1/StateChannelDiamondProxy/DisputeVerificationFacet.sol:10:1:
>   
>   
>   Warning: Contract code size is 25263 bytes and exceeds 24576 bytes (a limit introduced in Spurious Dragon). This contract may not be deployable on Mainnet. Consider enabling the optimizer (with a low "runs" value!), turning off revert strings, or using libraries.
>   --> contracts/V1/StateChannelDiamondProxy/StateChannelManagerProxy.sol:18:1:
> ```

Yeah I don't read the warnings for now :D - Luckily everything is partitioned following the Diamond Proxy pattern, so we'll add 1 more `Facet` and all of that will be taken care off :D

contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol
```diff
@@ -377,32 +386,194 @@ contract DisputeFraudProofFacet is StateChannelCommon {
...
-        (bool isFound,) = getBlockCallDataCommitment(
+        (bool isFound, bytes32 commitment) = getBlockCallDataCommitment(
             _getDisputeChannel(dispute),
             _getDisputeFork(dispute),
             dispute.input.timeout.blockHeight,
             dispute.input.timeout.participant
         );
-        if (!isFound) revert();
+        if (!isFound) return address(0); // the calling context may decide to slash the caller
+        bytes32 _commitment = keccak256(abi.encode(postedBlock, timeoutCalldataPostedProof.onChainTimestamp));
```
daiagi, line 419: unclear naming
I get what you are doing here
but comparing commitment to _commitment is not very revealing

contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol
```diff
@@ -377,32 +386,194 @@ contract DisputeFraudProofFacet is StateChannelCommon {
...
+        if (commitment != _commitment) return address(0); // the calling context may decide to slash the caller
+
+        // get previousTimestamp
```
daiagi, line 422: Suggestion to extract functions

```suggestion
(bool isPreviousTimestampValid, uint256 previousTimestamp) =
            _computePreviousTimestamp(dispute, timeoutCalldataPostedProof);
        if (!isPreviousTimestampValid) return address(0); // the calling context may decide to slash the caller
```



and:
```solidity
    function _computePreviousTimestamp(
        Dispute memory dispute,
        TimeoutCalldataPosted memory timeoutCalldataPostedProof
    ) internal view returns (bool, uint256) {
        (bool hasBlock, Block memory latestBlock) = _getLatestBlock(dispute.input.stateProof);
        uint256 previousTimestamp;
        if (!hasBlock) {
            bytes32 originForkId = timeoutCalldataPostedProof.auditingData.genesisStateSnapshotData.originForkId;
            (bool hasGenesis, uint256 genesisTimestamp) =
                getGenesisTimestamp(dispute.input.channelId, originForkId, dispute.input.forkId);
            require(hasGenesis, ErrorGenesisTimestampNotAvailable());
            previousTimestamp = genesisTimestamp;
        } else {
            (bool _isFound, bytes32 previousBlockCommitment) = getBlockCallDataCommitment(
                _getDisputeChannel(dispute),
                _getDisputeFork(dispute),
                latestBlock.transaction.header.transactionCnt,
                latestBlock.transaction.header.participant
            );
            if (!_isFound) {
                previousTimestamp = latestBlock.transaction.header.timestamp;
            } else {
                if (timeoutCalldataPostedProof.previousBlockOnChainTimestamp == 0) {
                    revert ErrorUnexpectedBlockCalldataPosted();
                }

                bytes32 _previousBlockCommitment = keccak256(
                    abi.encode(
                        timeoutCalldataPostedProof.previousBlockcalldata,
                        timeoutCalldataPostedProof.previousBlockOnChainTimestamp
                    )
                );
                if (previousBlockCommitment != _previousBlockCommitment) {
                    return (false, 0);
                }
                if (
                    keccak256(abi.encode(latestBlock))
                        == keccak256(timeoutCalldataPostedProof.previousBlockcalldata.encodedBlock)
                ) {
                    // only if the uploaded calldata matches the stateProof latest block, we grant extra time, otherwise the caller can forge a double sign or something else
                    previousTimestamp = timeoutCalldataPostedProof.previousBlockOnChainTimestamp;
                } else {
                    previousTimestamp = latestBlock.transaction.header.timestamp;
                }
            }
        }
        return (true, previousTimestamp);
    }
```

contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol
```diff
@@ -377,32 +386,194 @@ contract DisputeFraudProofFacet is StateChannelCommon {
...
+            timestamp: _block.transaction.header.timestamp
+        });
+
+        if (_block.stateSnapshotHash != keccak256(abi.encode(recomputedSnapshot))) {
+            return address(0); // the calling context may decide to slash the caller
+        }
         return dispute.input.disputer;
     }
 
+    function _handleDisputeInvalidBlockInStateProofApplyFraudProof(
```
daiagi, line 533: how about?
"_handleDisputeInvalidBlockInStateProof"
or simply
"_handleDisputeInvalidStateProof"

contracts/V1/StateChannelDiamondProxy/DisputeVerificationFacet.sol
```diff
@@ -46,6 +49,7 @@ contract DisputeVerificationFacet is StateChannelCommon {
         uint256 maxSlashCount;
         uint256 slashCount;
         uint256 selfRemovalCount;
+        bool latestBlockInitialized;
```
daiagi, line 52: an internal implementaion detail, and is not used anywhere
don't see why it should be returned from the function

contracts/V1/StateChannelDiamondProxy/DisputeVerificationFacet.sol
```diff
@@ -83,12 +91,22 @@ contract DisputeVerificationFacet is StateChannelCommon {
...
-                        >= reducedOutput.latestBlock.transaction.header.transactionCnt
-            ) {
-                reducedOutput.latestBlock = disputeLatestBlock;
+            if (hasBlock) {
+                uint256 candidateTxCount = disputeLatestBlock.transaction.header.transactionCnt;
+                if (!latestBlockInitialized) {
+                    reducedOutput.latestBlock = disputeLatestBlock;
+                    latestBlockInitialized = true;
+                } else {
+                    uint256 currentTxCount = reducedOutput.latestBlock.transaction.header.transactionCnt;
```
daiagi, line 100: why not call it block height?

contracts/V1/StateChannelDiamondProxy/DisputeVerificationFacet.sol
```diff
@@ -83,12 +91,22 @@ contract DisputeVerificationFacet is StateChannelCommon {
...
+                    if (candidateTxCount > currentTxCount) {
+                        reducedOutput.latestBlock = disputeLatestBlock;
+                    } else if (
+                        candidateTxCount == currentTxCount
```
daiagi, line 104: see my comment on the crdt fix PR, this could be clearer
https://github.com/peer3to/state-channels-plus/pull/229#issuecomment-3594642949

contracts/V1/AStateMachine.sol
```diff
@@ -60,6 +61,23 @@ abstract contract AStateMachine {
 
     function getZeroBalance() public pure virtual returns (Balance memory zeroBalance);
 
+    function processInboundMessage(Message calldata message) external _nonReentrant returns (bool) {
+        return _processInboundMessage(message);
+    }
+
+    function _processInboundMessage(Message calldata message) internal virtual returns (bool) {
+        if (message.messageType == MESSAGE_TYPE_JOIN) {
+            JoinChannel memory joinChannel = abi.decode(message.data, (JoinChannel));
```
daiagi, line 70: I get that naming things as inbound and outbound messages has elegance to it
but it seems a bit like a pointless abstraction

the only inbound message is join and the only outbound is exit

also, if there are plans to expand to more kinds of messages (what kinds?)

~~then it would make more sense for a general message to look like~~
~~`{type: MessgeType, payload: Struct}`~~

Edit:
I saw that the data field of `Message` is the payload





the feature we discussed, that a dispute is not needed in order to join, and only triggered if an inbound message was not digested, makes sense, but that does not require the added abstraction

contracts/V1/StateChannelDiamondProxy/FraudProofFacet.sol
```diff
@@ -117,21 +127,44 @@ contract FraudProofFacet is StateChannelCommon {
...
+                    stateMachineImplementation.addBalance(updatedTotalWithdrawals, outboundMessages[i].balance);
+            }
+
+            outboundHeight += 1;
+
+            MessageBlock memory outboundMessageBlock;
+            outboundMessageBlock.previousBlockHash = nextOutboundMessageBlockHash;
+            outboundMessageBlock.blockHeight = outboundHeight;
+            outboundMessageBlock.messages = outboundMessages;
+            outboundMessageBlock.totalBalance = updatedTotalWithdrawals;
```
daiagi, line 153: sure about this?
`outboundMessageBlock.totalBalance = updatedTotalWithdrawals;`?

not confused with somethin like

new_blance = balance - updatedTotalWithdrawals
??

contracts/V1/StateChannelDiamondProxy/LocalDiamond.sol
```diff
@@ -58,44 +59,43 @@ contract LocalDiamond is StateChannelManagerProxy {
...
+        channelBalance.totalDeposits = zeroBalance;
+        channelBalance.totalWithdrawals = zeroBalance;
+
+        bytes32 inboundMessageBlockHash = stateSnapshot.snapshotData.latestInboundMessageBlockHash;
+        channelBalance.latestInboundMessageBlockHash = inboundMessageBlockHash;
+        channelBalance.latestInboundMessageBlockHeight = stateSnapshot.snapshotData.latestInboundMessageBlockHeight;
+        channelBalance.latestOutboundMessageBlockHeight = stateSnapshot.snapshotData.latestOutboundMessageBlockHeight;
+        if (inboundMessageBlockHash != bytes32(0)) {
+            MessageBlock memory snapshotInboundBlock;
+            snapshotInboundBlock.previousBlockHash = bytes32(0);
```
daiagi, line 71: `snapshotInboundBlock.previousBlockHash = bytes32(0);`

why?

daiagi (commented):
Partial review
didn't  finish yet

contracts/V1/StateChannelDiamondProxy/StateSnapshotFacet.sol
```diff
@@ -148,15 +151,30 @@ contract StateSnapshotFacet is StateChannelCommon {
...
         }
     }
+
+    function _resolveTotalDeposits(bytes32 channelId, bytes32 blockHash) internal view returns (Balance memory) {
+        Balance memory zeroBalance = stateMachineImplementation.getZeroBalance();
+        if (blockHash == bytes32(0)) {
+            return zeroBalance;
+        }
+        MessageBlock storage inboundBlock = inboundMessageBlockMap[channelId][blockHash];
+        if (inboundBlock.timestamp == 0) {
```
daiagi, line 171:
```suggestion
        if (inboundBlock.timestamp == 0) {
        // missing inbound block, use snapshot totals instead
```

src/disputeManager/DisputeManager.ts
```diff
@@ -80,6 +81,25 @@ class DisputeManager {
                 fraudProofsToApply
             } = await this.constructDispute(forkId);
 
+            // Check if there's anything meaningful to dispute
```
daiagi: redundant and root cause fixed in 
https://github.com/peer3to/state-channels-plus/pull/230

src/disputeManager/DisputeManager.ts
```diff
@@ -156,7 +179,41 @@ class DisputeManager {
...
+                );
+            if (!disputeFraudProof) {
+                throw new Error("No dispute fraud proof found for dispute");
+            }
+            const txRespone =
+                await this.stateChannelManagerContract.applyDisputeFraudProofs([
+                    disputeFraudProof
+                ]);
+            txRespone.wait().then(() => {
+                this.logger.info("Dispute killed successfully", {
```
daiagi:
```suggestion
                this.logger.debug("Dispute killed successfully", {
```

src/disputeManager/DisputeManager.ts
```diff
@@ -156,7 +179,41 @@ class DisputeManager {
...
+                    forkId: dispute.input.forkId,
+                    errorDescription: error.errorDescription
+                });
+            } else {
+                this.logger.error("Error killing dispute", {
+                    forkId: dispute.input.forkId,
+                    error:
+                        error instanceof Error ? error.message : String(error)
+                });
+            }
```
daiagi, line 194:
```suggestion
            const errorMessage = isCustomEvmError(error)
                ? error.errorDescription
                : error instanceof Error
                  ? error.message
                  : String(error);
            this.logger.error("Error killing dispute", {
                forkId: dispute.input.forkId,
                error: errorMessage
            });
```

src/disputeManager/DisputeManager.ts
```diff
@@ -379,7 +446,10 @@ class DisputeManager {
                 milestoneSnapshots: milestoneSnapshots.map((snapshot) =>
                     snapshot.toStruct()
                 ),
-                exitChannelBlocks: exitChannelBlocks
+                inboundMessageBlocks:
+                    inboundMessageBlocks as MessageBlockStruct[],
+                outboundMessageBlocks:
+                    outboundMessageBlocks as MessageBlockStruct[]
```
daiagi: uneeded type "conversions"

src/disputeManager/DisputeManager.ts
```diff
@@ -267,15 +324,20 @@ class DisputeManager {
...
 
         const outputSnapshotData =
             await this.diamondStateMachine.localDiamondContract.computeDisputeOutputSnapshotData.staticCall(
                 disputeInput,
-                latestStateSnapshot.toStruct(),
-                latestStateMachineState,
-                auditingData.genesisStateSnapshotData.latestJoinChannelBlockHash
+                auditingData.latestStateSnapshot,
+                auditingData.latestStateStateMachineState,
+                auditingData.inboundMessageBlocks || []
```
daiagi: uneeded `|| []`

getMessageBlocksInRange will return [] when the iterator is empty

src/eventHandlers/EventHandler.ts
```diff
@@ -158,9 +162,13 @@ export class EventHandler {
             Type.Dispute
         );
         const forkId = dispute.input.forkId;
+        const disputeHash = hash(
```
daiagi, line 165: computationaly expensive just for log
does it really help debugging something

src/eventHandlers/EventHandler.ts
```diff
@@ -446,26 +451,23 @@ export class EventHandler {
...
-            return;
-        }
-        await this.diamondStateMachine.localDiamondContract.onJoinChannelProcessed(
-            channelId,
-            joinChannelBlock,
-            timestamp,
-            totalDeposits
+        const messageBlockHash = ethers.keccak256(
+            Codec.encode(messageBlock, Type.MessageBlock)
+        ) as Hash;
```
daiagi:
```suggestion
        const messageBlockHash = hash(
            Codec.encode(messageBlock, Type.MessageBlock)
        );
```

src/eventHandlers/EventHandler.ts
```diff
@@ -518,17 +520,18 @@ export class EventHandler {
             throw new Error(
                 `GenesisSnapshot not available for forkId: ${forkId}`
             );
-        const jcbs = this.storage.joinChannelBlocks.getBlocksInRange(
-            genesisSnapshot.snapshotData.latestJoinChannelBlockHash,
-            latestSnapshot.snapshotData.latestJoinChannelBlockHash
-        );
+        const inboundMessageBlocks =
+            this.storage.inboundMessages.getMessageBlocksInRange(
+                genesisSnapshot.snapshotData.latestInboundMessageBlockHash,
```
daiagi: wrong order
should be 
```ts
(
    latestSnapshot.snapshotData.latestInboundMessageBlockHash,
    genesisSnapshot.snapshotData.latestInboundMessageBlockHash
)
```

src/storage/BlockCalldataStorage.ts
```diff
@@ -0,0 +1,42 @@
+import { ForkId, BlockHeight, BlockCalldata, Address } from "@/types/types";
+import { Block } from "@/models";
+
+type CalldataCoordinateKey = string;
+
+export class BlockCalldataStorage {
+    // ====================================
+    // STORAGE MAPS
+    // ====================================
+    private coordinatesToBlockMap: Map<CalldataCoordinateKey, BlockCalldata>;
```
daiagi, line 10: same question like the one on `dispute-validation-pipeline` PR
why not use Hash as key?

src/storage/BlockCalldataStorage.ts
```diff
@@ -0,0 +1,42 @@
...
+    // ====================================
+    // CREATE
+    // ====================================
+
+    storeBlockCalldata(blockCalldata: BlockCalldata): CalldataCoordinateKey {
+        const block: Block = Block.fromSignedBlock(
+            blockCalldata.signedBlock,
+            blockCalldata.onChainTimestamp
+        );
+        const coordinateKey = `${block.forkId}:${block.height}:${block.author}`;
```
daiagi: altough only two places
but consider adding private function that converts a block into cooridantes

src/storage/MessageBlockStorage.ts
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    // ====================================
+    // CREATE / UPDATE
+    // ====================================
+
+    store(messageBlock: MessageBlockStruct, options?: StoreOptions): Hash {
+        const hash =
+            options?.hash ??
+            ethers.keccak256(Codec.encode(messageBlock, Type.MessageBlock));
```
daiagi:
```suggestion
            hash(Codec.encode(messageBlock, Type.MessageBlock));
```

src/storage/MessageBlockStorage.ts
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    getLatestMessageBlocks(limit?: number): MessageBlockStruct[] {
+        return this.getLatestEntries(limit).map((entry) => entry.messageBlock);
+    }
+
+    private normalizeBlockHeight(
+        height: MessageBlockStruct["blockHeight"]
+    ): BlockHeight {
+        if (height === undefined || height === null) return 0;
```
daiagi: this could lead to bugs
why not normalize it into undefined?

src/storage/MessageBlockStorage.ts
```diff
@@ -0,0 +1,138 @@
+import { ethers } from "ethers";
+import { MessageBlockStruct } from "@typechain-types/contracts/V1/StateChannelDiamondProxy/LocalDiamond";
+import { BlockHeight, Hash } from "@/types/types";
+import { Codec, Type } from "@/utils";
+export interface MessageBlockEntry {
```
daiagi: ha?
why is this needed?

why not store `MessageBlockStruct` directly?

src/storage/MessageBlockStorage.ts
```diff
@@ -0,0 +1,138 @@
...
+
+    // ====================================
+    // READ
+    // ====================================
+
+    getMessageBlock(blockHash: Hash): MessageBlockStruct | undefined {
+        return this.blockMap.get(blockHash)?.messageBlock;
+    }
+
+    getEntry(blockHash: Hash): MessageBlockEntry | undefined {
```
daiagi: uneeded

daiagi (commented):
another review round
not finished yet


Also, Do i understand correctly that this PR supersedes your previous two PR's @lukastanisic99 ?

src/utils/EthersResultProxy.ts
cyony95: so, I tested this and it works perfectly with my schedule tests *as they are*. It's also much more elegant than the additions I originally made to the staticCallProxy to make my schedule tests run. I stand by using this one from now on @lukastanisic99 @daiagi

src/disputeManager/DisputeManager.ts
```diff
@@ -156,7 +160,41 @@ class DisputeManager {
             this.mutex.unlock();
         }
     }
-
+    public async killDispute(dispute: DisputeStruct): Promise<void> {
```
daiagi, line 162: perhaps a name like "applyDispute" is better?

kill suggests that the dispute was invalid and is terminated
this here is the happy path progression, isn't it?

src/stateManager/utils/DisputeFraudProofService.ts
```diff
@@ -1,10 +1,273 @@
...
+            proofType: toSolidityDisputeFraudProofType(proof.type),
+            participant: dispute.input.disputer,
+            dispute: dispute,
+            encodedProof: Codec.encode(proof.struct, proof.type)
+        };
+
+        const proofHash =
+            this.storage.disputeFraudProofs.storeFraudProof(disputeFraudProof);
+
+        this.logger?.warn("Stored dispute fraud proof", {
```
daiagi: why warn?
debug is more suitable, no?

src/stateManager/utils/FraudProofService.ts
```diff
@@ -8,11 +8,15 @@ import {
     WrongGenesisProofStruct
 } from "@typechain-types/contracts/V1/types/FraudProofTypes";
 import { FraudProofStruct } from "@typechain-types/contracts/V1/types/ProofTypes";
-import { ZeroHash } from "ethers";
 import { Address, Hash } from "@/types/types";
 import { Codec, FraudStruct } from "@/utils/Codec";
 import { FraudProofType, toSolidityFraudProofType } from "@/types/sol-enums";
 
+const createEmptySignedBlock = (): SignedBlockStruct => ({
+    encodedBlock: "0x",
```
daiagi, line 16: use ethers zeroHash

src/stateManager/DisputeValidationService.ts
```diff
@@ -277,28 +296,161 @@ export default class DisputeValidationService {
...
+            let previousTimestamp = 0;
+            if (previousBlockOrSnapshot.stateSnapshot) {
+                previousTimestamp =
+                    previousBlockOrSnapshot.stateSnapshot.timestamp;
+            } else {
+                // previous block
+                const previousBlock = previousBlockOrSnapshot.block!;
+                const onChainSignature = dispute.input.timeout
+                    .participantSignatureOnPreviousBlock as Signature;
+                if (!onChainSignature || onChainSignature === "0x") {
```
daiagi, line 375: use ethers zeroHash

src/stateManager/DisputeValidationService.ts
```diff
@@ -277,28 +296,161 @@ export default class DisputeValidationService {
...
+                    // siganture not posted on-chain, so time is not forfeited
+                    previousTimestamp = previousBlock.currentTimestamp;
+                } else {
+                    // signature exists on-chain, verify it's from the timedout participant
+                    const retrievedAddress = previousBlock.signatureToAddress(
+                        dispute.input.timeout
+                            .participantSignatureOnPreviousBlock as Signature
+                    );
+                    if (retrievedAddress == dispute.input.timeout.participant) {
```
daiagi, line 384: extracting into a function would allow to simply return,avoiding the nested if/else

src/stateManager/DisputeValidationService.ts
```diff
@@ -277,28 +296,161 @@ export default class DisputeValidationService {
...
+                this.disputeFraudProofService.createTimeoutTooEarly(
+                    dispute,
+                    disputeAuditingData,
+                    previousBlockOrSnapshot?.block?.onChainTimestamp
+                );
+                return false;
+            }
+
+            // [check] N/N Threshold
+            if (block && block.didEveryoneSign(participants)) {
```
daiagi, line 411: better handle the case of block undefined at the top, right after `storage.blocks.getBlock`

src/stateManager/StateManager.ts
```diff
@@ -451,30 +440,29 @@ class StateManager {
         encodedState: Bytes,
         forkId: ForkId,
         genesisTimestamp: Timestamp,
-        exitChannelBlock?: ExitChannelBlockStruct
+        outboundMessageBlock?: MessageBlockStruct
     ): Promise<void> {
+        const normalizedGenesisTimestamp = Number(genesisTimestamp);
```
daiagi, line 445: uneeded.
it is already a number in all call sites, and TS will warn otherwise:
```ts
genesisTimestamp: Timestamp
```

src/stateManager/StateManager.ts
```diff
@@ -1405,50 +1394,66 @@ class StateManager {
...
-
-        if (exitChannels && exitChannels.length > 0) {
-            participants = await this.diamondStateMachine.getParticipants();
-            exitChannelBlock = {
-                exitChannels,
-                previousBlockHash: latestExitChannelBlockHash
-            };
+        let { latestOutboundMessageBlockHash, totalWithdrawals } =
+            previousSnapshotData;
+        let latestOutboundMessageBlockHeight = BigInt(
```
daiagi, line 1405: i dont think the BigInt call is needed
the type is already ethers.BigNumberish

daiagi (approved):
final round
After these, Approved

src/disputeManager/DisputeManager.ts
lukastanisic99: So this actually kills the dispute - only the soldidity code is named `applyDisputeFraudProof` since it's a 1:1 mirror of the `fraudProof` system, but a successful dispute fraud proof, kills the dispute

src/stateManager/utils/DisputeFraudProofService.ts
lukastanisic99: sure - will push that

src/stateManager/utils/FraudProofService.ts
lukastanisic99: yeah I'm not sure about this one - '0x' means it's empty, `zeroHash` has the same dataSize/length as a hash, but it's filled with `0s` - what I think would be correct here is to fill the data structure with 0s but in a representation where it would still decode correctly to the underlying structure - not sure will this be a problem - tests are needed

src/stateManager/StateManager.ts
lukastanisic99: Yeah I was looking at this too - but aparently the event listener has an `any` type so you can pass that and TS won't complain - Magic added this while testing at runtime - the root solution is to also cast all BigNumerish to Numbers as we did with ethers.result, but kinda not trying to do this now since for some cases we actually will need the big numbers and will need to use that in the end - it's a Trello card from long ago

src/stateManager/StateManager.ts
lukastanisic99: True - I'll remove it

lukastanisic99: image

Actually when you remove it a few lines bellow it complains - at runtime it's a BigInt, but typescript doesn't know it it's BigNumberish

lukastanisic99:
@daiagi didn't see you pushed stuff - I pushed just the debug thing - can you do a PR into this one with the rest of your suggestions from this PR, previous PRs and merge any other PRs that are ready, so I can quickly just go over it later and build on top of that tonight and hopefully finish by our call tomorrow - need to finish some other stuff now

PR #238, opened 2025-12-03
let OS allocate port for discovery server

daiagi:
let OS allocate port for Discvery server

stress tested a 100 rounds of openChannel, no collision observed


cyony95 (approved):
looks reasonable to have this for the discovery server.

PR #239, opened 2025-12-05
Added E2E tests for the spectate RPC

cyony95:
- For new channels (where originForkId == 0), getGenesisTimestamp() returns isAvailable=false because no dispute window exists yet, causing spectate sync to fail. Added a special case in Step 2.6 (genesis snapshot verification)

- Spectators were being disconnected and blacklisted when receiving blocks before completing the initial spectate sync, because validation failed. Modified all validation methods to return BlockValidationResult.NOT_READY and queue blocks if the spectator hasn't synced yet. All incoming blocks are queued until the initial spectate sync completes, preventing premature disconnections.

- When validation failed (keepConnection = false), the code called disconnectAndBlacklistPeer() for all peers, including spectators. This blacklisted spectators before they could complete the spectate sync. Added a check to skip blacklisting for non-participants. Spectators are not blacklisted on validation failures, allowing them to reconnect and complete the spectate sync.

### DIFF
--- a/contracts/V1/StateChannelDiamondProxy/StateChannelCommon.sol
+++ b/contracts/V1/StateChannelDiamondProxy/StateChannelCommon.sol
@@ -76,21 +76,27 @@ contract StateChannelCommon is StateChannelManagerStorage, StateChannelManagerEv
     {
         DisputeData storage _disputeData = disputeData[channelId];
         DisputeWindow storage disputeWindow = _disputeData.disputeWindowMap[originForkId];
+
+        // Check if dispute window exists first
+        if (disputeWindow.evidence.creationTimestamp == 0) {
+            // Dispute window doesn't exist - check on-chain snapshot for new channels
+            if (originForkId == bytes32(0)) {
+                StateSnapshot memory currentOnChainSnapshot = stateSnapshots[channelId];
+                // check if current on-chain snapshot.fork == forkId
+                if (
+                    currentOnChainSnapshot.forkId == forkId
+                        && UtilityFacet(utilityFacetAddress).isGenesisSnapshotWithoutTimeCheck(currentOnChainSnapshot)
+                ) {
+                    return (true, currentOnChainSnapshot.timestamp);
+                }
+            }
+            return (false, 0);
+        }
+
+        // Dispute window exists - check if kill period expired
         timestamp = disputeWindow.evidence.lastEvidenceSubmissionTimestamp + getEvidenceTime();
         (bool isExpired,) = _isKillPeriodExpired(disputeWindow, getEvidenceTime());
         if (!isExpired) {
-            return (false, timestamp);
-        }
-        if (timestamp == 0) {
-            // Dispute window doesn't exist
-            StateSnapshot memory currentOnChainSnapshot = stateSnapshots[channelId];
-            // check if current on-chain snapshot.fork == forkId
-            if (
-                currentOnChainSnapshot.forkId == forkId
-                    && UtilityFacet(utilityFacetAddress).isGenesisSnapshotWithoutTimeCheck(currentOnChainSnapshot)
-            ) {
-                return (true, currentOnChainSnapshot.timestamp);
-            }
             return (false, timestamp);
         }
         return (true, timestamp);
--- a/src/rpc/services/spectate/SpectateRpcMethods.ts
+++ b/src/rpc/services/spectate/SpectateRpcMethods.ts
@@ -9,6 +9,7 @@ import { ChannelId, Timestamp } from "@/types/types";
 import Clock from "@/Clock";
 import { Codec, hash, Type } from "@/utils";
 import { Block } from "@/models";
+import { ethers } from "ethers";
 
 class SpectateServiceRpcMethods extends ARpcMethods {
     service: SpectateService;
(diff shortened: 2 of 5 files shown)


src/rpc/services/spectate/SpectateRpcMethods.ts
```diff
@@ -190,18 +191,30 @@ class SpectateServiceRpcMethods extends ARpcMethods {
...
                         .originForkId,
                     finalForkId
                 );
-            isCorrectGenesis =
-                isCorrectGenesis &&
-                isAvailable &&
-                genesisTimestamp ==
-                    syncPayload.latestForkGenesisSnapshot.timestamp;
+
+            if (isNewChannel && !isAvailable) {
```
cyony95: isAvailable=false here, so it stops the pipeline

cyony95: getGenesisTimestamp was checking first if the killPeriod expired and then if the disputeWindow exists. I changed the order so first check naturally that  disputeWindow exists and only then check the killPeriod has expired.

cyony95: however, this doesn't handle the case where the snapshot is not the genesis. we'll come back to it.

contracts/V1/StateChannelDiamondProxy/StateChannelCommon.sol
```diff
@@ -76,23 +76,29 @@ contract StateChannelCommon is StateChannelManagerStorage, StateChannelManagerEv
     {
         DisputeData storage _disputeData = disputeData[channelId];
         DisputeWindow storage disputeWindow = _disputeData.disputeWindowMap[originForkId];
+
+        // Check if dispute window exists first
+        if (disputeWindow.evidence.creationTimestamp == 0) {
+            // Dispute window doesn't exist - check on-chain snapshot for new channels
+            if (originForkId == bytes32(0)) {
```
lukastanisic99, line 83: This origin check is not needed here, apart from that - this is good

src/rpc/services/stateTransition/StateTransitionRpcMethods.ts
```diff
@@ -22,8 +22,21 @@ class StateTransitionRpcMethods extends ARpcMethods {
...
+                const senderAddress = profile?.evmAddress;
+                let isParticipant = false;
+                if (senderAddress) {
+                    const participants =
+                        await this.p2pManager.stateManager.getParticipantsCurrent();
+                    isParticipant = participants.includes(senderAddress);
+                }
+
+                if (!isParticipant) return;
+
```
lukastanisic99, line 39: This for sure shouldn't be the case - spectators shouldn't send blocks -> this shouldn't change. You need to find the root of the problem - why are `honest` spectators sending blocks? The pipeline behaves as expected and disconnects from them as it should since they're not part of the leader election process

src/stateManager/validationStrategy/SpectatingValidationStrategy.ts
```diff
@@ -128,24 +181,48 @@ export default class SpectatingValidationStrategy extends AValidationStrategy {
...
 
     private disconnect() {
         this.p2pManager.disconnectAll();
     }
+
+    private isSynced(): boolean {
+        const forkId = this.p2pManager.stateManager.forkId;
+        if (!forkId) return false;
+        return this.storage.blocks.getNextBlockHeight(forkId) > 0;
+    }
```
lukastanisic99, line 227: You need to think more about this entire file - My quick though: future blocks should be queued under this strategy and maybe subjective failures. Any objective failure should result in disconnecting and aborting

src/rpc/services/stateTransition/StateTransitionRpcMethods.ts
cyony95: yes, one of the reasons I found is that the spectator broadcasts onBlockConfirmation, I dont know why that is the case but it shouldn't(I mean I know why, it s coded in the strategy). This part can be safely removed, you re right.

src/stateManager/validationStrategy/SpectatingValidationStrategy.ts
cyony95: you are correct, I haven't gotten yet on fixing this one after they meeting yesterday.

test/e2e/E2E-RPC.test.ts
```diff
@@ -1186,57 +1195,394 @@ describe("E2E: RPC Services", function () {
...
+                        ).signature as BytesLike
+                )
+            );
+
+            const tx = await harness.channelManager.open({
+                encodedOpenChannel: Codec.encode(openChannel, Type.OpenChannel),
+                signatures: signatures
+            });
+
+            await Promise.all([tx.wait(), sleep(100)]);
```
lukastanisic99, line 1237: open the channel with 3 peers as normal

```typescript
await harness!.setup(3);
const forkId = await harness!.openChannel();
```

progress the channel a bit 

```typescript
await harness!.submitNextTransaction((contract) => contract.add(1));
await harness!.submitNextTransaction((contract) => contract.add(2));
.
.
.
```

on 4th signer 
spectator = await createPeer(4thIndex, 4thSigner)
spectator.p2pInstance.p2pSigner.connectToChannel(channelId)

This should establish all the connections, so somewhere after the `initHandsake` we should have a check `if channel opened on-chain && i'm not participating -> spectateService.sync(...)`  

Currently in tests we `negotiate the channel open for everyone with `createOpenChannelTestObject`, there will be an RPC where they can negotiate it - I'll design it later and in general our RPCs will be made extendable, so a consumer implementation can inherit all our RPCs and add its own - all the entrypoints / what we expose, should be added to `p2pInstance` or `p2pSigner` (I think it's cleaner to add it to p2pInstance unless it's needed also within the SDK, since the p2pSigner is available there)

Since the spectator is joining already a live channel (peers exchanging many RPCs), many race conditions could happen, but not in these tests, since there is no concurrency and we're creating artificial conditions, so we should also have concurrent peers (WebWorkers) in these tests - you don't have to think about it now and it's something the test framework should be able to support eventually (@daiagi) 

These race conditions are really real in a distributed system which our system is, so I'll have to think a bit more today about this `how are other RPCs affected when spectators join, from the perspective of the spectator and from the perspective of the peers` and is there a good simple design solution so we `don't have to think too much, after adding an RPC`

daiagi:
outdated. closing

PR #241, opened 2025-12-06
L1 message production, L2 message consumption with tests and examples

lukastanisic99:


test/V1/DiamondProxy/StateChannelManager/OpenChannel.test.ts
```diff
@@ -190,6 +193,94 @@ describe("StateChannelManagerProxy", function () {
...
+                await consumerFacet.forceInboundJoin.staticCall(
+                    openChannel.channelId,
+                    thirdSigner.address,
+                    forcedAmount
+                );
+            await consumerFacet.forceInboundJoin(
+                openChannel.channelId,
+                thirdSigner.address,
+                forcedAmount
+            );
```
lukastanisic99, line 240: This is the first unit test that I like - created by magic

src/models/Block.ts
```diff
@@ -157,6 +158,10 @@ export default class Block {
         return this.block.transaction;
     }
 
+    get messageBlocks(): MessageBlockStruct[] {
+        return this.block.messageBlocks ?? [];
```
daiagi: uneeded `?? []`

src/stateManager/validationStrategy/AValidationStrategy.ts
```diff
@@ -43,6 +46,11 @@ export default abstract class AValidationStrategy {
         block: Block
     ): Promise<BlockValidationResult>;
 
+    public abstract forgedInboundMessageBlockDetected(
```
daiagi, line 49: need to implement it in CalldataCommittedStrategy

src/stateManager/StateManager.ts
```diff
@@ -1379,37 +1493,52 @@ class StateManager {
     }
 
     private async createStateSnapshot(
+        previousStateSnapshot: StateSnapshot,
         stateMachineStateHash: Hash,
         coordinates: BlockCoordinates,
         timestamp: Timestamp,
-        outboundMessages: MessageStruct[]
+        outboundMessages: MessageStruct[],
+        options: {
```
daiagi: not realy options if they are mandatory
only optional here is totalDepositsOverride
and that might as well be flat. i.e.  not wrapped in an "options" object
and finally all the calling context are passing the totalDepositsOverride. so why is it optional to begin with?

src/stateManager/StateManager.ts
```diff
@@ -1379,37 +1493,52 @@ class StateManager {
...
-        const latestInboundMessageBlockHeight = BigInt(
+        let latestInboundMessageBlockHeight = BigInt(
             previousSnapshotData.latestInboundMessageBlockHeight ?? 0n
         );
-        const participants = await this.diamondStateMachine.getParticipants();
+        const participants = options.participants;
+
+        if (options.inboundMessageBlocks.length > 0) {
+            const lastInboundBlock =
+                options.inboundMessageBlocks[
```
daiagi: a bit messy to pass an entire potential array just to take the last element

src/stateManager/StateManager.ts
```diff
@@ -1379,37 +1493,52 @@ class StateManager {
     }
 
     private async createStateSnapshot(
+        previousStateSnapshot: StateSnapshot,
```
daiagi: why?
previousStateSnapshot is a function of cooridantes. so why add it as an additional argument?

src/stateManager/StateManager.ts
```diff
@@ -1459,9 +1588,129 @@ class StateManager {
         };
     }
 
+    private getPreviousStateSnapshotOrThrow(
```
daiagi, line 1581: better design would be to pass some flag to storage, changing the behvior to throw rather then return undefined
or define `getPreviousStateSnapshot!` which does that

src/stateManager/StateManager.ts
```diff
@@ -1459,9 +1588,129 @@ class StateManager {
...
+        const previousStateSnapshot =
+            this.storage.getPreviousStateSnapshot(coordinates);
+        if (!previousStateSnapshot)
+            throw new Error(
+                "createStateSnapshot for block - previousStateSnapshot undefined"
+            );
+        return previousStateSnapshot;
+    }
+
+    private getPendingInboundMessageBlocks(
```
daiagi, line 1593: this also, i think is more suitable as a storage function of cooridnates, or s stateSnapshot hash

src/stateManager/StateManager.ts
```diff
@@ -1459,9 +1588,129 @@ class StateManager {
...
+            return [];
+        }
+
+        return this.storage.inboundMessages.getMessageBlocksInRange(
+            latestStoredHash,
+            previousHash ?? ethers.ZeroHash
+        );
+    }
+
+    private findBrokenInboundMessageChainBlock(
```
daiagi, line 1615: unclear name
what is a  "message chain block"?

src/stateManager/StateManager.ts
```diff
@@ -1459,9 +1588,129 @@ class StateManager {
...
+            );
+        }
+
+        return undefined;
+    }
+
+    private async applyInboundMessageBlocksToState(
+        inboundMessageBlocks: MessageBlockStruct[],
+        totalDeposits: BalanceStruct,
+        encodedState?: Bytes
```
daiagi: no need for optional, all call sites pass it

src/stateManager/StateManager.ts
```diff
@@ -1459,9 +1588,129 @@ class StateManager {
...
+            return {
+                encodedState:
+                    encodedState ?? (await this.diamondStateMachine.getState()),
+                totalDeposits: updatedTotalDeposits
+            };
+        }
+
+        for (const messageBlock of inboundMessageBlocks) {
+            for (const message of messageBlock.messages) {
+                const processed =
```
daiagi, line 1663: do i understand correctly?
processInboundMessage will process the `data` field
and the later `addBalance is the uniform accounting for all message types

test/V1/DiamondProxy/StateChannelManager/OpenChannel.test.ts
```diff
@@ -190,6 +193,94 @@ describe("StateChannelManagerProxy", function () {
...
+            const parsedOpenEvent = openReceipt?.logs
+                .map((log) => {
+                    try {
+                        return mathChannelManager.interface.parseLog(log);
+                    } catch (error) {
+                        return null;
+                    }
+                })
+                .find((parsed) => parsed && parsed.name === "ChannelOpened");
+            if (!parsedOpenEvent) {
```
daiagi: use `expect`

test/V1/DiamondProxy/StateChannelManager/OpenChannel.test.ts
```diff
@@ -190,6 +193,94 @@ describe("StateChannelManagerProxy", function () {
...
+            const openReceipt = await openTx.wait();
+
+            const parsedOpenEvent = openReceipt?.logs
+                .map((log) => {
+                    try {
+                        return mathChannelManager.interface.parseLog(log);
+                    } catch (error) {
+                        return null;
+                    }
+                })
```
daiagi, line 223:
```suggestion
                .map(mathChannelManager.interface.parseLog)
```

test/e2e/E2E-Core.test.ts
```diff
@@ -578,6 +582,74 @@ describe("E2E: Core Functionality", function () {
         });
     });
 
+    describe("Forced Inbound Joins", function () {
+        it("should propagate forced joins through event listeners and state updates", async function () {
+            await harness!.setup(3);
+            await harness!.openChannel();
+
+            const mathConsumerFacet = MathConsumerFacet__factory.connect(
```
daiagi, line 556: this is something the harness should provide an easy API for

daiagi (commented):
Didn't review Solidity yet

src/stateManager/StateManager.ts
lukastanisic99: yes, so the processInboundMessage is for the `internal stateMachine` accounting and STF - regardless of the internal accounting, we keep track `outside` of all the amounts that have been deposited in a general way. Mostly to not have to do `stateMachineImplementation.STF(...)` on-chain when creating the message, so we just do a cheap addition.

lukastanisic99:
@daiagi push some suggestions from above to the other PR, since it builds on top of this one

contracts/V1/StateChannelDiamondProxy/StateChannelManagerProxy.sol
```diff
@@ -53,6 +53,13 @@ contract StateChannelManagerProxy is StateChannelManagerInterface, StateChannelC
         gasLimit = 3_000_000;
     }
 
+    fallback() external {
```
daiagi, line 56: what for?
not used anywhere

daiagi (approved):

contracts/V1/StateChannelDiamondProxy/StateChannelManagerProxy.sol
lukastanisic99: Remind me tomorrow on the call to explain if it's still unclear

lukastanisic99: Remind me tomorrow on the call to explain if it's still unclear

PR #242, opened 2025-12-07
MathStateChannelManagerProxy removal - part of old implementation

lukastanisic99:
In the old implementation we forced the user to extend our base `StateChannelManagerProxy` with its own specific logic. This wasn't scalable (contracts are limited in size), wasn't clean, wasn't upgradable -> in the new implementation the use-case specific logic is in a user defined facet that extends `AConsumerFacet`. This is also in line with the direction that Solidity is taking and moving away from `inheritance` and suggesting to use `composition`.

How we'll migrate the rest of the implementation, since we use a lot of inheritance - TBD since it's not a priority at the moment


daiagi (approved):
Actually
`deployContractProxy` is not used anywhere
see 
- #244 
continuation of this PR

PR #245, opened 2025-12-08
Dispute-e2e-tests

daiagi:
the tests:
`should reject timeout dispute when accused participant is not next to write`
`should reject dispute when auditing data is partial and state proof invalid`

pass, but they trigger `Error killing dispute`

https://github.com/peer3to/state-channels-plus/actions/runs/20030357535/job/57438039686#step:8:107


lukastanisic99:
@daiagi pushed fixes - take a look 

Apart the logger stuff, the bugs were simple:
1) `!=` -> `==`
2) `isSubset` had wrong order `isSubest(X,Y)` - > `isSubset(Y,X)`

So the dispute was killed successfully, which appended the committer to on-chain slashes and triggered a new dispute that would include them in the slashes set -> the new dispute was committed correctly and as all disputes it goes through the dispute confirmation pipeline and since isSubest had a wrong order it triggered a dispute fraud proof incorrectly which the chain correctly rejected

PR #251, opened 2025-12-12
more dispute valiaditon tests

daiagi:


contracts/V1/StateChannelDiamondProxy/UtilityFacet.sol
```diff
@@ -300,7 +300,12 @@ contract UtilityFacet {
                 }
             } else {
                 //check if signedBlocks are linked, signed and built on genesis
-                if (!_areSignedBlocksLinkedAndVerified(dispute.input.stateProof.signedBlocks, dispute.input.forkId)) {
+                // HACK:Pass bytes32(0) to skip "linked to genesis" check for the first block
+                // propsoed solution:  convert first block's `previousBlockHash`  = forkId = keccak256(abi.encode(genesisSnapshotData))
+
+                bool linkedAndVerified =
+                    _areSignedBlocksLinkedAndVerified(dispute.input.stateProof.signedBlocks, bytes32(0));
```
lukastanisic99, line 307: After thinking for a bit - it makes sense to have `block.prevousBlockHash == hash(stateSnapshot)` in the case of genesis and `not hash(snapshotData)` - the forkID is available in all blocks regardless. The hash of the snapshot is useful, since it's a commitment on time, not just data.

I did make a mistake here of passing forkId, but the `genesisSnapshot` is available in this context from `auditing data`, so just hash that.

evainfeld (approved):

evainfeld (approved):

daiagi:
`rpc-guards` is becoming the new `dispute` branch
merging

PR #252, opened 2025-12-12
Rpc guards

lukastanisic99:
RPC Guards to protect RPC endpoints from processing RPCs before all conditions are satisfied. The guards are modular and services can easily reuse them. 

Also some bug fixes from before. 

Next is to test the RPCs and add more guards as needed.


src/rpc/guards/HandshakeCompletedGuard.ts
```diff
@@ -0,0 +1,50 @@
...
+import ATransport from "@/transport/ATransport";
+import { AGuard } from "@/rpc/guards/AGuard";
+import ARpcService from "@/rpc/ARpcService";
+import ARpcMethods from "@/rpc/ARpcMethods";
+
+export interface HandshakeCompletedGuardOptions {
+    onFailure?: (rpc: Rpc, transport: ATransport) => void;
+}
+
+export class HandshakeCompletedGuard extends AGuard<ARpcService<ARpcMethods>> {
```
daiagi, line 41: Any plans to add more guards?
how many more?
an abstract class just for a single guard?

src/rpc/guards/HandshakeCompletedGuard.ts
```diff
@@ -0,0 +1,50 @@
...
+        const profile =
+            this.service.p2pManager.profileManager.getProfileByTransport(
+                transport
+            );
+        const completed = profile?.getIsHandshakeCompleted() ?? false;
+        if (completed) {
+            return true;
+        }
+
+        return false;
```
daiagi: or simply
```ts
return profile?.getIsHandshakeCompleted() ?? false
```

src/rpc/guards/runGuards.ts
```diff
@@ -0,0 +1,23 @@
+import { AGuard } from "@/rpc/guards/AGuard";
+import Rpc from "@/rpc/Rpc";
+import ATransport from "@/transport/ATransport";
+
+/**
+ * Runs guards sequentially.
```
daiagi, line 6: I'm not a  fan of leaving these AI comments in the code
this comment is very "[duh](https://dictionary.cambridge.org/dictionary/english/duh)"

this here is a small, clear, self documenting function
comments explaining such a trivial thing just add vertical space and visual noise to code base

daiagi (approved):

src/rpc/guards/HandshakeCompletedGuard.ts
lukastanisic99: I'm adding isSynced to some endpoints when I submit the spectate test PR + I want to make RPC services extendable, so the consumer can add their services and these guards have a simple interface if they require protection on their endpoints

Later we might have different algos for rate limiting dependent on the RPC endpoint - a critical one like `stateTransitions` will be allocated higher bandwidth/priority while a non-critical one like "messages" can be cut under load

src/rpc/guards/HandshakeCompletedGuard.ts
lukastanisic99: Ye... not sure why magic likes adding extra stuff - maybe it thinks it's more readable

daiagi: because that's how it is

daiagi:
@lukastanisic99  fixes the compilation error and a failing test
there is still a failing e2e test

lukastanisic99:
> @lukastanisic99 fixes the compilation error and a failing test there is still a failing e2e test

Ok great - if it works for you for the poker stuff use that - I'll fix the remaining stuff tomorrow

lukastanisic99:
@daiagi please take a look why tests fail on the CI - there is some flakyness still

daiagi:
If I remember correctly, It 1 spectate test  that fails consistently on ci and local

lukastanisic99:
> If I remember correctly, It 1 spectate test  that fails consistently on ci and local 

On local for me it all passes even with parallel - is it different for you?

lukastanisic99:
@daiagi github CI tests are so flaky - I have to run then multiple times to get lucky

Parallel tests work well unless max cores are used, since the OS starves some processes and they don't regain context in time and by the time they resume execution the time has elapsed enough for the test to fail. If there is no starvation it should always pass deterministically. Github is running tests sequentially and it seems similar things are happening there since random tests randomly fail

PR #256, opened 2025-12-17
Snapshot update

lukastanisic99:
Snapshot update - same fork. There are some bug fixes in `verifyMilestones` so pushing now, before I also do the snapshot update on different forks - @daiagi take a look maybe you observed some of the same bugs


lukastanisic99:
@daiagi why do we keep `GeneratedArtifacts` in source if they're generated? The expectation is someone will run `yarn compile` and for `builds` they can be included if needed so `.npmignore` and `.gitignore` don't have to be the same

daiagi:
> @daiagi why do we keep `GeneratedArtifacts` in source if they're generated? The expectation is someone will run `yarn compile` and for `builds` they can be included if needed so `.npmignore` and `.gitignore` don't have to be the same

yes
i thought the same:

https://github.com/peer3to/state-channels-plus/pull/251/changes/4b5bf5890049809ea1ad049d33102167bb44ab98

lukastanisic99:
@daiagi added tests and fixed bugs for snapshot update both same fork and when forks need to change and a test with multicall - most of the tests are combinations of other tests, so we might want to refactor at some point

The mock tests I don't know what's going there - magic did some fixes and it was passing, so I didn't look too much

contracts/V1/StateChannelDiamondProxy/UtilityFacet.sol
```diff
@@ -335,30 +336,38 @@ contract UtilityFacet {
...
+    ) public pure returns (bool isFinal, bytes32 finalizedSnapshotHash) {
+        address[] memory expectedParticipants = thresholdSnapshotData.participants;
         address[] memory thresholdSet = new address[](expectedParticipants.length);
         uint256 thresholdCount = 0;
         bytes memory previousEncodedBlock;
         BlockConfirmation memory currentBlockConfirmation;
         Block memory currentBlock;
         address adr;
         bool isValid;
+        console.log("_isMilestoneFinal: forkId");
```
daiagi, line 357: please remove all the console logs

src/stateManager/StateManager.ts
```diff
@@ -1039,6 +1052,30 @@ class StateManager {
         const forkData = await this.prepareUpdateStateSnapshotFork();
         const sameForkData = await this.prepareUpdateSnapshotSameFork(forkId);
 
+        if (forkData) {
+            this.logger.debug(
```
daiagi, line 1149: As mentioned before,I don't think you should leave all these logs in the code once you are done finding whatever bug you are looking for.

test/e2e/E2E-Core.test.ts
```diff
@@ -156,6 +157,156 @@ describe("E2E: Core Functionality", function () {
 
             harness!.assertAllPeersInSync();
         });
+
+        // Arrange: Setup 3 participants, open channel, execute 3 state transitions
+        // Act: Post the latest state snapshot on-chain
+        // Assert: All peers observe StateSnapshotUpdated and on-chain snapshot matches latest local snapshot
+        it("should post updated state snapshot on-chain after 3 transitions", async function () {
```
daiagi, line 164: This test would be better if it was from the perspective of a user story
a peer never "posts snapshot", they join /exit the channel. advance the state observe disputes etc.
it would be better if the test was something like 

Act: a peer joins the channel
Assert: on chain snapshot is updated

test/e2e/E2E-Security.test.ts
```diff
@@ -386,6 +387,495 @@ describe("E2E: Advanced Security", function () {
                 "Removed peer should NOT receive next turn"
             );
         });
+
```
daiagi, line 351: // Arrange: Malicious participant removed via fork reduction after dispute
// Act: Post two independent state snapshot updates on the reduced fork
// Assert: Both snapshot updates succeed and on-chain snapshot reflects correct state

test/e2e/E2E-Security.test.ts
```diff
@@ -386,6 +387,495 @@ describe("E2E: Advanced Security", function () {
...
+            const maliciousPeer = harness!.peers[2];
+            const honestPeers = [
+                harness!.peers[0],
+                harness!.peers[1],
+                harness!.peers[3]
+            ];
+            const maliciousIndex = maliciousPeer.index;
+            const honestIndices = honestPeers.map((peer) => peer.index);
+
```
daiagi: // Trigger dispute by having malicious peer submit invalid block

test/e2e/E2E-Security.test.ts
```diff
@@ -386,6 +387,495 @@ describe("E2E: Advanced Security", function () {
...
+                "onDisputeCommitted",
+                harness!.peers.map((peer) => ({
+                    peerId: peer.index,
+                    expectedCount: 3
+                })),
+                8000,
+                { mode: "atLeast" }
+            );
+            expect(disputesCommitted).to.be.true;
+
```
daiagi: // Wait for fork reduction to complete

test/e2e/E2E-Security.test.ts
```diff
@@ -386,6 +387,495 @@ describe("E2E: Advanced Security", function () {
...
+                return allMoved && uniqueForks.size === 1;
+            }, 10000);
+            expect(forkSettled).to.be.true;
+
+            const newForkId = honestPeers[0].stateManager.forkId;
+            expect(newForkId).to.not.equal(
+                originalForkId,
+                "Expected to be on a new fork after reduction"
+            );
+
```
daiagi, line 383: // Verify malicious participant is removed

test/e2e/E2E-Security.test.ts
```diff
@@ -386,6 +387,495 @@ describe("E2E: Advanced Security", function () {
...
+            for (const peer of honestPeers) {
+                const participants =
+                    await peer.stateManager.diamondStateMachine.getParticipants();
+                expect(participants).to.have.lengthOf(honestPeers.length);
+                expect(participants).to.not.include(maliciousPeer.address);
+            }
+
+            // Ensure the on-chain snapshot is moved onto the reduced fork first.
```
daiagi, line 384:
```suggestion
            // Act - First snapshot update: Sync on-chain snapshot to reduced fork
```

test/e2e/E2E-Security.test.ts
```diff
@@ -386,6 +387,495 @@ describe("E2E: Advanced Security", function () {
...
+                await harness!.channelManager.getStateSnapshot(
+                    harness!.channelId
+                )
+            );
+            expect(onChainSnapshotAfterForkSync.forkId).to.equal(
+                newForkId,
+                "On-chain snapshot should be on the reduced fork before same-fork updates"
+            );
+
+            // From here, do the same 3 transitions as E2E-Core and post a same-fork snapshot update.
```
daiagi, line 396:
```suggestion
            // Advance state with additional transactions
```

src/stateManager/StateManager.ts
lukastanisic99: @daiagi Will come back to other comments later, but to comment on this one. My plan is to have as much useful logging as possible and to leave it there for a testnet and whenever some error occurs, we want the peers to send us their entire log, so we understand what's going on and why things failed - to understand better how things perform in the real world, so no we should not remove these logs if they're useful. You can improve them if you have better ideas. They don't affect performance nor are persisted (don't bloat the disk) and when things work you don't look at them. When things don't work you wish they're there.

PR #257, opened 2025-12-19
Custom RPC support

lukastanisic99:
This introduces a type safe way of allowing the consumer of the SDK to add their own RPC endpoints(services), while granting them access to interact with RPC endpoints(services) that are implemented within the SDK.

This is allows us to be less opinionated and to not implemented everything in the SDK related to communication. Mostly doing it now, since the design team always wants some features that are not security concerns but are good for UX (e.g. chat...).


test/e2e/E2E-PingService.test.ts
```diff
@@ -0,0 +1,260 @@
...
+import { ATransport } from "@/transport";
+import type P2PManager from "@/P2PManager";
+
+type PingPongFactories = {
+    pingService: (p2pManager: P2PManager<PingPongFactories>) => PingService;
+    relayService: (p2pManager: P2PManager<PingPongFactories>) => RelayService;
+};
+
+type PingPongP2PManager = P2PManager<PingPongFactories>;
+
```
lukastanisic99, line 18: This file shows how 2 custom RPCs are added and used

daiagi (approved):

PR #259, opened 2025-12-23
Logs

daiagi:


lukastanisic99:
I'll see to review this one today - my quick thoughts on using indexDB for this: I wouldn't use it since we're not actually interested in persisting logs on the client and we want to use indexDB for things we're interested in persisting + the browser limits us on the disk space, so better not have our logs eat up that space or trigger the browser to prune some other things we actually want persisted. I'd keep it in a simple in memory variable (works cross platform) that we append to and occasionally serialize and compress - that way tests are deterministic and simple cross platform and good enough for now

daiagi:
OK, I'll simplify to keep in memory.
That will reduce a big chunk of code. 

Edit:

Done

lukastanisic99:
I never actually finished the review of this, but here are my notes I made while on some plane without internet 

--------

Not a big fan of the StructuralLogger - there are some interesting ideas that I like, but feels like over engineering and I don’t like the new usage of it vs `.child()` from the previous design  
I like the traceId idea - it could be useful in some cases and it works well with .child().   The general problem in both designs is that the context is not updatable (I had problems with log level being fixed to the singleton at creation), so if the channelId changes you’d have to create a new logger everywhere. This is not a problem for standalone loggers when you want to use traceId for a nested execution, but in general for persisted loggers that are part of `someManager` or other relevant classes that get their own .child().   So in my opinion .child() should bi-directionally link the 2 instances `parent` and `children[ ]` so an update to the context in any logger can cascade the update across all loggers that are connected.   This should also be an `option` so someone can do .child({standalone = true}) which just creates the logger that’s not connected, which is useful for the traceId   I don’t like polluting the logger with blocks and other structs. What I think is better design when we want to print our structs to define a utils/LogFormatter that’s filled with stateless functions that take in a struct and return a `message` - not all fields in the struct are useful for debugging so we just extract / transform / format the useful ones. This way we can combine messages if needed  e.g. `logger.debug(log(block) + log(snapshot))` so the logger stays simple and just logs the message   What I also like is the idea behind logging tx execution into different levels based on `success` -> can also be a simple stateless function

daiagi:
outdated pr, closing

PR #260, opened 2025-12-24
Spectate test

lukastanisic99:
`Tests for spectating` - happy case - deviation from this should abort - needs to be tested

`InitHandShake Ack:`
Needed for the HandshakeGuard. Both peers do the handshake independently from their end and once they verify it, they acknowledge (ack) to the other peer completing the handshake. Only once the handshake is completed, does the connection get added to the p2pManager and used within the system. In reality the connection exists before the handshake completes, but honest peers won't use it to send messages. The handshake guard protects RPC endpoints, by just observing that the handshake is verified on our end (allowing receiving messages from the transport). This is safe, since there is no honest race condition, since our verification comes before our `ack`, so no honest disconnects will come from the handshake guard.

There were still some bugs and race conditions else where. The RPC test for the handhsake are sometimes flaky since they're now incorrect and to be honest I don't know why they're even passing - they need a rewrite. The handshake logic is tested in other places and for now it works as expected. 

One of the bigger bugs that we're not easy to spot was that our internal `Block and Snapshot models` were assignable to their typechain generated structures. They don't serialize in the same way, so a failure at runtime is detected, even tough typescript doesn't complain.


src/evm/P2pSigner.ts
```diff
@@ -135,15 +136,23 @@ class P2pSigner<TFactories extends RpcServiceFactoryMap = {}>
         return this.isLeader;
     }
 
-    public connectToChannel(channelId: Bytes) {
+    public async connectToChannel(channelId: Bytes) {
         this.setChannelId(channelId);
+
+        // Update status to NOT_OPENED/OPENED as soon as we know the channelId.
+        await this.p2pManager.stateManager.refreshOpenedStatusFromChain();
```
daiagi, line 132: I'm not following
1. the name: what does the function has to do with the chain?
2. don't you want to pass in the channel id?
3. won't it always set to`NOT_OPENED`?

src/evm/P2pSigner.ts
```diff
@@ -135,15 +136,23 @@ class P2pSigner<TFactories extends RpcServiceFactoryMap = {}>
...
     }
 
     public disconnectFromPeers() {
         this.p2pManager.disconnectAll();
     }
 
+    public async getChannelStatus(): Promise<Status> {
+        return await this.p2pManager.stateManager.getChannelStatus();
+    }
+
```
daiagi: Same same, but different ;)
```suggestion
    public getChannelStatus(): Promise<Status> {
        return this.p2pManager.stateManager.getChannelStatus();
    }

```

src/rpc/guards/HandshakeCompletedGuard.ts
```diff
@@ -41,10 +36,23 @@ export class HandshakeCompletedGuard extends AGuard<ARpcService<ARpcMethods>> {
...
+            "No peer profile for transport; disconnecting",
             {
-                peerAddress
+                peerAddress,
+                service: rpc.service,
+                method: rpc.method
             }
         );
+
+        // With the new connection semantics, any guarded RPC over an unverified
```
daiagi: avoid the word `new` in comments
new is only relevant during your code changes. a week from now, `new` means nothing

src/rpc/services/WebRTCSetup/WebRTCSetupService.ts
```diff
@@ -36,29 +36,30 @@ class WebRTCSetupService extends ARpcService<WebRTCSetupRpcMethods> {
                 this.p2pManager
             );
 
+            const adr =
+                this.p2pManager.profileManager.getProfileByTransport(
+                    transport
+                )?.evmAddress;
+            if (!adr)
+                return this.logger.error("initiateWebRTC - no EVM address");
```
daiagi: that's not a great return value
```
(method) WebRTCSetupService.initiateWebRTC(transport: ATransport): Promise<winston.Logger | undefined>
```

src/rpc/services/initHandshake/InitHandshakeRpcMethods.ts
```diff
@@ -5,6 +5,7 @@ import { Hash, Signature, Timestamp } from "@/types/types";
 import { ethers } from "ethers";
 import InitHandshakeService from "./InitHandshakeService";
 import PeerProfile from "@/PeerProfile";
+import { Status } from "@/types";
```
daiagi: unused

src/rpc/services/spectate/SpectateRpcMethods.ts
```diff
@@ -9,6 +9,8 @@ import { ChannelId, Timestamp } from "@/types/types";
 import Clock from "@/Clock";
 import { Codec, hash, Type } from "@/utils";
 import { Block } from "@/models";
+import { Status } from "@/types";
+import { BlockStruct } from "@/index";
```
daiagi: unused

src/rpc/services/spectate/SpectateRpcMethods.ts
```diff
@@ -27,38 +38,48 @@ class SpectateServiceRpcMethods extends ARpcMethods {
...
     }
 
     public async onSpectateResponse(
         channelId: ChannelId,
         syncPayload: SyncPayload
     ) {
-        try {
-            this.service.logger.debug(`onSpectateResponse - start`);
-            const senderTransport = this.senderTransport;
+        this.service.logger.debug(`Syncpayload received`, { syncPayload });
```
daiagi:
```suggestion
        this.service.logger.debug(`Sync payload received`, { syncPayload });
```

daiagi (commented):
Partial review. more to follow

test/setup.ts
```diff
@@ -4,4 +4,7 @@ import Clock from "@/Clock";
 // Global test setup to ensure Clock is always initialized
 before(async function () {
     await Clock.init(ethers.provider);
+
+    // Ensure the chain has enough history for tests that query logs "N blocks back".
+    await ethers.provider.send("hardhat_mine", ["0x64"]); // 100 blocks
```
daiagi, line 9: why use hex? why not simply  "100"?

test/e2e/E2E-Security.test.ts
```diff
@@ -292,7 +291,7 @@ describe("E2E: Advanced Security", function () {
         it("should remove malicious participant after fork and keep liveness", async function () {
             await harness!.setup(4, {
                 timeConfig: {
-                    p2pTime: 3,
+                    p2pTime: 30,
```
daiagi, line 294: how is this sensible for tests? 30 seconds??

test/e2e/E2E-Spectate.test.ts
```diff
@@ -0,0 +1,288 @@
...
+    ): boolean => {
+        const profile =
+            ownerPeer.stateManager.p2pManager.profileManager.getProfileByEvmAddress(
+                counterparty.address
+            );
+        return profile?.getIsHandshakeCompleted() ?? false;
+    };
+
+    it("should NOT allow spectate RPC before handshake completes", async function () {
+        this.timeout(30000);
```
daiagi: was the dafult of 40_000 not good? why?

test/e2e/E2E-Spectate.test.ts
```diff
@@ -0,0 +1,288 @@
...
+        const onChainSnapshot = StateSnapshot.from(
+            await h.channelManager.getStateSnapshot(h.channelId)
+        );
+        expect(onChainSnapshot.forkID).to.equal(forkId);
+    });
+
+    it("should spectate successfully even when it must traverse forks (dispute -> reduced fork)", async function () {
+        await harness!.setup(5, {
+            timeConfig: {
+                p2pTime: 30,
```
daiagi, line 185: same question

PR #261, opened 2026-01-04
Browser support with some fixes

lukastanisic99:
Working in browsers, everything except ganache console @daiagi take a look and see how we can just use the logging part of that precompile or something similar.

Some tests are now failing since I need to update the logic for them + to resolve some address normalization issues - will do that in next PR on top of this so it's not all in one PR

Summary:
The example suggest using `pnpm` instead of `yarn` for better system links, but in reality yarn also works (at least I didn't see a bug with that). What also needs to be tested is `ethers` as a peerDependency - with the system link it for some reason still uses ethers in the SDK `node_modules` even though it should use ethers from the consuming application - In the end I think I just exposed ethers from the SDK and used that within the consumer app - that works. Typechain also has some problems with types if generated with different ethers/hardhat versions. I've added many Trello cards, including that one - what we'll probably end up doing is pass an address instead of the ethers contract instance. To be discussed. Some contract calls within the SDK fail, due to not being able to estimate gas, so I just gave them 5M gas to consume and that works - to be discussed. 

The logger stuff from the other PR we'll chunk into this, so I'll review that PR eventually soon (I started and have some notes) 

I've changed the `config` to be passed as an optional argument to the SDK and to be loadable through 
json, the environment and as an argument override

Added a custom RPC for negotiation, changes to handhsake logic and webRTC initialization, transport upgrade etc etc - @daiagi take a look and we can discuss on the call - all your PR suggestions we'll come at the end as `cleanup`


daiagi:
`build:browser` command in `tic-tac-toe-vite` does not work without a `peer3.config.ts` file
which is not included in this PR

i have added my own to pass that step
assuming yours looks something like:

```ts
const config = {
    PROVIDER_URL: "http://localhost:8545",
    DEBUG_LOCAL_TRANSPORT: true,
    HOLEPUNCH_RELAYER_URLS: ["wss://dht1-relay.leet.ar:49443"]
};

export default config;
```

examples/TicTacToe/tic-tac-toe-vite/README.md
```diff
@@ -22,21 +22,31 @@ Copy the `typechain-types` directory to a place where it's accesable by the user
...
-yarn build:browser
+pnpm run build:browser
 ```
 This will generate the dist directory.
 
 ## Serve the UI
 Use a static server to serve the UI. For example, you can use `http-server`.
 ```shell
-npm install -g http-server
+pnpm add -g http-server
```
daiagi: no reason to install global dep for this
1. add `http-server` to dev dependencies
2. add package.json script `"serve": "http-server dist"`
3. update readme

or use the already existing `"preview": "vite preview"` script

examples/TicTacToe/contracts/TicTacToe/TicTacToeConsumerFacet.sol
```diff
@@ -0,0 +1,56 @@
...
+contract TicTacToeConsumerFacet is AConsumerFacet {
+    uint256 private constant DEFAULT_BET_AMOUNT = 50;
+
+    function openChannelGenesis(JoinChannel[] memory successfulJoinChannels, bytes memory optionalOpeningData)
+        external
+        pure
+        override
+        returns (bytes memory encodedGenesisState, address[] memory participants)
+    {
+        require(successfulJoinChannels.length >= 2, "TicTacToeConsumerFacet: at least 2 participants");
```
daiagi: 1. tic tac tow require exactly 2 partcipants, no?
2. isnt the `>=2` requirement already fulfilled by `StateChannelManagerProxy` : `require(processedJoins.length >= 2, ErrorAtLeastTwoParticipantsRequired())`

examples/TicTacToe/scripts/deployTicTacToeContractsProxy.ts
```diff
@@ -1,100 +1,134 @@
...
+        "Deployed TicTacToeConsumerFacet at ",
+        ticTacToeConsumerFacetAddress
     );
 
-    //Deploy MathStateChannelManager
-    let TicTacToeSmcFactory = await ethers.getContractFactory(
-        "TicTacToeStateChannelManagerProxy",
-        { libraries: { StateChannelUtilLibrary: libraryAddress } }
+    // Deploy core facets + StateChannelManagerProxy via the SDK helper.
+    const timeConfig = getTimeConfig();
```
daiagi: unused `timeConfig`

examples/TicTacToe/contracts/TicTacToe/TicTacToeStateChannelManagerProxy.sol
```diff
@@ -1,112 +1,42 @@
...
-        override
-    {}
-
-    function removeParticipant(bytes32 channelId, bytes[] calldata removeParticipantData, bytes[] calldata signatures)
-        public
-        virtual
-        override
+import "@peer3/state-channels-plus/contracts/V1/StateChannelDiamondProxy/StateChannelManagerProxy.sol";
+
+/**
```
daiagi, line 6: it's less the ideal no?
do we have to require all our customers to  do this wrapper?

examples/TicTacToe/tic-tac-toe-vite/src/stateChannel/rpc/openChannelNegotiation/OpenChannelNegotiationRpcMethods.ts
```diff
@@ -0,0 +1,156 @@
...
+            ? normalizeAddress(this.senderTransport.peerAddress)
+            : undefined;
+        if (!from) return;
+
+        if (this.service.state.negotiatingWith === from) {
+            this.service.resetNegotiation(`remote abort: ${reason}`);
+        }
+    }
+
+    // helper functions moved to the service
```
daiagi:
```suggestion
```

examples/TicTacToe/tic-tac-toe-vite/.gitignore
```diff
@@ -22,3 +22,6 @@ dist-ssr
 *.njsproj
 *.sln
 *.sw?
+
+# Local app-specific Peer3 config overrides
+src/peer3.config.ts
```
daiagi, line 27: qhy not include it?
and if not, then readme should explain what to config.
i was unable to play

examples/TicTacToe/tic-tac-toe-vite/pnpm-lock.yaml
```diff
@@ -0,0 +1,2099 @@
+lockfileVersion: '9.0'
```
daiagi: we need to choose and stick to a single option
either yarn or pnpm

scripts/V1/deploy.ts
```diff
@@ -45,15 +65,33 @@ export async function deployArtifact<T>(
     }
 ): Promise<{ address: string; contract: T }> {
     const linkedArtifact = linkLibraries(artifact, options?.libs || {});
-    const factory = new ContractFactory(
-        artifact.abi,
-        linkedArtifact.bytecode,
-        signer
+
+    // Important: do NOT bind the factory to `signer` here.
```
daiagi: remove the AI process comment
the code already enacts the change

src/disputeManager/DisputeManager.ts
```diff
@@ -133,7 +133,8 @@ class DisputeManager {
                     );
                 } else {
                     await this.stateChannelManagerContract.uploadDispute(
-                        disputeConfirmation
+                        disputeConfirmation,
+                        { gasLimit: 5000000 }
```
daiagi: proably want to move this to a clear contant at the top of the file

src/rpc/services/TESTJoinChannel/TESTJoinChannelService.ts
```diff
@@ -1,4 +1,5 @@
-import { ARpcService, MainRpcService } from "@/rpc";
+import ARpcService from "@/rpc/ARpcService";
+import MainRpcService from "@/rpc/MainRpcService";
```
daiagi: `MainRpcService` not used

src/rpc/services/WebRTCSetup/WebRTCSetupService.ts
```diff
@@ -1,4 +1,5 @@
-import { ARpcService, MainRpcService } from "@/rpc";
+import ARpcService from "@/rpc/ARpcService";
+import MainRpcService from "@/rpc/MainRpcService";
```
daiagi: `MainRpcService` not used

src/rpc/services/initHandshake/InitHandshakeService.ts
```diff
@@ -1,5 +1,6 @@
 import { ethers } from "ethers";
-import { ARpcService, MainRpcService } from "@/rpc";
+import ARpcService from "@/rpc/ARpcService";
+import MainRpcService from "@/rpc/MainRpcService";
```
daiagi: `MainRpcService` not used

src/rpc/services/joinChannel/JoinChannelService.ts
```diff
@@ -1,4 +1,5 @@
-import { ARpcService, MainRpcService } from "@/rpc";
+import ARpcService from "@/rpc/ARpcService";
+import MainRpcService from "@/rpc/MainRpcService";
```
daiagi: ..

src/rpc/services/stateTransition/StateTransitionService.ts
```diff
@@ -1,6 +1,7 @@
 import { BlockConfirmationStruct } from "@typechain-types/contracts/V1/types/DataTypes";
 
-import { ARpcService, MainRpcService } from "@/rpc";
+import ARpcService from "@/rpc/ARpcService";
+import MainRpcService from "@/rpc/MainRpcService";
```
daiagi: ..

src/utils/LocalDiscoveryServer.ts
```diff
@@ -34,9 +34,14 @@ type DiscoveryInfo = {
  *    - Connects directly to other Peers upon receiving announcements.
  */
 export class LocalDiscoveryServer {
-    private static logger: Logger = createLogger({
-        component: "LocalDiscovery"
-    });
+    private static _logger?: Logger;
```
daiagi, line 37: this change seems harmless, but also pointless
this is the only place that uses logger like this
did this cahnge solve any issue?

src/utils/PeerLogger.ts
```diff
@@ -1,8 +1,9 @@
-import winston from "winston";
-
+import type winston from "winston";
+import { config } from "./config";
 export interface LoggerContext {
     peerId?: number;
-    peerAddress?: string;
+    // Can be an EVM AddressLike or any identifier; keep flexible.
+    peerAddress?: any;
```
daiagi: so use `Address | string` (by address i mean the ethers Address struct, not our type alias)
not `any`

src/utils/PeerLogger.ts
```diff
@@ -36,72 +37,281 @@ const Colors = {
...
-        timestamp,
-        level,
-        message,
-        peerId,
-        peerAddress,
-        component,
-        ...meta
-    }) => {
-        let prefix = "";
+function isNodeRuntime(): boolean {
```
daiagi: already created somewhere else
create once and re-use

src/utils/PeerLogger.ts
```diff
@@ -36,72 +37,281 @@ const Colors = {
...
-                    ? Number(timestamp)
-                    : (timestamp as string | number | Date);
-            const time = new Date(timeValue).toLocaleTimeString("en-US", {
-                hour12: false,
-                hour: "2-digit",
-                minute: "2-digit",
-                second: "2-digit"
-            });
-            prefix += `${Colors.TIMESTAMP}[${time}]${Colors.RESET}`;
+class BrowserLogger implements Logger {
```
daiagi, line 52: should come after my logging pr's and be placed in a new file

src/utils/PeerLogger.ts
```diff
@@ -36,72 +37,281 @@ const Colors = {
...
+function getGlobalLogger(): Logger {
     if (!globalLogger) {
+        if (!isNodeRuntime()) {
+            globalLogger = new BrowserLogger({}, config.LOG_LEVEL);
+            return globalLogger;
+        }
+
+        // Lazy-load winston only in Node runtimes.
+        // eslint-disable-next-line @typescript-eslint/no-var-requires
+        const winstonImpl = require("winston") as typeof import("winston");
```
daiagi, line 305: if lazy load winston only in node
what is the point of creating `BrowserLogger`?

src/ProfileManager.ts
```diff
@@ -47,7 +47,12 @@ class ProfileManager {
         );
         if (!profile) return;
         const oldTransport = profile.getTransport();
-        if (oldTransport) this.removeTransport(oldTransport);
+        if (oldTransport) {
+            setTimeout(() => {
```
daiagi: use `TimeoutManager` ?

lukastanisic99:
> `build:browser` command in `tic-tac-toe-vite` does not work without a `peer3.config.ts` file which is not included in this PR
> 
> i have added my own to pass that step assuming yours looks something like:
> 
> ```ts
> const config = {
>     PROVIDER_URL: "http://localhost:8545",
>     DEBUG_LOCAL_TRANSPORT: true,
>     HOLEPUNCH_RELAYER_URLS: ["wss://dht1-relay.leet.ar:49443"]
> };
> 
> export default config;
> ```

Yes it has the same structure as the config - you have the relayer url in our poker app, so you can take it from there - those old public ones don't work anymore

examples/TicTacToe/tic-tac-toe-vite/README.md
lukastanisic99: preview also works, but had some problems where it caches old files, but yes it doesn't have to be installed globally

examples/TicTacToe/contracts/TicTacToe/TicTacToeConsumerFacet.sol
lukastanisic99: ye - I didn't care too much to double check tic-tac-toe - I just wanted to see stuff running :D 
I'll change this in `cleanup`

examples/TicTacToe/contracts/TicTacToe/TicTacToeStateChannelManagerProxy.sol
lukastanisic99: No the wrapper is not needed I was just too lazy to change the old TicTacToe implementation that used the TicTacToeProxy so magic just wrapped it so typechain generates the type - the consumer just implements the consumer facet and uses the exported proxy

examples/TicTacToe/tic-tac-toe-vite/.gitignore
lukastanisic99: So people don't assume our public relayers are for their testing :D our relayers will have the rate limit and will let you use them for poker -> they have to run their own infra for their apps 

Yes readme should explain that and give an example - I just didn't do it :D

examples/TicTacToe/tic-tac-toe-vite/pnpm-lock.yaml
lukastanisic99: yarn works we can stick to yarn

src/disputeManager/DisputeManager.ts
lukastanisic99: Ye - Ideally we want to get rid of it - it was just a quick fix since I didn't want to dig for the root cause and don't know is it only on the ganache node for testing or production nodes will need this also - I think I added a todo on this in trello

src/rpc/services/TESTJoinChannel/TESTJoinChannelService.ts
lukastanisic99: This entire service will be removed - I think it's not used anymore - it was used in TicTacToe but i did a custom RPC for it and will probably add that RPC also as a service to the SDK - will add a few RPCs that can be used as building blocks (optional)

src/utils/LocalDiscoveryServer.ts
lukastanisic99: Yes - the root problem is that the logger is a singleton that loads the config at first run. Since this is a `static field` it was instantiated immediately before `config` loads and now the instantiation happens on first read (lazy) which is after the config loads - the root problem is the singleton logger (globalLogger) but this was a quick and easy fix

src/utils/PeerLogger.ts
lukastanisic99: The BrowserLogger is for browsers where winston doesn't work - the logger is polymorphic so from the outside you just interact with a logger and don't know is it a browser one or not

PR #262, opened 2026-01-05
Normalized addresses over EIP-55 and fix tests

lukastanisic99:


examples/TicTacToe/tic-tac-toe-vite/src/stateChannel/rpc/openChannelNegotiation/OpenChannelNegotiationHelpers.ts
```diff
@@ -5,12 +5,12 @@ export type Address = string;
 export const DEFAULT_JOIN_AMOUNT = 500;
 export const NEGOTIATION_TIMEOUT_MS = 20_000;
 
-export const normalizeAddress = (addr: string): Address =>
+export const toChecksumAddress = (addr: string): Address =>
```
daiagi: this Address type is in conflict with what we have in `@/types`
1. the type should probably move there
2. and be called `ChecsummedAddress` or something

also this entire function belongs in utils and then re-used
- in `Block`
- in `WebRTCSetupService`

probably more places

src/models/Block.ts
```diff
@@ -139,7 +139,9 @@ export default class Block {
         this._onChainTimestamp = onChainTimestamp;
     }
     get author() {
-        return this.block.transaction.header.participant as Address;
+        return ethers.getAddress(
```
daiagi: reuse `toChecksumAddress`
make it be able to consume `ethers.AddressLike` in addition to string

examples/TicTacToe/tic-tac-toe-vite/src/stateChannel/rpc/openChannelNegotiation/OpenChannelNegotiationHelpers.ts
lukastanisic99: This is AI just creating wrappers (probably for clarity) - the entire function is just `ethers.getAddress(adr)` - we don't need to wrap it and can use that directly and I think that's the case in the SDK

src/models/Block.ts
lukastanisic99: Probably makes sense to add the normalization function internally that uses ethers under the hood, so if we ever change the normalization standard we don't have to change every `ethers.getAddress()` - not sure why ethers imposes `AddressLike` when all of the instances at runtime are strings (at least the stuff we use) and their typescript return types are strings

lukastanisic99:
@daiagi if you have no objections I'll merge my blockchain like PRs since `cleanup` is the latest among them

PR #263, opened 2026-01-06
fixed-ethere-version-at 6.13.7

daiagi:
Using pnpm exposed a problem
- pnpm did not respect yarn.lock (obviously)
- package.json ethers dependency was `"^6.4"`
- yarn.lock resolved it as 6.13.2
- pnpm was a fresh install - took the latest ethers 6.16

in 6.14 there was a breaking change to the `Signer` interface 

- https://github.com/ethers-io/ethers.js/blob/main/CHANGELOG.md#ethersv6140-2025-05-06-2202

- https://github.com/ethers-io/ethers.js/commit/db490e1afaaa139be9cb38e6b50a1bc4ef28e075#diff-d9258f9e2b48c5b480fd2607924f87da4c245cb49829fba2481e9def14e04e49

At first i added the two missing method to `P2PSigner` and  `LocalDiamondSigner`

but then test began  failing due to incompatibility with hardhat signer

so instead i took to easy route of fixing to latest version before that change: 6.13.7

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "eslint": "^9.35.0",
     "eslint-plugin-import": "^2.32.0",
     "eslint-plugin-unused-imports": "^4.2.0",
-    "ethers": "^6.4.0",
+    "ethers": "6.13.7",
     "hardhat-gas-reporter": "^1.0.8",
     "husky": "^8.0.3",
     "lint-staged": "^14.0.1",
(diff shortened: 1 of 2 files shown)


lukastanisic99 (approved):
Yes for now we're opinionated on ethers and it's fine to hardcode it like this and any `hacks` we do along the way - down the line we'll do this normally by having a local RPC node so we're not opinionated on ethers (at least from the outside - internally we still might be)

PR #266, opened 2026-01-12
ad hoc cahnges for poker

daiagi:
small
ad-hoc changes to integrate with poker


lukastanisic99:
Can I merge all the other stuff that's pending? This doesn't build on that?

daiagi:
Yes you can.

src/HolepunchRelay.ts
```diff
@@ -50,8 +52,6 @@ class HolepunchRelay {
                 this.removeAndConnectToRelayer(relayerUrl);
             };
             console.log("HolepunchRelay - set onError", relayerUrl);
-
-            this.updateCallback();
```
lukastanisic99, line 54: what was the problem here

PR #267, opened 2026-01-16
ad-hoc fixes

daiagi:


src/evm/EvmDiamondStateMachine.ts
```diff
@@ -481,14 +481,19 @@ class EvmDiamondStateMachine extends ADiamondStateMachine {
             stateManager.p2pManager.p2pSigner
         ) as T;
 
+        const wrappedP2pContractInstance =
+            createEthersResultProxy(p2pContractInstance);
```
lukastanisic99: why this? Problems with the end user using ethers normally? Probably not needed since you don't need to do encode/decode on the outside or it's actually needed?

package.json
```diff
@@ -111,7 +111,8 @@
     ".": {
       "import": "./dist/src/index.js",
       "default": "./dist/src/index.js"
-    }
+    },
+    "./contracts/*": "./contracts/*"
```
lukastanisic99: does this make a difference since I don't think that solidity tools know how to read package.json and it worked for me without for tic-tac-toe or am I missing something

lukastanisic99: @daiagi actually it makes it better to include this since the IDE marks it as an error even though the compiler compiles it without errors

daiagi:
irrelevant PR. closing

PR #271, opened 2026-01-25
Local-diamond-logs

daiagi:
image


contracts/log/Console.sol
```diff
@@ -0,0 +1,37 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.8;
+
+contract Console {
+    // Well-known console precompile address.
+    // The hex literal encodes "console.log" in ASCII and matches Hardhat’s console.
+    address constant CONSOLE_ADDRESS = 0x000000000000000000636F6e736F6c652e6c6f67;
+
+    function log(string memory message) external view {
```
lukastanisic99: Do we need a function for every selectors that is supported? Like it seems we can't do `console.log(bool)`, `console.log(address)` and other stuff

contracts/V1/StateChannelDiamondProxy/StateChannelCommon.sol
```diff
@@ -61,12 +61,12 @@ contract StateChannelCommon is StateChannelManagerStorage, StateChannelManagerEv
...
-                UtilityFacet(utilityFacetAddress)
-                    .concatAddressArrays(getSnapshotParticipants(channelId), getPendingParticipants(channelId)),
-                getOnChainSlashedParticipants(channelId)
-            );
+        return UtilityFacet(utilityFacetAddress).subtractAddressArrays(
+            UtilityFacet(utilityFacetAddress).concatAddressArrays(
+                getSnapshotParticipants(channelId), getPendingParticipants(channelId)
+            ),
+            getOnChainSlashedParticipants(channelId)
+        );
```
lukastanisic99, line 69: Why do we have again formatting difference

src/evm/ConsolePrecompile.ts
```diff
@@ -0,0 +1,167 @@
...
+    const data = calldata.slice(4);
+
+    try {
+        const handler = CONSOLE_SELECTORS[selector];
+        if (handler) {
+            // Decode ABI-encoded parameters
+            const args = abiCoder.decode(handler.types, ethers.hexlify(data));
+            // Format and log the message
+            const message = handler.format(args);
+            solidityLogger.info(message);
```
lukastanisic99: maybe better to use debug level - we're only going to look into solidity if we see something is reverting

lukastanisic99:
@daiagi take a look - I've removed our console.log implementation - it works with the hardhat one 

For some reason the pre-commit hooks stoped working? I ran `forge fmt` explicitly and formatting changes were made - I committed the changes

daiagi:
Can merge if you are happy with it

PR #272, opened 2026-01-27
Improve logs

daiagi:
Just merged the branch `fixes-logger ` into here and re-create a PR into `rpc-guards`


lukastanisic99:
I'm still pushing changes - will create a PR and tag you when ready

PR #273, opened 2026-01-27
Collect-logs-on-errors

daiagi:
@lukastanisic99 
not done yet with this PR
but i broke `PeerLogger` to smaller components (it was too large, too many stuff going on in it)

if you make farther changes to it, you better do it on top of this pr
i expect you to care only about `BrowserLogger` (by care i mean the only place i expect you may want to add/change stuff)

i've already included in you implicit trace and group change from `fixes-logger` branch


daiagi:
image

daiagi:
created a browser log re-player / viewer:


image

src/utils/logging/browser/BrowserLogger.ts
```diff
@@ -0,0 +1,219 @@
+import { config } from "../../config";
+import { Logger, LoggerContext } from "../types";
+import { createLogStore } from "../logStore";
+import { formatTime } from "../formatUtils";
+import { BROWSER_PEER_COLORS, BROWSER_LEVEL_CSS } from "./colors";
+
+export class BrowserLogger implements Logger {
```
lukastanisic99: I'll push a PR with changes later and we can compare - I'll quickly comment over stuff:

`Logger` should be a base `abstract` class that has a common `shared` implementation and defines what dervied implementations must override - no need to have the same code in both BrowserLogger and NodeLogger, only differences

scripts/crash-log-server.js
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    console.log(`[CrashLogServer] [${requestId}] Processing POST request`);
+
+    try {
+        // Get filename from header or generate one
+        const filename =
+            req.headers["x-filename"] ||
+            `crash-${Date.now()}-${Math.random().toString(36).slice(2, 8)}.ndjson.gz`;
+        console.log(`[CrashLogServer] [${requestId}] Filename: ${filename}`);
```
lukastanisic99: I'll tell magic to rewrite this with express so it's more modular and supports multiple endpoints, so we can also fetch our logs and as needed to other stuff like authentication + I think it will have a lot less boilerplate code related to http

src/utils/logging/browser/BrowserLogger.ts
```diff
@@ -0,0 +1,219 @@
...
+    public level?: string;
+    private context: LoggerContext;
+    private logStore: ReturnType<typeof createLogStore>;
+    private enableMemoryStorage: boolean;
+    private static sharedLogStore: ReturnType<typeof createLogStore> | null =
```
lukastanisic99: `Statefull` singeltons are not good - this will polute the logs if multiple instances (tables) are ran at the same time - storage should be per StateManager

src/utils/logging/logStore.ts
```diff
@@ -0,0 +1,111 @@
...
+        return {
+            message: err.message,
+            stack: err.stack
+        };
+    }
+    return undefined;
+}
+
+// Shared log storage helper
+export function createLogStore(maxSize: number, enabled: boolean) {
```
lukastanisic99: I'll write this as a class since I think it would be a lot more readable and than you wouldn't use `ReturnType<>` wherever this function is called - this function is literally a class constructor, it encapsulates 2 fields and exposes 3 methods

src/utils/logging/logStore.ts
```diff
@@ -0,0 +1,111 @@
...
+    [key: string]: any;
+};
+
+export type LogObserver = (entry: StoredLog) => void;
+
+let observer: LogObserver | null = null;
+
+export function setLogObserver(fn: LogObserver | null): void {
+    observer = fn;
+}
```
lukastanisic99: I'm not a big fan of the functional programing style in OO languages - for instance this is over engineered and confusing to follow:

random functions and type definitions floating arround not encapsualted/grouped, but logically together and used together, and used within `closures` (within the `store` function, which I call store `method` of the `createLogStore` constructor/class), so all of this together behaves like a hidden/sneaky class

src/utils/logging/globalLogger.ts
```diff
@@ -0,0 +1,118 @@
+import { config, isNodeRuntime } from "../config";
+import { setupCrashHandler, CrashUploadConfig } from "./CrashHandler";
+import { Logger } from "./types";
+import { NodeLogger } from "./node/NodeLogger";
+import { BrowserLogger } from "./browser/BrowserLogger";
+
+// Global singleton logger to prevent multiple process event listeners
+let globalLogger: Logger | null = null;
```
lukastanisic99: I'll see to remove the singleton - makes it hard to have isolated state - rarely do singletons make sense

src/utils/logging/globalLogger.ts
```diff
@@ -0,0 +1,118 @@
...
+
+    const eqFlag = args.find((arg) => arg.startsWith("--exclude-tags="));
+    if (eqFlag) {
+        addTags(eqFlag.split("=", 2)[1]);
+    }
+
+    return excludedTags;
+}
+
+export function getGlobalLogger(): Logger {
```
lukastanisic99: I never liked this global logger - having a general `create` function that deducts which consturctor to call and return a polymorphic instance - yes!, global shared state - no :D

lukastanisic99 (commented):
Most stuff look ok - I'll see to do a PR today and then we can comment over - mostly it will be sticking to hard explicit OO/Classes and not having sneaky classes + removing singletons

lukastanisic99:
just realized this doesn't include hardhat.console fixes and all the stuff we did after -.- 

Will have to manually merge a lot of stuff

daiagi:
there is nothing to manually merge.
all these changes are merged into `dispute` branch already.
i just changed the base of this PR to target `dispute`.

PR #274, opened 2026-01-29
More fixes and more logging

lukastanisic99:


lukastanisic99:
@daiagi take a look - some tests are failing (I know why) - I'll fix it in the morning after my calls, but good enough to take a look 

I've deployed to our testnet and it's working good - you can play if you want

src/eventHandlers/EventHandler.ts
```diff
@@ -152,23 +152,16 @@ export class EventHandler {
...
-        this.logger.group(`✅ Dispute received: ${meta.disputeHash}`);
-        this.logger.info(message, {
-            disputer: LoggerUtils.formatHash(meta.disputer),
-            isFinal: isFinal
+        const disputeMeta = LoggerUtils.getDisputeMetadata(dispute);
+        const formattedHash = LoggerUtils.formatHash(disputeMeta.disputeHash);
+        this.logger.info(`✅ Dispute received: ${formattedHash}`, {
+            dispute: disputeMeta,
+            isFinal: isFinal,
+            disputeConfirmation,
```
daiagi, line 160: Just saying, this is explictly opposite of what you've asked me to do when you asked to enhance logs
you have asked for:

- short info messages
- gradual reveal, hidden in debug level
- no full objects

src/evm/ConsolePrecompile.ts
```diff
@@ -0,0 +1,167 @@
+import { PrecompileInput } from "@ethereumjs/evm";
+import { ethers } from "ethers";
+import { createLogger } from "@/utils";
+
+/**
+ * Hardhat's console.log contract address.
```
daiagi, line 6: what is this file doing in the pr?
i thought you removed it

there are other changes in this pr that came from the precompile one

src/rpc/guards/HandshakeCompletedGuard.ts
```diff
@@ -86,6 +86,7 @@ export class HandshakeCompletedGuard extends AGuard<ARpcService<ARpcMethods>> {
 
             const timeoutMs =
                 this.service.p2pManager.stateManager.timeConfig.agreementTime *
+                2 *
```
daiagi: magic number
plese move to constant or at least comment what it is

src/rpc/services/initHandshake/InitHandshakeService.ts
```diff
@@ -101,9 +102,25 @@ class InitHandshakeService extends ARpcService<InitHandshakeRpcMethods> {
     }
 
     public isHandshakeCompletedForTransport(transport: ATransport): boolean {
-        const profile =
-            this.p2pManager.profileManager.getProfileByTransport(transport);
-        return !!profile && profile.getIsHandshakeCompleted();
+        const address = transport.peerAddress;
+        let profile: PeerProfile | undefined;
```
daiagi:
```suggestion
const profile = address == undefined ? getProfileByTransport(transport): getProfileByEvmAddress
```

src/stateManager/ValidationService.ts
```diff
@@ -43,6 +44,7 @@ export default class ValidationService {
             block.channelId != this.stateManager.channelId
         ) {
             this.logger.warn("validateBlockConfirmation - wrong channel", {
+                strategy: (strategy as any)?.constructor?.name,
                 stateManagerChannelId: String(this.stateManager.channelId),
```
daiagi: consider:
 `get name(): string` on `AValidationStrategy`

src/StateChannelEventListener.ts
```diff
@@ -37,8 +41,15 @@ class StateChannelEventListener {
             );
         }
         const filter = filterFactory();
-        await this.stateChannelManagerContract.on(filter, handler);
-        this.filters[key] = { filter, listener: handler };
+        const wrappedHandler = async (logObj: any) => {
+            this.logger.info("On-chain event received", {
```
daiagi, line 45: same "complaint" here
this is not gradual reveal and lot of data for an info message
if this is how you want it, so be it, I'm not asking to change it.
just mentioning

src/eventHandlers/EventHandler.ts
lukastanisic99: Yes true, but I saw this nice `grouping` that you added which I liked a lot -> I made it implicit with the stack trace -> kept your short human readable messages with the `emojis` and hid the objects in the groups, so you don't see them, but when you need them you have them. The human readable part is the object keys and enums being strings. You look at stuff you're interested in and don't expand if you don't have to. The stack trace tells you in most cases everything you need to know with the object. The debug is for more verbose stuff (but I don't like the debug too much, since groupings are `info` so if I use groupings all the `debug` shows with `info` and the stack trace in it is ugly since it's not collapsed...). Maybe our extension can actually fix it - but it's ok like this - only if I really need to go to debug/verbose I change and in 99% of cases I just read short error messages (labels of groupings).

The human messages for cause/reason were just polluting the code and the stack trace is better in my opinion and it's clearer and now everything is reusable and implemented in one place.

src/evm/ConsolePrecompile.ts
lukastanisic99: I merged with the precompile one - I removed the solidity implementation, since it works with the default hardhat one. This is the typescript precompile that catches and logs.

src/StateChannelEventListener.ts
lukastanisic99: Only the message part is visible - the meta (object) is collapsed and you don't see it unless you expand.

src/rpc/guards/HandshakeCompletedGuard.ts
lukastanisic99: yeah maybe this number is not needed, since the bug was somewhere else, but still I'll comment what it is

daiagi (approved):

PR #276, opened 2026-02-02
Simplify and cleanup log collection

lukastanisic99:
- No more singletons 
- Explicit classes
- Modular and reusable
- Exclusive and shared context for loggers
- Express server with upload & query (need to test query tomorrow)
- no `maybe compress` - cross platform full support (will test different browsers in the morning) - solid algo on average over 95% compression 
- clean request/response 
- No duplicate persistence - single format cross platform. Replay polymorphic with `logger.logEntry()`, but we'll build our simple web viewer since we can fine tune everything and have custom features. Rendering and style is up to the viewer.  

Not sure why formatting differences are again present - this is getting really annoying 

I'll merge tomorrow the other branches and finish what's left - I'll for sure have some manual merges since I've touched files I've touched in other branches too, but it should be straight forward. 

@daiagi take a look


daiagi:
Honestly, I don’t really know what to say here.

A lot of what I built is just gone. The encoders for the two viewing systems are missing, the browser replay format is gone, the plain-text encoder for the VS Code extension is gone, and the 2-second delayed capture after an error log isn’t there anymore. That means the browser replayer won’t work, and neither will the extension.

Yes, the code is cleaner now. I can see that. but some things should be allowed to exist without always pushing for classes.


For example, when I wrote `LoggerUtils`, I turned it into a class purely to match your strong preference for classes. In reality, its entire interface is static methods. Making it a class just makes the call sites longer and adds boilerplate without any real benefit.

I feel the same way about LogUploader. It’s over-engineered for what it does — there’s no real need for an abstract base class here. Something  simpler would work, e.g.:

```ts
type ListenerAttacher = (uploadFn: () => void) => void;

export class LogUploader {
    constructor(
        store: LogStore,
        config: Config,
        attachListener?: ListenerAttacher 
    ) {
        attachListener?.(() => this.uploadLogs());
    }
}
```

But I’m guessing this kind of approach isn’t something you’re comfortable with.



Also, this PR pulls in changes that are already part of the disputed branch. I’m not sure why those keep getting dragged in. I try to keep my PRs focused and clean, and I think it would help if this one did the same.

daiagi:
oh and the server
it was meant to be a one off for local dev
and it was working fine
it was never intended for prod, it was only intended so me/you can start the app and see the logs that are generated

what do you need a `get` endpoint for?
the production server for this purpose is suppose to upload to aws, not save it on disk

lukastanisic99:
> Honestly, I don’t really know what to say here.
> 
> A lot of what I built is just gone. The encoders for the two viewing systems are missing, the browser replay format is gone, the plain-text encoder for the VS Code extension is gone, and the 2-second delayed capture after an error log isn’t there anymore. That means the browser replayer won’t work, and neither will the extension.

Yes - having 2 formats was not ideal - the idea is persist only the ` raw compressed data`, this way it works cross platform in a single format and the renderer can decide how to display the data. Replay is now very simple, you basically call `logger.logEntry()` it will polymorphically replay the logs the same way cross platform. 

> Yes, the code is cleaner now. I can see that. but some things should be allowed to exist without always pushing for classes.

> For example, when I wrote `LoggerUtils`, I turned it into a class purely to match your strong preference for classes. In reality, its entire interface is static methods. Making it a class just makes the call sites longer and adds boilerplate without any real benefit.

I'm not saying push for classes unconditionally - where I'm pushing for classes is where the functions that were there are `sneaky` classes where `encapsulation` is not clear + namespace separation like `LoggerUtils`. When you see `LoggerUtils.XXX` its a lot more clear than just importing `XXX` - it's about reducing the cognitive load while looking at the code.  

> I feel the same way about LogUploader. It’s over-engineered for what it does — there’s no real need for an abstract base class here. Something simpler would work, e.g.:
> 
> ```ts
> type ListenerAttacher = (uploadFn: () => void) => void;
> 
> export class LogUploader {
>     constructor(
>         store: LogStore,
>         config: Config,
>         attachListener?: ListenerAttacher 
>     ) {
>         attachListener?.(() => this.uploadLogs());
>     }
> }
> ```
> 
> But I’m guessing this kind of approach isn’t something you’re comfortable with.

The `LogUploader` is just an abstraction on how to `upload logs` - it's minimal and it kept your features and just renamed from `crashHandler`. The only reason it's abstract is because the `listeners` are different for each environment. The code you suggested here I wouldn't have any objection writing it like that, but it's insufficient and I don't see over engineering in it - there are stuff that can be cleaned up further for sure and it's mostly there not to pollute the `Logger` class so things stay lean and modular and not like our `StateManager` where everything is in 1 large file 

> Also, this PR pulls in changes that are already part of the disputed branch. I’m not sure why those keep getting dragged in. I try to keep my PRs focused and clean, and I think it would help if this one did the same.

Yes I know that's why I made the comment - Those were things that I needed to change and I'm aware that merge conflicts will arise from those changes and I'll handle them manually. The only things I could have skipped is further pruning of the disconnect logs - but those are minimal and I mostly did them to not forget since I was looking at them already - but yes in general I agree with you the PR should stay clean but what's more annoying for the PR (in my opinion) is the constant formatting changes which I don't understand why we're getting when we have the formatting hooks to enforce style

lukastanisic99:
> oh and the server it was meant to be a one off for local dev and it was working fine it was never intended for prod, it was only intended so me/you can start the app and see the logs that are generated
> 
> what do you need a `get` endpoint for? the production server for this purpose is suppose to upload to aws, not save it on disk

My problems with the server were that there was a lot of boilerplate code instead of the simple api as with express + it exposed only 1 endpoint. I'm trying to think and plan 10 steps in advance and it wasn't a lot of time for me nor magic - this way it's extendable and the `get` is there to get the logs so you can retrieve them and view them - I'm for sure not going to look at them on the server. We're going with the react app for viewing so you and me can just go to a website and click a log and look at it from any device. If you want to use an extension or another viewer that's also fine - you can do it and you'll be able to retrieve the logs in the same way and put them in any viewer you like - that's the point the logs are not opinionated on the style

daiagi (approved):

PR #278, opened 2026-02-06
Refactor-e2e-tests

daiagi:
Refactored the E2E test suite from monolithic test files into a composable, block-based architecture for better maintainability and code reuse.

Key Changes:

- Composable Block System: Tests now built using reusable blocks (Assert, Byzantine, Transition, Event, RPC, Wait, Time) that can be composed and interchanged to create complex test scenarios from simple primitives

- Block Implementation Pattern: Each block encapsulates actions via dedicated action modules (AssertActions, ByzantineActions, TransitionActions, DisputeOrchestrator, NetworkController) or direct harness calls.

- Test File Reorganization: Renamed and restructured test files to better match their domains (E2E-DisputeManager, E2E-InitHandshake, E2E-IsForkDisputed, E2E-StateTransition, E2E-Timeouts, E2E-Spectate)

- Event-Driven Synchronization: Removed all polling-based waiters in favor of event barriers.

- High-Level Scenarios: Scenario class provides pre-built patterns (emptyChannel(), activeChannel(), byzantineScenarios()) while ScenarioRunner orchestrates block execution with automatic cleanup


src/transport/ATransport.ts
```diff
@@ -21,7 +21,6 @@ abstract class ATransport {
         if (!this.isClosed) {
             LoggerUtils.logTransportDisconnect(this, isExpected);
             this.isClosed = true;
-            this.p2pManager.disconnectConnection(this);
             this._close();
```
lukastanisic99, line 31: You probably removed this since in `p2pManager.disconnectConnection`
```typescript
// Fire onDisconnection hook if peer address is available
if (peerAddress && this.stateManager.p2pEventHooks?.onDisconnection) {
    this.stateManager.p2pEventHooks.onDisconnection(peerAddress);
}
```
was firing multiple times.

Better to move the event fire hire, since this is protected with a `if(!this.closed)` and keep `this.p2pManager.disconnectConnection(this)`, since not all disconnections go through the p2pManager (e.g. holepunch and webRTC `close` events), so this was added here to make sure that however the transpored is closed, p2pManager will remove the connection from its internal accounting

src/P2PManager.ts
```diff
@@ -160,6 +161,11 @@ class P2PManager<TFactories extends RpcServiceFactoryMap = {}>
         } catch {
             // ignore
         }
+
+        // Fire onDisconnection hook if peer address is available
+        if (peerAddress && this.stateManager.p2pEventHooks?.onDisconnection) {
+            this.stateManager.p2pEventHooks.onDisconnection(peerAddress);
+        }
```
lukastanisic99: move to ATransport close

You can write short without the check like this:
`...p2pEventHooks?.onDisconnection(peerAddress)`

test/e2e/E2E-DisputeManager.test.ts
```diff
@@ -0,0 +1,307 @@
...
+ */
+describe("E2E: Dispute Manager", function () {
+    describe("Dispute Initiation", function () {
+        it("should create dispute for double-sign detected", async function () {
+            await ScenarioRunner.execute(
+                Scenario.activeChannel(3, 2),
+                Byzantine.doubleSignFrom(1), // Peer 1 double-signs
+                Assert.disputeInitiatedBy({ peers: [0, 2] }), // Peers 0,2 detect
+                Assert.disputeCommitted()
+            );
```
lukastanisic99, line 32: very clean and reads like a story!

test/harness/actions/AssertActions.ts
```diff
@@ -0,0 +1,473 @@
...
+import { expect } from "chai";
+import { ForkId } from "@/types/types";
+
+export class AssertActions {
+    constructor(
+        private harness: PeerTestHarness<any, any>,
+        private logger: Logger
+    ) {}
+
+    async disputeInitiatedBy(
```
lukastanisic99: not used

test/harness/actions/AssertActions.ts
```diff
@@ -0,0 +1,473 @@
...
+            await this.harness.eventActions.waitForEventCounts(
+                "onDisputeCommitted",
+                expectedCounts,
+                timeout
+            );
+
+        expect(disputeCommitted).to.be.true;
+    }
+
+    async calldataPosted(
```
lukastanisic99: not used

test/utils/SyncCoordinator.ts
```diff
@@ -145,6 +154,92 @@ export class SyncCoordinator {
         return { inSync: allInSync, syncDetails };
     }
 
+    /**
+     * Assert all peers are in sync with state machine state validation
+     */
+    public assertAllInSync(
```
lukastanisic99: not used

src/utils/index.ts
```diff
@@ -18,3 +18,7 @@ export * from "./evmErrorHandler";
 export * from "./logging";
 export * from "./EthersResultProxy";
 export * from "./address";
+
+export function sleep(ms: number): Promise<void> {
+    return new Promise((resolve) => setTimeout(resolve, ms));
```
lukastanisic99, line 23: Feels like some stuff are being repeated in multiple places - harnes/blocks/Time.ts

lukastanisic99 (commented):
Very readable and modular - I like it a lot! Will continue tomorrow. Some things are not used and I feel some are implemented multiple times in multiple places. We need to think about how we can group stuff together that are interchangable so the `engine` can do all variations.

lukastanisic99:
Apart of the engine doing all variations we need a methodical approach for coming up and writing test cases. E.g. if we had a list of test cases, implementing them would be straight forward, so coming up with the list in a methodical way where we cover most is also something we have to do. For now most of our tests are testing 1,2 main things of a feature we implemented, while many failure and edge cases are not accounted for (yet).

test/harness/actions/AssertActions.ts
daiagi: do you mean the logger?

because `disputeInitiatedBy` is used

src/utils/index.ts
daiagi: i remove it from SyncCoordinator,  the `sleep` function  here is used (in a single place)

test/harness/actions/AssertActions.ts
lukastanisic99: `disputeInitiatedBy` has a different implementation that is used - this one has 0 references in the code (at least in my VSCode)

daiagi: you're right
i removed `disputeInitiatedByPeers` actions

src/P2PManager.ts
```diff
@@ -149,6 +149,7 @@ class P2PManager<TFactories extends RpcServiceFactoryMap = {}>
 
     public disconnectConnection(transport: ATransport) {
         const profile = this.profileManager.getProfileByTransport(transport);
+        const peerAddress = transport.peerAddress;
```
lukastanisic99: now this is not used and `onDisconnection` is also not used

src/transport/ATransport.ts
lukastanisic99: if(!isExpected)this.p2pManager.stateManager.p2pEventHooks?.onDisconnection(this.peerAddress)

src/P2PManager.ts
daiagi: restored
sorry. I misunderstood what you wrote

src/transport/ATransport.ts
```diff
@@ -21,6 +22,9 @@ abstract class ATransport {
         if (!this.isClosed) {
             LoggerUtils.logTransportDisconnect(this, isExpected);
             this.isClosed = true;
+            this.p2pManager.stateManager.p2pEventHooks?.onDisconnection?.(
```
lukastanisic99: `close` has a `isExpected = false` argument -> would do this as `if(!isExpected)...p2pEventHooks?.onDisconnection..` so this doesn't fire for transport upgrades

daiagi: done

lukastanisic99 (approved):

PR #280, opened 2026-02-11
State-snapshot-tests

daiagi:
New API tests for posting state snapshots
- in single fork
- across forks -> in fork
- multi call between forks

### DIFF
--- a/test/e2e/E2E-Spectate.test.ts
+++ b/test/e2e/E2E-Spectate.test.ts
@@ -178,7 +178,7 @@ describe("E2E: Spectate Service", function () {
                 // Create and resolve invalid state transition dispute
                 // This will reduce the fork (remove peer 2)
                 // Using extended settlement timing for 5-peer scenario
-                Scenario.forkResolutionWithSettlement({
+                Scenario.disputeWithReduction({
                     maliciousPeerIndex: 2,
                     forkSettleTimeoutMs: 15000,
                     disputesCommittedTimeoutMs: 10000
--- a/test/e2e/E2E-StateSnapshots.test.ts
+++ b/test/e2e/E2E-StateSnapshots.test.ts
@@ -1,0 +1,117 @@
+import {
+    ScenarioRunner,
+    Scenario,
+    Assert,
+    Event,
+    Transition,
+    Context,
+    PeerTestHarness
+} from "@test/harness";
+
+PeerTestHarness.setDefaultLogLevel("error");
+
+/**
+ * E2E Tests for State Snapshot Posting
+ *
+ * Maps to: src/stateManager/StateManager.ts (postStateSnapshot, multicall path)
+ *          State snapshot on-chain updates after transitions and fork resolution.
+ *
+ * Covers: posting snapshot after N transitions, 2 independent snapshot updates
+ * after fork resolution, and single multicall (fork + same-fork) update.
+ */
+describe("E2E: State Snapshots", function () {
+    const forkTimeConfig = {
+        p2pTime: 3,
+        agreementTime: 2,
+        chainFallbackTime: 2,
+        evidenceTime: 3
+    };
+
+    it("should post updated state snapshot on-chain after 3 transitions", async function () {
+        await ScenarioRunner.execute(
+            Scenario.emptyChannel(3),
+
+            // Execute the 3 specific state transitions
+            Scenario.advanceState(1),
+            Transition.valid((c) => c.leaveChannel()),
+            Scenario.advanceState(1),
+            Assert.allPeersInSync(),
+            Event.reset(),
+            Assert.channelBalanceMatchesSnapshot(),
+            // Prepare snapshot data and store in context for delta assertions
+            Context.capturePrePostSnapshotContext(),
+            Transition.postSnapshot(),
+            Assert.channelBalanceMatchesSnapshot(),
+            Assert.withdrawalDeltaMatchesExpected(),
+            Event.waitForAllPeers("onStateSnapshotUpdated", 1, {
+                mode: "atLeast"
+            }),
+            Assert.snapshotMatchesLocal()
+        );
+    });
+
+    it("should remove malicious participant after fork and then post updated state snapshot on the reduced fork - 2 independent snapshot updates", async function () {
+        await ScenarioRunner.execute(
+            Scenario.forkResolutionWithSnapshotMoved({
+                timeConfig: forkTimeConfig
+            }),
+
+            Transition.fromHonestPeersOnly((c) => c.add(1)),
+            Transition.fromHonestPeersOnly((c) => c.leaveChannel()),
+            Transition.fromHonestPeersOnly((c) => c.add(3)),
+            Assert.onlyHonestPeersInSync(),
+            Event.reset(),
+
+            Assert.channelBalanceMatchesSnapshot(),
+
+            // Prepare snapshot data and store in context for delta assertions
+            Context.capturePrePostSnapshotContext(),
+
+            Transition.postSnapshot(),
+
+            Assert.channelBalanceMatchesSnapshot(),
+            Assert.withdrawalDeltaMatchesExpected(),
+
+            Event.waitForHonestPeers("onStateSnapshotUpdated", 1, {
+                mode: "atLeast"
+            }),
+
+            Assert.snapshotMatchesLocal(),
+            Assert.maliciousPeerExcluded()
+        );
+    });
+
+    it("should remove malicious participant after fork and then post updated state snapshot on the reduced fork - multicall", async function () {
+        await ScenarioRunner.execute(
+            // Fork resolved locally; do NOT post snapshot yet so on-chain is still on disputed fork
+            Scenario.fourPeerForkResolution({ timeConfig: forkTimeConfig }),
+
+            // Three transitions on reduced fork (same as test 2) to have both fork update and same-fork update
+            Transition.fromHonestPeersOnly((c) => c.add(1)),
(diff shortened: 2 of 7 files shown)


test/harness/core/types.ts
```diff
@@ -122,6 +122,15 @@ export interface HarnessContext {
     /** Function to restore dispute construction after interception (set by Byzantine.interceptDisputeConstruction) */
     restoreDisputeConstruction?: () => void;
 
+    /** last milestone snapshot before posting snapshot (set by Context.captureContextForSnapshotSameFork) */
+    lastMilestoneSnapshot?: any;
+
```
lukastanisic99, line 130: why is there no typesafety here

test/harness/blocks/Scenario.ts
```diff
@@ -343,7 +343,7 @@ export class Scenario {
      * - Control over fork settlement timing (forkSettleTimeoutMs)
      * - More lenient dispute commit requirements (some peers may be slow)
      */
-    static forkResolutionWithSettlement(options: {
+    static disputeWithReduction(options: {
```
lukastanisic99, line 346: you didn't use VSCode rename, so the rename is only here and other parts of the code use the old name and this one has 0 references

test/harness/blocks/Assert.ts
```diff
@@ -540,6 +540,139 @@ export class Assert {
         });
     }
 
+    /**
+     * Assert that channel balance matches snapshot totalWithdrawals
+     */
+    static channelBalanceMatchesSnapshot(options?: { peerIndex?: number }) {
```
lukastanisic99: this only checks withdrawals - so name it in a way where it's clear that it doesn't check the entire invariant

test/harness/blocks/Assert.ts
```diff
@@ -540,6 +540,139 @@ export class Assert {
...
+        return new HarnessBlock(async (harness) => {
+            const peer = harness.peers[peerIndex];
+            if (!peer) {
+                throw new Error(`Peer ${peerIndex} not found`);
+            }
+
+            const stateMachine = peer.stateManager.diamondStateMachine;
+
+            // Get channel balance before (from context) and after
+            const channelBalanceBefore = harness.context.channelBalanceBefore;
```
lukastanisic99, line 597: type safety on the context

test/harness/blocks/Assert.ts
```diff
@@ -540,6 +540,139 @@ export class Assert {
...
+            }
+
+            const onChainSnapshot = StateSnapshot.from(
+                await harness.channelManager.getStateSnapshot(harness.channelId)
+            );
+
+            const stateMachine = peer.stateManager.diamondStateMachine;
+            const withdrawalsMatch = await stateMachine.areBalancesEqual(
+                onChainSnapshot.snapshotData.totalWithdrawals,
+                lastSnapshot.snapshotData.totalWithdrawals
```
lukastanisic99, line 661: this is only true if there are no withdrawals 

The `deposit` and `withdrawal` balances of a snapshot the total aggregated sum of deposits/withdrawals, so it is always increasing

`newSnapshot.withdrawals = oldSnapshot.withdrawals+delta`
There's a channelBalance that keeps track of processed depsoits and withdrawals regarldess of the snapshot, so the invariant that should hold for withdrawals is always `channelBalance.totalWithdrawals == newSnapshot.totalWithdrawals` that's checked at the end of the transaction after withdrawals are processed - I think that sdk solidity code does it already

test/harness/blocks/Assert.ts
```diff
@@ -540,6 +540,139 @@ export class Assert {
...
+            );
+            if (!withdrawalsMatch) {
+                throw new Error(
+                    "On-chain totalWithdrawals does not match last milestone snapshot totalWithdrawals"
+                );
+            }
+
+            const depositsMatch = await stateMachine.areBalancesEqual(
+                onChainSnapshot.snapshotData.totalDeposits,
+                lastSnapshot.snapshotData.totalDeposits
```
lukastanisic99, line 671: this is also true only if there are no deposits between the 2 snapshots - similar comment as to above

test/harness/blocks/Assert.ts
```diff
@@ -540,6 +540,146 @@ export class Assert {
...
+                );
+            }
+
+            return harness;
+        });
+    }
+
+    static onChainBalanceMatchesSnapshot(options?: { peerIndex?: number }) {
+        const { peerIndex = 0 } = options || {};
+
```
lukastanisic99, line 640: Better to do the context variables with setters and getters so you can quickly lookup the refference where they are read vs where they are written + the getter can throw the implicit error if you forgot to set them and than you can figure out did you forget to call a building block or create a new one if existings ones can't be reused for that scenario

lukastanisic99 (changes requested):
mostly stuff above

test/harness/core/types.ts
```diff
@@ -122,6 +125,15 @@ export interface HarnessContext {
...
     restoreDisputeConstruction?: () => void;
 
+    /** last milestone snapshot before posting snapshot (set by Context.captureContextForSnapshotSameFork) */
+    lastMilestoneSnapshot?: StateSnapshot;
+
+    /** Channel balance before posting snapshot (set by Context.captureContextForSnapshotSameFork) */
+    channelBalanceBefore?: ChannelBalanceStructOutput;
+
+    /** Expected withdrawals delta from prepared outbound messages (set by Context.captureContextForSnapshotSameFork) */
+    expectedWithdrawalsDelta?: BalanceStruct;
```
lukastanisic99, line 135: ok these are updated, but there are a lot of places that use any while there is a type that coresponds to it

test/e2e/E2E-Spectate.test.ts
```diff
@@ -178,7 +178,7 @@ describe("E2E: Spectate Service", function () {
                 // Create and resolve invalid state transition dispute
                 // This will reduce the fork (remove peer 2)
                 // Using extended settlement timing for 5-peer scenario
-                Scenario.forkResolutionWithSettlement({
+                Scenario.disputeWithReduction({
```
lukastanisic99, line 181: Image

Think this the entire family of similar names

test/e2e/E2E-StateSnapshots.test.ts
```diff
@@ -0,0 +1,117 @@
...
+
+            Assert.snapshotMatchesLocal(),
+            Assert.maliciousPeerExcluded()
+        );
+    });
+
+    it("should remove malicious participant after fork and then post updated state snapshot on the reduced fork - multicall", async function () {
+        await ScenarioRunner.execute(
+            // Fork resolved locally; do NOT post snapshot yet so on-chain is still on disputed fork
+            Scenario.fourPeerForkResolution({ timeConfig: forkTimeConfig }),
```
lukastanisic99, line 87: better the peerCount to be an argument to the dispute resolution than to have a method named `fourPeer***`

test/e2e/E2E-StateSnapshots.test.ts
```diff
@@ -0,0 +1,117 @@
...
+
+            // Three transitions on reduced fork (same as test 2) to have both fork update and same-fork update
+            Transition.fromHonestPeersOnly((c) => c.add(1)),
```
lukastanisic99, line 90: does this `fromHonest*` make a difference (here and in general) since the malicious ones are removed and never selected by the leader election algo

test/harness/blocks/Assert.ts
```diff
@@ -540,6 +540,146 @@ export class Assert {
...
+            );
+            if (!depositsMatch) {
+                throw new Error(
+                    "On-chain totalDeposits does not match last milestone snapshot totalDeposits"
+                );
+            }
+
+            return harness;
+        });
+    }
```
lukastanisic99, line 681: this file needs to be more split and maybe others too - this one has 28 static functions - we need quick binary search logic where you exactly know where to look to find what you want and click click and see if what you need is there and if not implement it, without reading all 28 functions that have things from are peers in sync to balance checking

lukastanisic99 (commented):
Namespace are probbaly not ideal in typescript, but we can do a convention e.g have an `assert` directory where things are split into AsserBalance.ts AssertSync.ts etc etc and then you just go to that directory and in the file you're interested it and see if what you're looking for is there and if not add it

test/harness/core/types.ts
daiagi: resolved in #282

test/harness/blocks/Scenario.ts
daiagi: resolved in #282

test/harness/blocks/Assert.ts
daiagi: resolved in #282

test/harness/blocks/Assert.ts
daiagi: resolved in #282

test/harness/core/types.ts
daiagi: resolved in #282

test/e2e/E2E-Spectate.test.ts
daiagi: resolved in #282

test/e2e/E2E-StateSnapshots.test.ts
daiagi: resolved in #282

test/e2e/E2E-StateSnapshots.test.ts
daiagi: it doesn't make a difference for writing
but it makes a difference in `TransitionsActions.submit`  which waits until only the honest peers are in sync

PR #290, opened 2026-03-06
[test] post state snapshot on block height 0, join -> leave sequence, block validation fraud proofs

daiagi:
1.
- test posting of state snapshot after a single block (block height 0)
- found and fixed some bugs
- ability to pass option flag to `assert.sync.peersInSyncWait` to wait for block finalization

2. `E2E-SpectatorStateProofPersistence.test.ts`
test for  leave -> join -> leave -> join sequence

3. `E2E-FraudProofsBlockConfirmation.test.ts` tests for all the fraud proofs  of  block confirmation pipeline (participant strategy)


contracts/V1/examples/MathStateMachine/MathStateMachine.sol
```diff
@@ -101,6 +115,7 @@ contract MathStateMachine is AStateMachine {
 
         if (success) {
             _addExitChannel(exitChannel);
+            emit NextToPlay(getNextToWrite());
```
lukastanisic99: Where is this used? I only see it's emitted?

src/agreementManager/AgreementManager.ts
```diff
@@ -163,10 +163,7 @@ class AgreementManager {
             for (const block of blockIterator) {
                 signedBlocks.push(block.signedBlock);
 
-                if (
-                    block.height === 0 ||
-                    block.stateSnapshotHash === previousThresholdSnapshot.hash
-                ) {
+                if (block.stateSnapshotHash === previousThresholdSnapshot.hash) {
```
lukastanisic99: does this really make a difference? The block is `first pushed` and than this `if` runs, so the `block.height = 0` block is added regardless, and since there are no blocks before block 0 (only genesis snapshot, which is not relevant here), the blockInterator on its own would terminate regardless of the block height check 

The bug is probably somewhere else

In fact this entire `if` is redundant, since signedBlocks are used only when there is `no finality` (no milestones), and they include all blocks down to genesis - since storage always contains correct blocks, we can just iterate backwars until the iterator stops on its own  

Tehnically I could have designed this without signedBlocks and only have milestones, but only did it since unlike in milestones that are explicitly final with the signatures, the genesis is implicitly final by being the genesis, so wouldn't be able to do the recursive milestone threshold check so I added a special case when building on genesis without explicit threshold finality

src/stateManager/StateManager.ts
```diff
@@ -1418,8 +1418,15 @@ class StateManager {
...
-                // Only include milestones that are newer than the current on-chain block height
-                if (snapshot.blockHeight > currentOnChainSnapshot.blockHeight) {
+                // Include milestones that are newer than current on-chain, or at blockHeight 0 with different hash
+                const snapshotIsForFirstBlock =
+                    snapshot.blockHeight === 0 &&
+                    currentOnChainSnapshot.blockHeight === 0;
+                if (
+                    snapshot.blockHeight > currentOnChainSnapshot.blockHeight ||
+                    (snapshotIsForFirstBlock &&
+                        snapshot.hash !== currentOnChainSnapshot.hash)
```
lukastanisic99: we should have a general solidity stateless function that comparesSnapshots `isSnapshotNewer`, since this is used in a few places. 

This one has a potential bug if the block 0 threshold is witheld by a malicous peer, but the malicious peer updates the snapshot to 0, so when we're here we'll compare hashes of the gensesis and block 0 snapshot, see it's different and try to update the snapshot to genesis - on-chain it will fail

so the implementation of `isSnapshotNewer` would be something like this:
if(newSnapshot.height > oldSnapshot.height) return true;
if(newSnapshot.height == 0 && oldSnapshot.height == 0 && isGenesisSnapsot(old) && old.hash != new.hash) return true;
return false;

daiagi: will come back for this one

daiagi:
the test `join/leave sequence and fork resolution` was passing before rebasing onto the dispute simplification branch
now it doesn't

I'll Investigate

src/stateManager/StateManager.ts

src/stateManager/StateManager.ts
```diff
@@ -2488,6 +2489,11 @@ class StateManager {
                 await this.diamondStateMachine.getParticipants();
             const isParticipant = participants.includes(this.signerAddress);
             if (isParticipant) this.setStatus(Status.PARTICIPATING);
+        } else if (
+            this.status === Status.PARTICIPATING &&
+            participantChanges.left.has(this.signerAddress)
+        ) {
+            this.setStatus(Status.SYNCED);
```
lukastanisic99: I've created this card in trello for joining https://trello.com/c/Ew4apxMd/295-add-statuspendingparticipant-so-we-dont-abort-a-channel-when-we-committed-on-chain-and-waiting-for-inclusion

This should fall under that task, since this is a bit more complex - you first have to make sure you exit and update the on-chain snapshot, before you can continue spectating, since if that operation fails, better to fallback on participating and try again until successful

daiagi: card done in #298

test/utils/SyncCoordinator.ts
```diff
@@ -18,24 +18,33 @@ export class SyncCoordinator {
...
             blockHashInStorage?: Hash;
+            waitForFinalization?: boolean;
+            expectedHeight?: number;
         }
     ): Promise<void> {
-        const { timeoutMs = 8000, blockHashInStorage } = options || {};
+        const {
+            timeoutMs = 8000,
+            blockHashInStorage,
+            waitForFinalization = false,
```
lukastanisic99, line 63: why not default `true`

test/utils/SyncCoordinator.ts
```diff
@@ -54,16 +63,48 @@ export class SyncCoordinator {
...
+            // Check block sync (same hash and height)
+            const blocksInSync = blocks.every(
                 (b) => b!.hash === firstHash && b!.height === firstHeight
             );
+
+            if (!blocksInSync) return false;
+
+            // If a minimum expected height was provided, ensure all peers have
+            // reached at least that height before considering them synced.
+            if (expectedHeight !== undefined && firstHeight < expectedHeight) {
```
lukastanisic99: why name it `expectedHeight` if it's used as `minHeight`

test/utils/SyncCoordinator.ts
```diff
@@ -54,16 +63,48 @@ export class SyncCoordinator {
...
+                } catch (error) {
+                    this.logger.error(`Failed to get participants`, { error });
+                    return false;
+                }
+
+                for (const block of blocks) {
+                    const actualSignatures = block!.allSignatures.size;
+
+                    if (actualSignatures < totalParticipants) return false;
+                }
```
lukastanisic99: @daiagi this doesn't take into account the union, since diamondStateMachine only has the latest state (if someone left, you won't take that into account) + since you're using expectedHeight/minHeight better to directly target the block you want

```typescript
for(let i =0; i<peers.length; i++){
    peers[i].stateManager.agreementManager.didEveryoneSignBlock(blocks[i]);
}
```

test/utils/SyncCoordinator.ts
```diff
@@ -74,11 +115,36 @@ export class SyncCoordinator {
...
+                waitForFinalization && block
+                    ? ` sigs=${block.allSignatures.size}`
+                    : "";
+            return `Peer ${peer.index}: ${base}${sigs}`;
         });
 
+        let reason = "";
+        if (expectedHeight !== undefined) {
+            const latest =
+                peers[0]?.stateManager.storage.blocks.getLatestBlock(forkId);
```
lukastanisic99: you already fetched the block above - use it here, since this is volatile (getLatestBlock may return something else (not the block you fetched the first time) dependent when it runs -> may stop working as expected with async code)

test/utils/SyncCoordinator.ts
```diff
@@ -74,11 +115,36 @@ export class SyncCoordinator {
...
+            reason = ` (expected height ${expectedHeight}, have ${latest?.height ?? "?"})`;
+        }
+        if (waitForFinalization) {
+            try {
+                const participants =
+                    await peers[0].stateManager.diamondStateMachine.getParticipants();
```
lukastanisic99: same comment as above + also doesn't do union

test/utils/SyncCoordinator.ts
```diff
@@ -74,11 +115,36 @@ export class SyncCoordinator {
...
+                const sigCount =
+                    peers[0].stateManager.storage.blocks.getLatestBlock(forkId)
```
lukastanisic99: also same comment

test/harness/actions/DisputeOrchestrator.ts
```diff
@@ -106,12 +106,13 @@ export class DisputeOrchestrator {
...
-                expect(participants).to.have.lengthOf(honestPeers.length);
-                expect(participants).to.not.include(maliciousAddress);
-            }
+            const participants = await this.harness
+                .getPeer(honestPeerIndices[0]!)
+                .stateManager.diamondStateMachine.getParticipants();
+            expect(participants).to.not.include(
+                maliciousAddress,
+                "malicious peer should be removed from channel participants"
+            );
```
lukastanisic99: what was the problem here?

test/e2e/E2E-SpectatorStateProofPersistence.test.ts
```diff
@@ -0,0 +1,135 @@
...
+        });
+        let peerIndices = [0, 1, 2, 3];
+        let spectatorIndices = [];
+
+        // blocks 0, 1
+        await h.transition.advanceState({ count: 2 });
+
+        // Leave peer 2, block 2
+        await h.transition.advanceState({ txFn: (c) => c.leaveChannel() });
+        //  hack until https://github.com/peer3to/state-channels-plus/pull/298 merged
```
lukastanisic99, line 26: Not sure why you don't like `blockchain like PRs` when you build on top and it's interconnected - esspecially for testing and fixing bugs it makes a lot of sense

test/e2e/E2E-SpectatorStateProofPersistence.test.ts
```diff
@@ -0,0 +1,135 @@
...
+describe("E2E: Join/Leave Sequence", function () {
+    it.only("join/leave sequence and fork resolution", async function () {
+        const h = TestSession.getHarness();
+
+        await h.lifecycle.start(4, 0, {
+            timeConfig: {
+                p2pTime: 30,
+                agreementTime: 2,
+                chainFallbackTime: 2,
+                evidenceTime: 10
```
lukastanisic99, line 15: this can be 5s - even that is generous

test/e2e/E2E-SpectatorStateProofPersistence.test.ts
```diff
@@ -0,0 +1,135 @@
...
+        let peerIndices = [0, 1, 2, 3];
+        let spectatorIndices = [];
+
+        // blocks 0, 1
+        await h.transition.advanceState({ count: 2 });
+
+        // Leave peer 2, block 2
+        await h.transition.advanceState({ txFn: (c) => c.leaveChannel() });
+        //  hack until https://github.com/peer3to/state-channels-plus/pull/298 merged
+        await h.byzantine.disconnect(2);
```
lukastanisic99, line 27: All these disconnects are the root cause for the test failure - The peer is disconnected, but is still `STATUS == PARTICIPATING` but is not replicating state and is `stuck in the past`, so `harness` `advanceState` even with your added guard (pick status == participating), picks a peer that is participating but has an `outdated` view on the state (not synced) -> from their perspective peerX is next when from the latest peers perspective peerY is next, so when you tell peerX (who is synced), `do the next state transition` it throws the error you observed in `playTransaction`

test/e2e/E2E-SpectatorStateProofPersistence.test.ts
```diff
@@ -0,0 +1,135 @@
...
+        //  I expect it to be resolved once a leaving partiapcnt triger snasphot update on chain which trigger status change to SYNCED
+        await h.transition.advanceState({
+            count: 2,
+            waitForPeers: peerIndices.concat(spectatorIndices),
+            waitForFinalization: true
+        });
+
+        // Join peer 5 as spectator
+        await h.addPeer();
+        await h.event.waitUntilEventOccurs("onConnection", 5000, [5]);
```
lukastanisic99, line 88: we should have variations of this test - the spectator should join 

Some variations:
1) This one - the spectator is jsut a spectator and never actually joins - what are we expecting the outcome to be?
2) Like 1) but with disputes
3) Like 1) but the spectator joins and is a full participant
4) like 2) but the spectator joins and is a full participant
5) spectators join pre-dispute
6) spectators join mid-dispute
7) spectators join post-dispute

8) The baove cases that have disputes with and without snapshot update post dispute
9) All from above with joining through InboundMessages (on-chain join) and all the above where joining is done on the L2 where a normal state transitions injects you into the participant set (aka `friend invites`)

Think how you extract this in `scenarios` so we can have everything modular and reusable

test/e2e/E2E-FraudProofsBlockConfirmation.test.ts
```diff
@@ -0,0 +1,191 @@
...
+        await h.lifecycle.start(3, 3, { waitForFinalization: true });
+
+        const maliciousPeerIndex = 1; // NOT the expected next leader
+        await h.byzantine.submitUnexpectedNextLeaderBlock(maliciousPeerIndex);
+
+        await h.assert.dispute.initiatedAndCommitedWait({
+            allowByzantineInitiation: true
+        });
+        //  this fails with a race condition, some peers see the fork as disputed and fail the validation pipleine on `isDisputedFork` check
+        //  which runs before the next-leader check -> they never store the fraud proof
```
lukastanisic99: this is valid and was part of the design, where the fraud proof doesn't have to be gossiped, but is acted upon immediately and some peers may not detect it - they just observe that the peer is in `onChainSlashes`. However `at least 1 honest peer` will store the fraud proof, so we can assert for that instead of `every honest peer`. 

There are other similar race conditions that for now, somehow are not triggered - for instance `initiatedAndCommitedWait` expects all honest peers to initiate and commit to a dispute and for some reason this never fails, while by design peers don't have to dispute if they audited a dispute and have nothing to add (e.g. they agree with the dispute they observed), so not all peer have to dispute and yet this is never triggered ?!

test/e2e/E2E-FraudProofsBlockConfirmation.test.ts
```diff
@@ -0,0 +1,191 @@
...
+
+        // Peer 2 submits a block with a valid transaction but a wrong
+        // stateSnapshotHash (ZeroHash).
+        const maliciousPeerIndex = 2;
+        await h.byzantine.submitInvalidStateTransitionBlock(maliciousPeerIndex);
+
+        await h.assert.dispute.initiatedAndCommitedWait({
+            allowByzantineInitiation: true
+        });
+        h.assert.storage.honestPeersStoredFraudProof({
```
lukastanisic99, line 178: seems all test in this file suffer from the same race conditon - I ran one of them and didn't trigger for me, but in that case I think it's best to add a function `atLeastOneHonestPeerStoredFraudProof` or `add an option` to the existing implementation

test/utils/SyncCoordinator.ts
daiagi: because, I did that initially and it crated failing tests that were previously passing.
I prefer My PR's to be focused on single topic, and you prefer that too for reviewing.
I don't want to create sprawling "fix everything PR's"

changing this to default true is appropriate, I agree, but can be a TODO

https://trello.com/c/408fAc2i

test/utils/SyncCoordinator.ts
daiagi: >this doesn't take into account the union

yes, This PR was written before you introduced the union


will edit

test/harness/actions/DisputeOrchestrator.ts
daiagi: don't remember. looks like I had a  problem where the honest peers from the harness perspective were not synced, so I read it directly from the state machine.

later when I created the dispute tests I didn't use this code,  but instead marked malicious peers better
so I'm reverting this change

test/e2e/E2E-SpectatorStateProofPersistence.test.ts
daiagi: That's not true. I added them as hack and it is acknowledged.
the test was 
- passing when i wrote it
- failing after your recent changes
- failing but in a later stage with the disconnects

and yes, pointing to that PR acknowledges that staying as status participating is the isue

test/e2e/E2E-SpectatorStateProofPersistence.test.ts
daiagi: I've put a link to this comment in trello

test/e2e/E2E-SpectatorStateProofPersistence.test.ts
lukastanisic99: Add a trello card for this or make it part of the pending participant PR so we don't forget

test/e2e/E2E-SpectatorStateProofPersistence.test.ts
lukastanisic99: I added a trello card for this already - it's above the card you added :D

test/e2e/E2E-SpectatorStateProofPersistence.test.ts
daiagi: GitHub doesn't mark is as outdated, but it's done

test/e2e/E2E-FraudProofsBlockConfirmation.test.ts
daiagi: image

test/e2e/E2E-SpectatorStateProofPersistence.test.ts
daiagi: ok. i removed my card and added link to yours

test/e2e/E2E-SpectatorStateProofPersistence.test.ts
daiagi: https://trello.com/c/5tvOPzar

test/harness/actions/StateQueryActions.ts
```diff
@@ -109,8 +110,13 @@ export class StateQueryActions {
      */
     async getNextPeerToWrite(): Promise<TestPeer> {
         try {
+            // Find the first participating peer to query
+            const participatingPeer = this.harness.peers.find(
+                (peer) => peer.stateManager.getStatus() === Status.PARTICIPATING
```
lukastanisic99, line 115: What's probably better, is to find the peer with the highest blockHeight on the current `activeForkId` - this makes it irrelevant are they a participant or not in this case

test/e2e/E2E-SpectatorStateProofPersistence.test.ts
lukastanisic99: I meant `evidenceTime`

test/utils/SyncCoordinator.ts
```diff
@@ -17,81 +48,112 @@ export class SyncCoordinator {
...
+            tipHeight: number
+        ): Block[] | undefined => {
+            const onlyMinHeight =
+                blockHashInStorage == null && minHeight !== undefined;
+
+            if (!onlyMinHeight || tipHeight === minHeight) {
+                return tipBlocks;
+            }
+
+            const atPinnedHeight = peers.map((peer) =>
```
lukastanisic99, line 97: Not sure what happened here - harder to read and seems over complicated - it's harness code, so don't care much but all this `onlyMinHeight`, `!onlyMinHeight`, `atPinnedHeight` and in general this entire function with all the comments seems like AI over complication where it hallucinated that the problem is more complex than it is:

This was supposed to be simple, make sure all the peers have the same latest block. While you're looping and have the latestBlock for each peer, you only have 2 extra checks
```ts
for(peer in peers){
const latestBlock = ... 
if(waitForFinalization && !agreementManager.didEveryoneSignBlock(latestBlock)return false;
if(minHeight && latestBlock.height < minHeight) return false;
}
``

daiagi: The comments are helpful, Yes they are Ai generated but I sign over it. It is not over complicated. 

Maybe just let things be and merge my prs so I can continue to other tasks rather then keep mulling over small comments? 



Also there are agian changes to this in the pr that sits on top of this

test/harness/actions/StateQueryActions.ts
daiagi: Can we get some velocity here and not leave things hanging forever? 

Yes, your idea probably is good. 
Trello card, assign to me, move on

lukastanisic99 (approved):

test/utils/SyncCoordinator.ts
daiagi: https://trello.com/c/zGwHn8bn

test/harness/actions/StateQueryActions.ts
daiagi: https://trello.com/c/yugwL07t

PR #293, opened 2026-03-13
kill (final) dispute if audit data hash is invalid

daiagi:
this covers the comment in  `E2E-DisputeManager.test.ts` -> `"should ignore incorrect auditing data commitment in dispute posted without auditing data and resolve normally"`

If I'm not mistaken, this is what you meant


contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol
```diff
@@ -126,6 +129,31 @@ contract DisputeFraudProofFacet is StateChannelCommon {
         return _invalid();
     }
 
+    function _handleDisputeInvalidAuditingDataHash(bytes memory encodedFraudProof, Dispute memory dispute)
```
lukastanisic99, line 132: We don't need a fraud proof for this, since we only use disputeAuditingDataHash when uploading the dispute with calldata and while it's being uploaded on-chain, the data is hashed and compared - if it fails, the dispute was never committed and if it succeeds the data is made available and the hash is checked during verifyStateProof - @daiagi

test/e2e/E2E-DisputeManager.test.ts
```diff
@@ -104,25 +104,18 @@ describe("E2E: Dispute Manager", function () {
     });
 
     describe("Fraud Proof Detection", function () {
-        it("should ignore incorrect auditing data commitment in dispute posted without auditing data and resolve normally", async function () {
```
lukastanisic99, line 107: @daiagi I named this correctly - `should ignore auditing data commitment...` so an incorrect commitment is not possible when calldata is uploaded and is irelevant when calldata is not uploaded

PR #295, opened 2026-03-14
fetch-logs, fixed milestoneBuilding, auditingData construction and other stuff

lukastanisic99:
- Created fetch-logs to fetch all logs for a channelId,
- fixed milestoneBuilding,
- fixed construction of AuditingData,
- extended SyncPayload to always provide the genesis state, so in the worst case the user can fallback on it as implicit finality,
- fixed some timeout too early bugs,
- fixed joining peer not signing the join block,
- fixed https agent only for node environments,
- fixed and stabilized some e2e tests and removed some unit tests that were causing the sdk to not compile/build


lukastanisic99:
wtf git didn't detect just the changes and thinks everything is a conflict while the difference is a single commit -.-

lukastanisic99:
@daiagi somehow yesterdays merge/rebase fucked git and it didn't deduct anything and again showed a bunch of conflicts - I resolved the conflict with an automatic merge that selects this branch in full

PR #296, opened 2026-03-15
Spam / No merit dispute -> killed

daiagi:
Kill Spam disputes, aka disputes with no merit

### DIFF
--- a/contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol
+++ b/contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol
@@ -134,12 +134,16 @@ contract DisputeFraudProofFacet is StateChannelCommon {
         pure
         returns (address)
     {
-        abi.decode(encodedFraudProof, (InvalidDisputeReason));
+        InvalidDisputeReason memory proof = abi.decode(encodedFraudProof, (InvalidDisputeReason));
 
-        if (!_hasDisputeReason(dispute.input)) {
-            return _valid(dispute.input.disputer);
+        if (!_isSnapshotLinkedToLatestBlock(dispute, proof.latestStateSnapshot)) {
+            return _invalid();
         }
-        return _invalid();
+
+        if (_hasDisputeReason(dispute.input, proof.latestStateSnapshot)) {
+            return _invalid();
+        }
+        return _valid(dispute.input.disputer);
     }
 
     function _handleDisputeInvalidOutputState(bytes memory encodedFraudProof, Dispute memory dispute)
--- a/contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol
+++ b/contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol
@@ -64,6 +64,9 @@ contract DisputeFraudProofFacet is StateChannelCommon {
         if (proofType == DisputeFraudProofType.DisputeLastMilestoneNotFinalAndNoAuditingData) {
             return _handleDisputeLastMilestoneNotFinalAndNoAuditingData;
         }
+        if (proofType == DisputeFraudProofType.InvalidDisputeReason) {
+            return _handleInvalidDisputeReason;
+        }
         return _handleInvalidDisputeFraudProofType;
     }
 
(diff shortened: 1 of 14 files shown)


contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol
```diff
@@ -126,6 +129,23 @@ contract DisputeFraudProofFacet is StateChannelCommon {
...
+        internal
+        pure
+        returns (address)
+    {
+        abi.decode(encodedFraudProof, (InvalidDisputeReason));
+
+        bool hasTimeout = dispute.input.timeout.participant != address(0);
+        bool hasOnChainSlashes = dispute.input.onChainSlashes.length > 0;
+        bool hasSelfRemoval = dispute.input.selfRemoval;
+
```
lukastanisic99, line 142: missing `isForcedInboundMessage` e.g. `dispute.input.lastInboundMessageBlockHeight>latestStateSnapshot.latestInboundMessageBlockHeight`

but for this you also have to upload the snapshot and verify it against the last block in stateProof

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
```diff
@@ -73,6 +73,16 @@ contract DisputeManagerFacet is StateChannelCommon {
...
                 RaceConditionDisputeEvidencePeriodExpired()
             );
+
+            if (hasNoCommitments) {
+                // Spam dispute: the window was opened while there were no on-chain commitments,
+                // so it was killed before any honest peer committed. The kill slashes the
+                // spammer at T > creationTimestamp. Reset creationTimestamp so the
+                // OnChainSlashesNotSubset check accepts that slash in the next round.
+                disputeWindow.evidence.creationTimestamp = block.timestamp;
+                delete disputeWindow.evidence.hasPosted;
```
lukastanisic99, line 83: Need to think about this, since I've intentionally not restarted the window creation timestamp, to have auditing be deterministic for on-chain slashes and some timeout stuff. The most logical thing to do would be to use the dispute creation timestamp, but would have to calculate the commitment as hash(dispute,timestamp) same as we do for blockcalldata and we did this b4 also for disputes, but it got removed later. 

The idea here was to have on-chain slashes that come after the dispute was created be applied in `reduce`, especially since we allow only 1 dispute per peer per fork, so if we really wanted to guarantee that everyone gets slashed in that dispute it would have to be in reduce, since it runs last

I'll think a bit and get back to you on this one

src/stateManager/DisputeValidationService.ts
```diff
@@ -440,6 +440,22 @@ export default class DisputeValidationService {
...
+
+        const hasOnChainSlashes = dispute.input.onChainSlashes.length > 0;
+
+        const hasSelfRemoval = dispute.input.selfRemoval;
+
+        if (!hasTimeout && !hasOnChainSlashes && !hasSelfRemoval) {
+            this.logger.warn("Dispute has no legitimate enforcement basis", {
+                dispute: LoggerUtils.getDisputeMetadata(dispute)
+            });
+            this.disputeFraudProofService.createInvalidDisputeReason(dispute);
```
lukastanisic99, line 458: usually we write a solidity function that's called on-chain and here to perform the check, so we reuse the logic and have deterministic execution

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
lukastanisic99: I've added a trello card to think about this once I'm done with more critical stuff - Fine to be like this for now

contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol
lukastanisic99: @daiagi this comment still not addressed

daiagi: Yes, not sure what you mean.
Yes I'm not sure what you want here. 

Who is supposed to upload a snapshot?  Dispute validation service? 

Not sure how uploading snapshot  fits in the life cycle of uploading dispute and validating it

lukastanisic99: A legit reason to upload a dispute is when no one wants to consume the inbound message from L1, so a pending participant creates a dispute with the latestInboundMessageBlocksHash to `force` inclusion. That's my first comment above, the snapshot is needed so you can verify it against the latest block in stateProof when you're doing the comparison from above.

daiagi: done

lukastanisic99 (approved):

PR #297, opened 2026-03-15
dispute fraud proofs tests

daiagi:
Test for dispute validation pipeline fraud proofs

### DIFF
--- a/contracts/V1/StateChannelDiamondProxy/StateProofFacet.sol
+++ b/contracts/V1/StateChannelDiamondProxy/StateProofFacet.sol
@@ -45,31 +45,25 @@ contract StateProofFacet is StateChannelCommon {
         virtual
         returns (bool)
     {
-        console.log("STATE PROOF 1");
+        // Not needed, done already in DisputeManagerFacet uploadDisputeWithCalldata
         require(
             dispute.input.disputeAuditingDataHash == keccak256(abi.encode(disputeAuditingData)),
             ErrorAuditingDataHashMismatch()
         );
-        console.log("STATE PROOF 2");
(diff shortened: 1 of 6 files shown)


contracts/V1/StateChannelDiamondProxy/StateProofFacet.sol
```diff
@@ -45,42 +45,34 @@ contract StateProofFacet is StateChannelCommon {
         virtual
         returns (bool)
     {
-        console.log("STATE PROOF 1");
+        // Not needed, done already in DisputeManagerFacet uploadDisputeWithCalldata
```
lukastanisic99: Ye I know, but it's added here since the auditor needs to provide the data - the dispute upload guaranteed that the data is available, but here they have to upload correct data (not junk) to prove that the stateProof is invalid, otherwise they could upload whatever they wanted and an honest stateProof could fail

test/harness/actions/DisputeTamperingActions.ts
```diff
@@ -161,6 +170,203 @@ export class DisputeTamperingActions {
...
+        dispute.input.disputeAuditingDataHash = hash(
+            Codec.encode(auditingData, Type.DisputeAuditingData)
+        ) as `0x${string}`;
+
+        this.logger.debug(
+            `Truncated state proof to height ${targetHeight} for disputer ${disputerPeerIndex}`
+        );
+    }
+
+    private getLatestBlockFromStateProof(
```
lukastanisic99: We have a solidity implementation for this and I think also in agreementManager (left a comment in agreementManager why there's a dual implementation instead of reusing solidity)

lukastanisic99 (approved):

PR #298, opened 2026-03-19
maybeExitOnChain and PENDING_PARTICIPANT

daiagi:
- Implement maybeExitOnChain, on self exit, update on chain snasphot or init a self removal dsipute
- status change in `onStateSnapshotUpdated`
-  promote SYNCED -> PENDING_PARTICIPANT when included in local state
- promote PENDING_PARTICIPANT → PARTICIPATING when included in on chain snapshot

### DIFF
--- a/src/eventHandlers/EventHandler.ts
+++ b/src/eventHandlers/EventHandler.ts
@@ -92,37 +92,28 @@ export class EventHandler {
             addressesEqual(p, signerAddress)
         );
 
-        // Detect when we've been included in the on-chain snapshot: PENDING_PARTICIPANT → PARTICIPATING
-        if (status === Status.PENDING_PARTICIPANT) {
-            if (snapshotHasSigner) {
-                this.logger.info(
-                    "onStateSnapshotUpdated - signer included in snapshot, transitioning PENDING_PARTICIPANT → PARTICIPATING",
-                    { channelId }
-                );
-                this.stateManager.setStatus(Status.PARTICIPATING);
-            }
-        }
-
         // Detect when we've fully left the channel: PARTICIPATING → SYNCED
         if (status === Status.PARTICIPATING) {
-            const pendingParticipants =
-                (await this.stateManager.stateChannelManagerContract.getPendingParticipants(
-                    channelId
-                )) as Address[];
             const localParticipants =
                 await this.stateManager.getParticipantsCurrent();
-            const inPending = pendingParticipants.some((p) =>
-                addressesEqual(p, signerAddress)
-            );
             const inLocal = localParticipants.some((p) =>
                 addressesEqual(p, signerAddress)
             );
-            if (!snapshotHasSigner && !inPending && !inLocal) {
-                this.logger.info(
-                    "onStateSnapshotUpdated - signer left channel, transitioning PARTICIPATING → SYNCED",
-                    { channelId }
+            if (!snapshotHasSigner && !inLocal) {
+                const pendingParticipants =
+                    (await this.stateManager.stateChannelManagerContract.getPendingParticipants(
+                        channelId
+                    )) as Address[];
+                const inPending = pendingParticipants.some((p) =>
+                    addressesEqual(p, signerAddress)
                 );
-                this.stateManager.setStatus(Status.SYNCED);
+                if (!inPending) {
+                    this.logger.info(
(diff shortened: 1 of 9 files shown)


src/eventHandlers/EventHandler.ts
```diff
@@ -76,6 +83,49 @@ export class EventHandler {
...
+                    { channelId }
+                );
+                this.stateManager.setStatus(Status.PARTICIPATING);
+            }
+        }
+
+        // Detect when we've fully left the channel: PARTICIPATING → SYNCED
+        if (status === Status.PARTICIPATING) {
+            const pendingParticipants =
+                (await this.stateManager.stateChannelManagerContract.getPendingParticipants(
```
lukastanisic99: In 99% of the cases when a snapshot is updated, we are a participant, so it's better to do this `on-chain querry last` since we'll in most cases return early on `snapshotHasSigner` and `inLocal`, so no need to consume resources on the RPC node in most cases

src/stateManager/StateManager.ts
```diff
@@ -1360,7 +1391,7 @@ class StateManager {
 
                     throw error;
                 });
-            DetachedPromises.collect(txResponsePromise);
+            await txResponsePromise;
```
lukastanisic99: This was aiated before and I changed it to detached, since it was blocking while there is no reason for it to be blocking - it was designed to function async. 

The problems of it being blocking, it's called in the BlockConfirmation pipeline, so it holds the mutext and doesn't let other blocks enter the pipeline. It laso was the cause of some failing tests, that blocks were getting timedout waiting for this. In production, you have no gurantee that the transaction will be mined 'instantly' and this can take some time -> should be non blocking, async and event driven as it was - most (if not all) our blockchain transactions should be detached

src/stateManager/StateManager.ts
```diff
@@ -2417,13 +2448,15 @@ class StateManager {
...
+        // step 9 - potentially change status: SYNCED → PENDING_PARTICIPANT
+        // Local state now includes us (inbound JOIN was processed), but we wait for
+        // on-chain snapshot confirmation before becoming fully PARTICIPATING.
+        // onStateSnapshotUpdated handles PENDING_PARTICIPANT → PARTICIPATING.
         if (this.status === Status.SYNCED) {
             const participants =
                 await this.diamondStateMachine.getParticipants();
             const isParticipant = participants.includes(this.signerAddress);
-            if (isParticipant) this.setStatus(Status.PARTICIPATING);
+            if (isParticipant) this.setStatus(Status.PENDING_PARTICIPANT);
```
lukastanisic99: This should be as it was 'Participating' - `Status.pending_participant` is just temporary, when we issue the join transaction until we observe it's successful, so we don't use the spectating strategy and don't abort on failure, but at the same time you don't want to use the participating strategy since you don't want to sign blocks too early. Here, where we're part of the participant set of the state, our signature is expected and we are a participant, so in order for the step bellow to sign, we have to have status.participating

src/types/flags.ts
```diff
@@ -12,5 +12,6 @@ export enum Status {
     NOT_OPENED,
     OPENED,
     SYNCED,
+    PENDING_PARTICIPANT, // joined on-chain, waiting for inclusion in the on-chain snapshot
```
lukastanisic99: initiated join on-chain, waiting for inclusion in the state (this has to happen before the on-chain snapshot), so it either happens during the state transition or after the dispute (at genesis) - both accounted for 

The point of this flag is to not be a spectator when we committed, and we're waiting to become part of the state

src/eventHandlers/EventHandler.ts
```diff
@@ -76,6 +83,49 @@ export class EventHandler {
...
+        const snapshotParticipants = stateSnapshot.snapshotData
+            .participants as Address[];
+        const status = this.stateManager.getStatus();
+
+        const snapshotHasSigner = snapshotParticipants.some((p) =>
+            addressesEqual(p, signerAddress)
+        );
+
+        // Detect when we've been included in the on-chain snapshot: PENDING_PARTICIPANT → PARTICIPATING
+        if (status === Status.PENDING_PARTICIPANT) {
```
lukastanisic99: This is not needed, since setState and state transitions already account when we should change from `pending` to `participating`

The trello card was about when to do `synced->pending` and `participating->synced`

lukastanisic99 (approved):

PR #299, opened 2026-03-23
Wait-for-finalization-true

daiagi:
Chain PR
~~sits on top of #290~~ Edit: Targets `dispute` branch
- Closes https://trello.com/c/408fAc2i

- Set `waitForFinaliztion` to `true` by default
- Added jsdocs to the various TransactionOptions in the harness, it is getting a bit much
- Some resolvers for the options, for example passing `waitForPeers` sets `waitForFinalization` to `false
- When doing more then single trnsaction at once, wait for minheight and finalization only after last tx
- [dev-ex]: Added `--grep` arg for `yarn:test:e2e`
- Minor cleanups and syntax improvemnts across the tests


daiagi:
@lukastanisic99  Note PR #300 that targets this PR

lukastanisic99 (approved):

PR #301, opened 2026-03-25
Join-leave-fixes

daiagi:
- Improve harness API for leave and join, both wait/assert until the peer status changes, leaving updates the harness context so leaving peer is nor included for finaliztion.

- "E2E: Join/Leave Sequence" intentially dumbed down to make a point- even in this very basic form (trivial boring test) it was failing on line 39:

```ts
      // Join peer 4 as spectator (`addPeer` waits for SYNCED)
        await h.addPeer();
```

the joining peer never reached SYNCED status

- Fixes follow:

    1. `getMessageBlocksInRange` direction was reversed
    2. getGenesisTimestamp check removed from isCorrectGenesis

        getGenesisTimestamp reads stateSnapshots[channelId] in the local diamond. After step 1 *(Fetch the onChainSnapshot and persist/update the local EVM with it)* writes the current on-chain snapshot there, the slot no longer holds the genesis, so the call returned (false, 0) and aborted valid syncs. The timestamp check was     redundant anyway: isGenesisSnapshotWithoutTimeCheck already proves forkId == keccak256(snapshotData)

    3. outboundMessageBlocksOfTheLatestFork started at the wrong offset in the multicall

- Left over fix - in StateManger, promote `Pending Participant`  -> `Participating`, in addition to the already exiting promotion of `Synced` -> `Participating`


src/rpc/services/spectate/SpectateRpcMethods.ts
```diff
@@ -204,38 +204,29 @@ class SpectateServiceRpcMethods extends ARpcMethods {
...
-                    channelId,
-                    syncPayload.latestForkGenesisSnapshot.snapshotData
-                        .originForkId,
-                    finalForkId
-                );
-            isCorrectGenesis =
-                isCorrectGenesis &&
-                isAvailable &&
-                genesisTimestamp ==
-                    syncPayload.latestForkGenesisSnapshot.timestamp;
```
lukastanisic99, line 225: @daiagi ping me for a call later on this one - can't explain quickly in a few sentences

src/rpc/services/spectate/SpectateRpcMethods.ts
```diff
@@ -204,38 +204,29 @@ class SpectateServiceRpcMethods extends ARpcMethods {
...
-            // 2.7) verify outboundMessageBlocks from onChainSnapshot to final genesisSnapshot
+            // 2.7) verify outboundMessageBlocks from final genesisSnapshot to onChainSnapshot
+            // generateSyncPayload collects blocks from genesis → onChain (genesis is earlier/from)
             let areValidExitBlocks =
                 await diamondStateMachine.localDiamondContract.verifyOutboundMessageBlocks(
                     syncPayload.outboundMessageBlocksUpToLatestGenesis,
-                    onChainSnapshot.snapshotData,
-                    syncPayload.latestForkGenesisSnapshot.snapshotData
+                    syncPayload.latestForkGenesisSnapshot.snapshotData,
+                    onChainSnapshot.snapshotData
```
lukastanisic99: The confusion here is that solidity and typescript are not consistent on the meaning `fromSnapshot` and `toSnapshot`

Typescript always iterates backwards and uses `from == max` and `to == min`, in solidity (at least) this function uses the convention iterate forward `from == min` and `to == max`

We should change some naming here so it's less confusing and more clear what's expected so there is no divergence in expectation from solidity and typescript

daiagi: - #303

src/rpc/services/spectate/SpectateService.ts
```diff
@@ -391,13 +397,32 @@ class SpectateService extends ARpcService<SpectateServiceRpcMethods> {
...
+                );
+                if (startIndex === -1) {
+                    // not found: nothing in the payload continues from this on-chain outbound head.
+                    // [] is not a valid proof to the milestone; staticCall reverts (reject sync).
+                    outboundBlocksForSameFork = [];
+                } else {
+                    outboundBlocksForSameFork =
+                        outboundBlocksForSameFork.slice(startIndex);
+                }
+            }
```
lukastanisic99: @daiagi  this is confusing to read + this needs to be written in solidity since in the worst case it has to run on-chain to account for race conditions. 

I accounted for this race condition when I was writing `verifyStateProof` and `verifyMilestones` (https://github.com/peer3to/state-channels-plus/blob/eb39fcb0416b2d5b1c0b2e74358dedc452debf35/contracts/V1/StateChannelDiamondProxy/StateProofFacet.sol#L294), but I didn't account for pruning outboundBlocks so we have to update `_verifyOutboundMessageBlocks` (https://github.com/peer3to/state-channels-plus/blob/eb39fcb0416b2d5b1c0b2e74358dedc452debf35/contracts/V1/StateChannelDiamondProxy/StateChannelCommon.sol#L322 ) to take pruning into account and return the `pruned array` alongside `success`. We could have the pruning implemented in a separate solidity function that's called prior to `_verifyOutboundMessageBlocks` in `StateSnapshotFacet` and in that case not touch `_verifyOutboundMessageBlocks` - fine with both options

Either way this has to run on-chain since off-chain you can never fully account for race condition and we want all of these functions to be best effort. If you decide to do pruning here (off-chain) too as an potential optimization to save on gas on-chain, just reuse the solidity function through the local diamond (it should be stateless)

test/harness/actions/TransitionActions.ts
```diff
@@ -190,6 +195,42 @@ export class TransitionActions {
...
+
+        await this.submitNext((c) => c.leaveChannel(), {
+            waitForTurn: true,
+            waitForSync: options?.waitForSync ?? true,
+            waitForPeers: options?.waitForPeers,
+            waitForFinalization: options?.waitForFinalization ?? true,
+            delayMs: options?.delayMs
+        });
+
+        await this.harness.event.waitUntilPeerStatus(
```
lukastanisic99, line 224: This is blocking, so there should be a `particiapntLeaveDetached` and `particiapntLeaveWait` implementation, or to not have the `assert` condtion here and just have the `state transition` like other state transition `helpers` in this file, and do the assertion for status change explicitly where needed either as `wait` or `detached` like other assert conditions

test/fixtures/PeerTestHarness.ts
```diff
@@ -252,9 +253,13 @@ export class PeerTestHarness<
...
     /**
      * Create a new peer after `setup()` has already run.
-     * If a channel is already open (i.e. `this.channelId` is set), the peer is also connected to that channel.
+     * If a channel is already open, connects to it and waits until that peer reaches `Status.SYNCED`
+     * (spectate path); the new node is appended to `peers` like any other harness peer.
      */
-    public async addPeer(signer?: Signer): Promise<TestPeer<TFactories>> {
+    public async addPeer(
+        signer?: Signer,
+        options?: { statusTimeoutMs?: number; statusTimeoutMessage?: string }
```
lukastanisic99: if the `signer` is optional, just add the signer to the options

test/fixtures/PeerTestHarness.ts
```diff
@@ -282,6 +287,13 @@ export class PeerTestHarness<
                 this.channelId,
                 peer.address
             );
+
+            await this.event.waitUntilPeerStatus(index, Status.SYNCED, {
```
lukastanisic99: also good to have `addPeerDetached` and `addPeerWait`, so we can test our more asyncrhonous joins

test/e2e/E2E-ParticipantLifecycle.test.ts
```diff
@@ -58,24 +50,18 @@ describe("E2E: Participant Lifecycle", function () {
...
-            await h.eventCountsBarrier.waitFor(
-                () => spectator.stateManager.getStatus() === Status.SYNCED,
-                {
-                    timeoutMs: 5000,
-                    timeoutMessage: "Spectator did not reach SYNCED status"
-                }
-            );
-
             // Build JoinChannelConfirmationStruct --------------------------
             const jc: JoinChannelStruct = {
```
lukastanisic99: extract JoinChannelStruct formation and signing into a utility/helper functions where they belong

maybe have `harness.join.*` where we put all the joining stuff (`addPeer`, etc...)

test/e2e/E2E-ParticipantLifecycle.test.ts
```diff
@@ -126,17 +112,11 @@ describe("E2E: Participant Lifecycle", function () {
 
             await h.transition.advanceState({ count: 1 });
```
lukastanisic99, line 87: Want to comment a on something a few lines up 'Fire joinChannel WITHOUT awaiting' - ye, as mentioned before we should have all our on-chain transaction run detached with .them().catch() like in other places, so we don't block other parts of the app waiting on block inclusion that may take a very long time

test/e2e/E2E-ParticipantLifecycle.test.ts
```diff
@@ -126,17 +112,11 @@ describe("E2E: Participant Lifecycle", function () {
...
-                () =>
-                    spectator.stateManager.getStatus() === Status.PARTICIPATING,
-                {
-                    timeoutMs: 15000,
-                    timeoutMessage:
-                        "Joiner did not reach PARTICIPATING after snapshot update"
-                }
+            // Joiner is now PARTICIPATING — promoted inside success() when the first
+            // block that includes them in the resulting participant set was processed.
+            expect(spectator.stateManager.getStatus()).to.equal(
```
lukastanisic99, line 90: Ye, but you might have a race condition here since you're `awaiting`/`observing` only on the joining peer, and if the peer that produces the block doesn't see the inboundMessage it will not include it -> this fails - better to have `assert.honestPeersObserveInboundMessage` `detached/wait`

test/e2e/E2E-SpectatorStateProofPersistence.test.ts
```diff
@@ -15,121 +15,108 @@ describe("E2E: Join/Leave Sequence", function () {
...
-        }
+        await h.assert.sync.blockHeight({ expectedHeight: 6 });
+
+        // // peer 0 is leaving the channel, block 7
+
+        // await h.transition.advanceState({
+        //     txFn: (c) => c.leaveChannel(),
+        //     waitForPeers: peerIndices.concat(spectatorIndices),
+        //     waitForFinalization: true
+        // });
```
lukastanisic99: Commented out since something is failing?

daiagi: yes. resolved already in later PR
https://github.com/peer3to/state-channels-plus/pull/306

test/e2e/E2E-ParticipantLifecycle.test.ts
daiagi: done already in later PR. will move into `h.join`

src/rpc/services/spectate/SpectateService.ts
daiagi: done in #309

test/harness/actions/TransitionActions.ts
daiagi: done in #309

test/fixtures/PeerTestHarness.ts
daiagi: done in #309

test/fixtures/PeerTestHarness.ts
daiagi: #309

test/e2e/E2E-ParticipantLifecycle.test.ts
daiagi: #309

test/e2e/E2E-ParticipantLifecycle.test.ts
daiagi: addresses in #309

PR #302, opened 2026-03-26
Fix-e2e-tests

daiagi:
## 📋 Description

- FraudProofFacet, introduce missing handler `        if (proofType == FraudProofType.InvalidTimestamp) return _handleInvalidTimestamp;`

- stored dispute fraud proofs, set `atLeastOneHonestPeer` to true

- `resolveDisputeWait`, make options object optional, mostly it is not need anymore, hanress keeps track of malicios/honest peers, assertions "just" work out of the box

- stabilize block and dispute fraud proofs tests suites


contracts/V1/StateChannelDiamondProxy/FraudProofFacet.sol
```diff
@@ -202,6 +203,40 @@ contract FraudProofFacet is StateChannelCommon {
...
+            previousTimestamp = previousStateSnapshot.timestamp;
+        } else {
+            if (!isBlockAuthentic(proof.previousBlock)) return _invalid();
+            Block memory previousBlock = abi.decode(proof.previousBlock.encodedBlock, (Block));
+            if (fraudBlock.previousBlockHash != keccak256(abi.encode(previousBlock))) return _invalid();
+            previousTimestamp = previousBlock.transaction.header.timestamp;
+        }
+
+        bool isValidTimestamp = fraudBlock.transaction.header.timestamp >= previousTimestamp
+            && fraudBlock.transaction.header.timestamp <= previousTimestamp + getP2pTime();
```
lukastanisic99: this doesn't take into account the `relevantTimestamp`, you have to check wether calldata is posted for the previous block (verify correct calldata), then you have to take into account the signature (did the peer forfeight extra time) - basically match all the checks as in typescript :D 

It's weird that I implemented the `soldity struct`, but didn't implement the fraud proof :D

test/e2e/E2E-DisputeValidationPipeline.test.ts
```diff
@@ -20,6 +19,8 @@ describe("E2E: Dispute Validation Pipeline", function () {
             const h = TestSession.getHarness();
             await h.scenario.preDisputeSetup();
 
+            h.tamper.delayDisputeForPeers([0, 2]);
```
lukastanisic99: Why do we need to delay? It's fine if all 3 peers create a dispute at the same time and N-1 are malicious - the honest peer can kill all N-1 without having to redispute later

test/e2e/E2E-DisputeValidationPipeline.test.ts
```diff
@@ -420,7 +371,10 @@ describe("E2E: Dispute Validation Pipeline", function () {
                 disputeFraudProofType:
                     DisputeFraudProofType.TimeoutNotLinkedToLatestState
             });
-            await h.dispute.resolveDisputeWait({ maliciousPeerIndex: 2 });
+            // mark peer 0 as malicious
+            h.contextApi.markMaliciousPeer({ maliciousPeerIndex: 0 });
```
lukastanisic99, line 544: there are other cases above that didn't `mark` - probably best to include this in the subbing as an option with default `true` since most stubs are malicious I guess

contracts/V1/StateChannelDiamondProxy/FraudProofFacet.sol
daiagi: #309

test/e2e/E2E-DisputeValidationPipeline.test.ts
daiagi: the stubbing includes it, the extra marking was  not needed

removed in #309

test/e2e/E2E-DisputeValidationPipeline.test.ts
daiagi: - tests with 3 peers, works without it

- test with more, not stable without it. I'm still not sure why.

daiagi: Yep, not needed.
removed in #309

daiagi:
ok. ready for approval i believe, all issues are addressed  in #309

PR #306, opened 2026-03-30
join-leave test passing

daiagi:
passing test:

join -> leave -> assert spectators have moved to new fork after dispure resoutions

- Closes https://trello.com/c/5tvOPzar


test/e2e/E2E-SpectatorStateProofPersistence.test.ts
```diff
@@ -35,90 +35,71 @@ describe("E2E: Join/Leave Sequence", function () {
...
+            "Fork should have changed after dispute resolution"
+        );
+
+        // spectators have movedto the new fork
+        for (const i of spectatorIndices) {
+            expect(h.getPeer(i).stateManager.forkId).to.equal(
+                postDisputeForkId,
+                `spectator peer ${i} should be on new fork`
+            );
+        }
```
lukastanisic99, line 107: we should also have a test where the spectators join (are participating) - we can have 2 version of the test `wait`/`detached` where 'wait' more syncrhonous and the 'detached' fully async - probably best to put this into a `scenarion` and wrap it with options {`justSpectate`,`joinDetached`, `joinWaited`}

lukastanisic99: Btw how come spectator peers didn't abort and persisted through disputes?

daiagi: >we should also have a test where the spectators join


I've put a Trello card

>Btw how come spectator peers didn't abort and persisted through disputes?

I've pushed a fix (in this PR)

what was happening:
the spectators **DID** disconnected from the p2p network, but still listened on `onDisputeCommited` event and updated their local fork id

What i changed:
1. added
```ts
            if (
                status !== Status.PARTICIPATING &&
                status !== Status.PENDING_PARTICIPANT
            )
                return;
```

to `onDisputeCommited` so that `Synced` peers will not update their fork id

2. made clearer assertions in the test

lukastanisic99: hmm better to have tis responsibility in spectate.abort() where they unsubscribe and dispose the channel and not to have spectating logic spill into other components, so if this changes and we make it easier to recover we don't have to remember to update other places

Plus this is added under `isFinal` not sure we ever hit final disputes at the moment (maybe if there's only 1 peer left)

daiagi: yep. moved it
the `spectate.abort` disposes of the listener
the finality was side effect, the channel is left with only 1 participating peer.

PR #308, opened 2026-04-03
Test-sync-across-forks

daiagi:
## 📋 Description

Test case

- Add two spectators to a channel.
1. before a dispute (spectator A)
2. after a dispute (spectator B)

Assert that:

1. A observes the dispute and disconnects
2. B syncs

## 💪 Related Issues

- Closes https://trello.com/c/tvxQix4R
- Closes https://trello.com/c/yugwL07t


src/rpc/services/initHandshake/InitHandshakeService.ts
```diff
@@ -254,32 +254,39 @@ class InitHandshakeService extends ARpcService<InitHandshakeRpcMethods> {
...
-                    `Skipping sync after handshake with peer ${completedPeerAddress} - not a participant`
-                );
-            }
+        // Only pull state from on-chain participants when we are not already
+        // PARTICIPATING.
+        const channelOpen = localStatus !== Status.NOT_OPENED;
+        if (
+            channelOpen &&
+            isPeerParticipant &&
+            localStatus !== Status.PARTICIPATING
```
lukastanisic99: what was the problem here?

src/rpc/services/spectate/SpectateRpcMethods.ts
```diff
@@ -206,27 +206,29 @@ class SpectateServiceRpcMethods extends ARpcMethods {
...
+            const isCorrectGenesis =
+                finalForkIdMatchesGenesisForkId &&
+                isGenesisValid &&
+                stateHashMatch;
+
             if (!isCorrectGenesis) return this.service.abort(peerAddress);
 
-            // 2.7) verify outboundMessageBlocks from final genesisSnapshot to onChainSnapshot
-            // generateSyncPayload collects blocks from genesis → onChain (genesis is earlier/from)
+            // 2.7) verify outboundMessageBlocks from onChainSnapshot (lower/older) to final genesisSnapshot (upper/newer)
```
lukastanisic99, line 226: hmm this is only correct if onChainSnapshot is on some older fork - think my comment on the first PR regarding pruning and verification of OutboundMessages won't help here if there's a race condition and the on-chain snapshot is updated to the same fork. It only helps in the ralation `[myPerception, ...onChainPerception, .... target]` -> it prunes it in place into `[onChainPerception,....target]`

But it can't help (currently) if `[myPerception, ...... target, ....onChainPerception]`, but since the `snapshots` contain the `block heights` of the `outboundBlockchain` we could make it succeed by pruning everything before `onChainPerception.outbound.height` and maybe we should do this? 

It's fine if the snapshot is updated to any state `<= latest state synced`, but if it's updated to some state that peers didn't sync us -> we should abort

src/rpc/services/spectate/SpectateService.ts
```diff
@@ -132,9 +132,10 @@ class SpectateService extends ARpcService<SpectateServiceRpcMethods> {
         const forkId = _forkId || stateManager.forkId;
 
         // -------- Collect what is needed to prove the latestForkGenesisSnapshot starting from the onChainSnapshot --------
-        // We'll do all the computation on our local state. If our local state is not synced we shouldn't even be syncing the spectator and we probably have bigger problems
+        // Anchor outbound block ranges to the same source the spectator uses in fetchAndPersistOnChainSnapshot
+        // (stateChannelManagerContract). LocalDiamond can lag or diverge slightly from RPC (e.g. late replay);
+        // using local getStateSnapshot here caused verifyOutboundMessageBlocks to fail against the verifier's chain snapshot.
```
lukastanisic99: seems this race condition problem is everywhere - hopefully solved with the other comments

src/rpc/services/spectate/SpectateRpcMethods.ts
daiagi: I've put a Trello card

src/rpc/services/initHandshake/InitHandshakeService.ts
daiagi: reverted, it was not needed.

daiagi:
@lukastanisic99 
I merged "higher" PR's so now this PR is big and not clean anymore

all you comments are either logged in Trello or resolved
So i think you can just approve it, 
if you prefer to review again, checkout the PR in you Vscode and checkout the commit `ab522d503a228774f3855d36018882390830783b`

this way you can see the "clean" changes of this PR again, without the extra changes from upper PR's

PR #313, opened 2026-04-14
Multicall-bug-fixes

daiagi:


lukastanisic99:
You're targeting the 'master' branch

contracts/V1/StateChannelDiamondProxy/DisputeVerificationFacet.sol
```diff
@@ -521,6 +532,20 @@ contract DisputeVerificationFacet is StateChannelCommon {
         emit DisputeKilled(dispute.input.channelId, forkId, dispute.input.disputer, commitment);
     }
 
+    function _inParticipantUnion(
```
lukastanisic99: could add this to the `UtilityFacet` so it can be used in other places, especially since it's `pure` so can run directly without a delegatecall

lukastanisic99 (approved):

contracts/V1/StateChannelDiamondProxy/DisputeVerificationFacet.sol
daiagi: moved

PR #314, opened 2026-04-17
dedicated test suite to dispute fraud proofs about latestStateSnapshotHash

daiagi:
from this test:

```ts
        it("DisputeInvalidStateProof: milestones, wrong hash, pipeline (missing last block)", async function () {

            await h.lifecycle.timeoutSetup(4, 2, {
                timeConfig: {
                    evidenceTime: 6,
                    agreementTime: 6
                }
            });
            // peer 2 leaves — all four peers (including peer 1) observe the milestone
            await h.transition.participantLeaveDetached({
                statusTimeoutMs: 20000
            });
            // peer 3, peer 0, peer 1 each take a turn; peer 1 is now NOT "next"
            await h.transition.advanceState({
                waitForPeers: [0, 1, 3],
                count: 3
            });

            await h.network.disconnectPeer(1);
            // peer 3 writes one more block — peer 1 misses exactly this one block
            await h.transition.advanceState({
                waitForPeers: [0, 3],
                count: 1
            });
            h.contextApi.captureOriginalFork();
            h.event.resetEventSpies();

            h.tamper.stubConstructDispute(3, (d) => {
                if (d.input.stateProof.milestones.length === 0) {
                    throw new Error(
                        "expected at least one milestone in stateProof for this setup"
                    );
                }
                DisputeTampering.tamperInvalidStateProof(d);
            });

            await h.byzantine.submitInvalidStateTransitionBlock(0);

            await h.assert.dispute.initiatedWait({
                peersIndices: [3],
                initiatedWithAuditingData: true
            });
            // peer 0 (full storage) and peer 1 (missing the latest block)
            // both audit via the calldata path; require peer 1 specifically to kill
            // the dispute to confirm pipeline detection with incomplete storage.
            await h.event.waitForPeers("onDisputeKilled", [1], 1, {
                mode: "atLeast"
            });
            await h.assert.storage.honestPeersStoredDisputeFraudProofDetached({
                disputeFraudProofType:
                    DisputeFraudProofType.DisputeInvalidStateProof,
                timeoutMs: 15000
            });
            await h.dispute.resolveDisputeWait();
        });
```

i get the error:

```sh
 Error: Missing latest finalized snapshot for hash 0x143bddfa67da905395b69d56e560d533ac4baf7118eefaa847aa7ce3c292b880
      at AgreementManager.getLatestFinalizedSnapshot (src/agreementManager/AgreementManager.ts:250:23)
      at DisputeValidationService.persistDisputeAuditingDataForPipeline (src/stateManager/DisputeValidationService.ts:787:35)
      at DisputeValidationService.validateDispute (src/stateManager/DisputeValidationService.ts:78:18)
      at async EventHandler.onDisputeCommitted (src/eventHandlers/EventHandler.ts:310:13)
      at async Proxy.<anonymous> (test/fixtures/PeerTestHarness.ts:475:25)
```

@lukastanisic99 
Does something pop to mind? have an idea why?


lukastanisic99:
@daiagi 
image
more specifically
image

peer2 left, peer3 submitted an invalid dispute and peer0 submitted an invalid state transition block, so only peer1 is honest and should transition, so why are 2 expected? 

probably because in `resolveDisputeWait` -> `getPeersForTransitionSyncBarrier` ->`this.context.leftChannelPeerIndices` is never updated as you can see from the screenshot bellow 

image

daiagi:
the two or one expectation is minor and hardly intresting. 

What I'm wondering is why 
`Missing latest finalized snapshot for hash... `

lukastanisic99:
> the two or one expectation is minor and hardly intresting.
> 
> What I'm wondering is why `Missing latest finalized snapshot for hash... `

@daiagi 
The bug is that `persistDisputeAuditingDataForPipeline` calls `agreementManager.getLatestFinalizedSnapshot` which does not get the snapshot from `auditingData` which is available, but tries to get it from `storage` and the peer that left does not have it in storage.

All peers log `✅ Dispute received:` for all disputes, even the one that left. Since the one that left isn't part of the threshold, the milestone doesn't include their signature and they don't have the block/snapshot locally. They got to that point since they could audit to that point (calldata available). 

Apart from the bug that this needs to be extracted from `auditingData` in this case, there is another bug - the peer that's in the process of leaving is auditing (ok we probably want this until the on-chain snapshot is updated and they leave), but they may not be able to audit in full - they can only audit up until the block at which they left, since future blocks are not guaranteed to be made available to them. So they should stop auditing gracefully after their block is final in the stateProof and they should continue updating the snapshot on-chain, but there's some annoying race condition guard which I added to protect people joining (the snapshot update always has to consume all inboundMessages) so if there's an inbound message injected, the snapshot would fail + they wouldn't be able to audit so they are stuck ..... Please add this to trello I need to think about this 

This is annoying and I need to think about it a bit more

daiagi:
> Please add this to Trello

https://trello.com/c/gLYGtJuN


Thanks. I fixed by simple re-order.
https://github.com/peer3to/state-channels-plus/pull/314/changes/f64e1120fd227ec99dfe4520b58cb1aff49bd618

test/e2e/E2E-DisputeValidationLatestStateSnapshotHash.test.ts
```diff
@@ -0,0 +1,264 @@
...
+    expectSignedBlocksOnlyStateProof,
+    PeerTestHarness,
+    TestSession
+} from "@test/harness";
+
+PeerTestHarness.setDefaultLogLevel("error");
+
+describe("E2E: latestStateSnapshotHash", function () {
+    describe("no calldata", function () {
+        describe("empty proof vs hash stub, non-signedBlocks", function () {
```
lukastanisic99: confusing naming - not sure what this should test from the name

test/e2e/E2E-DisputeValidationLatestStateSnapshotHash.test.ts
```diff
@@ -0,0 +1,264 @@
...
+                        disputeFraudProofType:
+                            DisputeFraudProofType.DisputeInvalidStateProof,
+                        timeoutMs: 10000
+                    }
+                );
+                await h.dispute.resolveDisputeWait();
+            });
+        });
+
+        describe("signedBlocks, full storage vs pipeline", function () {
```
lukastanisic99: again confusing - what do you mean `vs`

test/e2e/E2E-DisputeValidationLatestStateSnapshotHash.test.ts
```diff
@@ -0,0 +1,264 @@
...
+                    }
+                );
+                await h.dispute.resolveDisputeWait();
+            });
+        });
+    });
+
+    describe("calldata posted", function () {
+        describe("milestones, full auditors: hash stub | empty proof", function () {
+            it("DisputeInvalidStateProof:  wrong hash, auditors complete (hash stub)", async function () {
```
lukastanisic99: ye really not understanding this naming: what does `auditors complete (hash stub)` mean - I have to take a look at the code to understand

lukastanisic99: Easier for me to deducts what's going on from the code than from these names - it should be the other way around

PR #315, opened 2026-04-18
test suite for various ways of posting junk data in dispute

daiagi:
https://trello.com/c/C7nGc3bS


contracts/V1/StateChannelDiamondProxy/StateChannelCommon.sol
```diff
@@ -555,7 +555,81 @@ contract StateChannelCommon is StateChannelManagerStorage, StateChannelManagerEv
...
+            for (uint256 j = 0; j < bcs.length; j++) {
+                Block memory mb = abi.decode(bcs[j].signedBlock.encodedBlock, (Block));
+                if (mb.transaction.header.channelId != channelId) {
+                    revert ErrorDisputeStateProofHeaderChannelMismatch();
+                }
+            }
+        }
+    }
+
+    function _requireStateProofHeaderForkMatchesInput(Dispute memory dispute) internal pure {
```
lukastanisic99: no reason for this and _requireStateProofHeaderChannelMatchesInput to be separate functions and to run sequentially -  they use the same loops just check different things 

Plus didn't we say we're going to move all of this to auditing and not have it during dispute upload/commitment? 

Dispute upload should mostly check stuff O(1) as race condition guards and everything that can be in auditing should be in auditing

lukastanisic99: Ok I see #318 does it - I'll first take a look at everything to get familiar with the 'absolute' changes before commenting on PRs

PR #322, opened 2026-04-27
Generalize Harness [First effort]

lukastanisic99:
We're testing other state machines (e.g. `poker`), so instead of reinventing everything we did over the last few months with `harness`, we want to generalize it, so other state machines can easily plug and play. 

To a big degree the harness logic was already general (as the SDK is), but some parts (still) are `MathStateMachine` specific and as such can't be reused with other state machines for testing. 

Sadly `all` scenarios build on top of something MathStateMachine specific, so in the first iteration the scenarios aren't generalized. 

Most of the pollution comes from the implicit state transitions `add`/`increment` that are always called somewhere implicitly.


daiagi:
Not a fan of this design

- `MathByzantineActions`
- `MathDisputeOrchastrator`
etc


nothing "Mathy" about them, all these actions are testing SDK behaviors and the identity  of the StateMachine is irreverent
except for the fact that you need to create SOME transaction.



something along these lines seems more sensible to me:


```ts
// MathPeerTestHarness.ts
const encodeMathAdd = (peer: TestPeer) =>
    (peer.contractInstance as MathStateMachine)
        .interface.encodeFunctionData("add", [1]) as Bytes;

this.byzantine = new ByzantineActions(this, this.logger, encodeMathAdd);
```

lukastanisic99:
> Not a fan of this design
> 
> * `MathByzantineActions`
> * `MathDisputeOrchastrator`
>   etc
> 
> nothing "Mathy" about them, all these actions are testing SDK behaviors and the identity of the StateMachine is irreverent except for the fact that you need to create SOME transaction.
> 
> something along these lines seems more sensible to me:
> 
> ```ts
> // MathPeerTestHarness.ts
> const encodeMathAdd = (peer: TestPeer) =>
>     (peer.contractInstance as MathStateMachine)
>         .interface.encodeFunctionData("add", [1]) as Bytes;
> 
> this.byzantine = new ByzantineActions(this, this.logger, encodeMathAdd);
> ```

Not ideal I know, but didn't have a better idea and needed a pure base which I can reuse with poker - nothing that has 'add'/'increment' is usable in the general case, so I just put all the specific stuff under `math*` 

As for this:
`new ByzantineActions(this, this.logger, encodeMathAdd);` I assume you mean `new ByzantineActions(this, this.logger, defaultStateTransitionFunction);` ? 

For math it's easy since `add`/`increment` is always callable, while we don't have such a function in all state machines (e.g. poker). That's the fundamental problem here, since in many places 'add'/'increment' is implicitly called and we can't just substitute it for some `function X`, unless function X also satisfies the same property "can call it always"

PR #324, opened 2026-04-29
add test cases; fix `timoutTooEarly` bug

daiagi:
1. Closes https://trello.com/c/R2YM0yuU

2. From https://trello.com/c/ubJdNdev:

*The dispute pipeline validation pipeline detected TimeoutTooEarly (it wasn’t too early) → auditing failed → tried to kill dispute → rejected on chain (ErrorInvalidFraudProof)*

Fixed the issue (change `<=` into `<`)
and add test to assert, ran it in loop 10 times -> pass every time

before the fix i observed the faulty  behavior intermittently (test sometime failed sometime passed)

```sh
i=0; failures=0; while [ $i -lt 10 ]; do i=$((i+1)); echo "=== run #$i ==="; if yarn test:e2e:log-file --grep 'TimeoutTooEarly NOT raised when timeout dispute is valid' > /tmp/run.out 2>&1; then echo "PASS"; else echo "FAIL"; failures=$((failures+1)); fi; done; echo "==== $failures / $i failures ===="
```


daiagi:
Edit:

- Closes https://trello.com/c/OkSA7EjO

PR #325, opened 2026-05-02
Fixes, custom gas limit for disputes, more harness repackaging

lukastanisic99:
Added custom gas limit for disputes, since the hardcoded one is too low for some state machines.

Repackaged and exposed more harness stuff, so it's reused for testing other state machines.

bug fix: `await` setGenesisState while performing reduction
bug fix: remove `time` from onChainSlashes
bug fix: ignore `timeout` when there's a slash | this is/was a harness bug - there's should be a distinction between `maliciousPeers` and `timedoutPeers`


contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol
```diff
@@ -271,7 +259,7 @@ contract DisputeFraudProofFacet is StateChannelCommon {
             if (!found) return _valid(dispute.input.disputer);
         }
 
-        return _invalid();
+        revert RaceConditionOnChainSlashes();
```
daiagi, line 270: why?
i think this is the only dispute handler that reverts, rather then retutning `_valid` or `_invalid`

scripts/V1/deploy.ts
```diff
@@ -150,6 +150,8 @@ export type TimeConfig = {
     evidenceTime?: number;
 };
 
+export const DEFAULT_DISPUTE_EXECUTION_GAS_LIMIT = 3_000_000;
```
daiagi: not sure this should be hidden away on deploy script
mostly because impoting some constant from `/scripts/deploy` is odd

better placed in some `consts.ts` file

contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol
lukastanisic99: There are some other that also have RaceCondition guards 
e.g. 
image

Mostly to simplify things logically and to be less dependent on time - there was a time race condition bug, where you use your local clock which lags behind, so when you call it in a function of time you get an empty result even though it exists. Always best to not be dependent on time - the flakyness in our tests are related to time

scripts/V1/deploy.ts
lukastanisic99: ye this should be hidden - each application can/should specify its gas limit 

I'll change this locally and push later

PR #327, opened 2026-05-05
More-test-scenarios

daiagi:
- Closes https://trello.com/c/C7nGc3bS
- Closes the gap on two timeout tests that were left as a TODO


daiagi:
- Closes https://trello.com/c/tdMm5OoL

daiagi:
- Partial https://trello.com/c/foDA4Hn3

contracts/V1/StateChannelDiamondProxy/StateChannelCommon.sol
```diff
@@ -234,6 +234,11 @@ contract StateChannelCommon is StateChannelManagerStorage, StateChannelManagerEv
         onlySelf
         returns (MessageBlock memory messageBlock, Balance memory newTotalDeposits)
     {
+        StateSnapshot memory currentSnapshot = getStateSnapshot(channelId);
+        require(
+            !StateChannelManagerInterface(address(this)).isForkDisputed(channelId, currentSnapshot.forkId),
+            RaceConditionForceInboundJoinForkDisputed()
```
lukastanisic99: What's the logic behind this? 

This is general message appending (does not have to be `join`) - only for joining we did these checks and for the general case they are not really relevant + isForkDisputed in the join service is the latest fork that the joiner specifies (does not have to be the snapshot.forkId, since the snapshot may not be updated)

PR #328, opened 2026-05-12
Revive-PingService-test

daiagi:
https://trello.com/c/C63QDuJw


lukastanisic99:
I think I over engineered the custom RPC implementation pattern - I don't like the circular type dependency and I'm not sure how it even works type wise

I think I'll simplify with

`CustomRPC extends MainRPC` and having a generic like now<CustomRpc extends MainRpc> and use it the same way as mainRpc instead of it being hacked like this

Not sure why I didn't do it like this in the first place (maybe there were some problems, not sure)

PR #336, opened 2026-05-21
Concurrency improvements

lukastanisic99:
Commit messages mostly tell what's done - @daiagi take a look and we'll discuss tomorrow how to merge our diverged branches


src/stateManager/BlockDataAvailabilityService.ts
```diff
@@ -0,0 +1,190 @@
...
+                            blockAuthor,
+                            error
+                        }
+                    );
+                })
+                .finally(() => {
+                    this.pendingOnChainBlockValidationKeys.delete(
+                        validationKey
+                    );
+                    this.processedOnChainBlockValidationKeys.add(validationKey);
```
daiagi, line 111: move this to `.then` block

the `.catch` swallows rejection and the `.finally` adds to `processedOnChainBlockValidationKeys` regardless of if `onBlockCalldataPosted` succeeded or not

src/rpc/services/spectate/SpectateService.ts
```diff
@@ -435,92 +435,138 @@ class SpectateService extends ARpcService<SpectateServiceRpcMethods> {
...
         // for all milestones except the last persist all blocks
         for (let i = 0; i < stateProof.milestones.length - 1; i++) {
             for (const blockConfirmation of stateProof.milestones[i]
                 .blockConfirmations) {
-                storage.blocks.storeBlock(
-                    Block.fromBlockConfirmation(blockConfirmation)
-                );
+                const block = Block.fromBlockConfirmation(blockConfirmation);
+                if (this.hasBlockConflict(block)) {
+                    return { shouldAbort: true };
```
daiagi: none atomic
the write that were done before this stay commited in storage
i'd suggest pre-scanning all `hasBlockConflict` checks before doing any writes

so that `persistSyncPayload` will either write all or none, never partial

src/stateManager/BlockQueueManager.ts
```diff
@@ -0,0 +1,418 @@
...
+            await this.stateManager.tryMergeStoredBlockConfirmation(
+                block,
+                strategy,
+                sourcePeers[0]
+            );
+        if (validationResult === undefined) return;
+
+        const shouldKeepConnection =
+            await strategy.interpretFinalValidationResult(validationResult);
+        const shouldDisconnect = !shouldKeepConnection;
```
daiagi: Nit:

uneeded: can use `shouldKeepConnection` directly

src/stateManager/BlockQueueManager.ts
```diff
@@ -0,0 +1,418 @@
...
+        this.timeoutManager.cancelTask(timeout);
+        this.timeoutHandles.delete(blockHash);
+    }
+
+    private requestSyncIfFuture(entry: QueuedBlockEntry): void {
+        const block = entry.block;
+        const nextHeight = this.stateManager.storage.blocks.getNextBlockHeight(
+            block.forkId
+        );
+        if (block.height <= nextHeight) return;
```
daiagi: Nit:
Has guard here, and in the call site (line 204).
One of them is not needed, I'd keep the internal one

src/stateManager/BlockQueueManager.ts
```diff
@@ -0,0 +1,418 @@
...
+        }
+
+        this.queueBlock(block, {
+            senderAddress: options?.senderAddress
+        });
+
+        return true;
+    }
+
+    private queueBlock(
```
daiagi: Nit
Private function with single caller that does 3 operations, only one of them is actually queuing the block.
Consider deleting it and inline the content in `ingestBlockConfirmation` so the side effects are more clear

daiagi (commented):
tryMergeStoredBlockConfirmation runs without the StateManager mutex

Only writer to `storage.blocks` that doesn't hold the mutex. reads `this.status` then awaits `disputeManager.dispute()` - that's an on-chain tx, takes seconds, and status can flip during the await. 
ill send a code suggestion draft that splits it into 3 parts
1. commit under mutex
2.  notify + broadcast (no mutex - block is already persisted)
3. deferred force-join check, detached submission of the dispute


Code proposal in:
https://github.com/peer3to/state-channels-plus/blob/propsal-tryMergeStoredBlockConfirmation/PR-REVIEW-FIX-tryMerge.ts

daiagi:
dead functions: no SDK callers

in `QueueStorage`:

- getQueuedBlock
- getSourcePeers
- isBlockQueued
- tryDequeueAt

src/eventHandlers/EventHandler.ts
```diff
@@ -360,6 +364,93 @@ export class EventHandler {
         this.stateManager.setReductionTimeout(forkId, Number(killPeriodEnd));
     }
 
+    private async updateDisputeView(
```
daiagi: rename: this name implies that we are updating out local view of the dispute, but, as you said it is intended for the front end

lukastanisic99:
> tryMergeStoredBlockConfirmation runs without the StateManager mutex
> 
> Only writer to `storage.blocks` that doesn't hold the mutex. reads `this.status` then awaits `disputeManager.dispute()` - that's an on-chain tx, takes seconds, and status can flip during the await. ill send a code suggestion draft that splits it into 3 parts
> 
> 1. commit under mutex
> 2. notify + broadcast (no mutex - block is already persisted)
> 3. deferred force-join check, detached submission of the dispute
> 
> Code proposal in: https://github.com/peer3to/state-channels-plus/blob/propsal-tryMergeStoredBlockConfirmation/PR-REVIEW-FIX-tryMerge.ts

image
Deleting this code from the function, it's irrelevant. Collecting signatures on an already accepted block shouldn't affect your perception on force join

src/stateManager/BlockDataAvailabilityService.ts
lukastanisic99: Double checked it looks good to me - I also wrapped ingestBlockConfirmation in a try/catch so it should never throw and returns false under failure. I don't think it can throw even now, since it can only throw if the block is invalid and we check the block in the EVM  `isBlockConfirmationAuthentic` and if it's invalid it throws under the low level `staticcall` which is returned as `false`, so the above catch should never run

lukastanisic99:
> updateDisputeView

kept the last 2 since they're used in tests - maybe they're outdated, but didn't think too much

PR #339, opened 2026-05-26
Dedicated (E)VM Thread

lukastanisic99:
- Uniform deployment through a `signer`;
- Option to run a dedicated thread(worker) for the (E)VM execution. It's fully polymorphic. Done in a way where we can easily swap out the contract executor for `a node implementation` later when we do `light clients` (bigger project/task);
- The SDK builds both `ESM` and `CommonJS`, and browser/node dependencies are separated at build. Separation should be clear now; 
- 20-30 tests started failing, since event loop starvation is reduced and race condition is easier to achieve;
- Dedicated flag for the logger(tests) to throw if event loop scheduling delay exceeds expectation.


src/evm/contractExecutor/node/ContractExecutorWorkerRuntime.ts
```diff
@@ -0,0 +1,46 @@
...
+    const nodeRequire = typeof require === "function" ? require : undefined;
+    if (!nodeRequire) {
+        throw new Error("Node worker_threads require() is unavailable");
+    }
+
+    const path = nodeRequire("node:path") as typeof import("node:path");
+    const fs = nodeRequire("node:fs") as typeof import("node:fs");
+    const { Worker } = nodeRequire(
+        "node:worker_threads"
+    ) as typeof import("node:worker_threads");
```
daiagi: overly defensive, can simply be
```ts
import * as path from "node:path";
import * as fs from "node:fs";
import { Worker } from "node:worker_threads";
```

as the tsconfig's path `@platform/contractExecutorWorkerRuntime` selects this file for only node env anyways

src/evm/contractExecutor/ContractExecutorFactory.ts
```diff
@@ -0,0 +1,33 @@
...
+import noOpLogger from "./NoOpLogger";
+
+export type ContractExecutorFactoryOptions = {
+    logger?: Logger;
+    dedicatedThread: boolean;
+    customPrecompiles?: EvmCustomPrecompileManifest[];
+};
+
+export type ContractExecutorFactory = AContractExecutor;
+
```
daiagi: why need this type aliasing?

src/evm/contractExecutor/ContractExecutorWorkerHostCore.ts
```diff
@@ -0,0 +1,105 @@
...
+    global?: typeof globalThis;
+    window?: typeof globalThis;
+};
+
+workerGlobal.Buffer ||= Buffer;
+workerGlobal.global ||= globalThis;
+workerGlobal.window ||= globalThis;
+
+let evm: Awaited<ReturnType<typeof createEvm>> | undefined;
+let executor: ContractExecutor | undefined;
```
daiagi:
```suggestion
let evm: EVM | undefined;

```

src/evm/contractExecutor/ContractExecutorWorkerHostCore.ts
```diff
@@ -0,0 +1,105 @@
...
+    Buffer?: typeof Buffer;
+    global?: typeof globalThis;
+    window?: typeof globalThis;
+};
+
+workerGlobal.Buffer ||= Buffer;
+workerGlobal.global ||= globalThis;
+workerGlobal.window ||= globalThis;
+
+let evm: Awaited<ReturnType<typeof createEvm>> | undefined;
```
daiagi: not needed. it is only used once, in init, and then passed to `new ContractExecuter`
no need to hold "module level" variable

src/evm/contractExecutor/ContractExecutorWorkerHostCore.ts
```diff
@@ -0,0 +1,105 @@
...
+                })
+            )
+        },
+        noOpLogger
+    );
+    executor = new ContractExecutor(evm, noOpLogger);
+    return null;
+}
+
+function getExecutor(): ContractExecutor {
```
daiagi: this reads like class in functions.  executer is a class field

src/utils/moduleLoader/node/importModuleFromManifest.ts
```diff
@@ -0,0 +1,53 @@
+export async function importPrecompileModule(
+    moduleSpecifier: string
+): Promise<any> {
+    if (typeof require === "function") {
```
daiagi, line 4: will it ever not be a function?

src/evm/EvmFactory.ts
```diff
@@ -27,11 +60,70 @@ export async function createEvm(
...
+}
+
+export function normalizePrecompileAddress(address: Address | string): Address {
+    if (typeof address === "string") {
+        return Address.fromString(address.toLowerCase());
+    }
+
+    const stringValue = address.toString();
+    return Address.fromString(stringValue.toLowerCase());
+}
```
daiagi:
```suggestion
export function normalizePrecompileAddress(address: Address | string): Address {
    return Address.fromString(address.toString().toLowerCase());
}

```

is duplicate of toEvmAddress in ContractExecuter

belongs better is utils/address.ts

src/evm/EvmFactory.ts
```diff
@@ -1,23 +1,56 @@
...
+    EVMOpts["customPrecompiles"]
 >[number];
 
+export type EvmCustomPrecompile =
+    | EvmNativeCustomPrecompile
+    | EvmCustomPrecompileManifest;
+
+export interface EvmFactoryOptions extends Omit<EVMOpts, "customPrecompiles"> {
+    customPrecompiles?: EvmCustomPrecompile[];
+}
```
daiagi, line 41: 🤮 types

src/evm/EvmFactory.ts
```diff
@@ -27,11 +60,70 @@ export async function createEvm(
...
+    if (typeof address === "string") {
+        return Address.fromString(address.toLowerCase());
+    }
+
+    const stringValue = address.toString();
+    return Address.fromString(stringValue.toLowerCase());
+}
+
+async function resolveCustomPrecompiles(
+    customPrecompiles: EvmFactoryOptions["customPrecompiles"],
```
daiagi:
```suggestion
    customPrecompiles:  EvmCustomPrecompile[] | undefined
```

src/utils/browser/DeployUtils.ts
```diff
@@ -0,0 +1,19 @@
+import type { ContractFactory } from "ethers";
+
+type ReturnTypeOfT<T extends ContractFactory> = T extends {
+    deploy(...args: any): infer U;
+}
+    ? U
+    : never;
+
```
daiagi: not needed

can simply use `ReturnType<T["deploy"]`

src/utils/node/DeployUtils.ts
```diff
@@ -0,0 +1,46 @@
+import fs from "fs";
+import path from "path";
+import { ContractFactory } from "ethers";
+
+type ReturnTypeOfT<T extends ContractFactory> = T extends {
+    deploy(...args: any): infer U;
+}
+    ? U
+    : never;
```
daiagi: duplicate, and not needed

src/utils/Mutex.ts
```diff
@@ -19,7 +18,7 @@ export class Mutex {
     }
 
     public lock(options?: MutexLockOptions): Promise<void> {
-        const queuedAt = Clock.getTimeInSeconds();
+        const queuedAt = Math.floor(Date.now() / 1000);
```
daiagi, line 21: why change?

lukastanisic99: Since it was used in places where Clock is not instantiated, and like this it's not depended on it - we still have all the information since it's relative

PR #344, opened 2026-06-03
Thread-partitioned logging: SDK ↔ EVM worker crash-log uploads

saumyakaran:
Logger now broadcasts its messages to other threads through a generic `GossipNode` which just forwards to neighbours knowing nothing about logs.

Added 2 generic worker primitives: `WorkerClient` (spawner / main side) and `AWorkerHost` (spawnee / worker side) and a platform-specific `workerTransport` that wires them together.

The EVM `ContractExecutorWorkerHost` now extends the generic `AWorkerHost` which gives it gossip capabilities so the EVM worker thread's logs reach the main thread on report-bug instead of being lost.

> _**NOTE**_: Incremental uploads are addressed in the follow up PR #345.

### DIFF
--- a/src/evm/contractExecutor/ContractExecutorWorkerHostCore.ts
+++ b/src/evm/contractExecutor/ContractExecutorWorkerHostCore.ts
@@ -56,7 +56,7 @@ export class ContractExecutorWorkerHost extends AWorkerHost<
               )
             : noOpLogger;
         // Only wire gossip with a real logger; the no-op sink would swallow the ops.
-        if (loggerConfig) this.attachLogger(this.workerLogger);
+        if (loggerConfig) this.workerLogger.setGossipNode(this.gossipNode);
 
         this.evm = await createEvm(
             {
--- a/src/evm/contractExecutor/ContractExecutorWorkerHostCore.ts
+++ b/src/evm/contractExecutor/ContractExecutorWorkerHostCore.ts
@@ -2,6 +2,8 @@ import { Buffer } from "buffer";
 import ContractExecutor from "./ContractExecutor";
 import { createEvm } from "../EvmFactory";
 import noOpLogger from "./NoOpLogger";
+import { createLogger } from "@platform/createLogger";
+import type { Logger, LogLevel } from "@/utils/logging/Logger";
 import type {
     WorkerRequest,
     WorkerRequestPayload,
@@ -20,8 +22,28 @@ workerGlobal.window ||= globalThis;
 
 let evm: Awaited<ReturnType<typeof createEvm>> | undefined;
 let executor: ContractExecutor | undefined;
+let workerLogger: Logger = noOpLogger;
 
 async function init(request: Extract<WorkerRequestPayload, { type: "init" }>) {
+    const loggerConfig = request.loggerConfig;
+    // The worker's own config singleton is DEFAULT_CONFIG (createConfig only
+    // runs on the main thread), so logger config must be injected explicitly.
+    workerLogger = loggerConfig
+        ? createLogger(
+              loggerConfig.sharedContext,
+              { component: "ContractExecutorWorker" },
+              {
+                  logUploaderConfig: {
+                      uploadEndpoint: loggerConfig.uploadEndpoint,
+                      apiToken: loggerConfig.apiToken
+                  },
+                  level: loggerConfig.level as LogLevel | undefined,
+                  skipWriting: loggerConfig.skipWriting,
+                  attachErrorListener: true
+              }
+          )
+        : noOpLogger;
+
     evm = await createEvm(
         {
             allowUnlimitedContractSize: true,
@@ -34,9 +56,9 @@ async function init(request: Extract<WorkerRequestPayload, { type: "init" }>) {
                 })
             )
         },
-        noOpLogger
+        workerLogger
     );
-    executor = new ContractExecutor(evm, noOpLogger);
+    executor = new ContractExecutor(evm, workerLogger);
     return null;
 }
 
--- a/src/evm/contractExecutor/ContractExecutorWorkerHostCore.ts
+++ b/src/evm/contractExecutor/ContractExecutorWorkerHostCore.ts
@@ -63,6 +63,19 @@ async function init(request: Extract<WorkerRequestPayload, { type: "init" }>) {
     return null;
 }
 
+function diagnostics(
+    request: Extract<WorkerRequestPayload, { type: "diagnostics" }>
+): void {
+    if (request.op === "updateContext") {
+        workerLogger.updateSharedContext(request.context);
+    } else {
+        // Fire-and-forget: ack immediately, let the upload run detached.
+        void workerLogger.uploadLogs(
+            request.message ?? "worker report-bug flush"
+        );
+    }
+}
+
 function getExecutor(): ContractExecutor {
     if (!executor) {
         throw new Error("Contract executor worker has not been initialized");
@@ -96,9 +109,8 @@ async function handleRequest(request: WorkerRequest): Promise<WorkerResponse> {
             case "init":
                 result = await init(workerRequestPayload);
                 break;
-            case "uploadLogs":
-                // Fire-and-forget: ack immediately, let the upload run detached.
-                void workerLogger.uploadLogs("worker report-bug flush");
+            case "diagnostics":
+                diagnostics(workerRequestPayload);
                 result = null;
                 break;
             default:
--- a/src/evm/contractExecutor/ContractExecutorWorkerHostCore.ts
+++ b/src/evm/contractExecutor/ContractExecutorWorkerHostCore.ts
@@ -5,145 +5,93 @@ import { createEvm } from "../EvmFactory";
 import noOpLogger from "./NoOpLogger";
 import { createLogger } from "@platform/createLogger";
 import type { Logger, LogLevel } from "@/utils/logging/Logger";
-import type {
-    WorkerRequest,
-    WorkerRequestPayload,
-    WorkerResponse
-} from "./types";
+import { AWorkerHost } from "@/utils/worker/AWorkerHost";
+import type { WorkerRequestPayload } from "./types";
 
 const workerGlobal = globalThis as unknown as {
     Buffer?: typeof Buffer;
     global?: typeof globalThis;
     window?: typeof globalThis;
 };
-
 workerGlobal.Buffer ||= Buffer;
 workerGlobal.global ||= globalThis;
 workerGlobal.window ||= globalThis;
 
-let evm: Awaited<ReturnType<typeof createEvm>> | undefined;
-let executor: ContractExecutor | undefined;
-let workerLogger: Logger = noOpLogger;
+export class ContractExecutorWorkerHost extends AWorkerHost<
+    WorkerRequestPayload,
+    ContractExecutionResult | null
+> {
+    private evm?: Awaited<ReturnType<typeof createEvm>>;
+    private executor?: ContractExecutor;
+    private workerLogger: Logger = noOpLogger;
 
-async function init(request: Extract<WorkerRequestPayload, { type: "init" }>) {
-    const loggerConfig = request.loggerConfig;
-    // The worker's own config singleton is DEFAULT_CONFIG (createConfig only
-    // runs on the main thread), so logger config must be injected explicitly.
-    workerLogger = loggerConfig
-        ? createLogger(
-              loggerConfig.sharedContext,
-              { component: "ContractExecutorWorker" },
-              {
-                  logUploaderConfig: {
-                      uploadEndpoint: loggerConfig.uploadEndpoint,
-                      apiToken: loggerConfig.apiToken
-                  },
-                  level: loggerConfig.level as LogLevel | undefined,
-                  skipWriting: loggerConfig.skipWriting,
-                  attachErrorListener: true
-              }
-          )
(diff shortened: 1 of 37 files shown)


src/utils/logging/Logger.ts
```diff
@@ -144,8 +207,13 @@ export abstract class Logger {
     }
     public error(message: any, ...meta: any[]): void {
         this.log("error", message, meta);
-        const prommise = this.logUploader?.uploadLogs();
-        if (prommise) DetachedPromises.collect(prommise);
+        // Flush only this thread's own store — NO cross-thread gossip (error() has
```
saumyakaran, line 209: @lukastanisic99 — needs your sign-off on this design decision.

`error()` now flushes **only this thread's own store** — no cross-thread gossip. Why: there are 60+ `error()` call sites (incl. socket/validation hot paths), and the uploader never clears the store between uploads, so fanning out would re-POST every thread's full (MB-scale) buffer on **every** error. Each thread already auto-uploads its own store on its own errors, so the cross-thread fan-out was pure duplication here.

I kept the cascade on the explicit **report-bug** path (`uploadLogs`) only — so report-bug still pulls the EVM worker's logs even when it's stuck-but-not-crashed.

Does this split work, or did you actually want an all-threads flush on every `error()`?

lukastanisic99: Have the logger keep track of what it uploaded so it just uploads the difference, the server code should also account for it. The compression algorithm might lose some efficiency when data is split into chunks, but maybe there's a smart and SIMPLE way to do it so it does not - figue it out, but in general we want all the logs so we have the "entire picture", but yes no need to re-upload the same stuff

lukastanisic99:
> **Thread-partitioned crash-log uploads.** Each thread uploads its _own_ log store, partitioned `channel → peer → thread`, so the EVM worker thread is no longer a logging black hole — `report-bug` now collects the SDK thread _and_ the EVM worker, tagged so they can't clobber each other.
> 
> Base branch is `webrtc-worker` (this stacks on the WebRTC-in-worker PR).
> 
> ## Architecture
> Three generic, reusable building blocks (used internally here; structured to be exported for the front-end when needed):
> 
> * **`GossipNode`** (`src/utils/GossipNode.ts`) — a generic thread-gossip tree node. Skip-sender forwarding; loop-free with no ids/dedupe **on a tree** (documented precondition). Knows nothing about logs.
> * **`WorkerClient` / `AWorkerHost`** (`src/utils/worker/`) — reusable main-/worker-side base for any worker runtime: RPC envelope framing, ready handshake, requestId correlation, a `GossipNode`, logger attach, dispose. Inheritance-style — a runtime subclasses `AWorkerHost` and implements `handle` (mirrors the custom-RPC pattern).
> * **`@platform` worker transports** (`src/utils/{node,browser}/workerTransport.ts`) — spawn a worker with a dedicated gossip `MessageChannel` (worker end transferred via `workerData` on node, a sentinel init message on browser).
> 
> On top of those, the **Logger** gains a small cross-thread vocabulary: `LoggerOp` (`flush` | `updateContext`), `applyOp` (the single interpreter applied to _this_ thread's logger), and `setGossipNode`. A worker rebuilds an equivalent logger from `toSerializableConfig()` (a live `Logger` can't cross the structured-clone boundary).
> 
> ## Gossip semantics
> * **report-bug (`uploadLogs`)** fans a `flush` out to **all** threads (bidirectional cascade) — so a single report-bug captures the EVM worker's logs _even when it's stuck but hasn't crashed_.
> * **`error()`** flushes **only its own thread's store** — deliberately no cross-thread fan-out (see open question below).
> * **`updateContext`** propagates `channelId` (known only after the channel opens) to the worker, so its uploads tag the right channel instead of `ZeroHash`.
> 
> ## EVM worker cutover
> The contract-executor worker moved onto `WorkerClient`/`AWorkerHost`; its old `NoOpLogger` is replaced by a real logger built from the injected recipe, so the worker collects and uploads its own store. Precompile manifest, EVM-revert `.data`, and request/response semantics are preserved.
> 
> ## Why `error()` doesn't fan out — needs a call
> `error()` has 60+ call sites (incl. socket/validation hot paths) and the uploader never clears the store between uploads, so fanning out on every `error()` would re-POST every thread's full (MB-scale) buffer on every error. Each thread already auto-uploads its own store on its own errors, so the fan-out was pure duplication there. The cascade is kept only on the explicit report-bug path. **@luka — does this split work, or did you want an all-threads flush on every `error()`?**
> 
> ## Tests
> * `GossipNode` — skip-sender, **3-node tree loop-freedom** (middle/leaf origin, exactly-once delivery), `addNeighbour` idempotency.
> * `WorkerBase` — RPC round-trip, error propagation, **bidirectional gossip incl. worker→main cascade through real Loggers**.
> * `WorkerLoggerUpload` — spawns a real EVM worker, faults it (INVALID opcode), and asserts its **own** decoded store crosses the gossip edge tagged `evm`, carrying the late-arriving `channelId`.
> * `LoggerApplyOp`, `LogUploaderEnvelope`, `Config`.
> 
> **Suite: 272 passing.** The 1 failing test — `WebRTCTransport › rejects a data channel that is not open` — is **pre-existing on `webrtc-worker`** (fails identically with this branch's changes stashed); it's WebRTC work, unrelated to logging.
> 
> ## Open items / deliberate decisions
> * `error()` fan-out dropped pending the call above.
> * Reusable worker base + `GossipNode` are **not re-exported from the package root yet** — a trivial one-liner when the front-end wants to build threads on them.
> * `scripts/logging/crash-log-server.js` (local test server) diverges from the real server — left uncommitted, not part of this PR.

@saumyakaran no one is reading these long AI generated PR descriptions. 
Explain stuff simply in your own words and if you really want to use these, put at the bottom "AI SUMMARY" and bellow short AI description, but the important thing is for you to be able to simply explain your idea behind the PR - if you can't you need to go back and understand it better before posting it 

Take a look at this:
https://skeptics.stackexchange.com/questions/8742/did-einstein-say-if-you-cant-explain-it-simply-you-dont-understand-it-well-en

lukastanisic99:
@saumyakaran didn't take a look yet, but what would also be nice if we can extract device information (what's available in browsers/nodejs - see how we do polymorphism regarding that) and have that uploaded too as some header or log it once at the top.

Everything that we can read about the hardware/OS/browser/node would be useful

daiagi:
I'm not sure if `GossipNode` is needed at all. 
As far as i can see it can be deleted completely.
why?
it is infra for gossip tree /  gossip network. but we don't have that.
we have just single bidirectional connection `sdk ↔ evm`.


Am I missing something?

saumyakaran:
> it is infra for gossip tree / gossip network. but we don't have that.
we have just single bidirectional connection sdk ↔ evm

that was the initial implementation, but @lukastanisic99 had mentioned the requirement for this to be generic to account for future multiple worker threads (like parallel provers), hence the decision to create the gossip tree

src/evm/contractExecutor/ContractExecutorWorkerHostCore.ts
```diff
@@ -1,105 +1,98 @@
...
-        noOpLogger
-    );
-    executor = new ContractExecutor(evm, noOpLogger);
-    return null;
-}
+export class ContractExecutorWorkerHost extends AWorkerHost<
+    WorkerRequestPayload,
+    ContractExecutionResult | null
+> {
+    private evm?: Awaited<ReturnType<typeof createEvm>>;
```
daiagi, line 24: type is simply EVM

please guide you AI agent to not use these 🤮 types

src/evm/contractExecutor/ContractExecutorWorkerHostCore.ts
```diff
@@ -1,105 +1,98 @@
...
-        throw new Error("Contract executor worker has not been initialized");
+    protected async handle(
+        payload: WorkerRequestPayload
+    ): Promise<ContractExecutionResult | null> {
+        switch (payload.type) {
+            case "init":
+                return this.init(payload);
+            default:
+                return this.call(payload);
+        }
```
daiagi, line 36: style choice i guess, but switch for two cases is overkill
simple if else or ternary is better fit

src/evm/contractExecutor/ContractExecutorWorkerHostCore.ts
```diff
@@ -1,105 +1,98 @@
...
 
-async function handleRequest(request: WorkerRequest): Promise<WorkerResponse> {
-    const { requestId, workerRequestPayload } = request;
-    try {
-        const result =
-            workerRequestPayload.type === "init"
-                ? await init(workerRequestPayload)
-                : await call(workerRequestPayload);
+    private async init(
+        request: Extract<WorkerRequestPayload, { type: "init" }>
```
daiagi, line 40: maybe better use enum for the type, instead of "free" strings

daiagi: also, would not need Extract type if the WorkerRequestPayload would be composed of names types variants

src/utils/browser/workerTransport.ts
```diff
@@ -0,0 +1,77 @@
+import type { Neighbour } from "@/utils/GossipNode";
+import type {
+    WorkerClientTransport,
+    WorkerHostTransport,
+    WorkerResult
+} from "@/utils/worker/types";
+
+export type GossipInitMessage = { __gossipInit: true; port: MessagePort };
```
daiagi, line 8: uses only in single place in this file. not clear why it need named type. but even so, no need to export it

src/utils/logging/node/createLogger.ts
```diff
@@ -27,6 +25,11 @@ export const createLogger = (
             apiToken: config.CRASH_LOG_API_TOKEN || ""
         } as LogUploaderConfig);
 
+    // Store-enable follows the EFFECTIVE endpoint (injected recipe or global),
+    // so the worker — whose global config is DEFAULT_CONFIG — still collects logs.
```
daiagi, line 29: i dont get what this comment is trying to tell me

src/utils/logging/Logger.ts
```diff
@@ -75,8 +108,35 @@ export abstract class Logger {
...
+        while (logger) {
+            if (logger.node) return logger.node;
+            logger = logger.parent;
+        }
+        return undefined;
+    }
+
+    // Single interpreter for a LoggerOp on this thread; returns the flush promise to await/detach.
+    public applyOp(op: LoggerOp): Promise<void> | undefined {
+        switch (op.type) {
```
daiagi, line 133: same comment:
two cases don't need switch.  I think enum is better choice for the string types

src/utils/logging/Logger.ts
```diff
@@ -75,8 +108,35 @@ export abstract class Logger {
...
+        let logger: Logger | undefined = this;
+        while (logger) {
+            if (logger.node) return logger.node;
+            logger = logger.parent;
+        }
+        return undefined;
+    }
+
+    // Single interpreter for a LoggerOp on this thread; returns the flush promise to await/detach.
+    public applyOp(op: LoggerOp): Promise<void> | undefined {
```
daiagi, line 132: Promise<void> | undefined is a strnge return type
which is the happy path?

src/utils/logging/Logger.ts
```diff
@@ -144,8 +206,11 @@ export abstract class Logger {
     }
     public error(message: any, ...meta: any[]): void {
         this.log("error", message, meta);
-        const prommise = this.logUploader?.uploadLogs();
-        if (prommise) DetachedPromises.collect(prommise);
+        // Flush only this thread's own store — NO cross-thread gossip (error() has
+        // 60+ call sites; fanning out would re-upload every thread on every error).
+        // The report-bug path (uploadLogs) is what pulls all threads.
```
daiagi, line 211: i really dont like to see these "AI journy" comments in the code
comments should explain non trivial stuff, they should answer why
they should not log the journy you went through to understand why this code and not that code

src/utils/logging/Logger.ts
```diff
@@ -156,10 +221,22 @@ export abstract class Logger {
     }
 
     public async uploadLogs(message: any, ...meta: any[]): Promise<void> {
-        await LoggerUtils.logTimestamp(this);
+        // logTimestamp needs a live Clock; the worker has none — don't abort the flush.
```
daiagi, line 224: same "don't abort the flush"

ok i won't???

src/utils/logging/Logger.ts
```diff
@@ -156,10 +221,22 @@ export abstract class Logger {
...
+            /* no Clock on this thread — skip the timestamp diagnostic */
+        }
         const localTime = new Date().getTime() / 1000;
         this.warn(message, ...meta, localTime);
-        await this.logUploader?.uploadLogs();
+        await this.originate({ type: "flush" });
+    }
+
+    // Apply an op locally AND gossip it to the rest of the tree; returns the flush promise.
+    private originate(op: LoggerOp): Promise<void> | undefined {
```
daiagi, line 236: comment says return the flush promise
code return cryptic Promise<void> | undefined

daiagi (changes requested):
apologies for taking long to review. i should have done it sooner

saumyakaran:
@daiagi sorry for the delay in response, i had put this on the back burner, prioritizing the frontend changes for the last week

Luka had mentioned that #346 introduces the primitives needed for this so we can use that instead of me recreating it here, but your comments all make sense to me

perhaps i could port the changes once that's reviewed and merged

PR #343, opened 2026-06-02
[Test Harness] - peers run inside their own worker thread

daiagi:
@lukastanisic99  it is a big PR, please take a moment to read this, before you dive into review

## what
second harness mode: each test peer can run in its own `worker_thread`
(`HARNESS_DEDICATED_PEER_THREAD=true`), alongside the existing inline
(in-process) mode. same tests, same assertion API, both modes.

## why it's +7.8k / −2.1k
| +lines | files | area | what |
|---|---|---|---|
| 2712 | 21 | `test/harness/threaded` | RPC transport + worker spawn (new) |
| 2546 | 48 | `test/harness/core` | polymorphic `PeerHandle`: 1 interface (528) × 2 impls — inline 1091, worker 927 |
| 929 | 22 | `test/harness/actions` | actions call the handle, not `getPeer().stateManager` |
| 374 | 3 | `test/fixtures` | inline-vs-worker mode switch (`PeerTestHarness`) |
| 457 | 23 | `test/e2e` | tests adapted |
| **27** | **3** | **`src/` (PRODUCTION)** | **LocalDiscoveryServer multi-isolate + 2-line P2PManager — whole prod surface** |

## topology (worker mode)

image

## RPC (one postMessage channel, symmetric `req`/`res`/`push`; both ends are caller+handler)
4 shapes:

1. **read** — orch → worker req/res, query peer state
   `handle.blocks.queryBlockAt({ forkId, height: 4 })`
   → `req "query.blockAt" { forkId, height: 4 }` → `res { hash, height, author }`

2. **action** — orch → worker req/res, fixed route menu, side-effecting
   `h.transition.increment(1)`
   → `req "math.add" { value: 1 }` → worker runs `contract.add(1)` (applies the transition, produces the block)

3. **reverse-rpc (stubs)** — worker → orch req/res; stubbed code runs in the worker, the test closure lives in the orch
   worker's patched `constructDispute(forkId)` calls back:
   `rpcClient.call("harness.tamperDispute", { peerIndex, dispute, disputeConfirmation, auditingData })`
   → orch runs the test's tamper fn, returns the mutated `{ dispute, disputeConfirmation, auditingData }`

4. **push/events** — worker → orch, one-way, no response
   `peerHandler.push("spy", { name: "onDisputeCommitted", count: 3, lastArgs })`
   → orch `SpyMirror` → `eventCountsBarrier.signal()`   (also `"fork.changed" { forkId }`, `"lifecycle.ready"`)

## review focus
[...]

PR #346, opened 2026-06-16
Threaded SDK, request/response pattern for RPCs, harness updated

lukastanisic99:
Originally wanted to do 3 PRs, but since threading the SDK fundamentally changes how our test framework works needed to do all together in order to push something that’s tested.

The SDK can now run in `threaded mode` by simply setting a boolean during p2pSetup (`RUN_SDK_IN_THREAD
=true`). Since nodejs/browsers (V8) doesn’t really have threads with shared memory (shared array buffer is useless), but process abstractions (workers) that have to communicate over message passing, I’ve changed the architecture to reflect that in a uniform way where we do not have 1 implementation for inline(reference) and 1 implementation fork worker message passing - which would lead to bugs across both implementations. 

The core idea:
- The same way an `ethers contract instance` allows a client to simply interact with a remote RPC node by having the entire communication serializable going though the network and `not caring` where the RPC node is located, we did the same. 
`p2pSetup` basically returns a remote control, that sends messages and can subscribe to events over a `MessageChannel` transport not caring where the recipient is located (same thread, different process, different device). The MessageChannel port is `transferable` across V8 workers, meaning the only thing we need to do to `run in threaded mode` is to transfer 1 MessageChannel port to the worker which would run the same protocol as if it was running on the same thread or different device. All the communication is serializable and goes over the MessageChannel. 

Also added a type safe `request/response pattern` for RPCs so we can simply write:
`sum = await remoteRpc.mathService.sum(5,3).request(peerAddress)` which would invoke the operation on the remote peer with the result returned to us. Before this we only had a `fire and forget` communication protocol. Also if the recipient is omitted (`request()`) it invokes the operation on 'our peer' (useful for tests). 

The 2 above changes allow us to rewrite parts of our test framework into `customRPCs` that run on the peer and have the test framework control them easily over the `remote control` not caring is the peer in the same memory process or not. So our test framework is an orchestrator that can be used to orchestrate even remote peers out of the box. 

That's most of the important parts, for everything else AI can explain parts easily. 

Currently about 30 tests are failing which were `false positives` before and will be addressed on a case by case basis in a future PR.

Run tests with `yarn test:e2e:worker` - takes about 30min - Did not update the parallel script, so we can run tests in parallel - also needed in one of the future PRs. 

EDIT before I forget: 
The only thing that kinda breaks our abstraction of an isolated environment where the code runs are `SINGLETONS` which have to be handled on a case by case basis (think I did all of them) - mostly since on the same thread the same singleton lives on `BOTH sides of the MessageChannel port` while in worker and remote peers it's obviously isolated


src/evm/p2pRuntime/P2pRuntimeHost.ts
```diff
@@ -0,0 +1,363 @@
...
+        const p2pContractInstance = stateMachineContract.connect(
+            stateManager.p2pManager.p2pSigner
+        ) as unknown as AStateMachineContract;
+        evmDiamondStateMachine.setP2pContractInstance(p2pContractInstance);
+
+        evmDiamondStateMachine.setContractEventEmitter((name, args) =>
+            port.post({ type: "contractEvent", name, args })
+        );
+
+        forwardEventHandlerInvocations(stateManager, port);
```
daiagi: I think you could simplify a bit and pass just eventHandler as an arg, and not the entire StateManager

src/evm/p2pRuntime/node/P2pRuntimeWorkerRuntime.ts
```diff
@@ -0,0 +1,82 @@
...
+    RuntimePort,
+    WorkerBootstrapMessage
+} from "../types";
+import { adaptPort } from "./P2pRuntimeChannel";
+
+export function createP2pRuntimeWorker(): P2pRuntimeWorker {
+    const nodeRequire = typeof require === "function" ? require : undefined;
+    if (!nodeRequire) {
+        throw new Error("Node worker_threads require() is unavailable");
+    }
```
daiagi: this guard is not needed and will never fire in practice. can simply import normally at the top of the file
aka as parentPort is imported

src/evm/p2pRuntime/P2pRuntimeClient.ts
```diff
@@ -0,0 +1,238 @@
...
+        ) as T;
+
+        this.port.onMessage((message) =>
+            this.handleMessage(message as RuntimeHostMessage)
+        );
+        this.port.start();
+    }
+
+    /** Send a request to the host and await its correlated response. */
+    request<TResult>(request: RuntimeRequestInput): Promise<TResult> {
```
daiagi: consider addding a timeout here, like  P2Pmanager line 137 does.

src/evm/p2pRuntime/types.ts
```diff
@@ -0,0 +1,125 @@
...
+
+/**
+ * Transport definitions for the p2p runtime channel.
+ */
+
+/**
+ * A minimal transport surface implemented by both Node `worker_threads`
+ * `MessagePort` and the browser `MessagePort`.
+ */
+export interface RuntimePort {
```
daiagi, line 14: I think it is missing an `onClose` callback, to do client side cleanup if the worker thread dies

src/evm/p2pRuntime/P2pRuntimeClient.ts
```diff
@@ -0,0 +1,238 @@
...
+
+    /** Tear down the runtime: dispose the host, reject pending work, close. */
+    async dispose(): Promise<void> {
+        if (this.disposed) return;
+        try {
+            await this.request<void>({ type: "dispose" });
+        } catch {
+            // The host may already be gone; proceed with local teardown.
+        }
+        this.disposed = true;
```
daiagi, line 193: while waiting for the request above, new requests can still come in and see disposed = false.
move `this.disposed = true` line to the top of the  function

src/evm/p2pRuntime/host/EventForwarding.ts
```diff
@@ -0,0 +1,60 @@
...
+import type P2pEventHooks from "@/P2pEventHooks";
+import { EVENT_HANDLER_HOOK_NAMES } from "@/eventHandlers/EventHandlerHooks";
+import type { RuntimePort } from "../types";
+
+/** Best-effort structured-clone of handler args; `[]` if not cloneable. */
+function safeEventArgs(args: unknown[]): unknown[] {
+    try {
+        return structuredClone(args);
+    } catch {
+        return [];
```
daiagi: are there any cases the args are no clonable?
if yes, should it throw? or at least log an warning?

src/P2PManager.ts
```diff
@@ -89,8 +112,108 @@ class P2PManager<TCustomRpc extends MainRpcService = MainRpcService>
...
+        rpc: Rpc,
+        transport: ATransport,
+        options?: { timeoutMs?: number }
+    ): Promise<T> {
+        const requestId = `${++this.rpcRequestCounter}`;
+        // `agreementTime` is in seconds; the RPC timeout is in milliseconds.
+        const agreementTimeoutMs =
+            this.stateManager.timeConfig.agreementTime != null
+                ? this.stateManager.timeConfig.agreementTime * 1000
+                : undefined;
```
daiagi: uselss guard. 
`this.stateManager.timeConfig.agreementTime` is always defined

src/P2PManager.ts
```diff
@@ -89,8 +112,108 @@ class P2PManager<TCustomRpc extends MainRpcService = MainRpcService>
...
+        const timeoutMs =
+            options?.timeoutMs ??
+            agreementTimeoutMs ??
+            P2PManager.DEFAULT_RPC_REQUEST_TIMEOUT_MS;
```
daiagi: I don't think the 3rd option will ever be selected

src/utils/moduleLoader/node/importModuleFromManifest.ts
```diff
@@ -1,4 +1,4 @@
-export async function importPrecompileModule(
+export async function importModuleFromManifest(
```
daiagi, line 1: you dropped my change to use swc-node

which is a shame, swc-node is genuinely faster then ts-node

see here https://github.com/peer3to/state-channels-plus/pull/343/changes#diff-80423c1b895fee0b20857bd1e4e9567635ceca226932cfac89c609297b1874ebR55

src/utils/node/LocalDiscoveryServer.ts
```diff
@@ -73,6 +74,11 @@ export class LocalDiscoveryServer {
 
     private constructor() {}
 
+    private static getExternalRegistryUrl(): string | undefined {
+        const url = config.LOCAL_DISCOVERY_REGISTRY_URL?.trim();
+        return url ? url : undefined;
```
daiagi, line 79:
```suggestion
return config.LOCAL_DISCOVERY_REGISTRY_URL?.trim()
```

src/Holepunch.ts
```diff
@@ -55,6 +55,13 @@ class Holepunch {
     //Mark resources for garbage collection
     public async dispose() {
         this.leaveTopics();
+        if (typeof this.swarm?.destroy === "function") {
+            await Promise.resolve(this.swarm.destroy());
+            return;
+        }
+        if (typeof this.swarm?.close === "function") {
+            await Promise.resolve(this.swarm.close());
+        }
```
daiagi, line 64: why was this added? how is it related to the topic of this PR?
was there any problem?
and , same as i ask every time, is there any case where they are not a fucntion?

how about?
```
await this.swarm?.destroy?.()
```

src/evm/p2pRuntime/host/EventForwarding.ts
lukastanisic99: For now it's always clonable since all the event args are on-chain event args, which are returned by ethers (one way or the other), but yes it should throw I'll fix it

src/evm/p2pRuntime/P2pRuntimeClient.ts
lukastanisic99: We don't have agreementTime on `this side` of the port (we could, but don't want to pipe it, since I don't think it's needed just for this), so I added a default timeout that's very generous with the posibility to override. This is only usefull to potentially catch livelock/deadlock bugs on the host, both sides of this port are the same peer, so not unexpected malcious behavior and if the request is routed to another peer, it's tracked on the host side which knows about agreementTime and other concepts.

src/utils/moduleLoader/node/importModuleFromManifest.ts
lukastanisic99: I haven't, but I knew you were going to think I have - take a look at `tsconfig.json` `"swc":true` that should tell ts-node to use it while ts-node still uses its fileloader which swc on its own doesn't have - claude explained it like that - double check

src/evm/p2pRuntime/P2pRuntimeClient.ts
lukastanisic99: Implemented it and failed on testing, since request doesn't send a request if it's disposed, so just reverted it to this - concurrent requests will be disposed with rejectAllPending

lukastanisic99:
@daiagi pushed changes

daiagi (approved):

PR #347, opened 2026-06-18
Test-contracts

daiagi:
Story time:

Initially the plan was to fuzz Sol contracts methods, but after giving it some more thougt, I decided this is not a good approach.
Why?

many of the methods, consume structs, not simple type arguments, and most of the time the structs are correct only within a context (linked to previous block, to genesis state, proof linked to dispute via hash, etc..)

to fuzz them, i,e to create these structs with random field values, will yield none sense in the vast majority of cases.
and in general this idea is unlikely to expose protocol bugs.

only few util style pure function can be fuzzed, but the value is little, these type of functions are  simple and clear.
and it makes sense to simply add a few unit tests for them. 

What I did instead,, with help of AI, is to try and find attack vectors, block them and add regression tests
i.e the test was failing before thee fix, now it passes

In addition, added a few foundry based unit tests that test the contracts directly, without involving the TS SDK. these can be expended in later PR's

Finally, instead of fuzzing the Sol contracts, i think it is more valuable to "fuzz" in the e2e test level
i.e create semi random list of operations and make sure the protocol handles it properly (not a new idea, we discussed this since long time ago)
I have created an initial MVP for this, to be expended on later PR's.

the fuzzing exposed a test infra bug (see test/e2e/disputeValidation/reducedForkTimestampMismatch.test.ts)
which i would like your inout how you think is best to solve it
the bug i think is test artifact, not a bug in the sdk.


docs/reduced-fork-timestamp-mismatch.md
```diff
@@ -0,0 +1,87 @@
+# Why some multi-dispute tests fail: a clock timing mismatch (not a protocol bug)
```
daiagi: @lukastanisic99  this doc is (obviously) AI generated and i have committed it only to communicate with you the issue and ask for you input on how best to fix this.

Once we agree, I'll delete it

daiagi: deleted

lukastanisic99:
@daiagi should I review this?

daiagi:
> @daiagi should I review this?

yes

contracts/V1/StateChannelDiamondProxy/utils/DisputeUtils.sol
```diff
@@ -67,6 +67,9 @@ function _getUnfinalizedBlockConfirmationsFromStateProof(StateProof memory state
     BlockConfirmation[] memory blockConfirmations = new BlockConfirmation[](0);
     if (stateProof.milestones.length > 0) {
         MilestoneProof memory lastMilestone = stateProof.milestones[stateProof.milestones.length - 1];
+        if (lastMilestone.blockConfirmations.length == 0) {
+            return blockConfirmations;
```
lukastanisic99, line 71: We have a fraud proof that checks does a milestone have blockConfimrations, so this is an un needed check that's just increasing the bytecode (which we have to shrink and refactor to deploy on mainnets) + gas costs, but it's so small that it's fine to keep it - just pointing it out

contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol
```diff
@@ -35,6 +37,23 @@ contract DisputeFraudProofFacet is StateChannelCommon {
         return finalDisputes;
     }
 
+    function applyDisputeFraudProofs(DisputeFraudProof[] memory proofs) public {
```
lukastanisic99, line 16: why have 2 entry functions for applying DisputeFraudProofs? The above function was the entry - this is probably a better name, so the above one should be named like this. Is there another reason to have this one also?

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
```diff
@@ -156,7 +157,8 @@ contract DisputeManagerFacet is StateChannelCommon {
         DisputeData storage disputeData = disputeData[dispute.input.channelId];
         SnapshotData storage snapshotData = stateSnapshots[dispute.input.channelId].snapshotData;
         uint256 pendingCount = getPendingParticipants(dispute.input.channelId).length;
-        uint256 thresholdCount = snapshotData.participants.length + pendingCount - disputeData.onChainSlashes.length;
+        (, uint256 thresholdCount) =
+            Math.trySub(snapshotData.participants.length + pendingCount, disputeData.onChainSlashes.length);
```
lukastanisic99: Is this only theoretical or you were able to practically test having more on-chain slashes than participants? trySub returns `0` if b > a; Either way I don't think this is correct; 

The code should exclude participants that were slashed and the resulting set should always be >=1 (at least 1 honest peer);

contracts/V1/StateChannelDiamondProxy/DisputeVerificationFacet.sol
```diff
@@ -151,7 +151,7 @@ contract DisputeVerificationFacet is StateChannelCommon {
                         break;
                     }
                 }
-                if (!isAlreadySlashed) {
+                if (!isAlreadySlashed && slashCount < maxSlashCount) {
```
lukastanisic99, line 154: Seems like an unneeded check since maxSlashCount=participants+pending (the largest possible set) and isAlreadySlashed does not allow duplicate slashes -> slashCount<=maxSlashCount

Is there a test contradicting this?

contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol
lukastanisic99: Hmm ok so this function was `moved` from `DisputeVerificationFacet`, but it has it's own implementation/interpretation of `verifyDisputeFraudProofs` and the `verifyDisputeFraudProofs` from above is not wired anymore (not used);

contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol
```diff
@@ -197,6 +216,17 @@ contract DisputeFraudProofFacet is StateChannelCommon {
         returns (address)
     {
         DisputeInvalidStateProof memory proof = abi.decode(encodedFraudProof, (DisputeInvalidStateProof));
+
+        if (dispute.postedAuditingData) {
+            if (dispute.input.disputeAuditingDataHash != keccak256(abi.encode(proof.auditingData))) {
+                return _invalid();
+            }
+        } else if (
```
lukastanisic99: why `else if` and not `if`?

contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol
```diff
@@ -215,7 +245,7 @@ contract DisputeFraudProofFacet is StateChannelCommon {
 
         bytes memory data = abi.encodeCall(StateProofFacet.verifyStateProof, (dispute, proof.auditingData));
         (bool success, bytes memory returnData) = stateProofFacetAddress.delegatecall(data);
-        if (!success) return _valid(dispute.input.disputer);
+        if (!success) return _invalid(); // auditingData hash mismatch -> proof is invalid
```
lukastanisic99: I understand calude's reasoning of why this change makes sense, but I'm not convinced entirely since what if a stateProof can not be decoded, than an invalid one will be accepted

contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol
```diff
@@ -197,6 +216,17 @@ contract DisputeFraudProofFacet is StateChannelCommon {
         returns (address)
     {
         DisputeInvalidStateProof memory proof = abi.decode(encodedFraudProof, (DisputeInvalidStateProof));
+
+        if (dispute.postedAuditingData) {
+            if (dispute.input.disputeAuditingDataHash != keccak256(abi.encode(proof.auditingData))) {
+                return _invalid();
```
lukastanisic99: When you take a look at DisputeManagerFacet line 24 this is always dead code

lukastanisic99:
I'll see to review this by tomorrow - I need to do some other stuff now + need to think about many things here

contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol
daiagi: I don't think so. 
The code here is checking fraud proof, not the upload of a dispute.
It's the  handler that validates a fraud proof.
and there is a new test that asserts that 
it("[calldata posted] auditingData.latestFinalizedStateStateMachineState = random → DisputeInvalidStateProof; proof author slashed"


point is,  if  `hash(proof.auditingData)  !== (dispute.input.disputeAuditingDataHash`
the fraud proof is incorrect and the uploaded should be slashed, right?

previously this check was done as a `require` at the top of `StateProofFacet.verifyStateProof`

but that require reverts, which returns `success == false`
which PREVIOUSLY returns _valid => the honest disputer got slashed




in short, an attacker could corrupt some field of auditingData when uploading a fraud proof and get the honest disputer slashed

contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol
daiagi: verifyStateProof has only single reason why it can return `success == false`

```solidity
        require(
            dispute.input.disputeAuditingDataHash == keccak256(abi.encode(disputeAuditingData)),
            ErrorAuditingDataHashMismatch()
        );
```


which is the same case as above.

so if the hashes mismatch, how can it be `_valid`??

contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol
daiagi: This check already exists in `verifyStateProof`  but it returns false
which in turn returns `_valid(dispute.input.disputer)`
which once again, slashes the honest disputer if the proof uploader supplied rubbish genesisStateSnapshotData

contracts/V1/StateChannelDiamondProxy/utils/DisputeUtils.sol
daiagi: you're right. removed

contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol
daiagi: removed `verifyDisputeFraudProofs`

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
daiagi: theoretical. and changed for better implementation as  you suggested

contracts/V1/StateChannelDiamondProxy/DisputeVerificationFacet.sol
daiagi: see  ` it("dispute.input.onChainSlashes has > maxSlashCount distinct addresses → reduce must not OOB-panic, both offenders slashed",`

contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol
lukastanisic99: My point was that by design when you look at DisputeManagerFacet:24 (the only place where you can upload a dispute with calldata), it enforces this check or it reverts -> all disputes that are uploaded with calldata satisfy this check, in other words there can't exist a committed dispute that does not satisfy this check

Many of the things here got me thinking about the design of things. How things are fragile since there are many implicit assumptions carried out in the code which are true, but not obvious to the reader, so for instance this check. A dead check only because DisputeManagerFacet:24 exists, if that line was removed it would break the assumption here. So even if the code/facet is modular, the system as a whole is not. What's a good solution, to always have these redundant checks? Not sure. In general having tests, should catch this (e.g. removing line 24, would cause some test to fail), just wondering what's the cleanest design so when you look at things in isolation you don't have to bare the burden of understanding the entire system to reason with the small isolated chunk of code

daiagi: `in other words there can't exist a committed dispute that does not satisfy this check`


we  must be thinking along different lines
it is not about the dispute. it is about the fraud proof
all you have to do is upload a byzantine fraud proof and you get the honest disputer slashed.


https://github.com/peer3to/state-channels-plus/pull/347/changes#diff-0b84e6432412867557785c34204c5502e99eaa97f26cc552d04e750923a92d57R12

lukastanisic99: > `in other words there can't exist a committed dispute that does not satisfy this check`
> 
> we must be thinking along different lines it is not about the dispute. it is about the fraud proof all you have to do is upload a byzantine fraud proof and you get the honest disputer slashed.
> 
> https://github.com/peer3to/state-channels-plus/pull/347/changes#diff-0b84e6432412867557785c34204c5502e99eaa97f26cc552d04e750923a92d57R12

would you convert `StateProofFacet:48`

from

```solidity
require(
            dispute.input.disputeAuditingDataHash == keccak256(abi.encode(disputeAuditingData)),
            ErrorAuditingDataHashMismatch()
        );
```
to
```soldity
if(dispute.input.disputeAuditingDataHash != keccak256(abi.encode(disputeAuditingData))return true;
```

lukastanisic99: probably what makes most sense is this:
Have the check be as you did in the fraud proof (the fraud proof has to make sure the correct data is provided) and `verifyStateProof` should return true if valid and false in all other cases. verifyStateProof should internally catch any decoding problems and treat it as false the same way that `isBlockAuthentic` does. We can bubble up the reverts and take no action on them.

contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol
daiagi: I've pushed a new test: look for `"unlinked genesisStateSnapshotData against a valid genesis dispute`

image

see the screenshot, it fails if  the `else-if` is commented out

im not sure if it should be `if` or `else-if` tbh

test/harness/actions/assert/AssertActions.ts
```diff
@@ -31,4 +32,38 @@ export class AssertActions<
         this.sync = new AssertSyncActions(this.harness);
         this.storage = new AssertStorageActions(this.harness);
     }
+
+    async slashedOnChain(
```
lukastanisic99: This would be the only method in assert /root/ so better move it to assert.dispute or somewhere else

src/utils/LoggerUtils.ts
```diff
@@ -528,22 +528,37 @@ export class LoggerUtils {
         };
     }
 
+    // logging must tolerate attacker-supplied undecodable blocks (a malicious dispute can carry junk
+    // `encodedBlock`s). decode per-block defensively so a poison dispute can't crash a peer's logging.
+    private static safeBlockMetadata(decode: () => Block) {
```
lukastanisic99: This is a general problem for all our decode operations, not just for logging. 

We should consider our decode operations implicitly doing a try/cath and returning `T | null` instead of just `T` so we can account for this in the typesript code once. 

Then the loggrs just log null or whatever message/format they want to log when it's null

daiagi: I don't think that's a good idea.
It  would force null checks everywhere, including in inner trusted code paths.


I opted for different idea, inspired by https://cekrem.github.io/posts/parse-dont-validate-typescript/

parse the untrusted stateProof at the top of `validateDispute`,  not hidden away inside Logger 

see commit https://github.com/peer3to/state-channels-plus/pull/347/changes/c89d2029dcaffd350e2e18df1803b7c0c0b35b9e

test/harness/actions/assert/AssertActions.ts
daiagi: done

contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol
daiagi: see commit
https://github.com/peer3to/state-channels-plus/pull/347/changes/6251b74f52e64108b2edd15c4eccff1d4f611287


looks clean to me and handles all cases correctly

- state proof valid -> slash prover
- state proof invalid -> slash disputer
- bad reference (hash mismatch) -> slash prover
- out-of-gas, panic -> revert, no-op

contracts/V1/StateChannelDiamondProxy/DisputeVerificationFacet.sol
lukastanisic99: Took a look at the test and the test is not related to the comment above - the test is someone populating dispute.input.onchainSlashes maliciously and getting slashed and the dispute killed - this is no different than having any other dispute.input field be invalid.

dispute.input.onChainSlashes != onChainSlashes; onChainSlashes (the one that's actually on-chain) are all the addresses that actually got slashed, while dispute.input.onChainSlashes are your commitment to the set of addresses that are slashed at the moment you created the dispute. If you're honest your commitment is always a subset of the actual onChainSlashes and if you're not (as in the test), you get slashed and your dispute killed.

lukastanisic99: the check is harmless, just wont be relevant in the end, since only honest disputes make it to reduce (or at least that's the plan)

lukastanisic99:
Will finish the review tomorrow before the call and will also do a PR on top with some proposed changes - need to sleep now and will finish what's left when I wake up

contracts/V1/StateChannelDiamondProxy/DisputeVerificationFacet.sol
daiagi: Not quite true. Remove the code and run the test. It panics/reverts

daiagi: without the check:
image


with the check:

image

lukastanisic99: Hmm interesting - I'll investigate to see the root cause

lukastanisic99: Actually for me the test is not passing, but not for these reasons - for the unknown snapshot reason - will fix that and observe the result

daiagi: you left this TODO in `test/e2e/E2E-MaliciousUpdateSnapshot.test.ts`
(in case you forgot)

```
    // TODO(separate-PR): the test BODY passes, but the afterEach hook hits the
    // known product teardown bug `onStateSnapshotUpdated: unknown snapshot while
    // status=4` (EventHandler.apply) — a slashed/disputed peer (status=4)
    // receiving an unknown snapshot after resolution, same class as the deferred
    // E2E-Spectate case4 teardown. Not a harness-conversion issue; fix product
    // side (status=4 unknown-snapshot handling).
```

lukastanisic99: Yes

lukastanisic99: Found the root cause:
image
So essentially it's erroring on the byzantine peer because the byzantine peer reduces locally their byzantine dispute to compare `do they have more evidence` - this is not observed for honest peers and works as designed (observe invalid dispute -> kill -> no reduce call with it) 

So the deeper question which is more general - what do we do with errors coming from byzantine peers, since the errors we throw were designed for honest peers. In my opinion once a peer is marked byzantine, we should discard their further errors (ignore them in tests). I think this is straightforward to do, since now erorrs are also communicated back to harness over the `MessageChannel`, but I'll see - I'll see to include it in my PR later today

daiagi: >In my opinion once a peer is marked byzantine, we should discard their further errors (ignore them in tests)


agree. should not care for errors from byzantine peers

lukastanisic99: For peers running in a worker this should be straightforward since their global nodejs EventListener is still isolated and communicates it back to the main thread, but for 'inline' peers, the event listener is a global singleton and the global singleton pattern doesn't work well with the the isolation logic per peer as stated in the PR I did - each singleton needs to be covered on a case by case basis, but I'll think about it

PR #342, opened 2026-05-31
Added support for WebRTC in worker threads

lukastanisic99:
All of this just because `RTCPeerConnection` not available in worker threads. 

`In short (AI summary):`

## Summary

Adds support for establishing WebRTC when the SDK is running inside a browser worker.

Some worker runtimes do not expose `RTCPeerConnection`, so WebRTC setup now chooses a backend at runtime:

- Use the local WebRTC path when `RTCPeerConnection` is available.
- Use a main-thread bridge when running in a worker without `RTCPeerConnection`.
- Fail with a clear setup error if neither path is available.

## Main Flow

The main thread owns the `RTCPeerConnection`; the worker owns SDK signaling and transport usage.

1. The app installs the bridge with `installWebRTCMainThreadBridge(worker)`.
2. Worker WebRTC setup requests are forwarded to the main thread.
3. The main thread creates/accepts offers and applies ICE candidates.
4. ICE candidates discovered on the main thread are sent back to the worker, which forwards them through existing RPC signaling.
5. Once a data channel exists, the main thread transfers the `RTCDataChannel` to the worker.
6. The worker creates `WebRTCTransport` only after the data channel is open.

## Fallback

If `RTCDataChannel` transfer is not supported and `allowProxyFallback` is enabled, the channel stays on the main thread and the worker receives a proxy data channel instead. Proxy messages are keyed by peer address.

## Other Changes

- `WebRTCTransport` now expects an already-open data channel.
- WebRTC setup internals were moved into a `connection` subdirectory.
- Narrowed a few imports to avoid CommonJS circular import issues in tests.

PR #348, opened 2026-06-18
fix(webrtc): harden worker-thread WebRTC bridge resilience

sh3ll3x3c:
Builds on #342 (`webrtc-worker`). This PR targets that branch so the changes show up as a reviewable delta on top of the worker-thread WebRTC work, rather than as inline comments.

## Summary

Hardening for the worker→main WebRTC bridge introduced in #342:

- **Bound the bridge request lifecycle.** Worker→main requests (`createOffer`/`acceptOffer`/`applyAnswer`/`addIceCandidate`/`close`) had no timeout and nothing rejected in-flight requests when the `MessagePort` was replaced or the bridge was torn down. A dropped/closed bridge could leave the awaiting WebRTC setup call pending forever. Each request now has a timeout, and replacing the port rejects its in-flight requests via a `dispose()` on the bridge client.
- **Made channel delivery safe by default and deterministic.** Transferable `RTCDataChannel` is Chromium-only; on Firefox/Safari the transfer attempt throws. The bridge defaulted to *not* falling back, so channel setup failed there (and on `acceptOffer` it failed asymmetrically, after the answer was already returned). Proxy fallback now defaults on, the transfer-vs-proxy choice is explicit via a `channelMode` option (`auto` | `transfer` | `proxy`) plus a one-time capability cache, and a transfer-unsupported runtime is logged once instead of silently degrading.
- **Normalized local-path ICE candidates** via `toJSON()` so they serialize to a reconstructable init dict across engines (matching the bridge path).
- **Removed the unreachable `getState` bridge request path** and made the broker throw on an unknown method instead of resolving `undefined`.
- **Registered the worker bridge listener before latching its guard flag**, so a premature main-thread call can't suppress registration.
- **Documented** that the handshake-ack `challengeHash` is a diagnostic correlation id, not an authenticity check (authenticity is the signature check in `onInitHandshakeResponse`).

### Note for review
`challengeHash` is documented as diagnostic rather than turned into a validated check: real validation needs the responder to persist the request-side challenge and adds a new handshake failure path that isn't end-to-end verifiable from the current browser smoke. Happy to add it if you'd prefer — it's your handshake design call.

## Test Plan

- `tsc --noEmit` on both `tsconfig.json` and `tsconfig.browser.json` — clean.
- **Full `yarn test` (non-e2e hardhat suite) — 257 passing.** Includes the WebRTC unit tests:
  - Rewrote the transport test that asserted a constructor throw the rewrite no longer performs; it now asserts queue-then-flush-on-open, plus closed-drop and handshake-dedup.
  - Added a worker-bridge port-replacement rejection test.
- `yarn test:browser:worker` (real headless Chromium via Playwright) — pass, including a **new forced-proxy worker-bridge smoke** so the proxy path (the one Firefox/Safari always take) is exercised end-to-end; Chromium otherwise always transfers and never hits it.

### DIFF
--- a/src/rpc/services/WebRTCSetup/connection/WebRTCMainThreadBridge.ts
+++ b/src/rpc/services/WebRTCSetup/connection/WebRTCMainThreadBridge.ts
@@ -13,7 +13,24 @@ import type {
 } from "./WebRTCConnectionTypes";
 import { loadWebRTCProvider, type WebRTCProvider } from "./WebRTCProvider";
 
+export type WebRTCChannelMode = "auto" | "transfer" | "proxy";
+
 export type WebRTCMainThreadBridgeOptions = {
+    /**
+     * How a freshly-created RTCDataChannel is handed to the worker:
+     *  - "auto" (default): attempt to transfer the channel to the worker and,
+     *    if the runtime can't transfer it, fall back to proxying its messages
+     *    over the bridge.
+     *  - "transfer": only transfer; throw if the runtime can't (strict).
+     *  - "proxy": always proxy — use when transfer is known-unsupported
+     *    (e.g. Firefox/Safari) or in tests.
+     */
+    channelMode?: WebRTCChannelMode;
+    /**
+     * Whether "auto" mode may fall back to proxying when a transfer attempt
+     * fails. Defaults to true so the bridge works on runtimes that do not
+     * support transferable RTCDataChannels.
+     */
     allowProxyFallback?: boolean;
 };
 
@@ -32,6 +49,9 @@ type ConnectionRecord = {
 
 class WebRTCMainThreadBridgeBroker {
     private provider?: WebRTCProvider;
+    // Cached the first time "auto" mode attempts a transfer: once we learn the
+    // runtime can't transfer channels we go straight to proxy for the rest.
+    private channelTransferSupported?: boolean;
     private readonly connectionsByPeerAddress = new Map<
         WebRTCPeerAddress,
         ConnectionRecord
@@ -134,9 +154,11 @@ class WebRTCMainThreadBridgeBroker {
         if (request.method === "close") {
             return this.close(request.peerAddress);
         }
-        if (request.method === "getState") {
-            return this.getState(request.peerAddress);
-        }
+        throw new Error(
+            `Unknown WebRTC bridge request method: ${
+                (request as { method?: string }).method
+            }`
+        );
     }
 
     private async createConnectionRecord(
@@ -255,22 +277,16 @@ class WebRTCMainThreadBridgeBroker {
         record.connection.close();
     }
 
-    private getState(
-        peerAddress: WebRTCPeerAddress
-    ): WebRTCConnectionStateSnapshot {
-        const connection =
-            this.connectionsByPeerAddress.get(peerAddress)?.connection;
-        if (!connection) {
-            return { connectionState: "unknown", iceState: "unknown" };
-        }
-        return this.getSnapshot(connection);
-    }
-
     private postChannel(
         peerAddress: WebRTCPeerAddress,
         record: ConnectionRecord,
         channel: WebRTCDataChannelLike
     ): void {
+        if (this.shouldProxyChannel()) {
+            this.postProxyChannel(peerAddress, record, channel);
+            return;
+        }
+
         const transferredMessage: WebRTCBridgePortMessage = {
             namespace: WEBRTC_BRIDGE_NAMESPACE,
             type: "channel",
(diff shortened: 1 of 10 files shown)


src/rpc/services/WebRTCSetup/connection/WebRTCMainThreadBridge.ts
```diff
@@ -13,7 +13,24 @@ import type {
...
+    /**
+     * How a freshly-created RTCDataChannel is handed to the worker:
+     *  - "auto" (default): attempt to transfer the channel to the worker and,
+     *    if the runtime can't transfer it, fall back to proxying its messages
+     *    over the bridge.
+     *  - "transfer": only transfer; throw if the runtime can't (strict).
+     *  - "proxy": always proxy — use when transfer is known-unsupported
+     *    (e.g. Firefox/Safari) or in tests.
+     */
+    channelMode?: WebRTCChannelMode;
```
lukastanisic99, line 29: either this or `allowProxyFallback` is redundant

src/rpc/services/WebRTCSetup/connection/WebRTCMainThreadBridge.ts
```diff
@@ -281,13 +297,41 @@ class WebRTCMainThreadBridgeBroker {
...
+            this.channelTransferSupported = false;
+            const channelMode = this.options.channelMode ?? "auto";
+            const allowProxyFallback = this.options.allowProxyFallback ?? true;
+            if (channelMode === "transfer" || !allowProxyFallback) {
                 throw error;
             }
+            // Surface the reason once (subsequent channels skip the attempt via
+            // shouldProxyChannel) so a transfer-unsupported runtime is visible
+            // rather than silently degrading.
+            console.warn(
```
lukastanisic99: for your `claude.md` don't use console.* directly for logging use our internal logger. 

This brindge runs on the MainThread - we also return a logger during p2pSetup so the 'app' can log stuff. We collect those logs, send them to a server and analyze them inorder to fix bugs -> add our logger to `WebRTCMainThreadBridgeOptions` and use it here

src/rpc/services/initHandshake/InitHandshakeRpcMethods.ts
```diff
@@ -228,6 +228,11 @@ class InitHandshakeRpcMethods extends ARpcMethods {
     /**
      * Sent after a peer verifies our handshake response. We only treat the handshake
      * as complete once we have both: (1) verified the remote, and (2) received this ack.
+     *
+     * `challengeHash` is a diagnostic correlation id only — it lets a single log
+     * stream be followed across the two threads/peers. It is NOT an authenticity
+     * check and MUST NOT be trusted for any decision: peer authenticity is
+     * established solely by signature verification in `onInitHandshakeResponse`.
```
lukastanisic99, line 235: (not related to this PR) This entire service can be simplified with my latest PR that introduces request/response for RPC services, so insead of comunicating over dedicated endpoints (onInitHandshakeRequest->void, onInitHandshakeResponse->void) we can have just one endpoint that returns the response (e.g. onRequest->response) and write the service as `response = await remoteRpc.initHandshakeService.onRequest(request).request(peer)`

This also removes the explicit tracking of requests/responses in the service and reduces the public endpoint surface which is better from a security perspective (less things to attack/exploit)

Probably other RPC services can be simplified in the same way, since they we're all writen for a `fire and forget` communcation style

lukastanisic99 (approved):

src/rpc/services/WebRTCSetup/connection/WebRTCMainThreadBridge.ts
sh3ll3x3c: Agreed — removed `allowProxyFallback` in 4e2ff3b. `channelMode` is now the single knob: `transfer` = strict (throw, no fallback), `auto` = transfer then fall back to proxy, `proxy` = always proxy.

src/rpc/services/WebRTCSetup/connection/WebRTCMainThreadBridge.ts
sh3ll3x3c: Done in 4e2ff3b — added a `logger` to `WebRTCMainThreadBridgeOptions`; the transfer-unsupported notice now goes through `logger.warn(...)` instead of `console.warn`, and I documented the no-`console.*` rule in `AGENTS.md`. The SDK side accepts the logger now; I will wire the actual logger through at the consuming app's `installWebRTCMainThreadBridge(...)` call site in a follow-up there.

src/rpc/services/initHandshake/InitHandshakeRpcMethods.ts
sh3ll3x3c: Makes sense — leaving this for a follow-up since it depends on your request/response RPC change. The doc comment added here is just a stopgap clarifying that `challengeHash` is diagnostic-only; happy to collapse the handshake service into the single `onRequest -> response` pattern once that lands.

Issue #357, opened 2026-06-22
A participant missing one dispute can't validate a reduction and crashes


momoshell:
When a dispute is being reduced, each peer re-checks the disputes that make up the on-chain dispute window by looking them up in its own local storage. If even one of them isn't there, the code gives up on the spot:

```ts
const dispute = this.storage.disputes.getDispute(commitment);
if (!dispute) {
    //TODO - querry longs
    throw new Error(`Dispute not available for commitment: ${commitment}`);
}
```

That `//TODO - querry longs` is the missing piece — the intent was to fall back to fetching the dispute from chain logs, but it was never written. For a peer that isn't an active participant the error is swallowed (it just disconnects); for a participating peer it's rethrown, and since the event listener marks the log handled before dispatching it, there's no redelivery. The peer's reduction handling simply dies.

So any honest participant that happened not to observe and store one dispute in the window can't validate an otherwise-perfectly-valid reduction. It doesn't recover — it falls over.

**How to fix:** implement the log fallback — when a commitment isn't in local storage, pull that dispute from the chain's dispute events and hydrate it before validating. The same lookup already exists in the P2P event-hooks helper, where a miss is handled gracefully (skip, don't throw); that's the behaviour to converge on, ideally by extracting a single shared helper.

_Repro (pre-rewrite test branch): the "join/leave sequence and fork resolution" e2e scenario._

Issue #356, opened 2026-06-22
An honest fraud proof for a broken block link is rejected on-chain, so the bad dispute survives


momoshell:
When a state proof contains two blocks whose chain-link is broken — one block's "previous hash" doesn't match the block before it — the off-chain and on-chain sides disagree about how to punish it.

Off chain, the broken link is caught during block replay and packaged as a `BlockInvalidStateTransition`-backed proof. On chain, the matching handler accepts that proof shape, but then asks the underlying fraud proof to actually name someone to slash — and for a mere linkage break it can't, so it bails out as invalid:

```solidity
address adr = runFraudProof(fraudProof, dispute);
if (adr != address(0)) return _valid(dispute.input.disputer);
return _invalid();   // ← a broken link isn't a slashable state-transition fault, so we land here
```

`_invalid()` makes the apply transaction revert with an invalid-fraud-proof error, so every honest peer's attempt to kill the bad dispute fails and the malicious dispute survives. The off-chain pipeline is convinced it has a valid fraud proof; the on-chain handler refuses to act on it.

**How to fix:** reconcile the two sides for linkage breaks — either have the on-chain handler treat a proven broken link as slashable in its own right, or have the off-chain pipeline build the proof type the contract will actually act on. Touches both the off-chain fraud-proof construction and the on-chain handler.

_Repro (pre-rewrite test branch): the "previousBlockHash = random (inter-block linkage break)" case._

Issue #355, opened 2026-06-22
Inbound joins can be appended to a fork that's under dispute (guard was removed)


momoshell:
There used to be a guard that stopped an inbound join from being appended while the current fork was under dispute. It was removed in commit `029c6a82` ("remove race condition check from appendInboundMessages"):

```solidity
StateSnapshot memory currentSnapshot = stateSnapshots[channelId];
require(
    !StateChannelManagerInterface(address(this)).isForkDisputed(channelId, currentSnapshot.forkId),
    RaceConditionForceInboundJoinForkDisputed()
);
```

The error type is still declared but is now thrown nowhere, and the append path no longer checks whether the fork is disputed at all. So an inbound join can land on a fork that's actively being contested — the exact race the guard prevented. (There's a similar check on the normal join entry point, but it's a different error and doesn't cover the force-inbound path.)

**Decision needed:** either restore the guard — re-add the disputed-fork check centrally so it covers every append path — or, if dropping it was deliberate, delete the now-dead error type and write down why appending during a dispute is safe. Worth a quick word with whoever removed it.

_Repro (pre-rewrite test branch): "forceInboundJoin on disputed fork reverts" — it expects a revert that can no longer happen._

Issue #354, opened 2026-06-22
Can't confirm a pending joiner survives a dispute reduction (blocked upstream)


momoshell:
This started life as "a pending joiner's join isn't re-applied when a dispute reduces the fork." Tracing it shows we can't actually confirm that yet — the test never gets far enough to check.

What really happens: the reduction lands on chain fine, but no peer adopts the reduced fork (because of the two upstream gaps — the unimplemented local fork transition and the self-removed-peer crash). So the scenario times out waiting for the on-chain snapshot to change, long before it ever looks at whether the joiner made it across.

So the joiner question is real but **unanswerable until those two are fixed**. Once peers can actually follow a reduced fork, re-run this: if the joiner is then missing from the participant set on the reduced fork, it's a genuine "pending join dropped on reduction" bug to fix here; otherwise it just resolves on its own.

Depends on #350 and #353.

_Repro (pre-rewrite test branch): "pending joiner participates after dispute reduction"._

PR #358, opened 2026-06-22
Scale the e2e parallel runner to the host + optional deploy-once reuse

momoshell:
Reworks the e2e parallel runner to use the machine well and stop redeploying contracts on every test. Each piece is its own commit.

**Scheduling.** Costs each test by its peer/thread demand and admits against a `cores × factor` budget (RAM-capped, largest-first) instead of a flat worker count, then self-tunes toward ~0.9 load/core from metadata persisted between runs. The event-loop-delay watchdog stays the hard ceiling. Each result line shows its slot index; `--dry-run` prints the schedule.

**Shared infra (opt-in, `--per-slot-node`).** One discovery registry shared across all child processes, and each slot keeps a long-lived node so tests connect to an already-deployed manager — keyed by config, reused under their own channelId — instead of redeploying the diamond. Off by default; without the flag the runner is unchanged. Validated at scale: 42 tests serially on one node, 33 manager reuses, zero non-monotonic-timestamp throws.

**Worker topology.** Under `--per-slot-node` the SDK runs in a worker by default (the production setup), since the per-slot node supplies the provider and signer it needs.

Calibrated default factor 4 is ~42% faster than the old flat default with zero starvation.

Follow-ups, pre-existing and independent of the runner: [#350](https://github.com/peer3to/state-channels-plus/issues/350), [#351](https://github.com/peer3to/state-channels-plus/issues/351), [#352](https://github.com/peer3to/state-channels-plus/issues/352), [#353](https://github.com/peer3to/state-channels-plus/issues/353), [#354](https://github.com/peer3to/state-channels-plus/issues/354), [#355](https://github.com/peer3to/state-channels-plus/issues/355), [#356](https://github.com/peer3to/state-channels-plus/issues/356), [#357](https://github.com/peer3to/state-channels-plus/issues/357), [#368](https://github.com/peer3to/state-channels-plus/issues/368).


scripts/test-e2e-parallel.js
```diff
@@ -3,11 +3,27 @@ const { spawn } = require("child_process");
...
+// hosts. Override anytime with --workers/-w.
+const PER_WORKER_MEM_GB = 2; // rough budget per concurrent hardhat task
+function defaultWorkerCount() {
+    const cpuCount = os.cpus()?.length ?? 4;
+    const totalMemGb = os.totalmem() / 1024 ** 3;
+    const byCpu = cpuCount - 2; // leave 2 cores for the OS + this orchestrator
+    const byMem = Math.floor(totalMemGb / PER_WORKER_MEM_GB);
+    return Math.max(2, Math.min(byCpu, byMem));
+}
+const DEFAULT_WORKERS = defaultWorkerCount();
```
lukastanisic99: target the `dispute` branch, that's our 'dev' branch - I've collapsed all these branches into it yesterday 

As for this script, ok improvement to have the script implicitly deduct the number of workes from the hardware available on the machine, instead of us having to manually provide it with the --worker (-w) flag, but what I was thinking was to improve on the concurrency. 

# Some additional context: 
When this script was designed we didn't have worker threads, everything was running on the main thread so it was `1 thread per process` so we logically mapped `1 process to 1 core` and also did some buffer (e.g. run with 14 workers if you have 16 cores, so the other 2 can service other processes on your machine and you'd have less starvation). That worked fine, but we had a lot of tests that were `false positives` since the async task within the process we're being `starved` (in some cases by 9+ seconds), which lead us to 'thread' the implementation. So now we'd have more than 1 thread per process. `Each peer is a thread + their VM can also be threaded`, so the total number of threads per test/process is `2*N+1` where N is the number of peers running in the test. Some tests have 2 peers, some have 5+, so now there isn't an obvious mapping between a process and the number of threads needed. 

We're interested in optimizing our hardware resources in a smart way where we have max concurrency and have the shortest possible testing time. Not all tests take the same amount of time and not all tests have the same resource utilization rate (many of the long tests, just 'wait' for some time condition to be satisfied e.g. the dispute challenge period to expire). 

Also we do allow for some 'starvation' / delay in task scheduling, but any task that's delayed for more than 1s (configurable) will immediately cause the test to fail - which we use as a performance indicator to point out what we have to optimize.

docs/trds/e2e-reduced-fork-followups.md
```diff
@@ -0,0 +1,191 @@
...
+This register collects everything surfaced while greening the e2e suite for the parallel test
+runner (branch `test/add-hardhat-parallel-script`). **None of these were caused by the test
+changes on this branch** — each reproduces identically on the clean committed HEAD, run
+serially and in isolation. Each has a tracked GitHub issue (linked below).
+
+## How findings were verified
+
+For every skipped/failing test: run focused (`--grep`), as a whole file serially, and with the
+branch WIP both applied and stashed. Deterministic items fail identically on clean HEAD; flaky
+items were run 5–7×. Two failures (#354, #356) were traced end-to-end with `LOG_LEVEL=debug`.
```
lukastanisic99: Best to log test execution in its own file with `LOG_LEVEL=verbose` which the AI can analyze - take a look at `yarn test:e2e:log-file` or even how the parallel script creates log files;

scripts/test-e2e-parallel.js
momoshell: Repointed the PR at `dispute`, thanks.

Agreed the `cores-2` default is just a floor, not the real win — it's a safe stopgap so nobody has to pass `-w` by hand. The threaded model is the interesting part with mapping under- or over-subscribes depending on which tests happen to be running together.

---
That's a bigger change than this PR. 
Edit: As agreed, will fold your suggestion into this PR

docs/trds/e2e-reduced-fork-followups.md
momoshell: Good call. I'll standardize on per-task `LOG_LEVEL=verbose` log files (like `test:e2e:log-file`) for anything meant to be machine-analyzed, and note that in the doc's verification section.

lukastanisic99:
@momoshell will review on the plane, but on my side when running it seems like it's not parallel (will investigate). 
Btw the default behavior for the script should be to set `RUN_SDK_IN_THREAD=true` and `VM_DEDICATED_THREAD=true` also - that means both the SDK and the EVM run in a dedicated worker, achieving the 2*N+1. It should also be able to pass a flag to disable either of the 2 so we can run with all 4 variations. 

Would be also nice visually see the process_id/index next to the test so we see things actually running in parallel.

Also take a look should we spin up a dedicated process per test (more expensive) or should we spin up a dedicated worker per test(cheaper and faster startup times) - pros and cons to both, let's max out concurrency

lukastanisic99:
adding more ideas that might be useful:
We want to maximize hardware utilization and each machine is different (regardless of cpu cores or ram). I'd suggest storing small metadata on the machine after running the parallel tests and loading that metadata on the next run to try and scale up/down resource allocation to get close to 90% utilization. 
E.g. we persist
`{avgUtilization: ..., scalingFactor: ...}` 
So when the parallel script finishes it persists the average utilization during its lifespan and the scalingFactor it used. 
When the parallel script starts, it loads the metadata (if available) and deducts how much it can scale down/up e.g. `{avgUtilization: 0.5 , scalingFactor: 1}` -> it calculates scalingFactor: ~2 and runs the tests with. Probably best is to target 90% utilization and to scale down anything that's above 90% and to scale up anything that's bellow. 

It's hard to do this just by thread count, since you don't know how 'active' the threads are and it's different for each test. If we allocate All threads and they're sitting idle 50% of the time (waiting on something), we should scale up allocation. 

I think having an algorithm that's based on utilization and 'adapting to what's good on your machine` is probably better than doing this based on thread count alone which also gives us overhead to maintain tracking for how much does a test consume. 

Also print avgUtilization at the end of the tests.

lukastanisic99:
Further, after the scheduling algorithm, the biggest 'time waters' are:
1) starting a hardhat node & deploying contracts for every test; We should have 1 shared node and contracts should be deployed once - on-chain they are partitioned per channelId, so we just need to make sure the channelId is unique per test (add entropy my hashing system clock milliseconds)
2) The LocalDiscovery server should also be a shared singleton for all processes where they match over the unique channelId


1) and 2) alone should save a lot of time  + that's how the system works in production (a single global blockchain network and 1 DHT network for discovery) 

Also ideally we write 1) and 2) in a way where it just works regardless are we running a single isolated test directly with hardhat/mocha or running a dedicated orchestration script (parallel tests) - the tester should not care nor think 'did I setup the server?'

Getting startup costs down (need empirical data on this) - currently all the typescript code is transpiled at runtime with ts-node (`swc` is used so it's optimized, but still transpiling), maybe better to compile to .js and have it run .js directly with a source map enabled (source map for sure, so the logs show the correct stack trace and not some gibrish that's useless)

momoshell:
Thanks for the detailed notes — folding all of it into this PR. Where each piece stands and the plan:

**"Doesn't look parallel."** It is — the runner admits by a thread budget now, not a fixed worker count. 
I'd bet on a stale `node_modules`: after I switched to `dispute` I hit the same `@swc/core` miss, and with stale deps every task dies in <1s, so the whole run blurs past looking serial (and all-red). 

**Default `RUN_SDK_IN_THREAD=true` + `VM_DEDICATED_THREAD=true`, with toggles.** Will do. Heads-up: the cost model already derives threads-per-peer from those two envs, it just doesn't *set* them — so today it runs N+1.
I'll default both on (2N+1) and add flags to switch either off so we can sweep all four variants.

**Utilization-based scaling over thread count.** Agree — and you're right that thread count is the wrong axis for us. I'd target **load average per core (~90%)**, not CPU%: with so many wait-bound tests, CPU% reads low and would fool us into over-scheduling straight into the event-loop-starvation cliff. 

Loadavg counts *runnable* threads — it ignores peers idling on a timer and rises exactly when threads actually contend, it's host-portable (loadavg/cores), and it's free to read. The event-loop-delay watchdog stays the hard ceiling: push toward the load target, but if a run trips starvation, back off.

(For reference, the static factor is just calibrated for now: on a 16-core/64GB box, factor 4 → ~185s initial wall, ~42% faster than the old default, 0 starvation; starvation starts ~factor 6. It's host-specific, which is exactly why the loadavg layer is the real fix.)

**Shared hardhat node + shared LocalDiscovery, partitioned by `channelId`.** 
Agreed, including "just works" solo or orchestrated. Today every test boots its own node, redeploys, and runs its own discovery server — so this is the biggest startup win. I'll key the shared singletons on a `channelId`.

**Startup cost / compile vs ts-node.** Currently ts-node+swc transpiles at runtime; I'll get real boot numbers and compare against precompiled `.js` (source maps kept either way).

**Process- vs worker-per-test.** Process-per-test today; I'll lay out the tradeoffs and pick for max concurrency.

Pushing these incrementally to this branch.

momoshell:
- **Scheduler** now budgets by thread demand (per-test peer-count cost, largest-first, RAM-capped), forces `RUN_SDK_IN_THREAD`/`VM_DEDICATED_THREAD` onto children with `--no-*` toggles (so cost stays truthful), prints a per-task slot index, and self-tunes via a **load-average-per-core adaptive factor** (seeded by `--thread-factor`, watchdog as the hard ceiling). Calibrated default factor 4 (~42% faster than the old flat default, 0 starvation).
- **Shared discovery**: all child processes now rendezvous through ONE channelId-partitioned `LocalDiscovery` registry instead of N in-process ones — verified in default mode too. Opt out with `--no-shared-discovery`.
- Also dropped the `yarn` wrapper on each spawned task (spawn hardhat directly) — small, free, per-task.

**Where I landed on the shared node — and the new direction.** I went to measure this before building it, and the numbers redirected me:

- Sharing a hardhat node across tests only saves ~0.5s per test (the in-process node init). It also runs into the automine/time-control problem — every test drives node-global mining in the harness `beforeEach`, so concurrent tests on one node corrupt each other; that would need a per-channel time-control rework first.
- The cost that actually dominates a test is **redeploying the whole diamond on every test (~4s)** — far bigger than the node, and a shared node doesn't touch it.

So rather than chase the shared node (or N-node worker-mode), I'm pivoting to the: **deploy the contracts once and have each test reuse them under its own channelId**, instead of redeploying per test. That goes after the ~4s, and it avoids the global-automine blocker since tests don't need to share a live chain concurrently. That's what I'll scope next.

lukastanisic99:
You still need the shared node in order to save on the deployment cost.

The design flaw, was including hardhat into harness and having harness manage it. Luckily the hardhat surface is small so we should just plug it out and have it work as originally planed yesterday

momoshell:
> You still need the shared node in order to save on the deployment cost.

You're right, and that's what landed — I haven't updated the earlier work and notes from this morning. Each slot now keeps a persistent external hardhat node and tests connect to an already-deployed manager on it (deploy-once, reused per channelId). "Instead of a shared node" should have read "instead of a single global node," not "no node."

And agreed on the previous flaw — the external-node path is exactly "plug hardhat out of the harness": the harness no longer owns a node, it just connects over PROVIDER_URL. 

The main reason why I choose per-slot in the end rather than one global node is the harness's node-global time control — every test drives automine in `beforeEach`, so concurrent tenants on a single node corrupt each other. 

Per-slot serial reuse sidesteps that while still amortizing the ~4s deploy. A single global node is the next step once that time control is per-channel. It's behind `--per-slot-node` for now; making it the default means dropping the in-process harness path.

lukastanisic99:
> The main reason why I choose per-slot in the end rather than one global node is the harness's node-global time control — every test drives automine in `beforeEach`, so concurrent tenants on a single node corrupt each other. 
> 
> Per-slot serial reuse sidesteps that while still amortizing the ~4s deploy. A single global node is the next step once that time control is per-channel. It's behind `--per-slot-node` for now; making it the default means dropping the in-process harness path.

Just landed an hour ago after my 25h trip,so sorry if my concentration is low - what's the difference between "a shared global node" and "a shared per slot node"? It's still shared

As for the automine and beforeEach for the hardhat node,imo tests shouldn't control the node - tests should just get a provider url and should not care how the chain is configured.

Hardhat defaults to automine, we just disable it and set block intervals to 2s, but that's done once on the node level, not related to tests

momoshell:
Nice trip bruv, grats!

> what's the difference between "a shared global node" and "a shared per slot node"? It's still shared

Fair point. "Global" vs "per-slot" was a wrong distinction — both are shared; but what I've meant was that per-slot only buys serial access, which sidesteps tests fighting over mining state. 
And we completely agree, wasn't arguing against it, just used whats currently shipped as a workaround, because I wanted to avoid changes to the harness.

Since tests would just take a `PROVIDER_URL` and don't care how the chain's configured, this would be a relocation and not a re-design. 
I'll take a pass at it; if I manage to achieve the results with it, the single global node replaces the per-slot interim that's in here now.

lukastanisic99:
image
How does performance look on your machine? This is how it looks on mine -
Room to scale up

Btw, while the parallel script is running do you have access to the internet? This is a symptom I observed before both on my mac and windows device while running the script and it's still the case for me. Never investigated in detail, but I guess it's the localDiscovery server overwhelming the OS with a constant burst of n^2 connections

lukastanisic99:
See if you can split parts of the script into their files, so we don't have to look at everything at once e.g. (scheduler, argParser, logger, slotNode mode etc) 

I'll review the scheduler now, since that's probably the most interesting/critical thing

scripts/test-e2e-parallel.js
```diff
@@ -3,20 +3,243 @@ const { spawn } = require("child_process");
...
+const http = require("http");
+const net = require("net");
+const os = require("os");
 const path = require("path");
 const { Project, SyntaxKind } = require("ts-morph");
 
 const DEFAULT_LOG_DIR = "./logs";
-const DEFAULT_WORKERS = 8;
+
+const HARDHAT_CLI = require.resolve("hardhat/internal/cli/cli.js");
```
lukastanisic99: best to have 1 node script which we just call to configure and start the node and have that be external, so we can just use it here and in normal (non-parallel tests); 

This script, harness and tests in general just see a URL and don't care where the node is locally/remote - this way later we can also test a real devnet to see do our expectations match

scripts/test-e2e-parallel.js
```diff
@@ -3,20 +3,243 @@ const { spawn } = require("child_process");
...
+    }
+
+    const vmThread =
+        cli.vmThread !== undefined
+            ? cli.vmThread
+            : process.env.VM_DEDICATED_THREAD !== undefined
+              ? process.env.VM_DEDICATED_THREAD !== "false"
+              : true;
+
+    return { sdkThread, sdkThreadSource, vmThread };
```
lukastanisic99: at first glance this function is unclear and seems over engieered - not sure why we care about 'source' e.g. explicit or per-slot-node or default off 

Imo this should be simple as before, assume (true,true) unless explicitly overridden through flags

scripts/test-e2e-parallel.js
```diff
@@ -332,15 +672,221 @@ function formatResultLine({
...
     const duration = formatDurationMs(durationMs);
     if (code === 0) {
-        return `[${completed}/${total}] ${phasePrefix} ${status} (${duration})`;
+        return `[${completed}/${total}] ${phaseTag} ${status} (${duration})`;
+    }
+    return `[${completed}/${total}] ${phaseTag} ${status} ${label} (${duration})`;
+}
+
+/** Probe the OS for an available TCP port by binding to :0 and reading back. */
+function getFreePort() {
```
lukastanisic99: You're probably using this for the hardhat node or disovery server. Imo in order to keep things simple, we shouldn't think too much about these things here and all of the port assigment and connection estanblishment should be in those relevant files/scripts. This file should not worry about 'how do I start a server and manage a connection to it' it should worry about 'how do I schedule tasks to run in parallel to utilize hardware resources most effectively' 

We just need to pass urls to tests/tasks and they'll connect to the nodes and servers. This is best since you don't care if the server is local/remote or exlusive/shared. Passing all of this is straightforward through .env 

`PROVIDER_URL` for the RPC node
`LOCAL_DISCOVERY_REGISTRY_URL` for the registry 

This file and tests assume these things run independently. 
Our mental focus here should be just on the scheduler

scripts/test-e2e-parallel.js
```diff
@@ -332,15 +672,221 @@ function formatResultLine({
...
+function resetSlotCacheDir(dir) {
+    fs.rmSync(dir, { recursive: true, force: true });
+    fs.mkdirSync(dir, { recursive: true });
+}
+
+/**
+ * Spawn an external hardhat node for `slotId` on `port` and wait until its
+ * RPC endpoint is ready. Returns `{ proc, url, logStream }`.
+ */
+function startSlotNode(slotId, port, logPath) {
```
lukastanisic99: All of this we'll get rid of. The per slot (if you mean per test) node is only better for isolation and not competing with other transactions that are included in the block, but a 'shared node/network' is what's in production and after deployemnt (which we do once), all the transactions are light weight + 90% of it is off-chain + if the 10 tests that run in parallel consume so much gas that transactions are getting starved and test fail because of that, we're fucked since in production it will be way more than 10 parallel games - we need to test for all of this and make sure it's stable

scripts/test-e2e-parallel.js
```diff
@@ -391,219 +938,681 @@ async function main() {
...
-        if (result.code !== 0) {
-            rerunFailures.push(result.task.label);
-            markLogAsError(logDir, result.rerunLogName);
+        const totalFailing = rerunFailures.length;
+        const totalPassing = tasks.length - totalFailing;
+
+        const totalTime = ((Date.now() - startTime) / 1000).toFixed(2);
+
+        // Compute avg load per core from samples (or fall back to a single snapshot).
+        const avgLoadPerCore =
```
lukastanisic99: I don't know how relevant the cpu load is, since we don't know how 'expensive' the waiting threads are on average. Currenlty when I take a look at my Activity Monitor, the OS is managing 6000 threads and my utilization is constant ~10% with os.loadavg() returning [2.03,2.06,2.12] for the [1min,5min,15min] average

Claude tells me this does NOT count `threads that are sleeping — blocked waiting for something` - all our threads that are not executing are waiting for something 

I'd still try and target 90% utilization, since even if things burst the OS does a decent job time slicing + we do allow 1s starvation so we have wiggle room

momoshell:
> How does performance look on your machine? This is how it looks on mine - Room to scale up
Btw, while the parallel script is running do you have access to the internet? 

Dug into both with instrumentation.

**Internet:** ran it watching the sockets — it's 100% local. The mesh is all loopback (`127.0.0.1`, ~150 conns peak); public connections stayed flat at my idle baseline the whole run, no STUN, no mDNS burst. 

The only piece of code that should hit public servers is a dead STUN-detection helper (never called); real peer connections use `RTCPeerConnection` with no ICE servers. Zero external traffic, and on 16-core/64GB it doesn't exhaust anything (~500 ephemeral ports of 16k peak). 

If you still drop connectivity, could it be environment-specific (router NAT table / smaller ephemeral range / mDNS rate)?
Happy to packet-capture on a repro.

**"Room to scale up":** you're right, and I measured why we can't fill it yet. 
At the RAM ceiling (32 concurrent on 64GB) CPU is only ~0.6/core — the suite's wait-bound *and* we're capped by RAM, not CPU: each in-process worker carries its own hardhat EVM at ~1.4GB. Against an external node a worker drops to ~460MB — ~900MB lighter. 
So one shared node + light workers ≈ one third the RAM at scale, roughly tripling the concurrency ceiling. That's the lever — and it's the same shared-node move you've been pushing, so "scale up" and "shared node" are one thing.

I've scoped the refactor — the provider plumbing's already there. 
Just two real wrinkles to handle in phases: serial CI runs in-process so it needs an in-process mining path, and a couple of tests hardcode account indices that'd collide on a shared node. 
So I'll land it incrementally, `--per-slot-node` kept as a fallback.

lukastanisic99:
> The only piece of code that should hit public servers is a dead STUN-detection helper (never called); real peer connections use `RTCPeerConnection` with no ICE servers. Zero external traffic, and on 16-core/64GB it doesn't exhaust anything (~500 ephemeral ports of 16k peak).

Are you observing WebRTC in tests? Unless we're testing WebRTC explicitly, everything should be over the DisoveryServer/LocalTransport 

> 
> If you still drop connectivity, could it be environment-specific (router NAT table / smaller ephemeral range / mDNS rate)? Happy to packet-capture on a repro.
> 

> I've scoped the refactor — the provider plumbing's already there. Just two real wrinkles to handle in phases: serial CI runs in-process so it needs an in-process mining path, and a couple of tests hardcode account indices that'd collide on a shared node. So I'll land it incrementally, `--per-slot-node` kept as a fallback.

You can do it in a way if the URLs are not provided / not available, harness first spins up an in process node/discoveryServer & deploys the contracts and you use the URL and contracts given from that to pass forward to the rest of the harness setup. On the accounts, what problem are you explicitly hitting? Nonce collision?

momoshell:
> Are you observing WebRTC in tests?

**WS** — yup my bad while typing. But, I've just checked again and my conclusion still holds (all loopback WS, zero external traffic) — I just named the wrong transport.

> On the accounts, what problem are you explicitly hitting?

Yes, nonce collision. Concurrent processes on one shared node all sign from the same mnemonic accounts (peers + deployer), so txs collide on nonces (plus shared-balance drain). Already landed the fix: partition the mnemonic by slot — each concurrent slot gets a disjoint account slice keyed off the slot index the runner injects. Verified zero collisions across slots.

lukastanisic99:
> Yes, nonce collision. Concurrent processes on one shared node all sign from the same mnemonic accounts (peers + deployer), so txs collide on nonces (plus shared-balance drain). Already landed the fix: partition the mnemonic by slot — each concurrent slot gets a disjoint account slice keyed off the slot index the runner injects. Verified zero collisions across slots.

I think only the first 20 hardhat accounts have a balance - double check / keep that in mind.

Btw, I have this as a TODO in Trello, we have to figure out how to solve the nonce collision so signers can be reused - It's essentially a race condition where to independent transactions are created simultaneously and are assigned the same noce

momoshell:
> I think only the first 20 hardhat accounts have a balance - double check / keep that in mind.

True, by default (count 20). 
But I've just set count: 400, and I just verified all 400 are actually funded: 400 signers, accounts [0, 19, 20, 200, 399] all show 10000 ETH. 

So slots up to the pool max are funded — your warning is now covered 😄 
And good thing that you prompted me to double-check.

momoshell:
> See if you can split parts of the script into their files

Done — split it in 3a642ec

scripts/test-e2e-parallel.js
momoshell: Agreed on the principle, and the URL part already landed: in e32079a the runner injects HARDHAT_NODE_URL/PROVIDER_URL and runs the child with --network localhost, so the harness/SDK connect to the node purely by URL (and 661bd26 lets the harness reuse a deployment it's handed). 
They don't care where the node is, and fall back to in-process when no URL is given.

One important data point that bears on this: I measured the single-shared-node path, and a hardhat node caps at ~4–8 concurrent test processes before tx/event throughput collapses into timeouts (with interval mining, one node batches everyone's txs into a block every 2s). So "one node for everything" works great at low concurrency but won't carry our full parallelism on hardhat.

And for the devnet, that's actually where this point becomes the answer: the URL-based architecture is exactly the right foundation, and pointing it at a faster/real node (anvil or a devnet) — which doesn't have hardhat's single-node ceiling — is what makes one shared node viable at scale. 

So the plan now should be: 
 - keep the connect-by-URL design (done), 
 - unify the node-start script, and decide the node behind the URL (hardhat node-pool vs anvil/devnet) based on the throughput numbers. 
 
 I've parked the detail so we can pick the node strategy together.

lukastanisic99: We also have to measure gas, most of the operations are cheap and the contracts are deployed once, so they shouldn't really compete with other transactions. I'm surprised that only 4-8 concurrent tests can run on a node and tbh it worries me - it's not like a production node is 100x faster and 400-800 concurrent are very low numbers. What's the block gas limit on the hardhat node? Is it less than 30M gas? 

I think one of the problems is that within the SDK we hardcoded large gas limits like 5M gas for some calls even though they don't use that much gas, since ethers had a hard time estimating the gas consumption and the TXs were failing. So even if 5M is not used, the block accounts it as 5M out of the 30M gas limit, so if you have many such transactions you can only include 6 per block.. There's a Trello card on this 

Either way the URL path is the way to go, even if we need to have a 'pool of nodes' for testing, the scheduler can also assign the correct url, but I'm more interested into getting to the root cause, since everything was designed to be light weight on the blockchain node

scripts/test-e2e-parallel.js
momoshell: Fair on the 'source' bit now — that's just cosmetic leftover now, it feeds the footer/startup label so you can see why `sdkThread` came out on or off in a given run. 

Happy to drop it and just print the resolved booleans; the tracking isn't load-bearing.

momoshell: Fair on the 'source' bit — that's just cosmetic now, it feeds the footer/startup label so you can see why `sdkThread` came out on or off in a given run. 
Happy to drop it and just print the resolved booleans; the tracking isn't load-bearing.

Edit:
Dropped it in 0f2363b. Reads cleanly and behavior's unchanged

scripts/test-e2e-parallel.js
momoshell: Yes, these two parts here, both done.
> This file should not worry about...

Handled by the split (3a642ec): `scheduler.js` is pure scheduling now (no port/server code), and `getFreePort/startDiscoveryRegistry/startSlotNode` moved to `nodeInfra.js`. The entry file is just main() orchestration.

> We just need to pass urls

Done in fd9771b: the runner now reads PROVIDER_URL/HARDHAT_NODE_URL and LOCAL_DISCOVERY_REGISTRY_URL from env and, when present, passes them straight through to tasks and provisions nothing — no node/registry boot, no teardown of infra it doesn't own. 

Auto-provision stays only as the no-URL fallback. Verified: with HARDHAT_NODE_URL set, the runner uses it (boots nothing), tests pass against it, and it leaves the node running on exit. So the scheduler genuinely doesn't care local/remote/shared now.

This also sets up our devnet point: external infra behind a URL means we can point the runner at anvil or a real devnet (no hardhat single-node ceiling), which is what makes one shared node viable at scale.

scripts/test-e2e-parallel.js
momoshell: Agreed on the model — shared node/network is production, per-slot is just isolation, and "do tests stay stable when many run in parallel" is exactly the right thing to stress. 
Strong +1 on testing for it.

As I've said in the previous one, I measured the shared-node path, and there's an important nuance: at low concurrency (~4) it's clean, but at ~15+ concurrent it hangs — tests wait on txs/events and hit the 90s timeout. Turning sdk-thread off made it worse, and there were zero nonce/account collisions, so it's not the worker boundary or accounts — it's the one node under load.

The thing to pin down: is that hardhat's node saturating (it's single-threaded JS — serving ~15+ concurrent clients' RPC + event-polling + 2s-interval mining chokes it), or genuine protocol gas-starvation (txs actually competing for block inclusion)? Different responses — the first is a test-infra limit a real node wouldn't have; the second is the production signal you're describing and we'd want to fix in the protocol.

As per you most recent comment, gas-starvations is what I'm now looking at now.

scripts/test-e2e-parallel.js
momoshell: We're aligned here — the controller targets ~0.9 load/core (that "90% utilization"), and the sleeping-threads point is exactly why it's `loadavg` and not CPU%.

CPU% reads ~10% on the box precisely because the suite is wait-bound — peers idle on dispute timers, so most of those 6000 threads are blocked, not running. 

`loadavg` counts only runnable threads (on-CPU or ready-to-run), so it ignores the sleepers and rises only when threads actually contend for CPU. 
That's the signal we want; CPU% would fool us into massively over-scheduling straight into the starvation cliff. 

And we don't need to know how "expensive" a waiting thread is — while it waits it costs ~0 CPU and isn't in `loadavg`; the moment it wakes and competes, it shows up. So `loadavg/core` captures real demand without per-thread accounting.

The [2.03, 2.06, 2.12] over the core count is well under 0.9 → the controller scales up, which matches our "room to scale up" concern here. Hard ceiling stays the event-loop-delay watchdog (the 1s starvation trip) — `loadavg` is the soft target, the watchdog is the cliff, exactly the wiggle room you mention.

One honest caveat from measuring it: for this suite CPU load usually isn't the binding limit — even at max RAM-capped concurrency I saw only ~0.6/core, because each in-process worker carries its own hardhat EVM (~1.4 GB), so RAM caps concurrency before CPU saturates. 
So load/core is the right signal, but RAM (and, on a shared node, node throughput) is what we hit first. The shared-node / lighter-worker direction is what frees the CPU headroom to actually reach ~0.9.

scripts/test-e2e-parallel.js
momoshell: You nailed the root cause!
And I have it fixed and committed.
Block gas limit is hardhat's default 30M, and the SDK declares big per-tx limits — `OpenChannelNegotiationService` 5M, the dispute multicall in `StateManager` 10M, deploy 20M. At 30M/block that's only ~3–6 of those per block, and with one block every 2s the rest queue → the hangs.

Measured it (hardhat-gas-reporter, e2e dispute suite): 
- the declared limits are 3–8× over actual — open ~1.84M (declared 5M), 
- the dispute multicall ~0.5M avg / 1.2M max (10M), 
- dispute upload ~0.5M (DEFAULT_GAS_LIMIT 5M), 
- largest deploy ~6.9M (20M)

At a 30M block that's only ~3–6 of those per block, so with one block every 2s the rest queue → the hangs, exactly as you described.
Right-sized them in 3225d51 (→ 3M / 3M / 2.5M / 10M, generous headroom, no out-of-gas). This could be the production-relevant fix, since over-declaring reserves block gas a tx never actually uses.

One honest nuance from measuring: the right-size alone, on a realistic 30M block, only took the hangs from 12–14 → 5, not to 0. 
The runner crams ~15–32 concurrent games' txs into one node's 2s blocks, not production's load profile (prod spreads over time) so that residual is a test-concurrency artifact, not real gas-starvation. Bumped the test-node `blockGasLimit` (e2e-gated) in 5b3aebc to cover it.

Now with shared-node at 32 concurrent → 0 hangs, and 3 residual failures (all recovered on rerun):
1. reduce_invalid_state_transition_disputes... → onStateSnapshotUpdated: unknown snapshot while status=4
2. reject_dispute_when_auditing_data_is_partial... → REPLACEMENT_UNDERPRICED
3. InitHandshake ... RTT exceeds agreementTime → Event loop delay 5519ms exceeded threshold 1000ms

So the gas fix did eliminate the gas-starvation hangs (0 hangs — that part's solid), but these residuals seems to be a mix or sorts. 
One chronic reduced-fork test, one nonce-replacement race (looks like the signer-reuse issue from your Trello card, surfacing under load), and one event-loop-starvation from running 32-wide. 

One for your eyes: I sized 3225d51 to e2e-observed gas × margin, so worth a sanity check against production worst-case dispute sizes before we lean on those limits tightly.

momoshell:
Ok, here's the wrap-up and where we've landed now.

**The throughput wall was block-gas accounting** —  @lukastanisic99  you were right. The hangs weren't hardhat being RPC-bound; they were the SDK's over-declared per-tx gas limits not fitting the node's blocks under interval mining. Two fixes landed:
- 3225d51 right-sizes the SDK gas limits to measured values (via hardhat-gas-reporter): open ~1.84M, dispute multicall ~0.5M avg / 1.2M max, dispute upload ~0.5M, largest deploy ~6.9M. Limits are now 3M / 3M / 2.5M / 10M.
- 5b3aebc raises the test node's block gas limit to 1B, gated on E2E_INTERVAL_MINING so the realistic 30M stays for everything else.

With both, the gas-starvation hangs go to zero. Confirmed across a full-suite shared-node run, not a single out-of-gas in 66 initial failures.

**Cross-process nonce collisions** are fixed via account partitioning. Each parallel slot draws from a disjoint account slice (`SLOT_STRIDE = 10`, keyed off `E2E_SLOT_INDEX`), so two processes never share an account. Verified holding under load.

But, **shared-node-as-the-default** is blocked until we solve intra-process nonce issue.
I built the single-shared-node mode behind a `--shared-node` flag and ran the full 132-task suite against it at ~32-wide. 
It comes out 64 passing / 68 failing, and the dominant failure (37 of 66 initial) is `Replacement transaction underpriced ... to replace the existing transaction with nonce 0`. 
That's a peer firing a second transaction for nonce 0 before its first one mines inside the 2s interval window. 

It's not gas (zero OOG) and not partitioning (the peers use correct disjoint accounts); it's purely that race, blown up by concentrating ~32 concurrent tests' tx load onto one interval-mining node. The current default is that each child on its own lightly-loaded in-process node.

**So this was my decision**: the parallel default stays per-child in-process; shared-node stays opt-in behind the flag as the foundation for later. Once the signer-reuse nonce handling lands, re-running that validation should clear the way to make shared-node the default.

Also, while reviewing the gas right-size: the `2.5M DEFAULT_GAS_LIMIT` only caps `uploadDispute`. The heavier dispute paths (the fraud-proof + upload multicall, the calldata upload, and `applyDisputeFraudProofs`) carry no explicit limit and fall through to ethers estimation.
So worst-case disputes aren't actually pinned at 2.5M. That's reassuring for production sizing, but it's inconsistent and probably belongs in the same gas-sizing issue/task?

lukastanisic99:
> But, **shared-node-as-the-default** is blocked until we solve intra-process nonce issue. I built the single-shared-node mode behind a `--shared-node` flag and ran the full 132-task suite against it at ~32-wide. It comes out 64 passing / 68 failing, and the dominant failure (37 of 66 initial) is `Replacement transaction underpriced ... to replace the existing transaction with nonce 0`. That's a peer firing a second transaction for nonce 0 before its first one mines inside the 2s interval window.

This one is straightforward to solve, wrap the signer in a NonceManager during p2pSetup. 

Some conclusion on all of this:
- Users can have normal wallets as master keys/wallets, but SDK signers will be 'hot wallets' that the SDK controls/generates and 1 per concurrent channel/game. E.g. if players play multiple games in parallel, each SDK instance has its own distinct signer, even though it's linked to the same master key/wallet (identity) - this way we don't have nonce collision from parallel channels/games
- The hot wallet signers of each SDK instance are wrapped in a NonceManager, so they can issue multiple transactions without having to wait for each to be mined before. 
- Account abstraction is still an open topic, since we want the Master Wallet to sponsor the gas (or ideally the contract directly) without forcing the 'hot wallets' to be funded in production - the user shouldn't even know these hot wallets exists, they should only see their master wallet as their identity 


> Also, while reviewing the gas right-size: the `2.5M DEFAULT_GAS_LIMIT` only caps `uploadDispute`. The heavier dispute paths (the fraud-proof + upload multicall, the calldata upload, and `applyDisputeFraudProofs`) carry no explicit limit and fall through to ethers estimation. So worst-case disputes aren't actually pinned at 2.5M. That's reassuring for production sizing, but it's inconsistent and probably belongs in the same gas-sizing issue/task?

Ye this is strange that a more complex operation is estimated correctly, while the estimate fails on the simpler one. Either way, ok for now, but we'll come back to this, since this can be a real problem in production if only N transactions (N<100) can be included in a block, since an attacker can spam the on-chain node for 2,3 blocks and cause honest peers to miss their tx inclusion window. This slightly forces us to increase all `time windows`, to make this harder to pull off, but there's a trade-off, some parts will seem slower (reduced UX) - still in the best case everything is off-chain and we just have to make it `impractical` for an attacker to attempt any spam. 

Any other open questions? @momoshell  ping me when I should review - for now I've mostly followed along your comments

momoshell:
@lukastanisic99  just shipped latest changes.
**And the short version is**: the SDK-side fix with the suggested wrapper is in, it's the correct production fix, but it turned out the dominant races under shared-node at scale were a test-harness artifact, not the SDK — and that part resisted a clean fix. Details below.

What landed (029fcab): the `NonceManager` idea, done carefully. The fix wraps the runtime-host signer in a `ManagedNonceSigner` that serializes nonce assignment from an owned counter under a mutex. It's wired once in `startP2pRuntimeHost` (so it covers both the inline and the sdk-in-thread worker runtimes) and is also passed as the `StateManager` signer, so the dispute and error-retry paths use it too.

**The twist** — the residual races were the test harness, not the SDK. Once the SDK's own race was fixed, the remaining REPLACEMENT_UNDERPRICED failures were all in the adversarial tests, none in honest-flow. 
The harness's tampering actions (`DisputeTamperingActions`, `TransitionActions`) send real on-chain txs straight from a peer's account on the main thread (`channelManager.connect(peer.signer).uploadDispute(...)` etc.) — while that same peer's SDK is live and sending from the same account through the managed signer in its worker. Two uncoordinated nonce sources on two threads.

**Tried to coordinate them and it didn't pay off.** 
Three approaches, each traded one problem for a worse one: 
1. routing the harness send through the peer's worker stalled the worker's own block/handshake processing; 
2. reserving a coordinated nonce with an explicit `gasLimit` made reverting negative-tests mine-and-revert, and a mined revert loses the custom-error data, so those assertions broke; 
3. adding a `staticCall` preflight to recover the error reintroduced latency that blew the dispute time-windows. The shared-node @32 regime is also high-variance, so small gains aren't reliably measurable. 

So reverted all of it.

**Where that leaves shared-node-as-default.** 
Still deferred, now for a precise reason and it's blocked by: 
- this test-harness double-send (a test-only issue is that it recovers on rerun and doesn't affect production), 
- the chronic dispute-flaky cluster, and
- the single-node load ceiling at 32-wide. 

None of these are the SDK's fault, and the SDK nonce fix is a genuine standalone improvement regardless. 
A clean harness fix almost certainly needs a small test restructure (don't run the attacker peer's SDK concurrently with the injection) rather than nonce-juggling. 
Happy to pick that up as a separate piece.

Net for this PR: gas right-sizing (3225d51), the e2e block-limit bump (5b3aebc), the rerun log-name crash fix (f708a9a), and now the signer nonce fix (029fcab). Shared-node stays opt-in behind the flag.

lukastanisic99:
Ye I was aware of this that we reconstruct the same deterministic signer in multiple threads and that this could happen, but thought "only the sdk thread actually interacts with the node", but forgot about harness.. 

This is annoying since there needs to be a shared process that does the accounting (nonce assignment) for all of this. 

Probably best to only construct the signer in the SDK thread, that's the only place where signing and nonce management happens.

Nonce is only relevant on-chain, locally we don't use any nonce for our local signer, so only when firing the request we need to assign the nonce.

So if you take a look we have a `few` signer implementations that are just polymorphic substitutions for the base signer and they mostly just intercept messages/tx and route them somewhere else. So the easiest solution is to add 1 more of these signers (e.g. `ProxySigner`) for it to route raw tx data to the single thread that actually has the on-chain signer with the secret, and then that signer repackaged the transaction with the correct nonce and sends it out, it gets the response and returns it back to the proxy signer which returns it normally to the calling context. Take a look at the signers we have they all do something like this

That should be a holistic fix to all of this

momoshell:
Closing the loop on the shared-node nonce work.

**Landed**: the SDK concurrent-flow nonce fix (`ManagedNonceSigner`, 029fcab), plus the earlier gas right-size (3225d51), e2e block-limit bump (5b3aebc), and rerun log-name crash fix (f708a9a). The nonce race is solved, where it counts, in production.

**On the ProxySigner**: built it as you described @lukastanisic99  — single host-owned signer, main-thread proxy forwards the raw tx, host assigns the nonce and submits, and the caller `awaits wait()` (not the worker, so the worker never stalls); reverts surface at the host's `estimateGas` so the negative tests still decode. 
It works — nonce 0 races go to zero at every concurrency. 
But it doesn't pay off: routing the send through the worker adds dispute-EventBarrier latency that scales with load — roughly cancels the nonce benefit at 8-wide (88→91 passing) and is a clear regression at 32-wide (72→36, ~+69 barrier timeouts). 
The recover-on-rerun nonce race is cheaper, so I reverted it.

**Why shared-node-default is staying deferred:** I ran the full baseline curve (direct sends) to see whether fewer slots per node (a small node pool) would help — it doesn't. 
The `EventBarrier` timeouts are flat (~11–23) across 4/8/16/32-wide with no load trend (the most were actually at 4-wide), so they're a shared-node-mode artifact, not load: these dispute tests wait on tight 5–8s windows for peers to react, and sharing one node's event stream + 2s interval mining jitters that enough to blow them regardless of concurrency. (Per-child mode is green precisely because each test gets its own isolated node.) 
Plus the chronic reduced-fork cluster (~10–15), a separate all-modes flaky.

So: nonces solved; routing and concurrency-tuning both ruled out by data. What's left for shared-node-as-default is dispute-timing-under-a-shared-node + the chronic cluster — neither reachable by the nonce work, both their own investigation. 

Parking it there: per-child stays the default (green), shared-node stays opt-in behind the flag.

PR #361, opened 2026-06-23
fix(handshake): domain-separate init-handshake signatures to close signing oracle

sh3ll3x3c:
## Summary

The init-handshake responder signed the peer-supplied 32-byte `challengeHash` verbatim with the participant signer (`p2pSigner.signMessage(getBytes(challengeHash))`). Blocks and protocol messages are EIP-191 signatures over a **raw 32-byte keccak hash** (`SignatureUtils`, `Block.hash`; verified on-chain by `UtilityFacet`). So a pre-authentication peer could send `challengeHash = keccak256(encodedBlock)` and receive a signature that is a **valid block signature** attributable to the victim — i.e. the handshake was a signing oracle.

### Fix
- Sign/verify a **domain-separated string** `peer3:init-handshake:v1:<hexlify(challengeHash)>` instead of the raw hash. The EIP-191 preimage length and content then differ from any 32-byte hash the protocol signs, so a handshake signature can never be replayed as a block/dispute signature. The helper lives on the (non-routable) `InitHandshakeService` as a static so both sign and verify sides derive an identical canonical message.
- **Reject malformed requests before signing**: a non-32-byte challenge or a non-finite `time` (a `NaN` time otherwise slips past the `Math.abs(...) > agreementTime` skew check, since any comparison with `NaN` is `false`).
- Updated the test harness responder (`RPCActions.deliverResponse`) to sign the domain-separated message so it mirrors production.

No new oracle is introduced: the only other `p2pSigner.signMessage` path signs a locally-built raw 32-byte block hash, and no protocol message consumes a `peer3:init-handshake:v1:…` string.

## Test Plan
- New unit test `test/rpc/initHandshake/InitHandshakeChallenge.test.ts`:
  - round-trips a domain-separated signature back to the signer;
  - asserts a handshake signature does **not** verify over the raw challenge hash (the anti-collision / anti-oracle property);
  - asserts message derivation is canonical regardless of hash casing.
- `test/e2e/E2E-InitHandshake.test.ts` — all 7 E2E handshake tests pass (completion, WebRTC upgrade, time/RTT validation, unsolicited-response rejection).

### DIFF
--- a/src/rpc/services/initHandshake/InitHandshakeRpcMethods.ts
+++ b/src/rpc/services/initHandshake/InitHandshakeRpcMethods.ts
@@ -27,6 +27,24 @@ class InitHandshakeRpcMethods extends ARpcMethods {
                 messageTime: time
             }
         );
+        // Reject malformed inputs before signing. A non-32-byte challenge or a
+        // non-numeric time (NaN slips past the skew check below) would let a
+        // peer steer what gets signed.
+        if (!ethers.isHexString(challengeHash, 32) || !Number.isFinite(time)) {
+            LoggerUtils.logInitHandshakeMessage(
+                this.service.logger,
+                this.senderTransport,
+                {
+                    direction: "local",
+                    message: "rejected",
+                    challengeHash,
+                    messageTime: time,
+                    reason: "malformed handshake request (challenge/time)"
+                }
+            );
+            this.p2pManager.disconnectConnection(this.senderTransport);
+            return;
+        }
         const timeDifference = time - localTime;
         if (Math.abs(timeDifference) > agreementTime) {
             LoggerUtils.logInitHandshakeMessage(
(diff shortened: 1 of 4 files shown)


lukastanisic99:
Edit: really good finding

sh3ll3x3c:
The randomness assumption holds only for an *honest* requester. This is the pre-auth handshake: the **responder** signs before it knows who the requester is (identity is established by the requester verifying the response), so it will sign for any peer that connects. An attacker as the requester sends `challengeHash = keccak256(encodedBlock)` — indistinguishable from a random challenge — and gets back the victim's signature over it. Because blocks are signed as `signMessage(getBytes(keccak256(encodedBlock)))` and verified on-chain under the same EIP-191 prefix, that handshake signature *is* a valid block signature attributable to the victim (feeds the double-sign slash).

The domain tag keeps the random-challenge / RTT / clock-comparison purposes intact and only stops the signature from doubling as a block signature; a Schnorr/Fiat-Shamir PoK would close it too, since it isn't a reusable signature over attacker-chosen bytes. Happy to fold this into the request/response refactor, but the separation is needed independently of it.

lukastanisic99 (approved):

PR #363, opened 2026-06-23
fix(validation): don't treat a posted block calldata commitment as unavailable

sh3ll3x3c:
## Summary

`tryFetchOnChainBlockAndScheduleValidation` confirmed a block-calldata commitment exists on-chain (`getBlockCallDataCommitment().found === true`) and then called `fetchBlockCommitmentCalldata` to load the `BlockCalldataPosted` event. That fetch scanned only a **bounded recent window** (`latestBlock - blocksToLookBack`). If the event was older than the window, it returned `undefined` and the status collapsed to `NOT_FOUND`.

`NOT_FOUND` is **not** deferrable (`shouldDeferCurrentValidation` only deferred `SCHEDULED`/`ALREADY_SCHEDULED`), so the caller stopped waiting for the objective on-chain timestamp and fell through to the fraud-proof branch — raising a **wrongful timeout / `InvalidTimestamp` dispute** against a block that was genuinely posted (the disputer can then be slashed).

### Fix
- **Scan the full history** when resolving the commitment's event. The commitment is unique and both filter topics (`channelId`, commitment) are indexed, so `queryFilter(filter, 0, "latest")` resolves to a single event. A bounded recent window was the root cause of the miss.
- Add a deferrable **`UNAVAILABLE`** status: when a commitment provably exists but its event still can't be fetched this attempt (e.g. transient RPC error), return `UNAVAILABLE` (now part of `shouldDeferCurrentValidation`) so the caller queues + retries instead of concluding "never posted." `found === false` still correctly returns `NOT_FOUND`.
- Removed the now-unused `timeConfig`/`Clock` scan-window math (and the corresponding constructor parameter).

The new enum value is appended (existing numeric values unchanged), and no `switch` consumes the enum — both `ValidationService` call sites and the two `StateManager` retry sites branch only through `shouldDeferCurrentValidation`, so they now defer correctly.

## Test Plan
- New unit tests (`test/stateManager/BlockDataAvailabilityService.test.ts`):
  - commitment found on-chain but event not fetched → `UNAVAILABLE` and deferrable;
  - no commitment → `NOT_FOUND` and not deferrable;
  - commitment + event found → `SCHEDULED` and validation is scheduled.
- Full non-e2e unit suite passes (259 + 3 new). The one unrelated `UniversalDeployment` parallel-deploy nonce flake passes in isolation.

### DIFF
--- a/src/stateManager/EventSyncService.ts
+++ b/src/stateManager/EventSyncService.ts
@@ -9,22 +9,35 @@ import {
     Address,
     BlockCalldata,
     BlockHeight,
+    BlockNumber,
     ChannelId,
     ForkId,
-    Hash
+    Hash,
+    Timestamp
 } from "@/types/types";
-import { convertEthersValue, DetachedPromises, hash, Logger } from "@/utils";
+import {
+    convertEthersValue,
+    correctBlockScanStart,
+    DetachedPromises,
+    estimateBlockScanRange,
+    estimateBlocksForSeconds,
+    hash,
+    Logger
+} from "@/utils";
 
 type BlockState = { pending: number; complete: boolean; failed: boolean };
 type OnChainBlockValidationKey = string;
 type EventKey = string;
 type ChannelKey = string;
-type BlockNumber = number;
 type NormalizedDisputeCommitment = string;
 type EventPromise = Promise<void>;
 export type BlockCalldataRecoveryResult = {
     blockCalldata?: BlockCalldata;
     validationScheduled: boolean;
+    // The chain proves a commitment exists for this slot. Without
+    // `blockCalldata` that means "posted, but not recovered yet" - the caller
+    // retries; it is never proof the block was never posted.
+    commitmentFound: boolean;
 };
 type BlockStates = Map<BlockNumber, BlockState>;
 type StateChannelManagerEventName = keyof StateChannelManagerProxy["filters"];
@@ -147,10 +160,14 @@ export default class EventSyncService {
         return promise;
     }
 
+    // `anchorTimestamp` is the block's validity-window anchor (the previous
+    // block's or snapshot's relevant timestamp) - the log scan is bounded around
+    // it instead of around the chain head.
     async tryRecoverBlockCalldataAndScheduleValidation(
         forkId: ForkId,
         blockHeight: BlockHeight,
-        blockAuthor: Address
+        blockAuthor: Address,
+        anchorTimestamp: Timestamp
     ): Promise<BlockCalldataRecoveryResult> {
         const validationKey = `${String(forkId)}:${blockHeight}:${String(blockAuthor).toLowerCase()}`;
         const pending = this.pendingOnChainBlockValidations.get(validationKey);
@@ -162,7 +179,8 @@ export default class EventSyncService {
                     blockHeight,
                     blockAuthor
                 ),
-                validationScheduled: false
+                validationScheduled: false,
+                commitmentFound: true
             };
         }
 
@@ -170,7 +188,8 @@ export default class EventSyncService {
             validationKey,
             forkId,
             blockHeight,
-            blockAuthor
+            blockAuthor,
+            anchorTimestamp
         ).finally(() => {
             this.pendingOnChainBlockValidations.delete(validationKey);
         });
@@ -182,8 +201,12 @@ export default class EventSyncService {
         validationKey: OnChainBlockValidationKey,
         forkId: ForkId,
         blockHeight: BlockHeight,
-        blockAuthor: Address
+        blockAuthor: Address,
+        anchorTimestamp: Timestamp
     ): Promise<BlockCalldataRecoveryResult> {
+        // Survives the catch below: once the chain has proven the commitment, a
+        // later failure must not read back as "never posted".
+        let commitmentFound = false;
         try {
             const commitmentResult =
                 await this.stateChannelManagerContract.getBlockCallDataCommitment(
@@ -193,8 +216,9 @@ export default class EventSyncService {
                     blockAuthor
                 );
             if (!commitmentResult.found) {
-                return { validationScheduled: false };
+                return { validationScheduled: false, commitmentFound: false };
             }
+            commitmentFound = true;
             const commitment = commitmentResult.blockCalldataCommitment;
             const existing = this.storage.blockCalldata.getBlockCalldata(
                 forkId,
@@ -203,19 +227,10 @@ export default class EventSyncService {
             );
             let validationScheduled = false;
             if (!existing) {
-                const latest = await this.getProvider().getBlockNumber();
-                const fallback = this.getBlockFallbackFrom(latest, blockHeight);
-                const cursor = this.storage.eventSync.getLatestProcessedBlock(
-                    this.channelId
-                );
-                const fromBlock = Math.min(cursor ?? fallback, fallback);
-                const logs = await this.stateChannelManagerContract.queryFilter(
-                    this.stateChannelManagerContract.filters.BlockCalldataPosted(
-                        this.channelId,
-                        commitment
-                    ),
-                    fromBlock,
-                    latest
+                const logs = await this.queryBlockCalldataPostedLogs(
+                    commitment,
+                    blockHeight,
+                    anchorTimestamp
                 );
                 const scheduledChannelId = this.channelId;
                 // onBlockCalldataPosted stores before its first await, so the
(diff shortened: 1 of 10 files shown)


src/stateManager/BlockDataAvailabilityService.ts
```diff
@@ -81,8 +80,13 @@ export default class BlockDataAvailabilityService {
             );
 
             if (!fetchedCalldata) {
+                // The commitment provably exists (found === true) but we could
+                // not fetch its calldata event this attempt. Treat as deferrable
+                // so the caller retries, NOT as NOT_FOUND — concluding "never
+                // posted" here would skip the objective on-chain timestamp and
+                // can trigger a wrongful timeout/InvalidTimestamp dispute.
```
lukastanisic99: I do understand why gpt/claude flagged this, but they don't have the whole picture:

We deliberately bounder how far back we 'search/query' the node, since looking up the entire history is not practically feasible + not needed. The calldata has to be posted in a 'time window' for it to be valid. We only care to query in that window and if it's not in that window we know it's invalid and can proceed with timing out the other peer, knowing that if the peer tried to produce a challenge they would not be able to prove to the chain that the calldata was committed during the window and as such they wouldn't be able to slash us. 

If you can disprove this claim with a test, I'd take a closer look

lukastanisic99 (changes requested):

src/stateManager/BlockDataAvailabilityService.ts
sh3ll3x3c: Fair point that genuinely-late calldata is safe to time out — I confirmed the on-chain backstop you're relying on: `_handleTimeoutCalldataPosted` returns `_invalid()` for `onChainTimestamp > maxValidTimestamp` (`DisputeFraudProofFacet.sol:504-508`), so a late post can't defeat the timeout. No disagreement there.

The gap is that the safety argument quietly assumes **"not in the look-back window" == "posted outside the validity window."** Those are two *different* windows with *different anchors*:

- **Validity window** — `[previousTimestamp, previousTimestamp + p2p + agreement + chainFallback]`, anchored to the *previous block's timestamp* (when the block was due). Same formula off-chain (`ValidationService.getOnChainPostTiming` → `maxAllowedTimestamp`) and on-chain (`_handleTimeoutCalldataPosted` → `maxValidTimestamp`).
- **Scan window** — `[latestBlock - ceil(maxTime/avgBlockTime)*2, latest]`, anchored to the *current chain head* (when validation runs).

They only line up when validation runs ~immediately at the deadline. Whenever it runs late relative to the post — node catch-up after a restart, a confirmation delivered late after a partition heals, a dispute flow re-validating older blocks — an **on-time** commitment sits *outside* the scan window but still *inside* the validity window. `getBlockCallDataCommitment` (a direct storage read, not windowed) still returns `found === true`, so the data is provably on-chain; we just fail to fetch the event and collapse to the non-deferrable `NOT_FOUND`. The caller then raises a (forced) timeout — and because the post was on time, the peer's `TimeoutCalldataPosted` proof is `_valid(dispute.input.disputer)` (`:563`), `killDispute` fires (`applyDisputeFraudProofs:21-25`), and **we** get slashed. The forced path even skips the submission-time race check that would otherwise revert (`DisputeManagerFacet.sol:120-130` only runs for non-forced timeouts), so nothing upstream stops it.

Test as requested — `test/stateManager/BlockDataAvailabilityService.test.ts`, commit 3b3e4b78, describe `"… bounded look-back self-slash regression"`:

- `"an ON-TIME commitment can fall OUTSIDE the legacy look-back window …"` — the *same* commitment is simultaneously (a) on-time by the exact `maxValidTimestamp` formula the chain enforces and (b) excluded by the legacy `latestBlock - blocksToLookBack` window (a range-honoring `eth_getLogs` returns `[]`). So "outside the window" ≠ "invalid," and the peer **can** prove it to the chain and slash us.
- `"the fix keeps the on-time commitment retrievable …"` — scanning from genesis recovers it (`SCHEDULED`), never a non-deferrable `NOT_FOUND`.

On the practicality concern you raised: I agree scanning full history isn't free (RPC range caps, grows with the chain). If you'd rather keep it bounded, anchoring the scan to the block's own validity deadline (`previousTimestamp + maxTime`, converted to block numbers) instead of `latestBlock` is bounded *and* correct regardless of when validation runs — happy to switch to that. The `UNAVAILABLE`-defer-on-transient-RPC half holds either way. I can also add a Foundry test driving `_handleTimeoutCalldataPosted` end-to-end if you'd like the on-chain leg executed rather than asserted via the (identical) formula.

lukastanisic99: > On the practicality concern you raised: I agree scanning full history isn't free (RPC range caps, grows with the chain). If you'd rather keep it bounded, anchoring the scan to the block's own validity deadline (`previousTimestamp + maxTime`, converted to block numbers) instead of `latestBlock` is bounded _and_ correct regardless of when validation runs — happy to switch to that. The `UNAVAILABLE`-defer-on-transient-RPC half holds either way. I can also add a Foundry test driving `_handleTimeoutCalldataPosted` end-to-end if you'd like the on-chain leg executed rather than asserted via the (identical) formula.

Yes, you're right that at the moment `scan window` != `validity window` if the head advances too much. The scan window should scan enough to include the full validity window + some extra blocks as a buffer, so we can form a decisive action after the call (accept or reject/dispute). Practically all of the scanning happens up to 1 min later, so no one will scan blocks an hour ago, so all the limits are within reasonable RPC usage limits.

Technically you just need to transform [lowTimestamp,highTimestamp] -> [lowBlockHeight,highBlockHeight] and scan that (plus add some extra blocks on both side just to account for some error, no reason to scan from/to latest

test/stateManager/BlockDataAvailabilityService.test.ts
```diff
@@ -0,0 +1,245 @@
...
+    const contract = {
+        getBlockCallDataCommitment: async () => ({
+            found: overrides.found ?? false,
+            blockCalldataCommitment: COMMITMENT
+        }),
+        filters: {
+            BlockCalldataPosted: () => ({})
+        },
+        queryFilter: async () => overrides.logs ?? []
+    } as unknown as StateChannelManagerProxy;
```
lukastanisic99: Claude/GPT "Remember to use harness for unit tests too. Teleport to a scenario, test what you're interested in isolation, but without mocks or junk data. Full contract and all components connected. Update AGENTS.md"

test/stateManager/BlockDataAvailabilityService.test.ts
```diff
@@ -121,3 +121,125 @@ describe("BlockDataAvailabilityService - on-chain calldata availability", functi
...
+            0 as BlockHeight,
+            AUTHOR
+        );
+
+        // Scanning from genesis finds the on-time event the bounded window missed,
+        // so the block is scheduled for validation instead of being declared
+        // "never posted". A transient fetch failure would yield the deferrable
+        // UNAVAILABLE — never the non-deferrable NOT_FOUND that arms the timeout.
+        expect(status).to.equal(OnChainBlockStatus.SCHEDULED);
+        expect(status).to.not.equal(OnChainBlockStatus.NOT_FOUND);
```
lukastanisic99: same comment here (-||-) for unit tests we should also use harness (teleport + test)

test/stateManager/BlockDataAvailabilityService.test.ts
sh3ll3x3c: Migrated to a harness e2e in a9b7eeb5 (`test/e2e/E2E-OnChainBlockAvailability.test.ts`): it posts real block calldata on-chain and drives the live `BlockDataAvailabilityService` against the real deployed `StateChannelManagerProxy` — no hand-rolled contract/queryFilter mock. Covers SCHEDULED (a far-back post stays retrievable after the head advances past any bounded window), UNAVAILABLE (a transient empty fetch defers instead of NOT_FOUND), and NOT_FOUND (never posted).

test/stateManager/BlockDataAvailabilityService.test.ts
sh3ll3x3c: Same here — the whole mocked file is replaced by the harness e2e in a9b7eeb5.

src/stateManager/BlockDataAvailabilityService.ts
sh3ll3x3c: Done in `1efe0f1b`. Bounded exactly as you suggested — the scan is anchored to the block's validity window instead of `[0, latest]`:

- Window: `[previousTimestamp, previousTimestamp + p2pTime + agreementTime + chainFallbackTime]`, widened symmetrically (the anchor handed in can be the *previous* block's timestamp, a few seconds off).
- `[lowTimestamp, highTimestamp] → [fromBlock, toBlock]` by **binary-searching** the chain for the bracketing heights (+ a small block buffer). I started with an average-block-time estimate as you described, but it drifts under irregular block spacing — the harness mines with time jumps and it missed by hundreds of blocks — so the exact binary search is what keeps it correct.
- `previousTimestamp` is threaded from each call site (ValidationService + the timeout path in `StateManager`), where it's the trusted validity anchor already in scope.

Safety net: the commitment-found check is a direct storage read, decoupled from the event scan, so a window miss degrades to `UNAVAILABLE` (deferrable retry) — never a wrongful `NOT_FOUND`/timeout.

The harness e2e still goes `SCHEDULED` after mining 1024 blocks past the post (a head-anchored window would miss it), while `NOT_FOUND` stays reserved for a genuinely-never-posted commitment.

src/stateManager/BlockDataAvailabilityService.ts
```diff
@@ -77,12 +85,18 @@ export default class BlockDataAvailabilityService {
...
                 this.pendingOnChainBlockValidationKeys.delete(validationKey);
-                return OnChainBlockStatus.NOT_FOUND;
+                return OnChainBlockStatus.UNAVAILABLE;
```
lukastanisic99: The problem with this is, it's unlikely that the calldata will ever become available, so if this triggers it would be an an infinite retry loop. The only way out is to either:
1) Abort the channel (start forceful exit)
2) Switch RPC providers and try again, but once your provider pool is depleted, you'd have to fallback to 1) 

Add a TODO in the code or a Trello card for this, so we come back

src/stateManager/BlockDataAvailabilityService.ts
```diff
@@ -187,4 +197,81 @@ export default class BlockDataAvailabilityService {
...
+    }
+
+    // Lowest block height whose timestamp is >= targetTimestamp (or headBlock if
+    // none), via binary search over [0, headBlock].
+    private async findFirstBlockAtOrAfter(
+        provider: ethers.Provider,
+        targetTimestamp: Timestamp,
+        headBlock: number
+    ): Promise<number> {
+        let low = 0;
```
lukastanisic99: Ye I don't think binary search from genesis makes sense here since we already know that the range is in the latest few blocks. With more or less 100% probability you can 'gestimate' the range from knowing the average block time (which you have available in `Clock.ts`), so you can do this in 99% of cases in O(1) (1 querry to the RPC node), while with binary search you'd have O(logN) where N is the latest block height which, which if you take a look at Polygon has 89M blocks which would be 25 RPC calls which is slow, wastefull + most RPC nodes just have recent history (are not archive nodes).

lukastanisic99:
@sh3ll3x3c see how this fits in the latest stuff - maybe only the bounded 'guesstimate' query is needed out of this entire thing, since there were changes to this also

src/stateManager/BlockDataAvailabilityService.ts
sh3ll3x3c: Done in `ab6d19dc` — the binary search is gone. `estimateBlockScanRange` (`src/utils/blockScanWindow.ts`) converts the validity window to `[fromBlock, toBlock]` using `Clock.getAverageOnChainBlockTime()`: one head read, no search, and a symmetric buffer of a full window's worth of blocks on each side.

Irregular block spacing can still push the estimate off, so a scan that finds nothing does one correction pass: read the estimated start block, and if its timestamp is outside the window, re-derive the rate from that block to the head and re-scan once — still O(1) RPC calls, no archive-depth walk. A miss after that stays deferrable (the storage read still proves the commitment exists), never "never posted".

`estimateBlocksForSeconds` is now the single owner of the span math that `ensureDisputesProcessed` uses as well.

src/stateManager/BlockDataAvailabilityService.ts
sh3ll3x3c: Added the TODO on the retry path — `StateManager.deferTimeoutForUnrecoveredCalldata` (`src/stateManager/StateManager.ts:1859`): the retry currently has no exit; the escape hatches to implement are switching RPC provider, and once the provider pool is depleted, falling back to channel abort (forceful exit).

src/stateManager/BlockDataAvailabilityService.ts
sh3ll3x3c: Update after rebasing on latest `dispute`: the binary-search conversion described above is replaced by the average-block-time estimate (see the thread on the range conversion below). The validity-window anchoring and the deferrable-miss safety net carry over unchanged.

sh3ll3x3c:
Rebased on latest `dispute` (`610f60ea`). You were right — upstream absorbed most of this PR: `BlockDataAvailabilityService` is gone, calldata recovery now lives in `EventSyncService.tryRecoverBlockCalldataAndScheduleValidation`, and the storage-read commitment check plus a first average-block-time pass were already there. What remains is a single commit (`ab6d19dc`):

- the log scan is anchored to the block's validity window instead of the chain head (the actual self-slash fix),
- the bounded guesstimate range conversion via `Clock.getAverageOnChainBlockTime()` (details in the review thread),
- posted-vs-never-posted split: a commitment the chain proves exists but the scan missed stays retryable/deferrable; only a provably absent slot reads as never posted,
- the TODO on the retry path.

The window math is pure and unit-tested (`test/unit/blockScanWindow.test.ts`); three harness tests cover the `EventSyncService` integration.

Note on the red check: CI currently cannot run any harness test — every harness session dies at setup with `TypeError: this.swarm.removeAllListeners is not a function` (`Holepunch.setup` via `P2pRuntimeHost.buildRuntime`). The contracts-only #414 run fails identically, so it's base breakage (plus the pre-existing `LogUploader` failures), not this diff.

Flagged while in there, deliberately out of scope: `ValidationService.objectiveInvalidTimestampDetected` can still build a fraud proof from calldata that is committed on-chain but not yet recovered. Handling that needs a new deviation on `AValidationStrategy` implemented across all strategies — separate card.

PR #364, opened 2026-06-23
fix(open-channel): verify proposed terms match negotiation before co-signing

sh3ll3x3c:
## Summary

The `openProposal` receiver in `OpenChannelNegotiationService` verified only that the lower-sorting participant signed the peer-supplied `OpenChannel` bytes, then co-signed those same bytes with the local participant key and submitted them on-chain. A valid lower signature only proves the peer signed **those** bytes — not that they match what was locally negotiated. A lower-sorting handshaked peer could therefore make the local signer authorize attacker-chosen `channelId` / `participants` / `balances` / `data` / `deadline`.

(Impact is bounded — `open()` is non-payable and balances are funded from each participant's own on-chain deposit, so there's no fund redirection — but the victim could still be co-signed into an unwanted channel/terms.)

### Fix
Before signing, reconstruct the expected `OpenChannel` from local negotiation state (`getParticipantsAndBalances(peer)` + `getChannelId()`, `isAtomic: true`, empty `data`) and reject (disconnect + blacklist) if any security-sensitive field differs:
- `channelId`, `participants` (and order), each balance `amount` + `data`, `isAtomic`, `data`, and a sane `deadline` window.

Also:
- Require amounts to have actually been negotiated (`theirAmount` set) before co-signing, so the reconstructed balances can't silently fall back to defaults.
- Extracted the proposer's deadline offset into a shared `OPEN_CHANNEL_DEADLINE_SECONDS` constant used on both the build and validate sides; the receiver's deadline upper bound is generous (2x the offset) so modest clock skew can't reject a legitimate proposal.

The comparison lives in a pure helper (`getOpenChannelProposalMismatch`) for direct unit testing; the honest flow's reconstruction matches `maybeProgress` field-for-field.

## Test Plan
- New unit tests (`test/rpc/openChannelNegotiation/OpenChannelProposal.test.ts`): a matching proposal is accepted; tampered balance amount, balance data, participant set, channelId, `isAtomic`, `data`, and out-of-range deadlines are each rejected; address casing is normalized.

### DIFF
--- a/src/rpc/services/openChannelNegotiation/OpenChannelNegotiationService.ts
+++ b/src/rpc/services/openChannelNegotiation/OpenChannelNegotiationService.ts
@@ -20,7 +20,9 @@ import OpenChannelNegotiationRpcMethods, {
 import {
     DEFAULT_JOIN_AMOUNT,
     NEGOTIATION_TIMEOUT_MS,
+    OPEN_CHANNEL_DEADLINE_SECONDS,
     compareAddresses,
+    getOpenChannelProposalMismatch,
     type Address
 } from "./OpenChannelNegotiationHelpers";
 
@@ -181,7 +183,8 @@ export default class OpenChannelNegotiationService extends ARpcService<
         if (!isLower) return;
 
         if (haveAmounts && !this.state.proposalSent) {
-            const deadlineTimestamp = Clock.getTimeInSeconds() + 60;
+            const deadlineTimestamp =
+                Clock.getTimeInSeconds() + OPEN_CHANNEL_DEADLINE_SECONDS;
             const openChannel: OpenChannelStruct = {
                 channelId,
                 participants,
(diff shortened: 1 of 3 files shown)


src/rpc/services/openChannelNegotiation/OpenChannelNegotiationService.ts
```diff
@@ -252,6 +256,44 @@ export default class OpenChannelNegotiationService extends ARpcService<
...
+            }
+        );
+        if (mismatch) {
+            this.logger.warn(
+                `openProposal - proposed terms do not match negotiation; rejecting (${mismatch})`
+            );
+            this.p2pManager.disconnectAndBlacklistPeerByEvmAddress(peer);
+            this.resetNegotiation(`openProposal terms mismatch: ${mismatch}`);
+            return;
+        }
```
lukastanisic99, line 295: We'll probably redesign this service since it was kinda hacked together anyway so we can just easily open channels + with the request/response patter it will be simpler anyway. Either way the concern here is valid.

lukastanisic99 (approved):

PR #365, opened 2026-06-23
fix(rpc): bound the pre-handshake RPC queue

sh3ll3x3c:
## Summary

`HandshakeCompletedGuard` buffers guarded RPCs received while a handshake is still negotiating, in a per-transport `RpcQueue`. `enqueue` was a bare `push` with no bound, so an unauthenticated (still-negotiating) peer could grow the queue without limit during the negotiation window.

### Fix
Bound the queue by **count** (`MAX_PREAUTH_QUEUED_RPCS = 64`) and **total bytes** (`MAX_PREAUTH_QUEUED_BYTES = 1_000_000`, sized via the existing `serializeRpc`). `enqueue` now returns `false` on overflow, and the guard treats a full queue as abuse — it clears the queue and disconnects/blacklists the peer. `clear`/`drain` reset the byte counter.

The window was already time-bounded (`2 × agreementTime`) and per-transport, so this is defense-in-depth, but it removes the unbounded-growth primitive. Guarded RPCs reaching the queue always originate from `deserializeRpc` (JSON), so `serializeRpc` can't hit a bigint here.

## Test Plan
- New unit tests (`test/rpc/guards/RpcQueue.test.ts`): count cap, byte cap, capacity freed after `clear()` and after `drain()`, and the default count bound.

### DIFF
--- a/src/rpc/guards/HandshakeCompletedGuard.ts
+++ b/src/rpc/guards/HandshakeCompletedGuard.ts
@@ -1,4 +1,4 @@
-import Rpc from "@/rpc/Rpc";
+import Rpc, { serializeRpc } from "@/rpc/Rpc";
 import ATransport from "@/transport/ATransport";
 import { AGuard } from "@/rpc/guards/AGuard";
 import ARpcService from "@/rpc/ARpcService";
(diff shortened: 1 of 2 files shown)


src/rpc/guards/HandshakeCompletedGuard.ts
```diff
@@ -8,12 +8,36 @@ export interface HandshakeCompletedGuardOptions {
     onFailure?: (rpc: Rpc, transport: ATransport) => void;
 }
 
-class RpcQueue {
+// Bound the per-transport queue of RPCs buffered while a handshake is still
+// negotiating, so an unauthenticated peer can't grow it without limit.
+export const MAX_PREAUTH_QUEUED_RPCS = 64;
+export const MAX_PREAUTH_QUEUED_BYTES = 1_000_000;
```
lukastanisic99, line 14: In the TODOs we have a general rate limit left for implementation, so peers can't abuse other peers - left that for the end, so we first fix all the other obvious bugs without tripping over limits. That single rate limit should cover/protect this and all other RPC services, so no need to do this on a case-by-case basis

sh3ll3x3c: In that case I'll be closing this PR. Proper rate limit implementation will be done top-level.

PR #370, opened 2026-06-28
tryDecode solidity, general abort, fix unknown snapshot and other stuff

lukastanisic99:
I didn't do here `ignore errors from byzantine peers`, since I didn't find a clean way of doing that and it's not that critical at the moment - will find a clean way to do it in some follow up PR, added a Trello card for it. 

In some cases I left redundant checks in the code to preserve tests - no harm

### DIFF
--- a/contracts/V1/StateChannelDiamondProxy/StateProofFacet.sol
+++ b/contracts/V1/StateChannelDiamondProxy/StateProofFacet.sol
@@ -16,7 +16,10 @@ contract StateProofFacet is StateChannelCommon {
             return false;
         }
 
-        (bool hasBlock, Block memory latestBlock) = _getLatestBlock(dispute.input.stateProof);
+        (bool validLatestBlock, bool hasBlock, Block memory latestBlock) = _tryGetLatestBlock(dispute.input.stateProof);
+        if (!validLatestBlock) {
+            return false;
+        }
         if (hasBlock) {
             return (latestBlock.stateSnapshotHash == dispute.input.latestStateSnapshotHash);
         }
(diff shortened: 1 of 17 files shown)


contracts/V1/StateChannelDiamondProxy/StateProofFacet.sol
```diff
@@ -43,40 +46,59 @@ contract StateProofFacet is StateChannelCommon {
     function verifyStateProof(Dispute memory dispute, DisputeAuditingData memory disputeAuditingData)
         public
         virtual
-        returns (StateProofVerification)
+        returns (bool)
     {
         // reference check - is this the auditingData the dispute committed to?
         if (dispute.input.disputeAuditingDataHash != keccak256(abi.encode(disputeAuditingData))) {
```
daiagi, line 52: to return `false` (aka state proof is invalid) is to assume that the reference (disputeAuditingData) is valid
yes, it is checked in the handler in `DisputeFraudProofFacet` 
but it means that `verifyStateProof` carries in it an implicit assumption

lukastanisic99: yes, it's like any other `verify/isValid` function, either the input is valid or it's invalid. The calling context uses that as a black box. Like in other places in fraud proofs, the fraud proof has to check data integrity. Deducting who to slash is the fraud proofs responsibility. Having a 3rd return value just for that didn't feel natural

src/stateManager/DisputeValidationService.ts
```diff
@@ -130,8 +126,7 @@ export default class DisputeValidationService {
 
             this.persistDisputeAuditingDataForPipeline(
                 dispute,
-                onChainDisputeAuditingData,
-                stateProof
+                onChainDisputeAuditingData
```
daiagi, line 129: you kept the idea of parsing the stateProof at the top of the fucntion
but rejected the idea of re-using the already parsed struct down stream
why?

lukastanisic99: yes, since it didn't feel natural passing a superset and a subset at the same time (e.g. dispute already contains stateProof so writing a function as f(dispute,stateProof), didn't feel natural). If you really want the stateProof model in those functions you can compute it internally from dispute. But in general after we know `it's ok`, we can just access the relevant fields directly without it. The StateProof `model` is only used as an implicit structural check, unlike other models that actually do operations on the data

daiagi (approved):

PR #371, opened 2026-06-28
visibility for test coverage

daiagi:
We want to know, quickly and without reading through the test files, what is tested and - more importantly - what is not.

This PR adds a script, yarn coverage:grid, that generates a .md and an .html report giving that visibility.

The idea is  straightforward:

We focus on the structs defined in the Solidity contracts.
In scripts/coverage-grid/domain.ts we list, per struct/field, the ways that field can be corrupted - its "byzantine variants".
For each variant we want to know: is there a test that exercises it, or not? And we want that without hand-writing lots of metadata per test.
So tests are linked to the variants two ways:

Inference - the script (scripts/coverage-grid/scan.ts) reads the test bodies and infers what each test does: which byzantine helper it triggers, which proof/calldata path it sets up.
Explicit metadata - test/harness/scenario.ts, a thin wrapper over mocha's it(), lets a test declare its target / setup / invariant for the cases inference can't catch.
Together these tie each test to the cell(s) it covers, and the script renders a map of how well the space is covered and where the gaps are.

Note: the grid does not run the tests. It's pure static analysis of the meta tags and the test source, so it's fast.

That makes it cheap to wire into a GitHub Actions workflow and publish the HTML to a gh-pages branch , so wee can have a current and evolving visibility into coverage


lukastanisic99:
@daiagi can you write the sequence of operations on how you envision this tool to be used? Like a mini docs - ask claude/gpt to do it and just verify it

daiagi:
run yarn test:coverage, open the html or the .md doc

i prefer the html


i plan to publish it to our github pages from gh workflow, but no point doing that if you reject the idea


image




i think it is very clear, once you open the html and interact with it

daiagi:
> @daiagi can you write the sequence of operations on how you envision this tool to be used? Like a mini docs - ask claude/gpt to do it and just verify it


Added

lukastanisic99:
I'm slowly forming an opinion on this one and still want to wait for Ljuba and Momcilo to share some thoughts on this. 

From reviewing Ljubas recent PRs and seeing that Momcilo had similar comments, I'm more or less convinced we have to do unit tests but with harness (no mocks and no junk data). Fake components and junk data give false positives, especially with claude/gpt convincing you it's a sound test. 

As for this PR, what I think I like is the option to add metadata to tests. So we can define in one place our variations and perform static analysis to see which variation is missing.  
E.g. We have 5 test variables:
1) calldataPosted [true/false] (2 options)
2) milestones or signed blocks or none (3 options)
3) linked blocks or not linked blocks (2 options)
4) final or not final (2 options) 
5) direct signatures or virtual voting or a combination (3 options) 

There are 2 * 3 * 2 * 2 * 3 = (72) variations, which if I had to write 'english text' to describe I would miss some or it would require more cognitive load than seeing a unified format [calldata, milestones, linked, final, virtual voting] and the tool telling me e.g. [calldata, milestones, linked, `not final`, virtual voting] is missing - with this even GPT/Claude would know how to write the 72 tests. 

What I have also observed is that often it's easier for me to understand what the test does by looking at the code than by reading the `english test name`. The code is a familiar pattern and once you know what's under the hood, it shares `information` faster and less ambiguous than a `new english sentence`. 

Humans like patterns -> that's why I like file system for unit tests since if I see `A B C` in `src` I know to expect `A B C` in `unit tests` and that conveys so much information than explaining in english "this is tested" - If I don't see B, I know in O(1) that B is missing. If I open B and don't see some method, I know tests for that method are missing without having to read english sentences that I didn't write and that are subjective (same for the other person, the english sentences that are written by someone else mean less to you than the person who wrote them). 

Also later if we add a `variable`, we'll know exactly which tests to write (AI can write them). 

### NOT SCALABLE ###
Doing all possible variations is `NOT feasible` e.g. (Disputes has 72 variations and snapshot update has 50 variations - it's not feasible to do 72 * 50 * adding_another_variable), so we have to test the variations internally and use the higher level components as black boxes and assume they `just work`. 

Scenarios/e2e tests follow the same pattern, but on a higher level:
E.g. instead of 72 variations of doing a dispute and 50 variations of snapshot update, we have:
1) Dispute/no dispute (2 options)
2) snapshot update/ no snapshot update (2 options) 
.
.

Finding many ways to combine the tested black boxes. 
Further we can have `variations with repetition` e.g. [dispute, update snapshot, dispute] and we don't have to test all possible repetitions - in most cases we just need to test 1,2 and if it works, theoretically we can assume it recursively works for all N due to mathematical induction. 

----- 

What I don't like (for now) is trying to map `metadata` to `parts of code` and enforce it - I don't think that's easy to maintain and at the end of the day, whoever writes the test same as writing the test needs to fill the metadata - most of it AI will do, since they will see a pattern and it's easy to describe these systems to them.

My thoughts for now.

daiagi:
>I'm slowly forming an opinion on this one and still want to wait for Ljuba and Momcilo to share some thoughts on this.

Follow up PR coming soon, to address the points you've raised in the above comment an in our meet. namely:

- divide the system into sub systems, and list the variants to test in each (the so called unit tests)
-  no inference of code, the test tag will contain metadata linking it to the named variant it tests
organize the "unit" test dir based on the sub systems, each sub system will decalre its varaints
- a script wll parse the "metadata" tags and will link them to the named variants (type safe, no random strings allowed) and will create both human and AI oriented  summary files that will show the gaps.

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
```diff
@@ -33,12 +33,13 @@ contract DisputeManagerFacet is StateChannelCommon {
         );
     }
 
-    function commitToReducedResult(bytes32 channelId, bytes32 disputedForkId, bytes32 reducedForkId) public {
-        DisputeData storage disputeData = disputeData[channelId];
-        DisputeWindow storage disputeWindow = disputeData.disputeWindowMap[disputedForkId];
-        require(canParticipateInDisputes(channelId, msg.sender), ErrorCantParticipateInDispute());
-        _commitToDisputeReducedResult(channelId, disputeWindow, reducedForkId, block.timestamp);
-    }
+    // not used
```
sh3ll3x3c, line 36: This (and the matching block in `StateChannelManagerProxy.sol`) is a functional contract/ABI change with no connection to coverage visibility — it deserves its own PR where it can be reviewed on its merits. I checked the branch: there are no live callers, so it may well be dead — but then it should be *deleted* in that PR, not left as commented-out code (which we otherwise flag as a smell).

scripts/coverage-grid/domain.ts
```diff
@@ -0,0 +1,259 @@
...
+export const TEST_DIRS = [
+    path.join(ROOT, "test/e2e"),
+    path.join(ROOT, "test/V1")
+];
+export const OUT_MD = path.join(ROOT, "test/V1/COVERAGE_GRID.generated.md");
+export const OUT_HTML = path.join(ROOT, "test/V1/COVERAGE_GRID.generated.html");
+
+// proof enums whose members tests reference by name (input axis)
+export const INPUT_PROOF_ENUMS = ["DisputeFraudProofType", "FraudProofType"];
+export const OUTCOME_ENUM = "StateProofVerification";
```
sh3ll3x3c, line 18: `StateProofVerification` doesn't exist anywhere in the repo — not under `contracts/V1/types/` (the only dir `parseTypes` scans) nor anywhere else in `contracts/`. So `outcomeMembers` is always empty and **every** `outcome:` tag is auto-rejected: the branch's own run reports `DSL! unknown outcome "InvalidProof"` for case4_blockInjection. Either point this at the real enum (and its actual location) or drop the outcome axis until the enum exists.

scripts/coverage-grid/index.ts
```diff
@@ -0,0 +1,70 @@
...
+    console.log(`coverage-grid ->`);
+    console.log(`  ${path.relative(ROOT, OUT_MD)}`);
+    console.log(
+        `  ${path.relative(ROOT, OUT_HTML)}  (echarts: ${echartsSrc === "echarts.min.js" ? "vendored locally" : "CDN fallback"})`
+    );
+    console.log(
+        `  scenarios ${s.scenarios} · proof types ${s.proofsCovered}/${s.proofsTotal} · cells ${s.cellsCovered}/${s.cellsTotal}`
+    );
+    console.log(
+        `  gaps: ${model.gaps.length}   DSL errors: ${model.errors.length}   cross-check mismatches: ${model.crossChecks.length}`
```
sh3ll3x3c, line 60: DSL errors (and cross-check mismatches) are only printed — the process exits 0. The README promises "run fails with `DSL errors`", which is the whole tags-can't-drift guarantee. Suggest:

```ts
process.exit(model.errors.length ? 1 : 0);
```

(verified: with the 2 DSL errors currently on this branch, exit code is 0)

.github/workflows/coverage-grid.yml
```diff
@@ -0,0 +1,41 @@
+name: Coverage Grid → GH Pages
+
+on:
+  workflow_dispatch:
```
sh3ll3x3c, line 4: `workflow_dispatch`-only means nothing ever gates a PR — combined with the exit-0 issue in `index.ts`, tag drift can land silently. The tool needs no compile and runs in ~1s, so a cheap `pull_request`-triggered check job (just `yarn` + `yarn coverage:grid` with a non-zero exit on DSL errors) would make the README's guarantee real. Pages publishing can stay manual.

test/e2e/E2E-FraudProofsBlockConfirmation.test.ts
```diff
@@ -12,557 +14,565 @@ import type { BlockBundle } from "@test/fixtures/customRpc/harnessControl/servic
...
+            await h.assert.sync.onlyHonestPeersInSync();
+        }
+    );
+
+    // genesis block forging a forkId the channel never opened (previousBlockHash still links to
+    // the real genesis). KNOWN FAILURE (left red): createWrongGenesisProof looks up the genesis
+    // by the forged block.forkId -> undefined -> `!`.toStruct() throws in onBlockConfirmation, so
+    // no dispute is initiated.Currentlly, the attcker DOES NOT get slashed
+    scenario(
+        "wrong genesis forkId → WrongGenesis",
```
sh3ll3x3c, line 377: This test is documented as KNOWN FAILURE and left red — it will fail every suite run until the product bug is fixed. The find itself is great (forged genesis forkId → `createWrongGenesisProof` looks up genesis by the forged id → `undefined` → `.toStruct()` throws in `onBlockConfirmation` → no dispute, attacker escapes slashing) and is arguably the most valuable single output of this PR. But please land it as `scenario.skip` with the reason + a tracked issue, or together with the fix — a permanently-red test trains everyone to ignore failures.

test/e2e/E2E-StateSnapshots.test.ts
```diff
@@ -13,13 +13,22 @@ import { expect } from "chai";
  * after fork resolution, and single multicall (fork + same-fork) update.
  */
 describe("E2E: State Snapshots", function () {
+    // @scenario channel-close-na
```
sh3ll3x3c, line 16: These comment-form `// @scenario` tags are never parsed — `extractScenarioBlocks` in scan.ts only matches actual `scenario(` calls (the `// @scenario` regex there is just a body-window boundary). So all 7 annotations in this file are inert; the tests register as untagged loose refs only. Convert them to `scenario()` wraps like the rest of the migration, or drop the comments — as-is they look load-bearing but aren't.

scripts/coverage-grid/contracts.ts
```diff
@@ -0,0 +1,108 @@
+import * as fs from "fs";
+import * as path from "path";
+import { parse, visit } from "@solidity-parser/parser";
```
sh3ll3x3c, line 3: `@solidity-parser/parser` isn't in `package.json` — it currently resolves only transitively (hoisted from `solidity-coverage` / `eth-gas-reporter@0.14.5`). Since neither tsconfig excludes `scripts/`, this file is in both builds' compilation sets, so a transitive bump/removal breaks `yarn tsc` for everyone, not just this tool. Please add it to `devDependencies`.

scripts/coverage-grid/scan.ts
```diff
@@ -0,0 +1,346 @@
...
+    meta: string;
+    line: number;
+    bodyStart: number;
+}
+
+// every scenario() call in a file, with its meta object source and body offset
+function extractScenarioBlocks(lines: string[]): ScenarioBlock[] {
+    const blocks: ScenarioBlock[] = [];
+    for (let i = 0; i < lines.length; i++) {
+        if (!/^\s*scenario(?:\.(?:skip|only))?\(/.test(lines[i])) continue;
```
sh3ll3x3c, line 134: `scenario.skip(` matches here too, and unless the scenario also carries `unreachable:`, its targets land in the covered set — a green cell backed by a test that never runs. Today the one skip in the migration is correctly `unreachable`-tagged, but nothing enforces that pairing. Suggest excluding `.skip` scenarios from coverage unless `unreachable:` is present (or erroring on a skip without it).

.gitignore
```diff
@@ -38,3 +38,8 @@ src/utils/GeneratedArtifacts.ts
 *.tsbuildinfo
 
 /profiles
+
+# coverage-grid: vendored lib + heavy generated dashboard (regenerate via yarn coverage:grid)
+test/V1/echarts.min.js
+test/V1/COVERAGE_GRID.generated.html
+test/V1/COVERAGE_GRID.matrix.csv
```
sh3ll3x3c, line 45: Two inconsistencies: `COVERAGE_GRID.matrix.csv` is never generated by the tool (no writer for it), while `COVERAGE_GRID.generated.md` *is* generated but neither ignored nor tracked — so every `yarn coverage:grid` run leaves an untracked file in the working tree. Decide whether the .md is a committed snapshot or a build artifact and make this block match.

sh3ll3x3c (commented):
Reviewed the tooling in depth, ran `yarn coverage:grid` end-to-end, audited every migrated test file against `test-contracts-2`, and type-checked both builds. Overall: the tool has real merit — contract-derived axes, inference + explicit tags with a drift cross-check, and N/A modeling are the right design — and the test migration is safe (every `it`→`scenario` wrap preserves assertions verbatim; nothing deleted or weakened; 5 net-new tests). Both `tsc` builds pass with the tool in the compilation set.

Findings are inline. The short version:

- **Blocking:** the `commitToReducedResult` contract change is unrelated to coverage visibility and should be its own PR.
- **Tool guarantees that don't hold yet:** the `outcome:` axis validates against an enum that doesn't exist (the branch's own run emits 2 DSL errors), and the promised "run fails on DSL errors" gate exits 0 and never runs on PRs.
- **Suite health:** one net-new test is a documented known-failure left red — it will fail every run until the product bug it found is fixed (nice catch, by the way — that bug is arguably this PR's most valuable output).
- Plus: an undeclared direct dependency, dead comment-form `@scenario` tags in one file, `scenario.skip` counting as covered, and a `.gitignore` inconsistency.

Cosmetic nits left out.

scripts/coverage-grid/domain.ts
daiagi: it did exist in the time of writing the PR, but since then was  removed in `dispute` branch
anyway, this PR will largely be  superseded by  a follow up I'm working on, in which `outcome` will not exist anymore

scripts/coverage-grid/index.ts
daiagi: >are only printed 

yes, I have no intention of failing CI for these.
same goes for  comment below : "nothing ever gates a PR"

this is not the intention, we are not gating PRs on test failures atm, so it goes without saying we don't gate PR's  on coverage

test/e2e/E2E-FraudProofsBlockConfirmation.test.ts
daiagi: >a permanently-red test trains everyone to ignore failures.

arguably, skipped test gives impression of all green when it isn't
this is the line that has been set by @lukastanisic99  a few times, we prefer to see the failure rather then silence it.

but I do agree, worth putting it on separate PR

test/e2e/E2E-StateSnapshots.test.ts
daiagi: yes, these are dead left overs.
will be gone in the follow up PR

scripts/coverage-grid/contracts.ts
daiagi: ok. I don't think I need this one anymore in the follow up. but I'll make sure it's either removed or put into `devDependencies`

.gitignore
daiagi: all these files are removed or in altered form in the follow up. 
I'll make sure the coverage reports int he follow up are tracked.

daiagi:
Closing, The follow up I'm working on changes too many things for this PR to be valuable to review anymore.

test/e2e/E2E-FraudProofsBlockConfirmation.test.ts
daiagi: https://trello.com/c/XyyK0Yk6

PR #372, opened 2026-06-28
Request/response for RPC

lukastanisic99:
Before, all RPCs were implemented as `fire and forget` where each request/response had 2 different endpoints and state had to be tracked across the calls explicitly. With implementing a general request/response pattern for RPCs in one of my previous PRs all of this is implicitly available to all RPC Services. 
The result, we can write:
`const response = await remoteRpc.serviceName.serviceMethod(args).request(peerAddress)`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -4,6 +4,11 @@
 - For post-edit validation, run the narrowest relevant test first when available.
 - Run `yarn tsc --noEmit -p tsconfig.json` for TypeScript typechecking.
 - Run `yarn compile` for compile-level validation when changes affect the build or exported package surface.
+- Don't disturb the user's working tree or index to inspect another revision.
+  Prefer a throwaway worktree (`git worktree add <tmp> <ref>`) over `git stash`.
+  If you must stash, preserve staging: restore with `git stash pop --index` so
+  files the user had staged come back staged — never leave them re-reviewing
+  files they'd already staged.
 
 ## Conventions
 
@@ -26,6 +31,12 @@ Applies to `src/rpc/services/*` and `test/fixtures/customRpc/**`.
 ### Comments
 
 - Keep comments simple and to the point. No long essays.
+- When refactoring or moving code (extracting a method, moving a body to another
+  file/service), carry over all original comments verbatim — do not drop or
+  condense them, including large explanatory/strategy blocks and numbered step
+  comments. If a comment genuinely needs to change, call it out rather than
+  dropping it silently; flag stale commented-out dead code instead of removing
+  it without mention.
 
 ### Type safety
 
--- a/test/e2e/E2E-SpectateStaleProofGuard.test.ts
+++ b/test/e2e/E2E-SpectateStaleProofGuard.test.ts
@@ -49,7 +49,7 @@ describe("E2E: Spectate stale-proof guard", function () {
     it("aborts sync when a peer answers with undecodable junk bytes", async function () {
         const h = TestSession.getHarness();
 
-        await h.lifecycle.start(2, 0, {
+        await h.lifecycle.start(2, 2, {
             timeConfig: {
                 p2pTime: 5,
                 agreementTime: 2,
@@ -57,12 +57,6 @@ describe("E2E: Spectate stale-proof guard", function () {
                 evidenceTime: 10
             }
         });
-
-        await h.transition.advanceState({
-            count: 2,
-            waitForFinalization: true
-        });
-        await h.transition.postSnapshotWait();
 
         // Both participants reply to spectate requests with bytes that are NOT a
         // valid encoded SyncPayload, so the spectator's Codec.decode throws and
--- a/test/e2e/E2E-SpectateStaleProofGuard.test.ts
+++ b/test/e2e/E2E-SpectateStaleProofGuard.test.ts
@@ -45,4 +45,48 @@ describe("E2E: Spectate stale-proof guard", function () {
             await h.control(spectator).query.getOpenConnectionCount().request()
         ).to.equal(0, "Spectator should have 0 open connections after abort");
     });
+
+    it("aborts sync when a peer answers with undecodable junk bytes", async function () {
+        const h = TestSession.getHarness();
+
+        await h.lifecycle.start(2, 0, {
+            timeConfig: {
+                p2pTime: 5,
+                agreementTime: 2,
+                chainFallbackTime: 2,
+                evidenceTime: 10
+            }
+        });
+
+        await h.transition.advanceState({
+            count: 2,
+            waitForFinalization: true
+        });
+        await h.transition.postSnapshotWait();
+
+        // Both participants reply to spectate requests with bytes that are NOT a
+        // valid encoded SyncPayload, so the spectator's Codec.decode throws and
+        // it must abort every sync attempt instead of crashing/hanging.
+        await h.rpcStub.stubSpectateJunkPayload([0, 1]);
+
+        let threwTimeout = false;
+        try {
+            await h.join.addSpectatorWait({ statusTimeoutMs: 5000 });
+        } catch {
+            threwTimeout = true;
+        }
+
+        expect(threwTimeout).to.equal(
+            true,
+            "Spectator should not have reached SYNCED with junk payloads"
+        );
+
+        const spectator = h.getPeer(2);
+        expect(
+            await h.control(spectator).query.getOpenConnectionCount().request()
+        ).to.equal(
+            0,
+            "Spectator should have 0 open connections after aborting on junk"
+        );
+    });
 });
--- a/test/fixtures/customRpc/harnessControl/services/stub/StubRpcMethods.ts
+++ b/test/fixtures/customRpc/harnessControl/services/stub/StubRpcMethods.ts
@@ -7,6 +7,8 @@ import type { Address } from "@/types/types";
 import type SpectateServiceRpcMethods from "@/rpc/services/spectate/SpectateRpcMethods";
 import type { SyncRequest } from "@/rpc/services/spectate/SpectateService";
 import type IsForkDisputedRpcMethods from "@/rpc/services/isForkDisputedService/IsForkDisputedRpcMethods";
+import InitHandshakeRpcMethods from "@/rpc/services/initHandshake/InitHandshakeRpcMethods";
+import type { Hash, Timestamp } from "@/types/types";
 import type { HarnessControlRpc } from "../../HarnessControlRpc";
 import type { StubService } from "./StubService";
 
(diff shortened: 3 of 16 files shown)


sh3ll3x3c:
While reviewing this refactor I wrote two independent e2e checks for behavior the PR's own suite doesn't currently exercise. Both pass on `request-response-rpc`. One is a genuine coverage gap; the other is optional hardening.

**Coverage gap worth adding — Spectate concurrent-sync dedup.** The new closure-based `inFlightByPeerAddress` guard in `SpectateService.sync` (which replaces the old `requestMapByPeerAddress`) has no test. The case below asserts that two concurrent `sync()` calls for the same peer collapse to a single on-the-wire request — it counts requests on the responder and would fail (`count === 2`) if the dedup regressed.

**Optional hardening — duplicate handshake ack.** A second `onInitHandshakeAck` over an already-acked transport must disconnect + blacklist. That guard pre-dates this PR (the refactor didn't touch the ack path), so it's extra coverage rather than a gap this PR introduces — including it only because it pairs naturally with the file.

Drop-in if useful — `test/e2e/E2E-RequestResponseRefactor.test.ts`, green on the branch:

```ts
import { waitFor } from "@test/utils/waitFor";

/**
 * Independent adversarial tests for the request/response RPC refactor
 * (PR: "Request/response for RPC").
 *
 * These are deliberately NOT the PR author's tests — they target invariants of
 * the new model that the PR's own suite does not exercise:
 *   1. InitHandshake: a duplicate handshake ack is a protocol violation and must
 *      disconnect + blacklist the sender (`onInitHandshakeAck` dup detection).
 *   2. Spectate: two concurrent `sync()` calls to the same peer must be deduped
 *      to a single request (the new closure-based `inFlightByPeerAddress` guard
 *      that replaced the old per-peer request map).
 */
describe("E2E: Request/response refactor — independent checks", function () {
    describe("InitHandshake duplicate ack", function () {
        it("disconnects and blacklists a peer that sends a second handshake ack", async function () {
            await h.lifecycle.start(2, 0, { autoConnect: true });
            await h.event.waitUntilEventOccurs("onConnection", 5000);
            // Sanity: the handshake completed, so peer 0 has already recorded
            // peer 1's (first, legitimate) ack.
            await h.assert.rpc.handshakeCompleted({ peer1: 0, peer2: 1 });

            const peer0 = h.getPeer(0);
            const peer1Address = h.getPeer(1).address;

            expect(
                await h.control(peer0).query.getOpenConnectionCount().request()
            ).to.equal(1, "peer 0 should be connected to peer 1 before the dup ack");

            // Deliver a SECOND ack over peer 1's transport, host-side, exactly as
            // the dispatcher would on a crafted duplicate. The first ack is
            // already recorded (handshake complete), so this must be treated as a
            // protocol violation.
            await h.execOnHost(
                peer0,
                (sm, args) => {
                    const transport =
                        sm.p2pManager.profileManager.getProfileByEvmAddress(
                            args.peer1Address
                        )?.transport;
                    if (!transport) {
                        throw new Error("no transport to peer 1 after handshake");
                    }
                    void sm.p2pManager.localRpc.initHandshakeService
                        .createRPCMethods(transport)
                        .onInitHandshakeAck();
                },
                { peer1Address }
            );

            // The duplicate ack must drop the connection...
            await h.assert.rpc.peerDisconnectedFrom({
                peerIndex: 0,
                expectedFinalCount: 0
            });
            // ...and blacklist the offender.
            expect(
                await h
                    .control(peer0)
                    .query.isBlacklisted(peer1Address)
                    .request()
            ).to.equal(
                true,
                "peer 1 must be blacklisted after sending a duplicate handshake ack"
            );
        });
    });

    describe("Spectate concurrent-sync dedup", function () {
        it("issues only one request when sync() is called twice for the same peer", async function () {
            await h.lifecycle.start(2, 0, { autoConnect: true });
            await h.event.waitUntilEventOccurs("onConnection", 5000);
            await h.assert.rpc.handshakeCompleted({ peer1: 0, peer2: 1 });

            const peer0 = h.getPeer(0);
            const peer1 = h.getPeer(1);
            const peer1Address = peer1.address;

            // Let any handshake-triggered startup sync settle before we measure.
            await new Promise((r) => setTimeout(r, 750));

            // On peer 1 (the responder): replace onSpectateRequest with a stub
            // that counts every incoming request and then rejects. Counting from
            // here means only requests sent AFTER install are tallied.
            await h.execOnHost(peer1, (sm) => {
                const ss = sm.p2pManager.localRpc.spectateService;
                const stub = sm.p2pManager.localRpc.stub as unknown as {
                    spectateReqCount: number;
                };
                stub.spectateReqCount = 0;
                const orig = ss.createRPCMethods.bind(ss);
                ss.createRPCMethods = (t) => {
                    const m = orig(t);
                    m.onSpectateRequest = async () => {
                        (
                            sm.p2pManager.localRpc.stub as unknown as {
                                spectateReqCount: number;
                            }
                        ).spectateReqCount++;
                        // Stay in-flight briefly, then reject; the payload itself
                        // is irrelevant — we only care how many requests arrive.
                        await new Promise((r) => setTimeout(r, 300));
                        throw new Error("counting stub: reject");
                    };
                    return m;
                };
            });

            // On peer 0 (the spectator): clear any residual in-flight marker from
            // a startup sync, then fire two syncs for the same peer in ONE tick.
            // The second must hit the in-flight guard and never reach the wire.
            await h.execOnHost(
                peer0,
                (sm, args) => {
                    const ss = sm.p2pManager.localRpc.spectateService as unknown as {
                        inFlightByPeerAddress: Set<string>;
                        sync: (peer: string, channelId: unknown) => void;
                    };
                    ss.inFlightByPeerAddress.clear();
                    ss.sync(args.peer1Address, sm.channelId);
                    ss.sync(args.peer1Address, sm.channelId);
                },
                { peer1Address }
            );

            const readCount = () =>
                h.execOnHost(
                    peer1,
                    (sm) =>
                        (
                            sm.p2pManager.localRpc.stub as unknown as {
                                spectateReqCount?: number;
                            }
                        ).spectateReqCount ?? 0
                );

            // Exactly one request should arrive...
            await waitFor(async () => (await readCount()) >= 1, 5000);
            // ...and give a generous window for a (wrongly) un-deduped second
            // request to show up before asserting it never did.
            await new Promise((r) => setTimeout(r, 1500));
            expect(await readCount()).to.equal(
                1,
                "two concurrent sync() calls for the same peer must collapse to a single request"
            );
        });
    });
});

```

sh3ll3x3c (approved):

test/fixtures/customRpc/harnessControl/services/stub/StubRpcMethods.ts
```diff
@@ -410,6 +451,76 @@ export class StubRpcMethods extends ARpcMethods<P2PManager<HarnessControlRpc>> {
         };
         return true;
     }
+
+    /**
+     * Make this peer answer handshake challenges (`onInitHandshakeRequest`) with
+     * a faulty response so the *initiator* rejects it:
+     * - `responseTimeOffsetSeconds` skews the returned `responseTime` so the
+     *   initiator's response-timestamp check fails.
+     * - `delayMs` delays the reply so the initiator's request times out (no
```
daiagi: I'm confused about this comment, through out the callstack
`initiateHandshakeWithFaultyResponse` -> `RpcStubActions.stubHandshakeResponse` -> `StubRpcMethods.stubHandshakeResponse`

Not clear why delay means the request will time out, if the delay is big enough, sure. but what is big enough is for the calling context to know. if i give 1 ms delay, will the request time out?
to be clear, my complaint is about the comments, not the code. The code is clear.

test/harness/actions/RPCActions.ts
```diff
@@ -221,134 +219,65 @@ export class RPCActions<
...
-        );
+        const {
+            initiatorPeer,
+            responderPeer,
+            delaySeconds,
+            responseTimeOffsetSeconds,
+            corruptSignature
+        } = options;
+
+        const delayMs = delaySeconds === undefined ? 0 : delaySeconds * 1000;
```
daiagi: why ont use delayMs instead of `delaySeconds`

test/fixtures/customRpc/harnessControl/services/stub/StubRpcMethods.ts
```diff
@@ -410,6 +451,76 @@ export class StubRpcMethods extends ARpcMethods<P2PManager<HarnessControlRpc>> {
...
+                const response = await realRequest.call(
+                    this,
+                    challengeHash,
+                    time
+                );
+                if (responseTimeOffsetSeconds) {
+                    response.responseTime += responseTimeOffsetSeconds;
+                }
+                if (corruptSignature) {
+                    response.signature = "0xdeadbeef";
```
daiagi: better use real signature and corrupt it slightly, (flip one byte or something)
or valid signature over unmatching payload.
claude sure loves this `deadbeef` thingy

test/fixtures/customRpc/harnessControl/services/stub/StubRpcMethods.ts
```diff
@@ -410,6 +451,76 @@ export class StubRpcMethods extends ARpcMethods<P2PManager<HarnessControlRpc>> {
...
+                }
+                if (delayMs > 0) {
+                    await new Promise((resolve) =>
+                        setTimeout(resolve, delayMs)
+                    );
```
daiagi: nicer to read:

```ts
await sleep(delayMs)
```

test/e2e/E2E-SpectateStaleProofGuard.test.ts
```diff
@@ -45,4 +45,48 @@ describe("E2E: Spectate stale-proof guard", function () {
...
+                agreementTime: 2,
+                chainFallbackTime: 2,
+                evidenceTime: 10
+            }
+        });
+
+        await h.transition.advanceState({
+            count: 2,
+            waitForFinalization: true
+        });
```
daiagi:
```suggestion
```

simply chnage `0` -> `2` in `lifecycle.start` above

test/e2e/E2E-SpectateStaleProofGuard.test.ts
```diff
@@ -45,4 +45,48 @@ describe("E2E: Spectate stale-proof guard", function () {
...
+        await h.transition.postSnapshotWait();
```
daiagi: not sure why this is relevant, and the test passes without it

test/fixtures/customRpc/harnessControl/services/stub/StubRpcMethods.ts
```diff
@@ -280,6 +288,38 @@ export class StubRpcMethods extends ARpcMethods<P2PManager<HarnessControlRpc>> {
...
+        }
+        const original = this.service.stubOriginals.get(
+            "spectateCreateRpcMethods"
+        ) as typeof service.createRPCMethods;
+        service.createRPCMethods = (transport: ATransport) => {
+            const methods = original(transport);
+            methods.onSpectateRequest = async function (
+                this: SpectateServiceRpcMethods
+            ) {
+                return { encodedSyncPayload: "0xdeadbeef" };
```
daiagi, line 312: same, not a fan of the `0xdeadbeef` thing
i acknowledge it doesnt make much of a difference, but i find the test to be more convincing when the "junk" value more closly resemebles the correct value.

i.e no attacker will give this kind of value, but they may alter a bit here and there, and we want to see that being detected and rejected

AGENTS.md
```diff
@@ -4,6 +4,11 @@
 - For post-edit validation, run the narrowest relevant test first when available.
 - Run `yarn tsc --noEmit -p tsconfig.json` for TypeScript typechecking.
 - Run `yarn compile` for compile-level validation when changes affect the build or exported package surface.
+- Don't disturb the user's working tree or index to inspect another revision.
+  Prefer a throwaway worktree (`git worktree add <tmp> <ref>`) over `git stash`.
+  If you must stash, preserve staging: restore with `git stash pop --index` so
```
daiagi, line 9: 👍 i like that

daiagi (approved):
comment are small to nit -> approve with comments

PR #374, opened 2026-06-29
(fix) none calldata path `areSignedBlocksLinkedAndVerified`

daiagi:
proposed fix for #369

### DIFF
--- a/contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol
+++ b/contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol
@@ -202,11 +202,12 @@ contract DisputeFraudProofFacet is StateChannelCommon {
             if (!_isLastMilestoneFinalByEveryone(dispute)) return _invalid();
 
             if (dispute.input.stateProof.signedBlocks.length != 0) {
-                bytes memory linkData = abi.encodeCall(
-                    StateProofFacet.areSignedBlocksLinkedAndVerified, (dispute.input.stateProof.signedBlocks)
+                bytes memory linkReturnData = _delegatecall(
+                    stateProofFacetAddress,
+                    abi.encodeCall(
+                        StateProofFacet.areSignedBlocksLinkedAndVerified, (dispute.input.stateProof.signedBlocks)
+                    )
                 );
-                (bool linkSuccess, bytes memory linkReturnData) = stateProofFacetAddress.delegatecall(linkData);
-                if (!linkSuccess) return _invalid();
                 if (!abi.decode(linkReturnData, (bool))) return _valid(dispute.input.disputer);
             }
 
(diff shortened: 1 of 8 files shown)


contracts/V1/StateChannelDiamondProxy/StateChannelManagerProxy.sol
```diff
@@ -119,12 +119,9 @@ contract StateChannelManagerProxy is StateChannelManagerInterface, StateChannelC
...
         }
         // verify threshold signature - must be from all participants - this is deterministic - no race condition on-chain
-        (bool isValid, string memory reason) = UtilityFacet(utilityFacetAddress)
-            .verifyThresholdSigned(
-                openChannelData.participants,
-                openChannelConfirmation.encodedOpenChannel,
-                openChannelConfirmation.signatures
-            );
+        (bool isValid, string memory reason) = UtilityFacet(utilityFacetAddress).verifyThresholdSigned(
+            openChannelData.participants, openChannelConfirmation.encodedOpenChannel, openChannelConfirmation.signatures
```
lukastanisic99, line 123: @daiagi can you take a moment to see why are formatting pre-commit hooks are not working? More or less in every PR we're looking at formatting diffs

daiagi: ok

contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol
```diff
@@ -201,6 +201,15 @@ contract DisputeFraudProofFacet is StateChannelCommon {
         if (!dispute.postedAuditingData) {
             if (!_isLastMilestoneFinalByEveryone(dispute)) return _invalid();
 
+            if (dispute.input.stateProof.signedBlocks.length != 0) {
+                bytes memory linkData = abi.encodeCall(
+                    StateProofFacet.areSignedBlocksLinkedAndVerified, (dispute.input.stateProof.signedBlocks)
+                );
+                (bool linkSuccess, bytes memory linkReturnData) = stateProofFacetAddress.delegatecall(linkData);
+                if (!linkSuccess) return _invalid();
+                if (!abi.decode(linkReturnData, (bool))) return _valid(dispute.input.disputer);
```
lukastanisic99, line 211: isn't this already checked internally in the stateProof?

lukastanisic99: Is this only here for the case where signedBlocks can not decode? 
If so, this commit (https://github.com/peer3to/state-channels-plus/commit/ac40acd9e7a8741e09432a629169b7471340c00b) is already merged into dispute and adds the `tryDecode` helper

try merging with dispute and seeing if it works without this

daiagi: >isn't this already checked internally in the stateProof?

this check is under the `if (!dispute.postedAuditingData)` branch (line 201)
the other call in line 227 is the  `dispute.postedAuditingData == true` path

>Is this only here for the case where signedBlocks can not decode

removed

lukastanisic99 (approved):

PR #376, opened 2026-06-29
feat(webrtc): surface main-thread bridge port on P2pInstance for nested workers

sh3ll3x3c:
## Merge order — read first

2. contract-event delivery fix (#379) → `dispute`
4. 👉 **(this PR)** WebRTC main-thread bridge port → `dispute`
5. browser host-worker boot fixes (#380) → `dispute`

The SDK PRs are independent of each other. This PR and the browser host-worker
boot fixes (#380) both merge before the downstream app consumer (step 6); order
between them does not matter.

---

## Summary

Bubble up the **main-thread end** of the WebRTC bridge `MessagePort` so a worker — however nested — can negotiate WebRTC directly with the real main thread.

- The runtime host now mints the bridge `MessageChannel` when it can't run `RTCPeerConnection` locally (`workerNeedsMainThreadBridge()`), registers the worker end with `WorkerBridgeWebRTCConnectionFactory`, and posts the main-thread end up over the runtime port (transferred). `P2pRuntimeClient` stores it; `P2pInstance.webRTCBridgePort` exposes it.
- `installWebRTCMainThreadBridge(port)` now binds a broker to an externally-provided `MessagePort`, replacing the old top-down flow (main thread minted the channel and posted an `init` message into a direct child worker). The worker-side global init listener is removed.
- Each nesting layer just forwards the field upward — no depth-aware logic. `WebRTCBridgeProtocol` is unchanged.
- Bridge teardown is port-scoped, and the worker end is registered only after the transfer succeeds, so a failed transfer can't leave a half-installed bridge.

## Test Plan

- `test/utils/WebRTCMainThreadBridge.test.ts` — provided-port handshake + bidirectional proxy-channel relay.
- `test/evm/P2pRuntimeClient.test.ts` — client stores the bubbled-up port from the host message.
- `test/browser/webrtc-service-smoke.js` + worker — updated to the provided-port contract.
- `yarn tsc --noEmit` clean (node + browser); bridge/client/factory tests green.

## Notes

- Draft — downstream consumer wiring (separate private repos) follows separately; this is the SDK-side dependency.


src/rpc/services/WebRTCSetup/connection/WorkerBridgeWebRTCConnectionFactory.ts
```diff
@@ -277,9 +267,7 @@ class WorkerBridgeWebRTCConnectionFactory implements WebRTCConnectionFactory {
         return WorkerBridgeWebRTCConnectionFactory.instance;
     }
 
-    private constructor() {
-        this.ensureReceiverRegistered();
-    }
+    private constructor() {}
 
     registerPort(port: MessagePort): void {
```
momoshell, line 273: Every runtime host now registers its own bridge port, but this singleton replacement disposes any existing host’s bridge in the same worker. 

Can we make bridge ownership host-scoped or shared/ref-counted so multiple `P2pInstances` in one worker realm cannot tear each other down?

src/rpc/services/WebRTCSetup/connection/WebRTCMainThreadBridge.ts
```diff
@@ -396,38 +391,29 @@ class WebRTCMainThreadBridgeBroker {
...
-    worker.postMessage(initMessage, [channel.port2]);
+    // Installing twice on the same port (e.g. an effect re-run) must reuse the
+    // first broker — a second one would clobber its `onmessage` and a single
+    // dispose would close the port out from under it.
+    const existing = brokersByPort.get(port);
+    const broker = existing ?? new WebRTCMainThreadBridgeBroker(port, options);
+    if (!existing) brokersByPort.set(port, broker);
 
     return {
         dispose: () => {
```
momoshell, line 426: Duplicate installs reuse the broker, but each returned handle can still dispose it. Please make duplicate handles ref-counted or idempotent so one stale cleanup cannot close the bridge for the active owner.

src/rpc/services/WebRTCSetup/connection/WebRTCMainThreadBridge.ts
```diff
@@ -396,38 +391,29 @@ class WebRTCMainThreadBridgeBroker {
...
+/**
+ * Bind a WebRTC main-thread bridge to the `port` surfaced by `p2pSetup` on
+ * `P2pInstance.webRTCBridgePort`. Call this on the real main thread (where
+ * `RTCPeerConnection` lives) after forwarding the port up through any worker
+ * nesting; the broker then drives `RTCPeerConnection` for the worker — however
+ * deeply nested — that negotiates WebRTC over the channel.
+ */
 export function installWebRTCMainThreadBridge(
-    worker: WebRTCBridgeWorkerTarget,
+    port: MessagePort,
```
momoshell, line 409: This is a public API break. 
My guess it's only us that use it right now?

src/rpc/services/WebRTCSetup/connection/WebRTCProvider.ts
```diff
@@ -25,8 +25,21 @@ export function hasLocalRTCPeerConnection(): boolean {
...
-export function isRTCPeerConnectionUnavailableWorker(): boolean {
-    return isWorkerRuntime() && !hasLocalRTCPeerConnection();
+/**
+ * True when this runtime is a worker that cannot negotiate WebRTC itself and so
+ * must drive an `RTCPeerConnection` on the real main thread over a bridge port.
+ * Mirrors the decision {@link createWebRTCConnectionFactory} makes — it probes
+ * the lazy `get-webrtc` import too, so a worker that CAN load WebRTC locally
+ * (e.g. Node with `get-webrtc`) reports `false` and no bridge is surfaced.
+ */
+export async function workerNeedsMainThreadBridge(): Promise<boolean> {
```
lukastanisic99: for all the boolean checks we mostly use some prefix (e.g. is/does/has etc..) that indicates 'this function performs a check and returns a boolean' without us having to look at the return value or the body

src/evm/P2pInstance.ts
```diff
@@ -19,6 +19,16 @@ export default class P2pInstance<
...
     logger: Logger;
 
+    /**
+     * Main-thread end of the WebRTC bridge `MessagePort`. Present only when the
+     * host runs in a worker that can't drive `RTCPeerConnection` itself; forward
+     * it up any worker nesting (in each `postMessage` transfer list) to the real
+     * main thread and pass it to `installWebRTCMainThreadBridge(port)`.
+     */
+    get webRTCBridgePort(): MessagePort | undefined {
+        return this.client.webRTCBridgePort;
```
lukastanisic99, line 36: now that we have threaded the SDK, I assume most users (including us), will use it directly from the MainThread and we won't need to wrap it in a thread further. So, in `p2pSetup` if it's called from the main thread directly, we can isntall the port automatically during the setup and if not, then we can land the port here and the consumer app would have to continue bubbling it up

src/rpc/services/WebRTCSetup/connection/WorkerBridgeWebRTCConnectionFactory.ts
lukastanisic99: ye, this should be tied to the singleton - set only once. There is no need for a worker to have multiple ports or to override this. There is only 1 main thread. The main thread may have multiple workers.

test/browser/webrtc-service-smoke.js
```diff
@@ -298,7 +298,14 @@ async function runWebRTCWorkerBridgeSmoke(bridgeOptions = {}) {
         new URL("./webrtc-service-worker.js", import.meta.url),
         { type: "module" }
     );
-    const bridge = installWebRTCMainThreadBridge(worker, bridgeOptions);
+    // Mirror the SDK contract: hand the worker the bridge port (the host would
+    // surface this on P2pInstance.webRTCBridgePort), then bind the other end.
+    const bridgeChannel = new MessageChannel();
+    worker.postMessage({ type: "bridgePort" }, [bridgeChannel.port2]);
```
lukastanisic99, line 304: I did a poor job here with these browser tests, so of course claude/gpt built on junk and deducted the same. This isn't really testing the infra, it's not creating our workers nor it has the same direction of the communication (here it's main theread worker and in the sdk it's worker -> main thread). 

So we should here do:
1) our normal p2pSetup directly from the main thread (with the sdk thread flag = `true` and see will we automatically setup the connection)
2) inside a worker thread where we'd have to buble up p2pInstance.port manually and see can we establish a connection by using the bubble up port and calling install on the main thread.

test/evm/P2pRuntimeClient.test.ts
```diff
@@ -0,0 +1,55 @@
...
+            onClose: () => undefined,
+            close: () => undefined
+        },
+        deliver: (message) => handler?.(message)
+    };
+}
+
+function makeClient(port: RuntimePort): P2pRuntimeClient {
+    return new P2pRuntimeClient(port, {
+        signerAddress: "0x0000000000000000000000000000000000000001",
```
lukastanisic99: same as my comment on the other PR - use harness to teleport and test what you want with all real components connected and all correct data, unlike these mocks and junk data

test/utils/WebRTCMainThreadBridge.test.ts
```diff
@@ -0,0 +1,163 @@
+import { expect } from "chai";
+
+import { installWebRTCMainThreadBridge } from "@/rpc/services/WebRTCSetup/connection/WebRTCMainThreadBridge";
+import WorkerBridgeWebRTCConnectionFactory from "@/rpc/services/WebRTCSetup/connection/WorkerBridgeWebRTCConnectionFactory";
+import type { WebRTCDataChannelLike } from "@/rpc/services/WebRTCSetup/connection/WebRTCConnectionFactory";
+import { waitFor } from "./waitFor";
+
+class FakeDataChannel {
```
lukastanisic99: -||-

src/rpc/services/WebRTCSetup/connection/WorkerBridgeWebRTCConnectionFactory.ts
sh3ll3x3c: Done in 7ba0ffbd — `registerPort` now sets the port once (first host wins) and `disposeBridge` is ref-counted by registered port, so the worker-realm bridge is torn down only when the last registered host disposes. Concurrent hosts sharing a worker realm can no longer tear each other's bridge down.

src/rpc/services/WebRTCSetup/connection/WebRTCMainThreadBridge.ts
sh3ll3x3c: Done in 7ba0ffbd — `installWebRTCMainThreadBridge` now ref-counts handles per port: a handle's `dispose()` is idempotent, and only the last outstanding handle tears the broker down, so a stale cleanup can't close the bridge for the active owner.

src/rpc/services/WebRTCSetup/connection/WebRTCMainThreadBridge.ts
sh3ll3x3c: Yep, only us right now. The exported signature is unchanged. With the new auto-install the SDK calls this itself on the main thread during `p2pSetup`, so app code only needs it for the nested-worker bubble-up path.

src/rpc/services/WebRTCSetup/connection/WebRTCProvider.ts
sh3ll3x3c: Renamed to `doesWorkerNeedMainThreadBridge` (7ba0ffbd).

src/evm/P2pInstance.ts
sh3ll3x3c: Done in 7ba0ffbd — `p2pSetup` now auto-installs the bridge on the main thread via `P2pInstance.installMainThreadBridgeIfOnMainThread()` (guarded by `!isWorkerRuntime()`); inside a worker the port stays on `p2pInstance.webRTCBridgePort` for the consumer app to bubble up.

test/browser/webrtc-service-smoke.js
sh3ll3x3c: In progress — rewriting this as a real `p2pSetup` browser e2e, both ways: (1) `p2pSetup` from the main thread with the SDK-thread flag, asserting the bridge auto-installs; and (2) inside a worker, bubbling up `p2pInstance.webRTCBridgePort` and installing on the main thread to establish the connection (worker → main). Also fixed a pre-existing missing-`@platform`-alias bug that was blocking the browser runner (cd3f612a).

test/evm/P2pRuntimeClient.test.ts
sh3ll3x3c: The `webRTCBridgePort` surfacing will be covered by the worker bubble-up path of the browser `p2pSetup` e2e (in progress); this fake-port test is removed once that lands.

test/utils/WebRTCMainThreadBridge.test.ts
sh3ll3x3c: Removed in 7ba0ffbd — it asserted the now-replaced "override the port on every register" behavior and stubbed `RTCPeerConnection`. Real coverage comes from the browser `p2pSetup` e2e.

sh3ll3x3c:
## Base changed: this PR is now stacked on #380 (`fix/browser-host-worker-boot`)

**Why.** The browser test Luka asked for here — a real `p2pSetup` two ways (main-thread auto-connect + worker bubble-up) — is being built as a real 2-peer WebRTC e2e in headless Chromium. Each peer spawns an SDK **worker host**, and on this branch that worker throws **`process is not defined`** booting the EVM — exactly the bug **#380** fixes with `nodeGlobalsShim`. So this e2e literally cannot run without #380's worker-boot fix.

**The upside.** That same e2e is precisely the reproduce-first test #380 needs: booting the SDK worker in a real browser is what decides whether #380's shim (and worker-scope transport detection) is warranted. If the worker only boots with the shim, #380 is proven necessary; if it boots without, #380 should be trimmed. The test settles the open question rather than assuming an answer.

**Merge order becomes #380 → #376.**

**Rebase note.** This branch previously carried duplicate copies of #378's and #379's commits (same messages, different SHAs — the #379 copy still had the old mocked `RuntimeContractEvents.test.ts`). The rebase onto #380 dropped those; #376 now contains only its own bridge + browser-discovery work on top of #380.

Landed so far on this PR (see replies on the individual threads): the ref-count/set-once bridge-ownership fixes (momo), the main-thread auto-install (Luka), the `is/does/has` rename, plus the browser-side local discovery infra (LocalDiscoveryServer over a relay hub) the e2e needs. The 2-peer WebRTC e2e itself is in progress.

test/browser/webrtc-service-smoke.js
sh3ll3x3c: Landed both scenarios as a real browser e2e (`yarn test:browser:webrtc`), no mocks — real `p2pSetup`, EVM, handshake and WebRTC against a hardhat node:

1. **Main thread + SDK-thread flag:** `p2pSetup` runs on the page main thread with `RUN_SDK_IN_THREAD`. Each SDK worker surfaces its `webRTCBridgePort` and `p2pSetup` auto-installs it on the main thread. Two such peers rendezvous through a local-discovery relay hub, handshake over `BrowserLocalTransport`, and both `RTCPeerConnection`s reach `connected` on the main thread — the worker → main direction.
2. **Inside a worker (bubble-up):** `p2pSetup` runs inside an app worker (inline host), so it does **not** auto-install; the worker bubbles `p2pInstance.webRTCBridgePort` up to the page, which installs it by hand with `installWebRTCMainThreadBridge` and drives WebRTC on the workers' behalf. Both peers connect through the manually-installed bridges.

The assertion spies `RTCPeerConnection` on the page main thread: worker peers can't construct one, so a main-thread connection reaching `connected` proves the bridge carried the negotiation.

test/evm/P2pRuntimeClient.test.ts
sh3ll3x3c: Removed in 68a1b89d. The `webRTCBridgePort` surfacing is now covered for real by both browser e2e paths: path 1 asserts each worker-hosted peer surfaces the port and it drives an actual worker → main `RTCPeerConnection` to `connected`; path 2 bubbles that surfaced port up and installs it by hand. The fake-`RuntimePort` assertion is superseded.

lukastanisic99 (approved):

PR #379, opened 2026-06-30
fix(runtime): deliver contract events to main-thread subscribers

sh3ll3x3c:
## Merge order — read first

2. 👉 **(this PR)** contract-event delivery fix → `dispute`
4. WebRTC main-thread bridge port (#376) → `dispute`
5. browser host-worker boot fixes (#380) → `dispute`

The SDK PRs are independent of each other. The bridge port (#376) and the
browser host-worker boot fixes (#380) both merge before the downstream app
consumer in step 6 (which needs this fix, the bridge port, and the boot fixes).

## Summary

The main-thread runtime contract runs on `ClientP2pSigner`, whose `provider` was
`null`. ethers v6 `Contract.on(...)` requires a `runner.provider` to register a
subscription (`getSub` throws `UNSUPPORTED_OPERATION` otherwise), so the call
threw and was swallowed by callers' `void` — and host-forwarded contract events
re-emitted via `dispatchContractEvent` reached no listeners.

This adds a `NoopEventProvider` (an `AbstractProvider` whose subscription methods
are no-ops) as the client signer's provider, so `Contract.on(...)` registers a
local subscription without starting a `getLogs` poller; forwarded events are then
delivered. Affects both inline and worker runtime modes.

Note: events are forwarded by name, so name-based and unindexed
`contract.filters.X()` subscriptions are delivered; an indexed-arg filter
`contract.filters.X(indexedValue)` resolves to a different ethers tag and is not
delivered (documented in `dispatchContractEvent`).

## Test Plan

- New regression test `test/evm/RuntimeContractEvents.test.ts`: a host-forwarded
  contract event reaches a `.on` subscriber. Verified it **passes with the fix
  and fails without it** (`contract runner does not support subscribing`).
- `yarn tsc --noEmit -p tsconfig.json` clean.


test/evm/RuntimeContractEvents.test.ts
```diff
@@ -0,0 +1,71 @@
...
+        // provider, ethers rejects this `on(...)` ("contract runner does not
+        // support subscribing") and the subscription never registers, so the
+        // forwarded event below would reach no listener.
+        await contract.on(contract.filters.ValueChanged(), (value: bigint) => {
+            received.push(value);
+        });
+
+        // The host detected a contract event and forwarded it over the port;
+        // dispatchContractEvent re-emits it on the main-thread contract.
+        fake.deliver({
```
momoshell: Could this test use a real `RuntimePort/MessageChannel`, or have the fake structured-clone messages before delivery? 

This direct deliver path bypasses the runtime-port serialization step. The host forwards raw `Object.values(event.args)`, and ethers tuple/struct event args can fail structured clone, so this currently proves the re-emit path for primitive args rather than actual worker/runtime-port delivery for complex event args.

lukastanisic99: yes exactly - harness for unit tests! Real contract events! If the MathContract doesn't have an event (it probably does), just add it and test it. Same comment across all PRs for tests :D

sh3ll3x3c: Done.

sh3ll3x3c: Rewritten as a harness e2e in 835fc233 (`test/e2e/E2E-RuntimeContractEvents.test.ts`): a real `MathStateMachine` emits a real `Addition` event from a real `add()` transition, forwarded across the real runtime port to a main-thread `.on` subscriber — so it exercises the actual serialization step the hand-crafted `contractEvent` message skipped (and in worker mode crosses structured clone). It fails without the client signer's event provider.

lukastanisic99 (approved):

PR #380, opened 2026-07-01
fix(runtime): boot the SDK host worker in a browser under RUN_SDK_IN_THREAD

sh3ll3x3c:
## Merge order — read first

2. contract-event delivery fix (#379) → `dispute`
4. WebRTC main-thread bridge port (#376) → `dispute`
5. 👉 **(this PR)** browser host-worker boot fixes → `dispute`

Independent of the other SDK PRs (touches only `Holepunch.ts` and the runtime
worker entry). Like #376, it is required by the downstream **browser** app, so
it must merge before the app migration (step 6). Order relative to #376 does not
matter.

## Summary

Two independent gaps kept the SDK-owned host worker from starting in a browser
under `RUN_SDK_IN_THREAD`:

- **Node globals in the worker EVM.** The EVM stack (`@ethereumjs/evm` and its
  bundled stream/debug deps) reads `process` (`process.env.DEBUG`,
  `process.nextTick`) and `global`, neither of which exists in a Web Worker, so
  the worker threw `process is not defined` before the host could boot. A
  minimal `nodeGlobalsShim`, imported first in the worker entry, installs them
  with `??=`; on Node the assignments no-op and it stays inert.

- **Browser-vs-node detection in a worker.** `Holepunch` selected its transport
  path via `typeof window`, which is `undefined` in a Web Worker, so the
  in-worker host fell through to the node DHT path and `@hyperswarm/dht` threw
  "not supported in browsers". It now also detects the worker scope
  (`WorkerGlobalScope`), mirroring `isWorkerRuntime()` in the WebRTC provider,
  so it takes the browser relay path. The relay callback also reads
  `globalThis.Hyperswarm` instead of `window.Hyperswarm` (absent in a worker).

## Test Plan

- `yarn tsc --noEmit -p tsconfig.json` clean (node + browser).
- Verified end-to-end in a browser: two peers with `RUN_SDK_IN_THREAD` boot the
  SDK host worker, connect to the DHT relay, and establish a direct WebRTC data
  channel. Before these fixes the worker threw `process is not defined`, then
  `hyperdht is not supported in browsers`.


src/evm/p2pRuntime/worker/nodeGlobalsShim.ts
```diff
@@ -0,0 +1,19 @@
...
+// place before the host boots the EVM. On Node these already exist, so the
+// `??=` assignments no-op and this stays inert.
+const scope = globalThis as unknown as {
+    global?: unknown;
+    process?: unknown;
+};
+
+scope.global ??= globalThis;
+
+scope.process ??= {
```
momoshell: This covers the “no process global” case, but not the common browser/bundler case where `process` already exists as a partial shim. Since this worker depends on `process.nextTick`, can we patch `env` and `nextTick` independently on an existing `process` object?

```suggestion
if (typeof scope.process !== "object" || scope.process === null) {
    scope.process = {};
}
const processShim = scope.process as {
    env?: Record<string, string | undefined>;
    nextTick?: (callback: (...args: unknown[]) => void, ...args: unknown[]) => void;
    browser?: boolean;
};

processShim.env ??= {};
processShim.nextTick ??= (
    callback: (...args: unknown[]) => void,
    ...args: unknown[]
) => queueMicrotask(() => callback(...args));
processShim.browser ??= true;
```

src/evm/p2pRuntime/worker/P2pRuntimeWorkerEntry.ts
```diff
@@ -1,3 +1,6 @@
+// Must run before any EVM/stream import pulls in Node globals.
+import "./nodeGlobalsShim";
```
momoshell, line 2: Can we add a browser-worker smoke for this entry path? The current browser worker coverage does not appear to exercise `RUN_SDK_IN_THREAD`/`P2pRuntimeWorkerEntry`, so the boot regression fixed by this PR would not be caught.

lukastanisic99:
> * **Node globals in the worker EVM.** The EVM stack (`@ethereumjs/evm` and its
>   bundled stream/debug deps) reads `process` (`process.env.DEBUG`,
>   `process.nextTick`) and `global`, neither of which exists in a Web Worker, so
>   the worker threw `process is not defined` before the host could boot. A
>   minimal `nodeGlobalsShim`, imported first in the worker entry, installs them
>   with `??=`; on Node the assignments no-op and it stays inert.

Not sure I follow/agree, since I didn't have problems running `@ethereumjs/evm` in the browser nor on the main thread, nor in a worker thread. Before this 'adapting the UI repo for the new SDK architecture` the UI explicitly created a thread and that thread ran the SDK normally - so I'm not really sure what's the problem you're hitting here 

> * **Browser-vs-node detection in a worker.** `Holepunch` selected its transport
>   path via `typeof window`, which is `undefined` in a Web Worker, so the
>   in-worker host fell through to the node DHT path and `@hyperswarm/dht` threw
>   "not supported in browsers". It now also detects the worker scope
>   (`WorkerGlobalScope`), mirroring `isWorkerRuntime()` in the WebRTC provider,
>   so it takes the browser relay path. The relay callback also reads
>   `globalThis.Hyperswarm` instead of `window.Hyperswarm` (absent in a worker).

Also here, it should use the relayed connection for browsers, regardless is it main thread or worker - I only had a problem of getting WebRTC to work in worker threads.

lukastanisic99:
> > * **Node globals in the worker EVM.** The EVM stack (`@ethereumjs/evm` and its
> >   bundled stream/debug deps) reads `process` (`process.env.DEBUG`,
> >   `process.nextTick`) and `global`, neither of which exists in a Web Worker, so
> >   the worker threw `process is not defined` before the host could boot. A
> >   minimal `nodeGlobalsShim`, imported first in the worker entry, installs them
> >   with `??=`; on Node the assignments no-op and it stays inert.
> 
> Not sure I follow/agree, since I didn't have problems running `@ethereumjs/evm` in the browser nor on the main thread, nor in a worker thread. Before this 'adapting the UI repo for the new SDK architecture` the UI explicitly created a thread and that thread ran the SDK normally - so I'm not really sure what's the problem you're hitting here
> 
> > * **Browser-vs-node detection in a worker.** `Holepunch` selected its transport
> >   path via `typeof window`, which is `undefined` in a Web Worker, so the
> >   in-worker host fell through to the node DHT path and `@hyperswarm/dht` threw
> >   "not supported in browsers". It now also detects the worker scope
> >   (`WorkerGlobalScope`), mirroring `isWorkerRuntime()` in the WebRTC provider,
> >   so it takes the browser relay path. The relay callback also reads
> >   `globalThis.Hyperswarm` instead of `window.Hyperswarm` (absent in a worker).
> 
> Also here, it should use the relayed connection for browsers, regardless is it main thread or worker - I only had a problem of getting WebRTC to work in worker threads.

But this is a great example of why we should use harness and actual SDK components for testing even unit tests and browser tests and all of the mocks and junk data is useless. If we did it like that, this divergence would have been caught and if you stumbled upon a new case that's not working, you write a test with harness to reproduce and patch

sh3ll3x3c:
Cross-ref: #376 has been rebased on top of this branch and is building the real 2-peer WebRTC `p2pSetup` browser e2e on top of it. That e2e boots the SDK worker host in a real headless-Chromium browser, so it directly exercises this fix — it's the reproduce-first test that decides whether the `nodeGlobalsShim` / worker-scope transport detection here is warranted (the worker booting only with the shim is the evidence). Will report the outcome back here once the e2e runs.

src/evm/p2pRuntime/worker/nodeGlobalsShim.ts
sh3ll3x3c: Done in `6b71f5b9` — each field (`env` / `nextTick` / `browser`) is now patched independently with `??=` on the existing (possibly partial) `process`, so a bundler shim that provides `env` but no `nextTick` gets `nextTick` filled in rather than left undefined. Also extracted the logic into a pure `applyNodeGlobalsShim(scope)` and added unit tests covering the partial-`process` case (existing `env` preserved, `nextTick` added, an existing `nextTick` not overwritten).

src/evm/p2pRuntime/worker/P2pRuntimeWorkerEntry.ts
sh3ll3x3c: Good call. `P2pRuntimeHost` eagerly pulls in the whole EVM/runtime graph (`@ethereumjs/evm`, the contract-executor factory, hyperswarm), so a real boot of `P2pRuntimeWorkerEntry` under `RUN_SDK_IN_THREAD` needs the full harness — a live node, deployed contracts, and the bundled worker — which the existing `test:browser:worker` harness (ContractExecutor / WebRTC-setup surfaces only) can't be cheaply extended to.

That end-to-end coverage lands in the PR stacked directly on top of this one (#376): `yarn test:browser:webrtc` (`test/browser/run-p2p-webrtc-e2e.mjs`) spins up a hardhat node, deploys, and boots the SDK host worker via `RUN_SDK_IN_THREAD` in real headless Chromium (`p2pWebRTCE2E.js` / `p2pWebRTCAppWorker.js`) — so the boot regression this PR fixes is caught in the stack rather than duplicated here. I did add a focused unit test for the shim itself (the field this PR actually fixes) in `6b71f5b9`.

lukastanisic99 (approved):

PR #377, opened 2026-06-30
avoid throw on legtimate race condition

daiagi:
there is a race condition between
- `StateManager.performaReduce` multicall to `updateStateSnapshotFork` -> event emitted -> `EventHandler.onStateSnapshotUpdated`

- `StateManager.performaReduce` -> `setGenesisState` -> `setLatestState` -> `unsafeSetLatestState` -> `storage.stateSnapshots.storeStateSnapshot`

sometimes, `EventHandler.onStateSnapshotUpdated` runs before `unsafeSetLatestState`    => the state snapshot is not yet in loacl storage -> `onStateSnapshotUpdated` throws

this shold not throw. 

this caused the test `"dispute.input.timeout.blockHeight = block whose calldata is on-chain; isForced=true → TimeoutCalldataPosted"`
to fail, due to honest peer (peer index 1) throwing

PR #381, opened 2026-07-01
refactor(e2e): overhaul parallel test runner + harness infra

lukastanisic99:
# Overhaul the e2e parallel test runner and harness infrastructure

Reworks how e2e tests are scheduled, how chain/discovery infra is provisioned, and how worker/thread health is observed — after chasing OOM-restarts and event-loop starvation under load.

## Scheduler (`scripts/e2e-parallel/`)

- Replaced the thread-cost model + persisted adaptive-factor metadata with a simple **1-admission-per-tick controller** gated on **live CPU utilization** (`os.cpus()` tick deltas, not the lagging `loadavg`) and **measured memory** (summed RSS of our processes with a running per-test average), bounded by `--target-load` (default `0.8`) and a memory fraction of total RAM.
- **Removed reruns** — failures surface immediately.
- New flags: `--slots N` (default 4), `--mem-limit-gb`, `--target-load`. Removed `--thread-factor`, `--thread-budget`, `--per-slot-node`, `--shared-node` (topology collapsed to `--slots`).
- Fixed a teardown double-count bug (idempotent infra teardown).

## Slots + harness infra

- **One uniform slot** = its own hardhat node + discovery + deploy-cache, provisioned in parallel before scheduling, assigned round-robin; account partitions (`E2E_SLOT_INDEX`) acquired per running test.
- **Harness is now indifferent to where the node comes from**: it uses passed-in `PROVIDER_URL` / `LOCAL_DISCOVERY_REGISTRY_URL`, or **starts its own** via the same shared `test/utils/nodeInfra.js` (one file, `.d.ts`-typed for the harness). Decoupled from `hre.ethers` — builds its own **WebSocket provider** + signers from the test mnemonic.
- Parallelized `createPeer`.

## SDK / workers

- Worker threads get **heap caps** (`resourceLimits`, `SCP_*_WORKER_MAX_OLD_SPACE_MB`) so a runaway test fails cleanly instead of OOM-restarting the machine; VM worker runs **swc transpile-only**.
- **All threads (main / sdk / vm) run the existing logger event-loop monitor**, gated by `EVENT_LOOP_DELAY_ERROR_THRESHOLD_SECONDS` (via config, not env). VM worker now gets real config + logger.

## Diagnostics / output

- Per-test line + run summary now report **startup / deploy / worker-boot times, per-thread event-loop delay peaks, CPU (avg/peak), memory, and per-slot block-gas** — communicated via `##E2E_TIMING##` stdout markers parsed by the runner.
- Colored scheduling output; **starvation failures print yellow** (load-induced) vs red (real bugs).

## Housekeeping

- Node-only files moved under `src/evm/node/` with `@platform` aliases so they never enter the browser build (both `tsc` configs pass); documented the rule in `AGENTS.md`.
- Added `.env.example` cataloguing all config + runner env vars.

## ⚠️ Behavior changes

- Every thread now **throws** past `EVENT_LOOP_DELAY_ERROR_THRESHOLD_SECONDS` (1s in tests), so worker starvation under heavy concurrency now **fails** the affected tests (yellow) instead of passing silently. Raise the threshold or lower `--target-load` / `--slots` if too aggressive.
- Reruns are gone; flaky tests aren't retried.

## Testing

- `tsc` clean for node and browser builds.
- Smoke-verified all modes (`--slots 0/1/N`, `±--sdk-thread`): default in-process, slot-provisioned, and harness-self-provisioned paths pass with clean teardown.


daiagi (approved):
looks good. I've got many failing tests 

image


After including the fix of PR #377 ,  Ive got:

image

PR #383, opened 2026-07-06
(test infra) ignore errors thrown from byzntine peers

daiagi:
### what

a peer we deliberately make misbehave throws expected errors in its detached
work, and those were failing the tests. now each error carries the address of
the peer it came from, and the harness ignores errors from peers the test
marked byzantine. honest-peer errors still fail the test.

https://trello.com/c/K6byL6EU

### how

each inline peer's handlers run inside an `AsyncLocalStorage` scope holding its
EVM address. when a detached promise is collected, we read the address back
from that scope and stamp it onto the eventual rejection error.
`TestSession.setFirstDetachedError` reads the stamp and skips errors from
marked peers. Worker peers skip the tracing:
one thread = one peer, so the client stamps its own address.

### also fixes

`expectFirstDetachedError` registered its ignore pattern before waiting ->
late-arriving errors were dropped by their own pattern. now waits first.

### tests

new `E2E-ByzantineErrorAttribution` (ignored byzantine error + recorded honest
error), passing inline and in worker mode.

### DIFF
--- a/test/fixtures/PeerTestHarness.ts
+++ b/test/fixtures/PeerTestHarness.ts
@@ -1027,7 +1027,6 @@ export class PeerTestHarness<
         return new Set(
             [
                 ...this.context.maliciousPeerIndices,
-                ...this.context.leftChannelPeerIndices
             ].map((i) => this.getPeer(i).address)
         );
     }
--- a/test/fixtures/PeerTestHarness.ts
+++ b/test/fixtures/PeerTestHarness.ts
@@ -27,8 +27,11 @@ import {
     LocalDiscoveryServer,
     Logger,
     EventBarrier,
-    createEthersResultProxy
+    createEthersResultProxy,
+    getErrorPeerAddress,
+    maybeStampErrorWithPeerAddress
 } from "@/utils";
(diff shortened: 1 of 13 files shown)


test/fixtures/PeerTestHarness.ts
```diff
@@ -986,20 +992,52 @@ export class PeerTestHarness<
...
+    /**
+     * EVM addresses of peers a scenario marked byzantine/malicious or
+     * left-channel. Errors originating on these peers are expected and
+     * suppressed.
+     */
+    private byzantinePeerAddresses(): Set<string> {
+        return new Set(
+            [
+                ...this.context.maliciousPeerIndices,
+                ...this.context.leftChannelPeerIndices
```
lukastanisic99: We should not include peers who left for 2 reasons:
1) this is set in advance and we want to observe errors that happen during leaving (if any)
2) We want proper cleanup, a peer that leavs should dispose and tairdown - if it's not the case we want to see the error

daiagi:
```suggestion
```

lukastanisic99 (approved):

PR #384, opened 2026-07-07
Part 2 of getting back to green

lukastanisic99:
# Wrong-fork block recovery, exact-target sync pinning, VM-rollback guard

## Summary

Three source fixes plus the test infrastructure and coverage to prove them:

1. **Wrong-fork block recovery** — a height-0 block on a fork we have no genesis
   snapshot for (a peer reduced before us, a join we haven't processed, or plain
   junk) used to crash-drop in the validation pipeline
   (`wrongGenesisDetected` → `createWrongGenesisProof` →
   `getGenesisSnapshotByForkId(...)!` → TypeError: no proof, no dispute, no
   disconnect, no requeue). Wrong-fork handling now lives in `BlockQueueManager`,
   outside the validation mutex.
2. **Exact-target sync pinning** — `generateSyncPayload` proved the responder's
   *latest* height instead of the requested target when the requested height was
   `0` (`_blockHeight || latestBlockHeight`, and `0` is falsy). One-line
   `??` fix.
3. **VM-rollback-after-commit guard** — `onBlockConfirmation` could restore the
   VM to pre-block state while storage already held the advanced block, if a
   post-commit side effect threw.

### Wrong-fork recovery (`BlockQueueManager`, `ValidationService`, strategies, `FraudProofService`)

- **Ingest fork gate.** A mismatched block first tries `tryRecoverForkMismatch`,
  pre-gated on the *current* fork being disputed (`didIDispute || on-chain` —
  `reduceLocally` does not self-guard, so junk gossip must not trigger on-chain
  probes or dispute attempts). It awaits `reduceLocally`; if the block's fork
  became ours (the late-to-reduce race), it proceeds through the normal
  queue → pipeline path. Otherwise the block is **quarantined**: queued under
  its own fork → timeout scheduled → sync requested from suppliers → never
  scheduled → connection kept. It drains only if a fork transition catches us up
  (post-transition `tryExecuteFromQueue`); the timeout evicts it otherwise, and
  peers who can't prove the fork are cut by the sync flow.
- **Fork gates at every feed point**, ending with a last-instant re-check inside
  `executeQueuedEntry` (a transition can land between scheduling and execution).
  **Sync only points forward**: a stale-fork entry — one whose fork we've moved
  *past* — is dropped silently (transitions are forward-only, so there's nothing
  to sync, and a sync request would fail against honest stragglers and blacklist
  them). Sync is requested only at the unknown-fork ingest quarantine.
- **`scheduleQueueTimeout` → remaining-window semantics.** It schedules the
  remainder of `firstSeenAt + agreementTime` and returns whether it scheduled,
  so duplicate copies and restores can't stretch an entry to ~2× the window; on
  a `false` return it leaves any live handle untouched.
- **`wrongGenesisDetected(entry)`** across all strategies (`checkConflictingBlock`
  is entry-based to thread source attribution). The block strategy adds a
  missing-genesis guard: no proof, no dispute (nothing to commit to), blacklist
  the entry's source peers, `DISCONNECT`, and a loud "missing genesis reached
  validation despite fork gate" warn.
- **`createWrongGenesisProof`** replaces the `!` with a descriptive throw,
  protecting the dispute/spectating callers from the silent TypeError.

### VM-rollback guard (`StateManager`)

`success()` gains an `onBlockCommitted` hook, invoked immediately after
`storeBlock`. `onBlockConfirmation` clears its rollback marker there, so a throw
in the post-commit side effects (gossip, `startMaybeExitOnChain`,
`getNextToWrite`) can no longer rewind the VM behind the stored block.

### Exact-target sync (`SpectateService`)

`generateSyncPayload` pins the state proof with `_blockHeight ?? latestBlockHeight`
(both the lookup and the debug log). A requested height of `0` is a valid target
and must be proved exactly, not fall through to the responder's latest.

## Test infrastructure

- **Parallel runner** logs each run to a fresh `./logs/run-N/` (history retained
  for cross-run comparison); only the current run's dir is cleared, an explicit
  `--logDir` is used/cleared as-is, and dirs outside `./logs` need
  `--allow-logdir-purge`.
- **Harness (reusable staging on `stub`/`rpcStub`/`byzantine`/`spectate`)** —
  `spectate.startSync` (drives the real host-side sync; the concurrent-dedup
  test folded onto it) and `generateSyncPayload`'s named `{ encodedSyncPayload }`
  return; `stub` hold/record/pause methods for reduction-race staging; `byzantine`
  bogus-fork-block-zero crafting; `rpcStub` combos; `execOnHost` now takes RPC
  options (timeout override); an `EventActions` barrier fix (deadline flag
  instead of re-reading `Date.now()`).

## Tests

- **`E2E-BlockQueueManager`** — recover-at-ingest race; unknown-fork quarantine
  (never validated, evicted at the deadline, the failed sync punishes the
  supplier); recovery with a still-bogus block; drain-after-transition; plus
  host-side unit-scope tests for the queue-timeout window math, the stale-fork
  gate (drops, never validates/syncs), and both `wrongGenesisDetected` branches
  through the real strategy (which also covers the fraud-proof descriptive
  throw).
- **`E2E-Spectate`** — exact-target payload pinning at heights 0 and 1 (the
  regression for the `??` fix, verified failing pre-fix / passing post-fix);
  the concurrent-sync-dedup test on the new `startSync` endpoint.
- **`E2E-SpectateStaleProofGuard`** — a participant requesting a target behind
  the on-chain snapshot is aborted by the stale bound and blacklists the
  responder.
- **`E2E-StateSnapshots`** — two reduction-race tests proving setState isn't
  re-emitted when an old-fork reduction (held timer, or already-entered/paused)
  completes after the snapshot-event reduction.

## Docs

`AGENTS.md` — AI PR reviews must verify adherence to all root + test guidelines;
the run-N log layout; a `scripts/` CLI exception to the no-`console.*` rule.
`test/AGENTS.md` — the chai-comparator-on-contract-objects pitfall (recursive
ethers stringify → multi-second stall); record-only host probes as the one
sanctioned stub shape; "test files contain tests, not helpers" (reusable staging
lives on the control services); no per-test restore hooks (sessions are
per-test).

## Notes on the exact-target regression boundary

The full sync pipeline normalizes an over-proved height (a requester-side
acceptance test passes with the bug and the fix alike), so the height-0
regression is pinned at the source — `generateSyncPayload` height pinning
(A1/A2) — rather than as a full-sync e2e. Requester-side verifier behavior is
covered by the stale-bound test plus the existing spectate suite. Adjacent
follow-up (does not block this change): the verifier's exact-height check
(step 6) appears ineffective in practice; whether targeted syncs should enforce
the exact height downstream is a separate decision.

## Status

- Typecheck clean; unit suite green.
- New e2e tests pass in isolation and (for the queue tests) under
  `test:e2e:parallel`. Dispute-heavy e2e remain load-sensitive per the chronic
  epicenter tracked separately; no regression attributable to these changes.


sh3ll3x3c (commented):
Reviewed via four parallel verification passes against the PR head (core recovery, StateManager/Spectate, harness, tests/scripts/docs), every finding checked against the actual code.

**Verdict:** all five claimed source behaviors are implemented as described, and the key regression tests genuinely fail without their fixes (the height-0 pin test, the ingest-recovery tests, the missing-genesis guard test). The inline comments flag 3 blocking issues (a mutex re-entrancy deadlock through `tryRecoverForkMismatch`, pre-genesis blocks blacklisting honest peers at channel open, and an empty `--logDir=` wiping the CWD), 2 majors, and a set of minors. Nits omitted by request.

**Out-of-diff findings** (can't be anchored inline):

- `src/agreementManager/AgreementManager.ts:95-126` — exact-target pinning is still not exact across participant-set change points: the milestone loop consumes `getChangePointsInRange(forkId)` unbounded by the requested height, so a pin to height 0 on a fork with a change point at height 5 returns a payload whose `latestFinalizedSnapshot` proves height ≥ 5. Bound the change-point range by `blockHeight`.
- `src/stateManager/StateManager.ts:1678` — the authoring path (`playTransaction`) uses `success()` with no rollback marker, so a throw before `storeBlock` (e.g. `block.sign`) leaves the VM advanced past the last stored block — the same dirty-VM hazard this PR fixes on the confirmation path. Worth a follow-up.

**Verified clean, for the record:** the remainder-window timeout math (including the `<= 0` boundary refusal); the timeout-owns-the-entry dequeue-first design (no double execution); the rollback marker lifecycle (per-invocation, cleared exactly once, `storeBlock`-throw and post-commit-throw paths both consistent, mutex serializes confirmations); the EventActions deadline-flag fix (closes the early-timer-fire race, no lost-signal window); per-test session teardown makes cross-test stub leakage structurally impossible; the crafted byzantine block-zero is genuinely authentic and crosses the port encoded; run-N allocation is concurrent-race-safe and the `..`-traversal gating is correct; RPC method/service split, `encoded*` naming, and the no-`console.*` rule are followed throughout.

src/stateManager/BlockQueueManager.ts
```diff
@@ -112,6 +112,36 @@ export default class BlockQueueManager {
                 return true;
             }
 
+            if (
+                block.forkId !== this.stateManager.forkId &&
+                !(await this.tryRecoverForkMismatch(block))
+            ) {
```
sh3ll3x3c: **[blocking]** Blocks arriving before local genesis is set (`stateManager.forkId` still `NULL` — the case the pre-existing `channelNotOpened` handler exists for) now take this unknown-fork quarantine and fire `requestSync` at the honest sender/author. `SpectateService.sync` blacklists on *any* failure, including a plain `agreementTime` timeout, so a co-participant who authors block 0 slightly before our `setGenesisState` completes can get disconnected and blacklisted at channel open — and a *successful* sync races `applySyncResponse`'s `setLatestState` against our own join-time genesis setup.

Suggest treating `forkId === NULL` as "not open yet" here: queue without sync, like the old path.

src/stateManager/BlockQueueManager.ts
```diff
@@ -112,6 +112,36 @@ export default class BlockQueueManager {
...
+                // timeout evicts it otherwise. Suppliers must prove the fork
+                // via sync or get disconnected there.
+                const hash = this.stateManager.storage.queues.queueBlock(
+                    block,
+                    { senderAddress: options?.senderAddress }
+                );
+                this.scheduleQueueTimeout(hash);
+                const entry =
+                    this.stateManager.storage.queues.getQueuedEntry(hash);
+                if (entry) this.requestSync(entry);
```
sh3ll3x3c: **[major]** This branch runs again on every received *copy*: duplicates merge in `queueBlock` and re-enter here, so each copy re-fires `requestSync` to **all** accumulated `sourcePeers` plus the author. `inFlightByPeerAddress` only dedups while a sync is in flight — each completed round re-arms a fresh sync exchange and another blacklist-on-timeout roll against every past supplier.

Only request sync when the entry was newly created — e.g. check `getQueuedEntry` before `queueBlock`, or consume `scheduleQueueTimeout`'s currently-unused boolean return.

src/stateManager/BlockQueueManager.ts
```diff
@@ -125,8 +155,42 @@ export default class BlockQueueManager {
...
+     * - junk gossip must not trigger on-chain probes or dispute attempts.
+     */
+    private async tryRecoverForkMismatch(block: Block): Promise<boolean> {
+        const currentForkId = this.stateManager.forkId;
+        if (
+            !(await this.stateManager.isForkDisputed(
+                currentForkId,
+                block.channelId
+            ))
+        ) {
```
sh3ll3x3c, line 193: **[minor]** Two notes on this gate:

1. `isForkDisputed` resolves against the **local mirror** diamond (`localDiamondContract` on the in-process executor), not the chain — if the mirror hasn't ingested the dispute event yet, recovery is skipped and an honest ahead-of-us peer gets quarantine-sync-probed instead (with the blacklist-on-timeout hazard). The freshness assumption deserves a comment here; the PR description's "on-chain" wording overstates it.
2. Once the fork *is* genuinely disputed — a public on-chain event — every mismatched-fork junk block from any peer triggers a fresh `reduceLocally` whose `isKillPeriodExpired` is a real chain read; the single-flight map only merges concurrent runs, so a byzantine peer flooding unique self-signed blocks drives ~1 chain RPC per block for the whole kill window. Caching the "not expired until `killPeriodEnd`" result would make the probe O(1) per window.

src/stateManager/BlockQueueManager.ts
```diff
@@ -125,8 +155,42 @@ export default class BlockQueueManager {
...
+            return false;
+        }
+        try {
+            await this.stateManager.reduceLocally(currentForkId);
```
sh3ll3x3c: **[blocking]** This can deadlock the whole StateManager. `reduceLocally` → `performLocalReduction` → `setGenesisState` → `setLatestState` → `withMutex` takes the non-reentrant StateManager mutex — but `ingestBlockConfirmation` is awaited *while that mutex is held* from the deferral re-ingest paths in `ValidationService` (the two `ingestBlockConfirmation` re-ingests inside `onBlockConfirmation`, which locks at `StateManager.ts:1194`).

Concrete path: the dispute pipeline (`runStateProofBlocksThroughPipeline` → `onBlockConfirmationStruct`) validates state-proof blocks whose `forkId` can differ from the current fork; `DisputeValidationStrategy` skips the disputed-fork early-out; the deferral branch re-ingests under the held mutex; this gate fires; and if the current fork is disputed with the kill period expired, the inner `setLatestState` lock never resolves — every future `onBlockConfirmation`/`playTransaction` queues forever.

Fix options: make the ValidationService deferral re-ingests scheduled tasks, or thread a "no recovery" option through re-ingest so `tryRecoverForkMismatch` is skipped on that path.

src/stateManager/BlockQueueManager.ts
```diff
@@ -320,6 +411,26 @@ export default class BlockQueueManager {
     }
 
     private async executeQueuedEntry(entry: QueuedBlockEntry): Promise<void> {
+        if (entry.block.forkId !== this.stateManager.forkId) {
```
sh3ll3x3c: **[minor]** This re-check isn't quite last-instant: `onBlockConfirmation` waits on the StateManager mutex *after* this gate, so a fork transition completing during that wait still sends a stale-fork entry through full validation. Block/Spectating strategies get rescued downstream by the disputed-fork check only because dispute-driven transitions leave the old fork disputed — `DisputeValidationStrategy` skips that check entirely (this residual window is what enables the re-entrancy deadlock flagged at `tryRecoverForkMismatch`). A fork re-check after mutex acquisition inside `onBlockConfirmation` would close it.

src/stateManager/validationStrategy/BlockValidationStrategy.ts
```diff
@@ -144,8 +144,31 @@ export default class BlockValidationStrategy extends AValidationStrategy {
...
+                "Missing genesis reached validation despite fork gate - blacklisting sources",
+                {
+                    blockAuthor: block.author,
+                    blockForkId: block.forkId,
+                    sourcePeers: Array.from(entry.sourcePeers)
+                }
+            );
+            for (const peer of entry.sourcePeers) {
+                this.p2pManager.disconnectAndBlacklistPeerByEvmAddress(peer);
+            }
```
sh3ll3x3c: **[minor]** The guard blacklists only `entry.sourcePeers` — the signer of the junk height-0 block (`block.author`) escapes if it isn't among the transport sources. `requestSync` includes the author; this path should too.

src/stateManager/validationStrategy/DisputeValidationStrategy.ts
```diff
@@ -146,8 +146,9 @@ export default class DisputeValidationStrategy extends AValidationStrategy {
         return BlockValidationResult.DISPUTE;
     }
     public async wrongGenesisDetected(
-        block: Block
+        entry: QueuedBlockEntry
     ): Promise<BlockValidationResult> {
+        const block = entry.block;
         // TODO - here we have to kill the dispute, since the dispute contains incorrect state
         const hash = this.fraudProofService.createWrongGenesisProof(block);
```
sh3ll3x3c, line 240: **[minor]** This path is still unguarded: it reaches `createWrongGenesisProof` without the missing-genesis check the block strategy gained, so a missing genesis here now throws the new descriptive error instead of being handled. Acceptable as defense-in-depth, but inconsistent with the guard's rationale — worth either mirroring the guard or noting why this path can't see a missing genesis.

src/rpc/services/spectate/SpectateService.ts
```diff
@@ -572,15 +572,18 @@ class SpectateService extends ARpcService<SpectateServiceRpcMethods> {
         const latestBlockHeight =
             stateManager.storage.blocks.getNextBlockHeight(forkId) - 1;
 
-        // There are blocks, so we can do a same-fork update
+        // There are blocks, so we can do a same-fork update.
+        // `??` not `||`: a requested height of 0 is valid and must be pinned,
+        // not fall through to the latest block.
+        const targetBlockHeight = _blockHeight ?? latestBlockHeight;
```
sh3ll3x3c, line 598: **[major]** `targetBlockHeight` is remote-supplied (`syncRequest.blockHeight`) and reaches `tryGetStateProof` → the DESC iterators in `BlockStorage`, which loop `for (let height = start; height >= 0; height--)` regardless of `maxHeight`. A malicious spectator sending `blockHeight: Number.MAX_SAFE_INTEGER` drives a ~9e15-iteration synchronous loop that hangs the responder's event loop. Pre-existing (`||` forwarded any nonzero height too), but since this line is the fix site: clamp to `latestBlockHeight` and reject negatives/non-integers.

Related, just above the diff: line 455 `const forkId = _forkId || stateManager.forkId;` is the same falsy pattern — not exploitable today (fork IDs are hashes), but worth `??` for consistency with this fix.

scripts/e2e-parallel/argParser.js
```diff
@@ -64,6 +68,7 @@ function parseCliArgs(argv) {
             arg.startsWith("--dir=")
         ) {
             options.logDir = arg.split("=").slice(1).join("=");
```
sh3ll3x3c: **[blocking]** An empty value slips through here: `--logDir=` (e.g. `--logDir="$E2E_LOG_DIR"` with the variable unset in CI) yields `""`, `path.resolve("")` is the CWD, and `safeEmptyDir` then `rmSync`s **every entry of the repo root** when `--allow-logdir-purge` is passed; `--logDir .` does the same. The space form above (`if (next)`) also swallows a following flag as the dir name (`--logDir --allow-logdir-purge`).

Reject empty/`.`-resolving values and dash-prefixed `next` (like the `--grep` handling does).

scripts/e2e-parallel/logging.js
```diff
@@ -92,12 +92,45 @@ function parseTimings(text) {
     };
 }
 
+// The default ./logs dir and anything under it (run-N dirs) are safe to
+// purge/clean without the explicit allow flag.
+function isWithinDefaultLogDir(resolved) {
+    const expected = path.resolve(DEFAULT_LOG_DIR);
+    return resolved === expected || resolved.startsWith(expected + path.sep);
```
sh3ll3x3c: **[minor]** Lexical prefix check: if `./logs` is a symlink (scratch volume, `$HOME`), an explicit `--logDir logs/x` auto-purges the symlink's *target* with no `--allow-logdir-purge` required. `fs.realpathSync` both sides before comparing.

scripts/e2e-parallel/logging.js
```diff
@@ -115,8 +148,7 @@ function safeEmptyDir(dirPath, allowLogdirPurge) {
 
 function cleanupNonErrorLogs(logDir, allowLogdirPurge) {
     const resolved = path.resolve(logDir);
-    const expected = path.resolve(DEFAULT_LOG_DIR);
-    const canAutoPurge = resolved === expected;
+    const canAutoPurge = isWithinDefaultLogDir(resolved);
     const allowUnsafe = allowLogdirPurge === true;
 
     if (!canAutoPurge && !allowUnsafe) {
```
sh3ll3x3c, line 235: **[minor]** Both gate warnings (here and in `safeEmptyDir` above) say "Set ALLOW_LOGDIR_PURGE=1 to allow", but nothing reads that env var — the only gate is the `--allow-logdir-purge` CLI flag (which is also what AGENTS.md documents). Fix the message or honor the env var.

test/fixtures/customRpc/harnessControl/services/stub/StubRpcMethods.ts
```diff
@@ -565,6 +574,357 @@ export class StubRpcMethods extends ARpcMethods<P2PManager<HarnessControlRpc>> {
...
+            );
+        }
+
+        this.service.pausedTryReduce = {
+            targetForkId: forkId,
+            entered: false,
+            released: false,
+            settled: false,
+            inside: false
+        };
```
sh3ll3x3c: **[minor]** A second `stubPauseTryReduceAtKillPeriod` call replaces `pausedTryReduce` wholesale while a previous call may be parked awaiting the old state's `release` resolver — `releasePausedTryReduce` only resolves the *new* state's resolver, stranding the old paused `tryReduce` until the bounded dispose wait. Guard against re-install while `entered && !settled`, or release the old state before replacing it.

test/fixtures/customRpc/harnessControl/services/stub/StubRpcMethods.ts
```diff
@@ -565,6 +574,357 @@ export class StubRpcMethods extends ARpcMethods<P2PManager<HarnessControlRpc>> {
...
+                    state.error =
+                        error instanceof Error ? error.message : String(error);
+                }
+            );
+            return promise;
+        }) as PrivateTryReduceHost["tryReduce"];
+
+        localDiamond.isKillPeriodExpired = (async (...args: unknown[]) => {
+            const state = this.service.pausedTryReduce;
+            const requestedForkId = String(args[1]);
```
sh3ll3x3c: **[minor]** `isKillPeriodExpired` is a public typechain-typed surface, but this wrapper casts to `(...args: unknown[])` and extracts the fork positionally via `String(args[1])` — if a parameter is ever added/reordered before `forkId`, this compiles silently and the pause never engages (the test dies at its `waitFor` timeout instead of at compile time). Type the wrapper's params off the real method (`Parameters<typeof localDiamond.isKillPeriodExpired>`), keeping the cast only on the assignment.

test/fixtures/customRpc/harnessControl/services/stub/StubRpcMethods.ts
```diff
@@ -565,6 +574,357 @@ export class StubRpcMethods extends ARpcMethods<P2PManager<HarnessControlRpc>> {
...
+        }) as KillPeriodHost["isKillPeriodExpired"];
+
+        return true;
+    }
+
+    public startPausedTryReduce(forkId: ForkId): boolean {
+        if (!this.service.pausedTryReduce) return false;
+        const host = this.service.sm as unknown as PrivateTryReduceHost;
+        void host.tryReduce(forkId);
+        return true;
```
sh3ll3x3c: **[minor]** A `forkId` that doesn't match the staged `targetForkId` still returns `true` (claiming a paused reduction was started) and fire-and-forgets an unwrapped-path `tryReduce`, whose rejection surfaces as an unhandled host rejection blamed on an unrelated point in the test. Check `forkId === this.service.pausedTryReduce.targetForkId` and return `false` on mismatch.

test/fixtures/customRpc/harnessControl/services/stub/StubRpcMethods.ts
```diff
@@ -565,6 +574,357 @@ export class StubRpcMethods extends ARpcMethods<P2PManager<HarnessControlRpc>> {
...
+        if (replay) {
+            for (const args of held) {
+                void (restored as (...a: unknown[]) => Promise<void>)(...args);
+            }
+        }
+        return true;
+    }
+
+    /** `reduceLocally` counts calls and resolves undefined (nothing to reduce). */
+    public stubReduceLocallyNoop(): boolean {
```
sh3ll3x3c: **[minor]** Unlike the sibling record stubs (`stubCountSpectateRequests`, `stubRecordDisputeAckRequest`, `stubRecordUnsafeSetLatestState`), this installer — and `stubRecordReduceLocally` / `stubRecordSpectateSync` below — never zeroes its counter on install, so an install→restore→reinstall within one test asserts against accumulated counts. Reset the counters in the installers.

test/e2e/E2E-BlockQueueManager.test.ts
```diff
@@ -285,4 +304,936 @@ describe("E2E: BlockQueueManager", function () {
...
+
+        it("recovers the first reduced-fork block by reducing locally at ingest", async function () {
+            this.timeout(90000);
+
+            const h = TestSession.getHarness();
+            const targetPeerIndex = 1;
+            const maliciousPeerIndex = 2;
+            const reducingHonestPeers = [0, 3];
+
+            await h.lifecycle.start(4, maliciousPeerIndex, {
```
sh3ll3x3c: **[minor]** `maliciousPeerIndex` is being passed as `MathLifecycleActions.start`'s `transitionCount` parameter — it only works because the index happens to equal the 2 warm-up transitions the dispute staging needs, and changing the index would silently change the staging. Same pattern at lines 555 and 662 here, and in `E2E-StateSnapshots.test.ts` (119, 198). Pass a literal `2` (or a named `transitionCount`).

test/e2e/E2E-BlockQueueManager.test.ts
```diff
@@ -285,4 +304,936 @@ describe("E2E: BlockQueueManager", function () {
...
+                waitForProcessed: false
+            });
+
+            // Queued unscheduled: the pipeline never ran, sync was requested.
+            expect(
+                await h
+                    .control(observer)
+                    .query.isBlockQueued(bogusBlock.hash)
+                    .request()
+            ).to.equal(true);
```
sh3ll3x3c, line 506: **[minor]** This immediate `isBlockQueued === true` assertion races the entry's own eviction timer: with `agreementTime: 2` it fires 2s after ingest, so a >2s event-loop stall between the ingest RPC returning and this query (exactly the multi-second starvation this PR documents in test/AGENTS.md) evicts first and fails spuriously under parallel load. Use `agreementTime: 4` like the sibling test below.

lukastanisic99:
@sh3ll3x3c 


# Timeout-driven block-queue sync, detached fork recovery, and PR #384 review fixes

## Summary

Reworks the block-queue / spectate-sync paths so that fork recovery and sync probing happen safely and
without amplification, and folds in the review comments from PR #384. Two headline behavior changes:

1. **Sync moves to the queue timeout.** A block on a non-current fork is queued but never sync-probed
   on arrival; the timeout is the *sole* sync probe and decides per fork class: drop a fork we've
   provably moved past, sync-probe an unknown one. This removes the arrival-time false-positive that
   could blacklist an honest co-participant at channel open.
2. **Spectate sync is mutual cooperation.** A sync request must be answerable with a valid proof; any
   other response (or absence) mutually blacklists both peers. There is no no-penalty "unprovable"
   escape — a peer that won't/can't prove what it was asked is cut (and cuts back).

## What changed

### Block queue (`BlockQueueManager`, `StateManager`, `QueueStorage`, strategies)
- **Detached, gated, coalesced fork recovery.** A mismatched block may just mean we're late to reduce
  our own disputed current fork. Reducing takes the StateManager mutex, and ingest can run *under* that
  mutex (dispute-pipeline re-ingest), so recovery now runs **detached** — awaiting it inline would
  deadlock. A per-fork gate (`recoveryScheduledForFork` + `recoverySuppressedUntil`) plus a memoized
  kill-period read bound a junk-fork flood to **one task + one chain read per fork per window**.
- **O(1) stale-fork detection.** `isKnownStaleFork` treats a non-current fork as known-past if it is
  disputed, or we hold its genesis snapshot or any of its blocks — no chain-ancestry walk. Ambiguity
  defaults to sync, never a silent drop.
- **Pure queue storage + manager-owned restore.** `QueueStorage` only mutates data; `restoreQueuedEntry`
  owns (re)scheduling — if the block became stored while dequeued it merges instead of waiting the
  timeout; a past-deadline restore schedules an async eviction task (never inline under the held mutex).
  Strategy not-ready hooks route through it (and no longer sync directly).
- **After-mutex fork re-check as a pipeline side effect.** A transition can land while a queued entry
  waits for the mutex; `onBlockConfirmation` handles that inline (restore unknown / drop known-past)
  rather than returning a special result.
- **Order-safe structural source cap.** `QueueStorage` bounds retained `sourcePeers`/`signatureSources`
  per entry with an overflow marker; a valid block is never invalidated by a junk-first flood.

### Spectate sync (`SpectateService`, `SpectateRpcMethods`)
- Mutual-blacklist model (see summary): `generateSyncPayload` returns a valid `SyncPayload` or
  `undefined`; the responder cuts the requester on `undefined` and the requester cuts the responder on
  any rejection.
- **Remote-height DoS hardening:** cheap shape-validation of the requested height *before* any chain
  work; the `getIterator` DESC branch is clamped to `maxHeight`; an above-latest target is never
  silently downgraded to a `latest` proof.

### Dispute validation (`DisputeValidationStrategy`)
- `wrongGenesisDetected` throws `"Unexpected genesisSnapshot missing"` when the genesis is absent — an
  invariant violation, since the upstream Solidity header-mismatch check guarantees block[0] is on the
  disputed fork by the time we reach here.

### Destructive-tooling guards (`scripts/e2e-parallel`)
- `--logDir` rejects empty/`.`/dash-prefixed values; `safeEmptyDir`/`cleanupNonErrorLogs`/`nextRunDir`
  refuse the repo root / fs root, checking the **real** (symlink-followed) path so a `./logs -> .`
  symlink can't purge or scatter runs at the repo root.

### Conventions (`AGENTS.md`)
- Added a "Class layout" rule: all fields at the top, then all methods.

## Testing

- **Unit** (`yarn test`): structural source cap incl. junk-first-then-valid; `getIterator` DESC clamp;
  argParser/logging/`nextRunDir` destructive-tooling guards (incl. the `./logs -> repo root` symlink).
- **E2E** (`yarn test:e2e:parallel`): `E2E: BlockQueueManager` re-timed to timeout-sync (unknown-fork
  punishment, drain-after-transition, stale-fork gate); a new "unprovable target mutually blacklists
  both peers" spectate test; a new `signedBlocks[0].header.forkId = random → DisputeStateProofHeaderMismatch`
  dispute test. Targeted runs green; the only deterministic e2e reds reproduce on the base branch
  (pre-existing dispute-heavy sandbox flakes).

## Follow-ups (tracked in `docs/tests/todo.md`)

- Same-pass e2e for three dispute-heavy invariants (junk-flood → one recovery/kill-period read;
  moved-past fork dropped silently; honest co-participant not punished at channel open) — deferred by
  explicit acceptance because they trip the sandbox event-loop watchdog; run on a sustainable-load box.
- Participant-aware CRDT attribution (the structural cap bounds memory but can drop a late supplier's
  attribution under a junk-first flood).
- Byzantine-spam rate-limiter for sync requests.

sh3ll3x3c (commented):
Re-checked at 9f77bcd9, run locally end to end this round.

**Prior findings: all 3 blocking + 2 major are genuinely fixed** — each traced through the code, and the claims hold: recovery never awaits `reduceLocally` under the StateManager mutex (errors double-caught); arrival never sync-probes — the queue timeout is the sole probe, and the `??` pin lets an honest block-0 author prove height 0; the `--logDir` guards hold including the realpath/symlink cases (guard tests run); one sync probe per entry per agreement window; height shape-validation + DESC clamp + no above-latest downgrade. Minors all addressed except the requested freshness comment on the mirror-backed `isForkDisputed` (comment-only).

**Local verification:** typecheck clean on both tsconfigs; unit suite 309 green; targeted e2e (BlockQueueManager / Spectate / stale-proof guard / StateSnapshots / the case4 header-mismatch tests) 41/45 — I re-ran the 4 reds in isolation on this head and on `dispute` base: identical failures on base, so they're the pre-existing dispute-heavy epicenter, nothing attributable to this PR.

**New findings:** 3 inline comments — 2 major, 1 convention. The two majors are not just code-traced: the runtime witness below stages the real dispute/recovery flow and observes both failures live against this head.

**Carried over from last round, sharper now (out of diff):** the milestone loop in `AgreementManager.getStateProof` (src/agreementManager/AgreementManager.ts:96) still consumes `getChangePointsInRange(forkId)` unbounded by the requested height, so a pin below a participant-set change point over-proves. As of this PR the requester's step-6 exact-height check (SpectateService.ts:423-433) aborts on that, and `abort()` blacklists when participating — an honest responder cut for answering correctly-but-over. The PR description says step 6 "appears ineffective in practice", which conflicts with the code as written; worth resolving explicitly. Bounding the change-point range by `blockHeight` is the cheap fix either way. (Code-traced only — a runtime repro needs join-mid-fork staging. Also not yet tracked in docs/tests/todo.md.)

Non-blocking notes for completeness: a crafted authentic `forkId = ZeroHash` block ingested during the pre-genesis window can bounce restore → past-deadline-evict → execute in 0-delay tasks until genesis lands (bounded CPU churn); a failed detached recovery clears its gate without setting suppression, allowing serial retries.

<details>
<summary>Runtime witness test (throwaway — asserts the buggy behavior, passes on this head; invert into regressions after the fix)</summary>

Staging: 4 peers, dispute + resolve with the target peer's reduction race held (same staging as the existing recover-at-ingest test). Phase A stubs only `stateChannelManagerContract.isKillPeriodExpired` to reject and asserts `keepConnection=false` for the honest writer's real block-0 delivery. Phase B gates the VM `setState` inside the recovery transition (after the new fork's genesis is stored, before `forkId` flips), probes `isKnownStaleFork(newForkId) === true` while `forkId` is still the old fork, shows a copy queued in the window is dropped by its own timeout with zero sync probes, and after release the target sits on the new fork permanently missing a block the network holds.

```ts
import { Codec, Type } from "@/utils";
import { waitFor } from "@test/utils/waitFor";

/**
 * THROWAWAY runtime witnesses for two review findings on PR #384 @ 9f77bcd9.
 * These tests PASS while the defects exist (they assert the buggy behavior),
 * and are never to be committed.
 *
 * Finding A: a transient chain-RPC failure inside maybeScheduleForkRecovery
 *   propagates out of ingestBlockConfirmation as `false`, which the transport
 *   layer translates into disconnect+blacklist of the (honest) sender.
 *
 * Finding B: during a fork transition, the target fork's genesis snapshot is
 *   stored before `forkId` flips (unsafeSetLatestState awaits the VM setState
 *   in between). queueTimeout classifies without the StateManager mutex, so a
 *   timeout firing in that window sees the genesis and misclassifies the
 *   about-to-be-current fork as "known stale" -> silently drops the entry.
 */
describe("E2E: PR384 repro - wrong-fork ingest and transition-window races", function () {
    const forkTimeConfig = {
        p2pTime: 3,
        agreementTime: 4,
        chainFallbackTime: 2,
        evidenceTime: 3
    };

    it("witnesses finding A (transient read failure cuts honest sender) and finding B (transition window silently drops the reduced-fork block)", async function () {
        this.timeout(150000);

        const targetPeerIndex = 1;
        const maliciousPeerIndex = 2;
        const reducingHonestPeers = [0, 3];

        // 2 = warm-up transitions the dispute staging needs, not the peer index.
        await h.lifecycle.start(4, 2, { timeConfig: forkTimeConfig });
        await h.assert.sync.peersInSyncWait();

        const originalForkId = h.activeForkId!;
        const targetPeer = h.getPeer(targetPeerIndex);

        // Keep the target un-reduced: hold the snapshot-event reduction race.
        const race = await h.rpcStub.holdReductionRace(targetPeerIndex);
        // Record sync probes without forwarding them (keeps phase B inert).
        const restoreSync = await h.rpcStub.recordSpectateSync(targetPeerIndex, {
            forward: false
        });

        h.event.resetEventSpies();
        const { newForkId } = await h.scenario.disputeAndResolve({
            maliciousPeerIndex,
            forkId: originalForkId,
            honestPeerIndices: reducingHonestPeers,
            resetEventSpies: false,
            forkSettleTimeoutMs: 20000,
            disputesCommittedTimeoutMs: 10000
        });

        await waitFor(
            async () => (await race.heldSnapshotEventCount()) > 0,
            10000,
            50
        );

        // ------------------------------------------------------------------
        // Phase A: make the target's chain read reject (as a flaky provider
        // would), BEFORE the first reduced-fork block can arrive.
        // ------------------------------------------------------------------
        await h.execOnHost(targetPeer, (sm) => {
            const smAny = sm as unknown as {
                killPeriodExpiryCache: Map<string, unknown>;
                stateChannelManagerContract: {
                    isKillPeriodExpired: (...a: unknown[]) => Promise<unknown>;
                };
                __reproOrigKillRead?: (...a: unknown[]) => Promise<unknown>;
            };
            smAny.killPeriodExpiryCache.clear();
            const contract = smAny.stateChannelManagerContract;
            smAny.__reproOrigKillRead =
                contract.isKillPeriodExpired.bind(contract);
            contract.isKillPeriodExpired = () =>
                Promise.reject(
                    new Error("repro: simulated transient provider failure")
                );
        });

        const sourcePeer = h.getPeer(reducingHonestPeers[0]);
        const nextWriterAddress = await h
            .control(sourcePeer)
            .query.getNextToWrite()
            .request();
        const writerPeer = h.peers.find(
            (p) => p.address === nextWriterAddress
        );
        expect(writerPeer, "no writer peer found").to.not.be.undefined;
        expect(writerPeer!.index).not.to.equal(
            targetPeerIndex,
            "test setup requires a different peer to write the first reduced-fork block"
        );

        await writerPeer!.p2pInstance.p2pContractInstance.add(1);

        await waitFor(async () => {
            const block = await h
                .control(sourcePeer)
                .query.getBlockByHeight(newForkId, 0)
                .request();
            return Boolean(block?.encodedBlockConfirmation);
        }, 10000);
        const firstBlock = await h
            .control(sourcePeer)
            .query.getBlockByHeight(newForkId, 0)
            .request();
        expect(firstBlock).not.to.equal(null);

        const blockConfirmation = Codec.decode(
            firstBlock!.encodedBlockConfirmation,
            Type.BlockConfirmation
        );
        const block0Hash = Block.fromBlockConfirmation(blockConfirmation).hash;

        // FINDING A WITNESS: the same delivery that succeeds in the shipped
        // test suite now returns keepConnection=false purely because OUR OWN
        // provider read failed. StateTransitionRpcMethods translates false
        // into disconnectAndBlacklist of the sender.
        await h.transition.ingestBlockConfirmationWait({
            peerIndex: targetPeerIndex,
            blockConfirmation,
            ingestOptions: { senderAddress: writerPeer!.address },
            keepConnection: false,
            waitForProcessed: false
        });

        // The block was not even queued - the throw preempted queueBlock.
        expect(
            await h
                .control(targetPeer)
                .query.isBlockQueued(block0Hash)
                .request()
        ).to.equal(false, "block should not be queued after failed ingest");
        expect(
            await h.control(targetPeer).query.getForkId().request()
        ).to.equal(originalForkId);

        // Bonus observation (transport path): if gossip delivered the block
        // first, the real RPC layer already cut the honest writer. Recorded,
        // not asserted (gossip timing is not deterministic).
        let writerBlacklistedViaGossip = false;
        try {
            await waitFor(
                async () =>
                    await h
                        .control(targetPeer)
                        .query.isBlacklisted(writerPeer!.address)
                        .request(),
                3000,
                100
            );
            writerBlacklistedViaGossip = true;
        } catch {
            writerBlacklistedViaGossip = false;
        }
        // Surfaced via the test name on failure; kept as a plain expect(true)
        // so the witness stays green either way.
        expect(typeof writerBlacklistedViaGossip).to.equal("boolean");

        // ------------------------------------------------------------------
        // Phase B: restore the chain read, gate the VM setState so the
        // transition parks inside unsafeSetLatestState AFTER the new fork's
        // genesis snapshot is stored but BEFORE forkId flips.
        // ------------------------------------------------------------------
        await h.execOnHost(targetPeer, (sm) => {
            const smAny = sm as unknown as {
                stateChannelManagerContract: {
                    isKillPeriodExpired: (...a: unknown[]) => Promise<unknown>;
                };
                __reproOrigKillRead?: (...a: unknown[]) => Promise<unknown>;
            };
            smAny.stateChannelManagerContract.isKillPeriodExpired =
                smAny.__reproOrigKillRead!;
        });

        await h.execOnHost(
            targetPeer,
            (sm, args: { newForkId: string }) => {
                const smAny = sm as unknown as {
                    diamondStateMachine: {
                        setState: (...a: unknown[]) => Promise<unknown>;
                    };
                    storage: {
                        stateSnapshots: {
                            getGenesisSnapshotByForkId: (f: string) => unknown;
                        };
                    };
                    __setStateGate?: {
                        entered: boolean;
                        release?: () => void;
                    };
                    __reproOrigSetState?: (
                        ...a: unknown[]
                    ) => Promise<unknown>;
                };
                const dsm = smAny.diamondStateMachine;
                const original = dsm.setState.bind(dsm);
                smAny.__reproOrigSetState = original;
                const gate: { entered: boolean; release?: () => void } = {
                    entered: false
                };
                smAny.__setStateGate = gate;
                dsm.setState = async (...a: unknown[]) => {
                    // Hold only the transition's setState: the one that runs
                    // after the new fork's genesis snapshot is stored.
                    if (
                        smAny.storage.stateSnapshots.getGenesisSnapshotByForkId(
                            args.newForkId
                        ) === undefined
                    ) {
                        return original(...a);
                    }
                    gate.entered = true;
                    await new Promise<void>((resolve) => {
                        gate.release = resolve;
                    });
                    return original(...a);
                };
            },
            { newForkId: String(newForkId) }
        );

        // Trigger the detached recovery: same block, chain read now healthy.
        await h.transition.ingestBlockConfirmationWait({
            peerIndex: targetPeerIndex,
            blockConfirmation,
            ingestOptions: { senderAddress: writerPeer!.address },
            keepConnection: true,
            waitForProcessed: false
        });

        // Wait until the transition is parked inside the window.
        await waitFor(
            async () =>
                await h.execOnHost(targetPeer, (sm) =>
                    Boolean(
                        (
                            sm as unknown as {
                                __setStateGate?: { entered: boolean };
                            }
                        ).__setStateGate?.entered
                    )
                ),
            30000,
            200
        );

        // FINDING B WITNESS (premise): inside the window the about-to-be-
        // current fork is misclassified as known-stale while forkId is still
        // the old fork.
        const probe = await h.execOnHost<
            { forkId: string; misclassifiedAsStale: boolean },
            { newForkId: string }
        >(
            targetPeer,
            async (sm, args) => ({
                forkId: String(sm.forkId),
                misclassifiedAsStale: await sm.isKnownStaleFork(args.newForkId)
            }),
            { newForkId: String(newForkId) }
        );
        expect(probe.forkId).to.equal(
            originalForkId,
            "forkId must not have flipped yet inside the window"
        );
        expect(probe.misclassifiedAsStale).to.equal(
            true,
            "isKnownStaleFork must misclassify the about-to-be-current fork"
        );

        // FINDING B WITNESS (consequence): a fresh copy of the reduced-fork
        // block queued during the window gets silently dropped by its own
        // queue timeout - no sync probe, no storage, no punishment.
        const syncCountAtHold =
            await h.rpcStub.spectateSyncCallCount(targetPeerIndex);

        await h.transition.ingestBlockConfirmationWait({
            peerIndex: targetPeerIndex,
            blockConfirmation,
            ingestOptions: { senderAddress: writerPeer!.address },
            keepConnection: true,
            waitForProcessed: false
        });
        expect(
            await h
                .control(targetPeer)
                .query.isBlockQueued(block0Hash)
                .request()
        ).to.equal(true, "fresh copy should be queued during the window");

        // Its queue timeout (agreementTime=4s) fires inside the held window.
        await waitFor(
            async () =>
                !(await h
                    .control(targetPeer)
                    .query.isBlockQueued(block0Hash)
                    .request()),
            (forkTimeConfig.agreementTime + 4) * 1000,
            200
        );
        expect(
            await h
                .control(targetPeer)
                .query.getBlockByHash(block0Hash)
                .request()
        ).to.be.null;
        expect(
            await h.rpcStub.spectateSyncCallCount(targetPeerIndex)
        ).to.equal(
            syncCountAtHold,
            "the drop must be silent - no sync probe fired"
        );

        // Release the transition and show the loss: the target lands on the
        // new fork with the first block gone, while the network has it.
        await h.execOnHost(targetPeer, (sm) => {
            const smAny = sm as unknown as {
                diamondStateMachine: {
                    setState: (...a: unknown[]) => Promise<unknown>;
                };
                __setStateGate?: { release?: () => void };
                __reproOrigSetState?: (...a: unknown[]) => Promise<unknown>;
            };
            smAny.diamondStateMachine.setState = smAny.__reproOrigSetState!;
            smAny.__setStateGate?.release?.();
        });

        await waitFor(
            async () =>
                (await h.control(targetPeer).query.getForkId().request()) ===
                newForkId,
            20000
        );
        expect(
            await h
                .control(targetPeer)
                .query.getBlockByHash(block0Hash)
                .request()
        ).to.equal(
            null,
            "LOSS: target transitioned to the new fork but the dropped block never arrives"
        );
        expect(
            await h
                .control(sourcePeer)
                .query.getBlockByHash(block0Hash)
                .request()
        ).to.not.equal(null, "the network holds the block the target lost");

        await restoreSync();
        await race.release();
    });
});

```

</details>

src/stateManager/BlockQueueManager.ts
```diff
@@ -112,6 +130,20 @@ export default class BlockQueueManager {
                 return true;
             }
 
+            if (block.forkId !== this.stateManager.forkId) {
+                await this.maybeScheduleForkRecovery(block.channelId);
```
sh3ll3x3c, line 134: **[major]** A transient chain-RPC failure on this await gets the honest sender blacklisted. On a cache miss `maybeScheduleForkRecovery` performs a real chain read (`isKillPeriodExpiredCached` → `stateChannelManagerContract.isKillPeriodExpired`, StateManager.ts:826) with no catch between it and this method's catch-all, which returns `false` — and `StateTransitionRpcMethods.onBlockConfirmation` translates `false` into `disconnectAndBlacklistPeerByEvmAddress(sender)`. On base, the equivalent read ran inside `tryRecoverForkMismatch`'s try/catch and was logged-and-swallowed (ingest still returned `true`).

The scenario is exactly the recovery case this PR builds: we're the laggard on the disputed fork, the honest writer gossips the reduced fork's block 0, our provider blips once → we cut the one peer we need, before the block is even queued.

**Runtime-witnessed** (test in the review body): with only that read stubbed to reject, ingest of the writer's real block returned `keepConnection=false` and nothing was queued; the same delivery with the read healthy recovers and converges.

Fix: catch inside `maybeScheduleForkRecovery` — a gating failure should degrade to "don't schedule recovery this round", never to a `false` ingest verdict.

src/stateManager/BlockQueueManager.ts
```diff
@@ -189,6 +298,40 @@ export default class BlockQueueManager {
             return;
         }
 
+        if (entry.block.forkId !== this.stateManager.forkId) {
+            if (await this.stateManager.isKnownStaleFork(entry.block.forkId)) {
```
sh3ll3x3c, line 309: **[major]** This classification races the fork transition and can silently drop the about-to-be-current fork's blocks. Both "known past" signals are persisted before `forkId` flips: `unsafeSetLatestState` stores the genesis snapshot (StateManager.ts:946), then awaits the VM `setState` (:959), and only then sets `forkId` (:963); `persistSyncPayload` additionally stores the target fork's finalized blocks before that call. `queueTimeout` runs without the StateManager mutex and dequeues first, so a timeout firing inside that await sees genesis-present + fork-not-current → "known stale" → the dequeued entry is dropped with no sync probe, and the post-transition drain finds nothing. The sender keeps the connection and believes the block delivered.

The after-mutex re-check in `onBlockConfirmation` is immune precisely because it runs under the mutex — this call site needs the same protection (classify under the mutex), or the transition should flip `forkId` before persisting the new fork's artifacts.

**Runtime-witnessed** (test in the review body): parking the transition at `setState` gave `isKnownStaleFork(newForkId) === true` while `forkId` was still the old fork; a copy queued during the window was silently dropped (zero sync probes), and after release the peer sat on the new fork permanently missing a block the rest of the network holds.

src/stateManager/StateManager.ts
```diff
@@ -1198,6 +1249,28 @@ class StateManager<
...
             block = entry.block;
 
+            // A fork transition can land while we wait for the mutex, so the
+            // block's fork may no longer be current. Don't validate against the
+            // wrong fork: restore the entry for a timeout sync if we don't
+            // recognize its fork as known-past, or drop it if we've moved past
+            // it. The dispute strategy runs disputed/other-fork state-proof
+            // blocks by design, so it is exempt.
+            if (
+                !(strategy instanceof DisputeValidationStrategy) &&
```
sh3ll3x3c, line 903: **[convention]** This is a validation deviation handled inline, with per-pipeline behavior selected by `instanceof` at the call site — the shape the block-validation rule in AGENTS.md names: every deviation is a method on `AValidationStrategy`, side effects live in the strategy implementations. A `forkChangedWhileAwaitingMutex(entry)` hook (dispute strategy returning "proceed", the others restoring/dropping) would keep the deviation catalog in one place and drop the `instanceof`. I realize the commit message declares the inline handling deliberate, and there are pre-existing `instanceof` sites in this file — flagging for an explicit justify-or-refactor rather than as a bug.

src/stateManager/BlockQueueManager.ts
lukastanisic99: The concern the AI found was "What if the RPC node is faulty or our connection breaks and our request throws, we'll blacklist the other peer even if honest" -> we have much bigger problems if our on-chain endpoint is faulty, it's the premiss of our shared security. There's a Trello card to have redundant connections to RPC nodes so at least 1 is available.

src/stateManager/BlockQueueManager.ts
lukastanisic99: More theoretical than practical and the solution would kill concurrency for something that wouldn't ever trigger in the honest case and even if it does there's a fallback in place to catch the block later as calldata. 

Honest peers send blocks `on-time` where they can run immediately. This code schedules blocks that `can not` run immediately (for whatever reason) to execute in the future, giving them `time` to meet the condition. If the condition is not met, they're discarded and/or used to initiate sync with other peers which may lead to other peers being blacklisted if they sent junk blocks. 

The problem here is that when we reduce, we have:
store..
store..
`await`...
setForkId

The await makes it not atomic, to make it atomic we'd have to add a mutex and to have every call-site wait on the mutex, so we would `not have concurrent` operations which the QUEUE is designed to have - the entire queue by design is a CRDT with no mutex and full concurrency. 

This runs at the end of the timeout (the task `had enough time`), no different than arriving 'slightly late' and being discarded. This entire system (like most blockchain systems) follows a partially synchronous model, where it's expected to converge in time.

src/stateManager/StateManager.ts
lukastanisic99: Ye I'm aware of this, my agent flagged it also - We have it in a couple of places and a TODO to make them cleaner

daiagi (approved):
see tracking issue:
- https://github.com/peer3to/state-channels-plus/issues/391

src/stateManager/validationStrategy/CalldataCommittedStrategy.ts
```diff
@@ -91,9 +91,9 @@ export default class CalldataCommittedStrategy extends AValidationStrategy {
         );
     }
     public async wrongGenesisDetected(
-        block: Block
+        entry: QueuedBlockEntry
     ): Promise<BlockValidationResult> {
-        return this.blockValidationStrategy.wrongGenesisDetected(block);
+        return this.blockValidationStrategy.wrongGenesisDetected(entry);
```
momoshell, line 99: This delegates straight into `BlockValidationStrategy.wrongGenesisDetected`, which (new in this PR) blacklists + `DISCONNECT`s on missing genesis, as a live-gossip-transport reaction?

`DisputeValidationStrategy`'s sibling override instead throws ("a bug, not a peer fault and fail loudly"). Since calldata-committed is a replay-style pipeline like dispute rather than live gossip, would it make sense for it to throw too rather than inherit the live disconnect/blacklist side effect?

src/stateManager/BlockQueueManager.ts
```diff
@@ -125,8 +157,85 @@ export default class BlockQueueManager {
...
+            const { isExpired, killPeriodEnd } =
+                await this.stateManager.isKillPeriodExpiredCached(forkId);
+            if (!isExpired) {
+                this.recoverySuppressedUntil.set(forkId, killPeriodEnd);
+                return;
+            }
+            // Idempotent single-flight: on success `setGenesisState` transitions
+            // the fork and the post-transition drain picks up queued blocks.
+            await this.stateManager.reduceLocally(forkId);
+        } catch (error) {
```
momoshell, line 230: Just a nit: `runForkRecovery`'s catch logs and swallows, and the `finally` unconditionally clears `recoveryScheduledForFork`. 

If `reduceLocally` fails persistently (e.g. `reduce.staticCall` reverting) while the fork stays disputed with kill-period-expired, there's no cooldown. 
Every subsequent junk-fork block re-triggers a fresh recovery attempt. Bounded impact (single-flight, local static calls only), but a short suppression window after a failure would avoid the repeated-attempt/log-spam pattern under a sustained flood.

test/fixtures/PeerTestHarness.ts
```diff
@@ -978,11 +979,12 @@ export class PeerTestHarness<
     >(
         peer: TestPeer<TCustomRpc, TStateMachine>,
         fn: (sm: StateManager<HarnessControlRpc>, args: A) => T | Promise<T>,
-        args: A = {} as A
+        args: A = {} as A,
+        options?: RpcRequestOptions
```
momoshell, line 1000: Small nit; the new `options?: RpcRequestOptions` param isn't passed at any `execOnHost` call site in this PR (or anywhere else in the repo currently). Not wrong, just unexercised.

Confirming it's intentional forward-looking plumbing rather than dead surface.

momoshell (approved):
Nice work closing out these races, especially the fork-recovery quarantine/drain flow and the remaining-window timeout math, both of which are easy to get subtly wrong. I verified all three fixes (wrong-fork recovery, exact-target sync pinning, VM-rollback guard) against the actual races they target and they hold up.

One documentation nit: the PR description names a `tryRecoverForkMismatch` function and describes it as awaiting `reduceLocally` inline on the ingest path. That function doesn't exist, the actual implementation (`maybeScheduleForkRecovery`) schedules reduction as a detached task and always quarantines the block until a fork transition drains it. 

That's actually the more correct design (an inline await would deadlock on the state-manager mutex, per the comment at `BlockQueueManager.ts:33-39`), but worth updating the description so it matches what's shipped.

A few small inline comments below, nothing blocking.

PR #390, opened 2026-07-12
fix(p2p): jitter HolepunchRelay's retry/backoff to avoid thundering herds

momoshell:
## Why

Follow-up to #386. That PR fixed the relayer pool permanently exhausting itself after repeated failures, but left two synchronized-retry gaps:

1. Single-relayer failover retried the next relayer immediately, with zero delay.
2. Full-pool-exhaustion backoff retried at a deterministic mark (`BACKOFF_BASE_MS * 2^n`).

Either way, if many clients fail over at the same moment (e.g. a relayer actually going down, or a wave of clients reconnecting after backgrounding), they all retry in lockstep — amplifying load on whichever relayer survives right when it's under the most pressure.

## What changed

`src/HolepunchRelay.ts`:
- Single-relayer failover now waits a small randomized delay (0–250ms, `FAILOVER_JITTER_MAX_MS`) before retrying, instead of retrying synchronously.
- Full-pool-exhaustion backoff now applies full jitter — uniform random in `[0, cappedBackoff]` — instead of retrying at the deterministic capped value.

Success path (exclusions + backoff counter resetting on `ws.onopen`) is unchanged.

## Test plan

- Extended `test/utils/HolepunchRelay.test.ts` (3 → 5 tests): two new tests simulate several concurrent clients failing over/exhausting at the same instant and assert the resulting delays are bounded, non-identical, and not pinned at the old deterministic marks — not just "a delay exists."
- The two pre-existing tests that relied on the old zero-delay synchronous failover were updated to advance the fake clock past the new jitter window, without changing what they actually assert (exclusion/backoff reset behavior is unchanged).
- `yarn hardhat test --no-compile test/utils/HolepunchRelay.test.ts` — 5/5 passing.

### DIFF
--- a/src/HolepunchRelay.ts
+++ b/src/HolepunchRelay.ts
@@ -10,6 +10,10 @@ import { Logger } from "@/utils";
 // success), so a fully-down network doesn't tight-loop hammering reconnects.
 const BACKOFF_BASE_MS = 1000;
 const BACKOFF_CAP_MS = 30000;
+// Upper bound for the randomized delay before retrying after a single
+// relayer failure, so many clients failing over at the same instant don't
+// all pile onto the next relayer at once (thundering herd).
+const FAILOVER_JITTER_MAX_MS = 250;
 
 class HolepunchRelay {
     relayerUrls: string[];
@@ -131,14 +135,18 @@ class HolepunchRelay {
     }
 
     private scheduleRetryAfterExhaustion(): void {
-        const delayMs = Math.min(
+        // Full jitter (AWS-style): pick uniformly in [0, cappedBackoff] rather
+        // than retrying at the deterministic cappedBackoff mark, so clients
+        // that exhaust the pool at the same moment don't retry in lockstep.
+        const cappedBackoffMs = Math.min(
             BACKOFF_BASE_MS * 2 ** this.backoffAttempt,
             BACKOFF_CAP_MS
         );
+        const delayMs = Math.random() * cappedBackoffMs;
         this.backoffAttempt++;
         this.logger.warn(
             "All holepunch relayers failed, retrying pool after backoff",
-            { delayMs, relayerUrls: this.relayerUrls }
+            { delayMs, cappedBackoffMs, relayerUrls: this.relayerUrls }
         );
         this.excludedRelayers.clear();
         setTimeout(() => this.connectToRelayer(), delayMs);
@@ -151,7 +159,10 @@ class HolepunchRelay {
             excludedCount: this.excludedRelayers.size,
             total: this.relayerUrls.length
         });
-        this.connectToRelayer();
+        // Randomized delay so many clients failing over off the same
+        // relayer at once don't all hit the next relayer simultaneously.
+        const delayMs = Math.random() * FAILOVER_JITTER_MAX_MS;
+        setTimeout(() => this.connectToRelayer(), delayMs);
     }
 }
 
--- a/src/HolepunchRelay.ts
+++ b/src/HolepunchRelay.ts
@@ -159,6 +159,13 @@ class HolepunchRelay {
             excludedCount: this.excludedRelayers.size,
             total: this.relayerUrls.length
         });
+        // Branch immediately so a pool-exhausting failure schedules only the
+        // exhaustion backoff, never the failover jitter as well - one timer
+        // owner per failure, not two stacked delays.
+        if (this.isRelayerPoolExhausted()) {
+            this.scheduleRetryAfterExhaustion();
+            return;
+        }
         // Randomized delay so many clients failing over off the same
         // relayer at once don't all hit the next relayer simultaneously.
         const delayMs = Math.random() * FAILOVER_JITTER_MAX_MS;
--- a/test/utils/HolepunchRelay.test.ts
+++ b/test/utils/HolepunchRelay.test.ts
@@ -215,13 +215,19 @@ describe("HolepunchRelay", function () {
             "wss://relay-b.example",
             "wss://relay-c.example"
         ];
-        const setTimeoutSpy = sinon.spy(global, "setTimeout");
         const delays: number[] = [];
 
         // Simulate several independent clients failing over off one relayer
         // at essentially the same instant (a reconnect storm), and capture
-        // the randomized retry delay each one schedules.
+        // the randomized retry delay each one schedules. Give each client
+        // its own fake-timer clock, spy, and instance list - otherwise a
+        // timer left pending by an earlier "client" can fire during a later
+        // one's clock.tick() and get mistaken for that client's own retry.
         for (let i = 0; i < 8; i++) {
+            clock.restore();
+            clock = sinon.useFakeTimers();
+            const setTimeoutSpy = sinon.spy(global, "setTimeout");
+            FakeWebSocket.instances = [];
             HolepunchRelay.init(relayerUrls, () => undefined, createLogger());
             const ws =
                 FakeWebSocket.instances[FakeWebSocket.instances.length - 1];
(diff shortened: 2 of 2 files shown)


lukastanisic99:
Ok just rebase, but I don't think there will be any conflicts on this nor do I think we touched this part of the code

lukastanisic99:
@momoshell 

**Bottom line:** rework before landing. Compilation and both TypeScript builds pass, and the focused relay run passed 5/5, but the exact-head canonical suite finished with 588 passing / 1 failing. The failure is a normal assertion in a changed test, not a worker-process failure: the two jitter stages can leave an exhaustion timer pending after the test declares a connection successful, then that stale timer creates a second connection during the next failover.

## 🧱 Fundamental

- 🟠 **[FO1] A full-pool failure waits through two independent jitter stages.** [`excludeAndConnectToRelayer()` always schedules the 0–250 ms failover timer](https://github.com/peer3to/state-channels-plus/blob/f6b81c5b6ec51f4abeff2ce473bc59059c892ab3/src/HolepunchRelay.ts#L155-L165), even when that failure has exhausted the pool. Only when that timer fires does [`connectToRelayer()` notice exhaustion and schedule the exponential backoff](https://github.com/peer3to/state-channels-plus/blob/f6b81c5b6ec51f4abeff2ce473bc59059c892ab3/src/HolepunchRelay.ts#L56-L65). The effective exhaustion delay is therefore `failover jitter + backoff jitter`, so a nominal 30-second cap can take up to 30.25 seconds.

  > **Fix FO1-FIX**
  >
  > After adding the exclusion, branch immediately: schedule the exhaustion backoff directly when the pool is exhausted; otherwise schedule the short failover jitter. Keep one timer owner for both paths.

- 🟢 **[SG1] No new authentication, remote-input validation, or secret-handling issue found.**

  > **Fix SG1-FIX**
  >
  > No change required.

## 🏁 Race conditions

- 🔴 **[RR1] Retry callbacks are not single-flight and are never invalidated.** Both [`onerror` and `onclose` schedule a retry](https://github.com/peer3to/state-channels-plus/blob/f6b81c5b6ec51f4abeff2ce473bc59059c892ab3/src/HolepunchRelay.ts#L78-L97), but the class stores neither a timer handle nor a connection generation. A socket can emit `error` followed by `close`, producing two pending callbacks. If either reconnect succeeds, the other still runs and creates another WebSocket, DHT, and swarm, overwriting `this.swarm`. Reinitializing the singleton through [`init()`](https://github.com/peer3to/state-channels-plus/blob/f6b81c5b6ec51f4abeff2ce473bc59059c892ab3/src/HolepunchRelay.ts#L35-L46) also leaves timers from the old instance active. The canonical failure exercised the same stale-callback shape inside the fake clock.

  > **Fix RR1-FIX**
  >
  > Track one retry timer and one active connection generation. Treat a socket's terminal failure once, cancel pending retries on success, reinitialization, and disposal, ignore stale socket callbacks, and destroy superseded network resources. Add deterministic `error`→`close`, success-before-timer, late-open, and reinitialization tests.

## ⚡ Performance

- 🟠 **[PO1] Stale callbacks can multiply live network stacks during an outage.** Each callback constructs a new WebSocket, DHT, and Hyperswarm before assigning `this.swarm`, so RR1 can leave duplicate sockets and swarms alive while only the last one remains reachable.

  > **Fix PO1-FIX**
  >
  > Make retries single-flight and dispose the previous socket/swarm before replacing it. Assert that one logical relay owns at most one connection attempt and one pending retry.

- 🔴 **[TR1] The exact-head canonical gate fails in a test changed by this PR.** `yarn test:parallel` completed with 588 passing / 1 failing. [`resets exclusion/backoff state on a successful connection`](https://github.com/peer3to/state-channels-plus/blob/f6b81c5b6ec51f4abeff2ce473bc59059c892ab3/test/utils/HolepunchRelay.test.ts#L173-L210) expected three connections and observed four; the preserved log is [`run-471`](../../../logs/run-471/error_HolepunchRelay.test__HolepunchRelay__resets_exclusion_backoff_state_on_a_successful_connection.ansi). After the second close, the test advances 1000 ms, but production first waits up to 250 ms and only then schedules up to 1000 ms. There is a 12.5% chance that their sum exceeds the tick. In that path the test calls `emitOpen()` on the already closed socket while the exhaustion timer remains pending; the next 250 ms tick fires both that stale timer and the newly scheduled failover timer.

  > **Fix TR1-FIX**
  >
  > Fix RR1 and FO1, then make the test deterministic by controlling the random source and advancing the exact scheduled delay. The full gate must pass on the PR's exact head before merge.

- 🔴 **[TR2] The exhaustion-jitter test asserts the failover timer, not the exhaustion backoff.** After the second relay fails, [the test immediately reads the first `setTimeout` call](https://github.com/peer3to/state-channels-plus/blob/f6b81c5b6ec51f4abeff2ce473bc59059c892ab3/test/utils/HolepunchRelay.test.ts#L244-L279). At that point production has scheduled only the 0–250 ms timer in `excludeAndConnectToRelayer()`; the exhaustion backoff does not exist until that timer fires. The `< 1000` assertion therefore passes even if full jitter is absent or broken.

  > **Fix TR2-FIX**
  >
  > Prefer FO1's single exhaustion timer and assert it directly. If both stages remain, advance through the failover timer before inspecting the separately scheduled backoff.

- 🟠 **[TO3] The timing tests sample real randomness and duplicate a private production constant.** The new tests call `Math.random()` repeatedly and assert that eight samples differ while [copying `FAILOVER_JITTER_MAX_MS` into the test](https://github.com/peer3to/state-channels-plus/blob/f6b81c5b6ec51f4abeff2ce473bc59059c892ab3/test/utils/HolepunchRelay.test.ts#L34-L38). This is probabilistic and misses exact boundary, attempt, cap, and reset behavior.

  > **Fix TO3-FIX**
  >
  > Inject the random source and retry policy. Test zero, midpoint, upper boundary, successive attempts, cap, and successful reset with exact delays.

- 🔴 **[TR4] There is no real-boundary coverage for the changed relay behavior.** The only changed test replaces WebSocket, DHT, and Hyperswarm through the module cache. It cannot show that concurrent clients spread failover attempts, recover through a surviving relay, or clean up sockets and swarms during outage recovery.

  > **Fix TR4-FIX**
  >
  > Add a local-relay E2E covering failover and full outage/recovery, including the common `error`→`close` sequence and teardown with a pending retry.

## ♻️ Code reuse

- 🟡 **[CY1] Retry policy has two owners.** Production closes over private delay constants, while the test copies one of them and reconstructs the expected timing. That duplication already let the test describe a one-stage exhaustion delay while production implements two stages.

  > **Fix CY1-FIX**
  >
  > Put the delay calculation behind one injected retry-policy boundary. Production supplies the constants and RNG; tests supply deterministic values without restating policy.

## 📏 AGENTS.md adherence

- 🔴 **[AR1] The `src/` change does not ship the required E2E coverage.** Repository rules require representative E2E coverage in the same pass for a source change that crosses an integration boundary; this PR adds only `test/utils/HolepunchRelay.test.ts`.

  > **Fix AR1-FIX**
  >
  > Add the real relay-boundary coverage described in TR4.

- 🟠 **[AO2] The changed unit tests replace every collaborator with local fakes.** [`FakeWebSocket`, fake DHT/swarm classes, module-cache replacement, logger construction, and singleton mutation](https://github.com/peer3to/state-channels-plus/blob/f6b81c5b6ec51f4abeff2ce473bc59059c892ab3/test/utils/HolepunchRelay.test.ts#L4-L113) all live in the test file. `test/AGENTS.md` requires unit tests with real collaborators through the harness and reusable staging outside test files.

  > **Fix AO2-FIX**
  >
  > Move deterministic relay controls into the harness and exercise the production wiring. Keep a small direct unit test only after extracting a pure retry-policy/state-machine boundary.

## 🧹 Miscellaneous (scope)

- 🟢 **[MG1] `git diff --check` passes, and there were no substantive reviewer comments to re-verify.**

  > **Fix MG1-FIX**
  >
  > No change required.

## ❓ Open questions

- 🟢 **[OG1] None.** Timer ownership, deterministic timing, and real-boundary coverage can be resolved without a product decision.

  > **Fix OG1-FIX**
  >
  > No decision required.

momoshell:
Thanks for the thorough pass — verified all of this against the code.

**FO1/TR1/TR2 fixed** in 806a8eff: `excludeAndConnectToRelayer` now branches immediately after adding the exclusion, so a pool-exhausting failure schedules only the backoff timer, never the failover jitter stacked on top. I reproduced TR1 locally (2/40 runs failing with the exact `4 to equal 3` you saw) and confirmed it's gone — 80/80 stable after the fix, and 589/589 on the full canonical gate. This also fixes TR2 for free: the timer the test captures is now genuinely the exhaustion backoff, not the failover jitter.

One thing I found while verifying: the fix exposed a *different* pre-existing flake in the "applies full jitter" test — it wasn't isolating fake-timer state between its 8 simulated-client loop iterations, so a stale timer from client 3 could get misattributed to client 6 once the timing shifted. Fixed alongside (each iteration gets its own clock/spy/instance-list).

On RR1/PO1/TR4/AR1/AO2/CY1 — I agree these are real, and I traced RR1 back further: the double `onerror`/`onclose` → shared-retry-path pattern already existed in #386 (synchronous back then, this PR just moved it onto a timer). Same for the stub-based test style AO2/TR4 flag. Opened #404 to track it properly rather than bolting it onto a jitter fix.

lukastanisic99 (approved):
Reviewed and left my agents to fix the 'red points' and pushed to a branch https://github.com/peer3to/state-channels-plus/tree/holepunch-relay
but it's probably over engineered and I don't think we'll have problems with this code, so I'll part that on the side and we'll go with this for now

PR #392, opened 2026-07-12
feat(p2p): add resumeFromBackground for mobile reconnect

momoshell:
## Why

Stacked on #388 (persistence port) — part of the mobile background/reconnect work - [Trello issue](https://trello.com/c/PjDUGE4c). This is the core reconnect mechanism: when a backgrounded client comes back, it needs to actually catch up and resume authoring, not just reconnect the transport.

## What was broken about the existing sync path

`SpectateService.sync()` is fire-and-forget: every one of its ~20 failure exits and its success path all returned `void` through `abort()`, so a caller had no way to tell success from failure. Its per-peer in-flight guard also returned early *before* anything awaitable could be constructed — if you tried to build a `resumeFromBackground` naively on top of it, a resume attempt that hit the guard would hang forever, which is exactly the failure mode this whole initiative exists to fix.

This design went through two rounds of adversarial review before implementation (documented in `.claude/dev-team/specs/be-02.yaml` if useful context): round 1 found the mechanism was underspecified, round 2 found the hung-promise path and the success/abort collapse as concrete bugs in the first fix attempt. An independent architect pass confirmed the resulting design (the guard-hit-resolves-immediately approach) was sound before this was implemented.

## What changed

**`SpectateService.ts`**
- `applySyncResponse` now returns a discriminated `SyncOutcome` (`{ok:true, confirmedForkId, confirmedHeight, localWasAhead}` | `{ok:false, reason}`) instead of `void`. Every former `abort()` call site routes through a new `failSync(peerAddress, reason, resumeMode)` helper.
- New `syncAwaitable()` entry point (used only by the resume path): checks the in-flight guard *synchronously first* and resolves `{ok:false, reason:'in-flight-collision'}` immediately on a hit, rather than ever attaching to whatever fire-and-forget sync is already running for that peer.
- `abort()` takes an optional `resumeMode` that suppresses both the hard `stateManager.abort()` branch (reachable pre-`PARTICIPATING` mid-resume — would nuke the whole state manager during a resume attempt) and peer blacklisting — neither is safe against a peer the reconnecting client still needs.
- The existing fire-and-forget `sync()` and its four existing reactive callers are untouched — same behavior, same side effects.

**`P2PManager.ts`**
- New `resumeFromBackground(channelId): Promise<ResumeResult>`. Dedups concurrent calls per channel. Rejoins the channel's topic — never touches the shared `Holepunch` swarm singleton, which is shared across every open channel (confirmed by reading the transport code directly; tearing it down here would disrupt other concurrently-open channels). Hydrates local storage, then tries up to 3 connected peers via `syncAwaitable`, 500ms apart, within a budget capped at 50% of the on-chain timeout window — leaving room for the self-defense/back-off work landing in a separate PR. Returns `{status:'restored', hydrated}` or `{status:'failed', reason, retryable}`.

**`P2pEventHooks.ts`**
- New optional `onResumePhase?: (phase: "reconnecting" | "resyncing") => void` hook so the frontend can drive reconnect UI off real state-machine progress.

## Test plan

- `yarn mocha test/e2e/E2E-Spectate.test.ts`, `E2E-SpectateStaleProofGuard.test.ts`, `E2E-Timeouts.test.ts` — ran multiple times against both this branch and a clean control checkout of #388 with none of this PR's changes. Both sides show non-deterministic failures (different specific test failing each run, same "provider destroyed"/nonce-contention signature), matching this repo's already-documented history of e2e flakiness under concurrent/automining conditions (see `devops-notes.md`). Every actual test assertion that ran passed; failures are exclusively teardown-hook or hardhat-automining races, confirmed present on the control branch too — not attributable to this change.

### DIFF
--- a/src/P2PManager.ts
+++ b/src/P2PManager.ts
@@ -468,7 +468,6 @@ class P2PManager<TCustomRpc extends MainRpcService = MainRpcService>
                 await sleep(Math.min(RESUME_PEER_SWITCH_DELAY_MS, remainingMs));
                 continue;
             }
-            attemptedPeers.add(peer);
 
             const remainingMs = deadline - Date.now();
             if (remainingMs <= 0) break;
@@ -492,7 +491,14 @@ class P2PManager<TCustomRpc extends MainRpcService = MainRpcService>
                 { channelId, peer, reason: outcome.reason }
             );
 
-            if (attemptedPeers.size >= MAX_RESUME_SYNC_PEERS) break;
+            // A collision with an already-running sync to this same peer is
+            // transient - that sync may finish and resync us within a beat.
+            // Don't burn a peer slot on it; retry the same peer after the
+            // delay instead of permanently giving up on it.
+            if (outcome.reason !== "in-flight-collision") {
+                attemptedPeers.add(peer);
+                if (attemptedPeers.size >= MAX_RESUME_SYNC_PEERS) break;
+            }
             const remainingAfter = deadline - Date.now();
             if (remainingAfter <= 0) break;
             await sleep(Math.min(RESUME_PEER_SWITCH_DELAY_MS, remainingAfter));
--- a/src/P2PManager.ts
+++ b/src/P2PManager.ts
@@ -16,12 +16,34 @@ import { DebugProxy, getChecksumAddress, LocalDiscoveryServer } from "@/utils";
 import type { Logger } from "@/utils";
 import { Buffer } from "buffer";
 import { config, isNodeRuntime } from "@/utils/config";
-import { Address } from "./types/types";
+import { Address, ChannelId } from "./types/types";
 import { isInstanceOfRpcService } from "./utils/ObjectChecks";
 import type ARpcService from "@/rpc/ARpcService";
 import RemoteRpcProxy, { RemoteRpcProxyType } from "./rpc/RemoteRpcProxy";
 import type { CustomRpcConstructor } from "./rpc/registry";
 import { LoggerUtils } from "@/utils/LoggerUtils";
+import { sleep } from "@/utils";
+
+/** Max distinct connected peers tried, one at a time, before giving up. */
+const MAX_RESUME_SYNC_PEERS = 3;
+/** Fixed delay between peer-switch attempts. */
+const RESUME_PEER_SWITCH_DELAY_MS = 500;
+/**
+ * Per-attempt request timeout for a resume sync. Deliberately NOT
+ * timeConfig.agreementTime (5s) - a placeholder pending devops SPIKE-0's
+ * measured RTT figures for the resume path specifically.
+ */
+const RESUME_SYNC_TIMEOUT_MS = 8000;
+/** Resume must leave window for be-03's self-defense/back-off. */
+const RESUME_BUDGET_FRACTION = 0.5;
+
+export type ResumeResult =
+    | { status: "restored"; hydrated: boolean }
+    | {
+          status: "failed";
+          reason: "transport" | "sync-timeout" | "no-peers" | "unknown";
+          retryable: boolean;
+      };
 
 class P2PManager<TCustomRpc extends MainRpcService = MainRpcService>
     implements IOnMessage
@@ -39,6 +61,14 @@ class P2PManager<TCustomRpc extends MainRpcService = MainRpcService>
     holepunch: Holepunch;
     self = config.DEBUG_P2P_MANAGER ? DebugProxy.createProxy(this) : this;
     preferredTransport: TransportType = TransportType.HOLEPUNCH;
+
+    // Dedups concurrent resumeFromBackground(channelId) calls - keyed by the
+    // stringified channelId (BytesLike has more than one valid runtime
+    // representation, so normalize before using it as a Map key).
+    private resumeInFlightByChannelId = new Map<
+        string,
+        Promise<ResumeResult>
+    >();
 
     private rpcRequestCounter = 0;
     private pendingRpcRequests = new Map<
@@ -348,6 +378,143 @@ class P2PManager<TCustomRpc extends MainRpcService = MainRpcService>
             }
         }
         return addresses;
+    }
+
+    /**
+     * Mobile reconnect: recover a backgrounded channel without tearing down
+     * the process-global Holepunch swarm (it's shared across every open
+     * channel - see Holepunch.ts/HolepunchRelay.ts). Channel-scoped recovery
+     * is topic rejoin + re-handshake + an awaited resync; no swarm
+     * dispose/destroy/close. Dedups concurrent calls for the same channel by
+     * returning the in-flight promise.
+     */
+    public resumeFromBackground(channelId: ChannelId): Promise<ResumeResult> {
+        const key = String(channelId);
+        const inFlight = this.resumeInFlightByChannelId.get(key);
+        if (inFlight) return inFlight;
+
+        const resumePromise = this.doResumeFromBackground(channelId).finally(
+            () => {
+                this.resumeInFlightByChannelId.delete(key);
+            }
+        );
+        this.resumeInFlightByChannelId.set(key, resumePromise);
+        return resumePromise;
+    }
+
+    private async doResumeFromBackground(
+        channelId: ChannelId
+    ): Promise<ResumeResult> {
+        const deadline =
+            Date.now() +
+            this.stateManager.getTimeoutWaitTimeSeconds() *
+                1000 *
+                RESUME_BUDGET_FRACTION;
+
+        this.stateManager.p2pEventHooks.onResumePhase?.("reconnecting");
+
+        try {
+            // Topic rejoin (Holepunch.rejoinTopics via join()) + whatever
+            // re-handshake the existing connection flow drives. Never touches
+            // the shared swarm singleton itself.
+            await this.tryOpenConnectionToChannel(String(channelId));
+        } catch (e) {
+            this.logger.warn("resumeFromBackground - transport rejoin failed", {
+                channelId,
+                error: e instanceof Error ? e.message : String(e)
+            });
+            return { status: "failed", reason: "transport", retryable: true };
+        }
+
+        try {
+            await this.stateManager.storage.hydrate();
+        } catch (e) {
+            this.logger.warn("resumeFromBackground - storage hydrate failed", {
+                channelId,
+                error: e instanceof Error ? e.message : String(e)
+            });
+            return { status: "failed", reason: "unknown", retryable: true };
+        }
+
+        this.stateManager.p2pEventHooks.onResumePhase?.("resyncing");
+
+        try {
+            return await this.resyncWithPeerSwitch(channelId, deadline);
+        } catch (e) {
+            this.logger.warn("resumeFromBackground - unexpected error", {
+                channelId,
+                error: e instanceof Error ? e.message : String(e)
+            });
+            return { status: "failed", reason: "unknown", retryable: true };
+        }
+    }
+
+    /**
+     * Try up to MAX_RESUME_SYNC_PEERS connected peers (one at a time, 500ms
+     * apart) via the awaitable resume sync, within the remaining budget. Also
+     * sidesteps a concurrent reactive (non-resume) sync blacklisting a given
+     * peer - a collision/failure just switches to the next candidate peer.
+     */
+    private async resyncWithPeerSwitch(
+        channelId: ChannelId,
+        deadline: number
+    ): Promise<ResumeResult> {
+        const attemptedPeers = new Set<Address>();
+        let attempts = 0;
+
+        while (
+            Date.now() < deadline &&
+            attemptedPeers.size < MAX_RESUME_SYNC_PEERS
+        ) {
+            const peer = this.pickNextResumePeer(attemptedPeers);
+            if (!peer) {
+                const remainingMs = deadline - Date.now();
+                if (remainingMs <= 0) break;
+                await sleep(Math.min(RESUME_PEER_SWITCH_DELAY_MS, remainingMs));
+                continue;
+            }
+            attemptedPeers.add(peer);
+
+            const remainingMs = deadline - Date.now();
+            if (remainingMs <= 0) break;
+            attempts++;
+
+            const outcome = await this.localRpc.spectateService.syncAwaitable(
+                peer,
+                channelId,
+                {
+                    resumeMode: true,
+                    timeoutMs: Math.min(RESUME_SYNC_TIMEOUT_MS, remainingMs)
+                }
+            );
+
+            if (outcome.ok) {
+                return { status: "restored", hydrated: !outcome.localWasAhead };
+            }
+
+            this.logger.debug(
+                "resumeFromBackground - peer sync failed; switching peer",
+                { channelId, peer, reason: outcome.reason }
+            );
+
+            if (attemptedPeers.size >= MAX_RESUME_SYNC_PEERS) break;
+            const remainingAfter = deadline - Date.now();
+            if (remainingAfter <= 0) break;
+            await sleep(Math.min(RESUME_PEER_SWITCH_DELAY_MS, remainingAfter));
+        }
+
+        return {
+            status: "failed",
+            reason: attempts === 0 ? "no-peers" : "sync-timeout",
+            retryable: true
+        };
+    }
+
+    private pickNextResumePeer(attempted: Set<Address>): Address | undefined {
+        for (const peer of this.getConnectedPeers()) {
+            if (!attempted.has(peer)) return peer;
+        }
+        return undefined;
     }
 }
 
--- a/src/P2PManager.ts
+++ b/src/P2PManager.ts
@@ -503,10 +503,10 @@ class P2PManager<TCustomRpc extends MainRpcService = MainRpcService>
             this.stateManager.p2pEventHooks.onResumePhase?.("reconnecting");
 
             // Storage must be hydrated strictly before the transport/topic
-            // join (same ordering Storage.attachPersistence and
-            // P2pRuntimeHost's connect path require) - otherwise a handshake
-            // completing on the rejoined transport can fire a reactive sync
-            // that mutates state on top of a half-replayed store.
+            // join (same ordering Storage.bind and P2pRuntimeHost's connect
+            // path require) - otherwise a handshake completing on the
+            // rejoined transport can fire a reactive sync that mutates state
+            // on top of a half-replayed store.
             const hydrateBudgetMs = Math.min(
                 Math.max(0, deadline - Date.now()),
                 totalBudgetMs * RESUME_HYDRATE_BUDGET_FRACTION
--- a/src/P2PManager.ts
+++ b/src/P2PManager.ts
@@ -23,9 +23,13 @@ import RemoteRpcProxy, { RemoteRpcProxyType } from "./rpc/RemoteRpcProxy";
 import type { CustomRpcConstructor } from "./rpc/registry";
 import { LoggerUtils } from "@/utils/LoggerUtils";
 import { sleep } from "@/utils";
+import type { SyncFailReason } from "@/rpc/services/spectate/SpectateService";
 
 /** Max distinct connected peers tried, one at a time, before giving up. */
 const MAX_RESUME_SYNC_PEERS = 3;
+/** Caps TOTAL attempts (including collision-deferred retries) so a peer that
+ * always collides can't spin the whole budget without progress. */
+const MAX_RESUME_SYNC_ATTEMPTS = MAX_RESUME_SYNC_PEERS * 2;
 /** Fixed delay between peer-switch attempts. */
 const RESUME_PEER_SWITCH_DELAY_MS = 500;
 /**
@@ -36,14 +40,53 @@ const RESUME_PEER_SWITCH_DELAY_MS = 500;
 const RESUME_SYNC_TIMEOUT_MS = 8000;
 /** Resume must leave window for be-03's self-defense/back-off. */
 const RESUME_BUDGET_FRACTION = 0.5;
+/** Share of the total resume budget given to hydrate and to transport rejoin
+ * each; the rest goes to peer rotation. No single correct split - chosen so
+ * neither phase alone can exhaust the budget before rotation gets a turn. */
+const RESUME_HYDRATE_BUDGET_FRACTION = 0.25;
+const RESUME_REJOIN_BUDGET_FRACTION = 0.25;
+
+export type ResumeFailReason =
+    | "transport"
+    | "hydration-failed"
+    | "no-peers"
+    | "deadline-expired"
+    | SyncFailReason;
 
 export type ResumeResult =
-    | { status: "restored"; hydrated: boolean }
-    | {
-          status: "failed";
-          reason: "transport" | "sync-timeout" | "no-peers" | "unknown";
-          retryable: boolean;
-      };
+    | { status: "restored"; localWasAhead: boolean }
+    | { status: "failed"; reason: ResumeFailReason; retryable: boolean };
+
+/** Raised by withDeadline() when the wrapped phase outran its own sub-budget
+ * (not the phase itself failing). */
+class ResumePhaseTimeoutError extends Error {}
+
+/** Races `promise` against `timeoutMs`; on timeout throws
+ * ResumePhaseTimeoutError but does not cancel the underlying operation. */
+function withDeadline<T>(promise: Promise<T>, timeoutMs: number): Promise<T> {
+    return new Promise<T>((resolve, reject) => {
+        const timer = setTimeout(
+            () => {
+                reject(
+                    new ResumePhaseTimeoutError(
+                        `resume phase exceeded ${timeoutMs}ms`
+                    )
+                );
+            },
+            Math.max(0, timeoutMs)
+        );
+        promise.then(
+            (value) => {
+                clearTimeout(timer);
+                resolve(value);
+            },
+            (error) => {
+                clearTimeout(timer);
+                reject(error);
+            }
+        );
+    });
+}
 
 class P2PManager<TCustomRpc extends MainRpcService = MainRpcService>
     implements IOnMessage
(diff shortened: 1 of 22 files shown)


src/P2PManager.ts
```diff
@@ -347,6 +377,143 @@ class P2PManager<TCustomRpc extends MainRpcService = MainRpcService>
...
+
+    /**
+     * Mobile reconnect: recover a backgrounded channel without tearing down
+     * the process-global Holepunch swarm (it's shared across every open
+     * channel - see Holepunch.ts/HolepunchRelay.ts). Channel-scoped recovery
+     * is topic rejoin + re-handshake + an awaited resync; no swarm
+     * dispose/destroy/close. Dedups concurrent calls for the same channel by
+     * returning the in-flight promise.
+     */
+    public resumeFromBackground(channelId: ChannelId): Promise<ResumeResult> {
```
daiagi: drop `channelId` arg
it is both unverified (can be junk) and uneeded
use `const channelId = this.stateManager.getChannelId();` instead

and then you can replace `resumeInFlight` to be `private resumeInFlight?: Promise<ResumeResult>`

daiagi (changes requested):
1. Missing tests
2. Unreachable from consumer API surface, i.e there is no resume method on `P2pInstance`

3. 2 bugs, described and caught by the tests below:

```ts

/**
 * E2E Tests for P2PManager.resumeFromBackground
 *
 * Maps to: src/P2PManager.ts (resumeFromBackground / resyncWithPeerSwitch)
 *          src/rpc/services/spectate/SpectateService.ts (syncAwaitable)
 *
 * resumeFromBackground isn't reachable from P2pInstance, so drive it host-side
 * via execOnHost.
 */
describe("E2E: resume from background", function () {
    // BUG: resume gives up on a peer that was only momentarily busy.
    //
    // setup - 2 peers, and a sync to peer B is already running. that's the
    // normal case, not a corner: a peer back from background is behind, and
    // being behind is what makes the validation pipeline start a sync at B.
    //
    // expect - resume waits the ~1 RTT for that sync to drain, retries B, done.
    // actual - resume marks B "attempted" before it ever talks to B, sees the
    //          collision, and never retries it. B is the only peer, so it then
    //          sleeps in 500ms slices for the whole 6s budget and returns
    //          {status:'failed', reason:'sync-timeout'} - nothing timed out,
    //          and the sync it collided with has long since synced us.
    it("resumes through its only peer when a reactive sync is already in flight", async function () {
        // resume budget = 50% of (p2p + agreement + chainFallback) = 6s.
        // control port timeout = agreementTime = 10s -> the port outlives the
        // budget, so the assertion fails, not the harness.
        await h.lifecycle.start(2, 0, {
            timeConfig: {
                p2pTime: 1,
                agreementTime: 10,
                chainFallbackTime: 1,
                evidenceTime: 4
            }
        });

        const resuming = h.getPeer(0);
        const onlyPeer = h.getPeer(1);

        const outcome = await h.execOnHost(
            resuming,
            async (sm, args) => {
                const channelId = sm.getChannelId();

                // same call the reactive path makes -> holds the in-flight
                // guard for this peer
                sm.p2pManager.localRpc.spectateService.sync(
                    args.peerAddress,
                    channelId
                );

                // resume races that sync, like on a real foreground
                const spectate = sm.p2pManager.localRpc
                    .spectateService as unknown as {
                    inFlightByPeerAddress: Set<string>;
                };
                const guardHeldAtStart = spectate.inFlightByPeerAddress.has(
                    args.peerAddress
                );

                const startedAt = Date.now();
                const result =
                    await sm.p2pManager.resumeFromBackground(channelId);
                const elapsedMs = Date.now() - startedAt;

                return { result, elapsedMs, guardHeldAtStart };
            },
            { peerAddress: onlyPeer.address }
        );

        // if the sync had already drained there was no collision -> the test
        // proved nothing
        expect(
            outcome.guardHeldAtStart,
            "precondition: the reactive sync must still hold the in-flight guard when the resume starts"
        ).to.equal(true);

        expect(
            outcome.result.status,
            `resume against a healthy, in-sync peer must not report failure ` +
                `(got ${JSON.stringify(outcome.result)} after ${outcome.elapsedMs}ms)`
        ).to.equal("restored");
    });

    // BUG: a resume lets provable fraud go unpunished.
    //
    // peer B answers the spectate request with bytes that aren't a SyncPayload.
    // that's B's fault and nothing else's - our own state can't make B's reply
    // undecodable. on the normal sync path we blacklist B for it (see
    // E2E-SpectateStaleProofGuard). on a resume, abort() returns early and B
    // keeps its connection and its clean record, free to serve junk forever.
    //
    // the two halves run the SAME fraud down both paths, so the contrast is the
    // assertion: same input, punished one way, unpunished the other.
    it("blacklists a peer that serves an undecodable payload - on both the sync and resume paths", async function () {
        await h.lifecycle.start(2, 0, {
            timeConfig: {
                p2pTime: 1,
                agreementTime: 10,
                chainFallbackTime: 1,
                evidenceTime: 4
            }
        });

        const victim = h.getPeer(0);
        const fraudster = h.getPeer(1);

        await h.rpcStub.stubSpectateJunkPayload([1]);

        // resume first, while B still has a clean record
        const afterResume = await h.execOnHost(
            victim,
            async (sm, args) => {
                const result = await sm.p2pManager.resumeFromBackground(
                    sm.getChannelId()
                );
                return {
                    result,
                    blacklisted: sm.p2pManager.isBlacklisted(args.peerAddress)
                };
            },
            { peerAddress: fraudster.address }
        );

        // control - the plain fire-and-forget path, same junk, same peer.
        // established behaviour: junk payload -> blacklist.
        const afterSync = await h.execOnHost(
            victim,
            async (sm, args) => {
                sm.p2pManager.localRpc.spectateService.sync(
                    args.peerAddress,
                    sm.getChannelId()
                );
                await new Promise((r) => setTimeout(r, 2000));
                return sm.p2pManager.isBlacklisted(args.peerAddress);
            },
            { peerAddress: fraudster.address }
        );

        expect(
            afterSync,
            "control: the normal sync path must blacklist a peer serving junk"
        ).to.equal(true);

        expect(
            afterResume.blacklisted,
            `resume must punish the same fraud the sync path punishes ` +
                `(resume returned ${JSON.stringify(afterResume.result)})`
        ).to.equal(true);
    });
});

```

lukastanisic99:
@momoshell I commented on #388 and on Trello - hope that's enough and it covers everything you need for this PR too, if not ping me with explicit questions

lukastanisic99:
@momoshell 

Five 🔴s block: transport is exposed before hydration, a partially applied sync can later report a false restore, a colliding reactive sync keeps destructive failure policy, well-formed fraudulent payloads remain eligible on future resumes, and the core background-recovery workflow has no test.

## 🧱 Fundamental

- 🔴 **[FR1] Resume exposes transport before durable state is hydrated.** [`doResumeFromBackground`](https://github.com/peer3to/state-channels-plus/blob/9f8967e4f8a773befd5baf3dc5cf5f7499713b0c/src/P2PManager.ts#L413-L435) rejoins first and hydrates second, despite the [`Storage` contract](https://github.com/peer3to/state-channels-plus/blob/9f8967e4f8a773befd5baf3dc5cf5f7499713b0c/src/storage/Storage.ts#L143-L160) requiring hydration “strictly before any transport/gossip join” and the [initial connection path](https://github.com/peer3to/state-channels-plus/blob/9f8967e4f8a773befd5baf3dc5cf5f7499713b0c/src/evm/p2pRuntime/P2pRuntimeHost.ts#L614-L623) enforcing that order. A completed handshake can start a reactive sync while hydration is replaying storage; if hydration fails, the peer remains joined and continues on stale or partially replayed state.

  > **Fix FR1-FIX**
  >
  > Hydrate before `tryOpenConnectionToChannel`. If a backgrounded transport can remain live, gate inbound handshake, RPC, and gossip dispatch until hydration completes, and undo or quarantine the connection when hydration fails. Hold hydration in a deterministic test and prove no inbound handler runs before release.

- 🔴 **[FR2] A failed restore can poison storage and make the next peer report a false success.** [`persistSyncPayload`](https://github.com/peer3to/state-channels-plus/blob/9f8967e4f8a773befd5baf3dc5cf5f7499713b0c/src/rpc/services/spectate/SpectateService.ts#L995-L1034) writes snapshots and finalized blocks before `unsafeSetLatestState`. If that final call throws, resume suppresses cleanup and switches peers. The next payload at the same height hits the [`localLatestHeight >= finalizedHeight` shortcut](https://github.com/peer3to/state-channels-plus/blob/9f8967e4f8a773befd5baf3dc5cf5f7499713b0c/src/rpc/services/spectate/SpectateService.ts#L972-L985), skips the failed VM/fork restoration, returns `ok`, and [`P2PManager` reports `restored`](https://github.com/peer3to/state-channels-plus/blob/9f8967e4f8a773befd5baf3dc5cf5f7499713b0c/src/P2PManager.ts#L480-L490).

  > **Fix FR2-FIX**
  >
  > Make sync application atomic: validate before commit and either transactionally commit storage plus active state or roll back/rebuild after any post-write failure. The “already ahead” path must prove the active fork, state hash, and VM are restored, not only that a block exists. Inject a failure in `unsafeSetLatestState`, retry through a second peer, and assert resume cannot return `restored` until all runtime state converges.

- 🟠 **[FO3] The advertised eight-second resume attempt is still rejected at `agreementTime`.** The request is allowed up to [`RESUME_SYNC_TIMEOUT_MS = 8000`](https://github.com/peer3to/state-channels-plus/blob/9f8967e4f8a773befd5baf3dc5cf5f7499713b0c/src/P2PManager.ts#L31-L36), but response application rejects any RTT above [`timeConfig.agreementTime`](https://github.com/peer3to/state-channels-plus/blob/9f8967e4f8a773befd5baf3dc5cf5f7499713b0c/src/rpc/services/spectate/SpectateService.ts#L252-L265). With a five-second agreement window, a valid six-second resume response fails despite being inside the dedicated resume timeout.

  > **Fix FO3-FIX**
  >
  > Carry the resume attempt’s absolute deadline into response validation and use that bound in resume mode; keep `agreementTime` for ordinary reactive sync. Test valid responses on both sides of the agreement-time and resume-time boundaries.

- 🔴 **[SR1] A Byzantine peer is not punished for well-formed but cryptographically invalid resume data.** Resume blacklists only ABI-undecodable bytes in [`failSync`](https://github.com/peer3to/state-channels-plus/blob/9f8967e4f8a773befd5baf3dc5cf5f7499713b0c/src/rpc/services/spectate/SpectateService.ts#L1115-L1124), while `abort` suppresses every side effect in resume mode. Wrong fork responses, invalid milestone proofs, state-hash mismatches, invalid outbound blocks, and invalid balances all collapse to `aborted` at [`applySyncResponse`](https://github.com/peer3to/state-channels-plus/blob/9f8967e4f8a773befd5baf3dc5cf5f7499713b0c/src/rpc/services/spectate/SpectateService.ts#L461-L505). An authenticated peer can therefore serve proven fraud on every resume and remain connected.

  > **Fix SR1-FIX**
  >
  > Split peer-provable invalidity from local races, timeouts, and internal errors. Blacklist and disconnect for every request-contradicting or cryptographically invalid response in both modes; suppress punishment only for genuinely ambiguous local failures. Add resume tests with valid encoding but invalid genesis, milestone proof, state hash, outbound block, and balance.

## 🏁 Race conditions

- 🔴 **[RR1] An in-flight reactive sync keeps destructive failure policy during resume.** [`syncAwaitable`](https://github.com/peer3to/state-channels-plus/blob/9f8967e4f8a773befd5baf3dc5cf5f7499713b0c/src/rpc/services/spectate/SpectateService.ts#L125-L151) returns `in-flight-collision`, but the existing run remains `resumeMode: false`. A handshake can start exactly that reactive run, and its request or validation failure still blacklists the peer or hard-aborts the channel through [`runSync`](https://github.com/peer3to/state-channels-plus/blob/9f8967e4f8a773befd5baf3dc5cf5f7499713b0c/src/rpc/services/spectate/SpectateService.ts#L176-L211) and [`abort`](https://github.com/peer3to/state-channels-plus/blob/9f8967e4f8a773befd5baf3dc5cf5f7499713b0c/src/rpc/services/spectate/SpectateService.ts#L1127-L1157). Sleeping and retrying in `P2PManager` cannot make that concurrent run resume-safe.

  > **Fix RR1-FIX**
  >
  > Store the actual per-peer in-flight promise and a policy that can be promoted to resume-safe, or quiesce/cancel reactive sync before resume starts. Hold a colliding reactive request, make it fail while resume is active, and assert it neither blacklists nor aborts before recovery continues.

- 🟠 **[RO2] A collision on peer A prevents trying healthy peer B.** Collision deliberately leaves A out of [`attemptedPeers`](https://github.com/peer3to/state-channels-plus/blob/9f8967e4f8a773befd5baf3dc5cf5f7499713b0c/src/P2PManager.ts#L498-L505), and [`pickNextResumePeer`](https://github.com/peer3to/state-channels-plus/blob/9f8967e4f8a773befd5baf3dc5cf5f7499713b0c/src/P2PManager.ts#L518-L522) always returns the first unattempted connection. A long reactive sync on A therefore consumes the entire budget in 500 ms retries while B is never contacted.

  > **Fix RO2-FIX**
  >
  > Track colliding peers in a temporary cooldown set, try every other connected candidate, then revisit collisions after the other peers or after their in-flight run settles. Hold A in flight, keep B healthy, and assert recovery goes through B.

- 🟠 **[RO3] `dispose` can tear the runtime down underneath an active resume.** Runtime-port messages launch concurrent [`handleRequest` promises](https://github.com/peer3to/state-channels-plus/blob/9f8967e4f8a773befd5baf3dc5cf5f7499713b0c/src/evm/p2pRuntime/P2pRuntimeHost.ts#L758-L767), while resume has no disposal guard before or between rejoin, hydration, and sync. A concurrent `dispose` marks the manager disposed and closes P2P/EVM resources, but resume can continue later and post a response from torn-down state.

  > **Fix RO3-FIX**
  >
  > Serialize lifecycle requests, or cancel and join `resumeInFlight` as part of disposal with disposed checks before every resume side effect. Pause resume in hydration, call `dispose`, then prove no late join, sync, hook, or port response occurs.

## ⚡ Performance

- 🟠 **[PO1] The half-window deadline does not bound the recovery operation.** The deadline is computed before rejoin and hydrate, but those waits are unbounded and only [`resyncWithPeerSwitch`](https://github.com/peer3to/state-channels-plus/blob/9f8967e4f8a773befd5baf3dc5cf5f7499713b0c/src/P2PManager.ts#L407-L441) reads it. Inside sync, `timeoutMs` bounds only the RPC request; response validation and application are unbounded after bytes arrive. The public client adds a separate fixed [`30 s request timeout`](https://github.com/peer3to/state-channels-plus/blob/9f8967e4f8a773befd5baf3dc5cf5f7499713b0c/src/evm/p2pRuntime/P2pRuntimeClient.ts#L249-L254), so a host budget above 30 seconds rejects at the consumer while host mutation continues.

  > **Fix PO1-FIX**
  >
  > Own one absolute deadline and cancellation signal across rejoin, hydrate, request, validation, and atomic application. Align or disable the generic client timeout for this host-bounded operation. Test a stalled stage and a channel window above 60 seconds; the consumer and host must settle once with the same result and no background mutation.

- 🟠 **[PO2] Every resume registers the same Holepunch topic again.** Resume calls `tryOpenConnectionToChannel`, which reaches [`Holepunch.join`](https://github.com/peer3to/state-channels-plus/blob/9f8967e4f8a773befd5baf3dc5cf5f7499713b0c/src/Holepunch.ts#L74-L83); `join` unconditionally pushes the topic. Every foreground cycle grows the array, and every relay reconnect and disposal replays all duplicates.

  > **Fix PO2-FIX**
  >
  > Make topic membership idempotent by topic hex, or add a rejoin operation that does not register ownership again. Resume the same channel repeatedly, reset the relay, and assert one stored topic, one rejoin, and one leave.

- 🔴 **[TR1] The new E2E never performs the background recovery the PR ships.** Crediting daiagi’s still-partial review feedback in timeline row 2: both cases start already-connected, already-synced peers, and the test config makes [`tryOpenConnectionToChannel` a Node no-op](https://github.com/peer3to/state-channels-plus/blob/9f8967e4f8a773befd5baf3dc5cf5f7499713b0c/src/P2PManager.ts#L277-L295). The success case asserts only [`status === "restored"`](https://github.com/peer3to/state-channels-plus/blob/9f8967e4f8a773befd5baf3dc5cf5f7499713b0c/test/e2e/E2E-ResumeFromBackground.test.ts#L40-L48); it does not disconnect, restore persisted state, catch up, or prove authoring resumes.

  > **Fix TR1-FIX**
  >
  > Restart or background a persistence-backed peer, let the surviving peer advance, invoke `P2pInstance.resumeFromBackground`, then assert transport recovery, hydrated storage and live-state convergence, both phase hooks, and a subsequent authored transition. Add a representative non-local transport workflow so topic rejoin is exercised.

- 🟠 **[TO2] `P2PManager` has no exhaustive black-box unit coverage.** There are no component cases for concurrent-call dedup, phase order, transport failure, hydration failure, in-flight reset after every outcome, no peers, deadline boundaries, remaining-time capping, three-peer exhaustion, or unexpected resync failure, despite the [repo rule requiring unit and E2E coverage for every `src` change](https://github.com/peer3to/state-channels-plus/blob/9f8967e4f8a773befd5baf3dc5cf5f7499713b0c/AGENTS.md#L53-L64).

  > **Fix TO2-FIX**
  >
  > Add harness-backed component tests through the public resume surface for every listed normal, failure, retry, boundary, and concurrency outcome. Assert calls, phases, result reason, retryability, and that a later call starts fresh after settlement.

- 🟠 **[TO3] `SpectateService` has no unit matrix for its new outcome and side-effect contract.** The diff changes roughly twenty exits into `SyncOutcome` values and makes punishment depend on resume mode, but no direct tests cover success fields, collision, request failure, RTT timeout, objective invalidity, malformed bytes, unexpected exceptions, guard cleanup, or participating versus pre-participating abort behavior.

  > **Fix TO3-FIX**
  >
  > Exercise `syncAwaitable` through the real harness with domain payloads for every outcome and both policies. Assert the returned discriminant and the exact blacklist, disconnect, abort, and in-flight-guard side effects.

- 🟠 **[TO4] The “peer-switch” test has only one candidate and never switches peers.** [`E2E-ResumeFromBackground`](https://github.com/peer3to/state-channels-plus/blob/9f8967e4f8a773befd5baf3dc5cf5f7499713b0c/test/e2e/E2E-ResumeFromBackground.test.ts#L5-L20) retries the same peer after a collision. It cannot catch RO2, a first-peer ordinary failure followed by second-peer success, sequential ordering, or the three-distinct-peer cap.

  > **Fix TO4-FIX**
  >
  > Add deterministic multi-peer cases: hold A in collision while B succeeds, fail A normally while B succeeds, fail three of four peers, and exhaust all candidates. Record request order and prove no parallel sync attempts.

- 🟠 **[TO5] The malformed-payload test claims both sync and resume paths but runs only resume.** The test invokes only [`resumeFromBackground`](https://github.com/peer3to/state-channels-plus/blob/9f8967e4f8a773befd5baf3dc5cf5f7499713b0c/test/e2e/E2E-ResumeFromBackground.test.ts#L65-L82), does not assert the returned failure reason, and does not prove participant state survives. It also covers only undecodable junk, not SR1’s well-formed invalid payloads.

  > **Fix TO5-FIX**
  >
  > Split direct reactive-sync and resume cases. Assert failure reason, blacklist/disconnect, preserved channel status, no hard abort, and the objective-invalid versus transient-failure policy.

- 🟠 **[TO6] The collision E2E is timing-dependent and probes a private guard instead of controlling the request.** It launches a healthy sync, makes another control round-trip, then reads [`inFlightByPeerAddress` through a structural cast](https://github.com/peer3to/state-channels-plus/blob/9f8967e4f8a773befd5baf3dc5cf5f7499713b0c/test/fixtures/customRpc/harnessControl/services/spectate/SpectateControlRpcMethods.ts#L96-L102). The sync can finish between those operations, and final success does not prove a collision occurred and was retried.

  > **Fix TO6-FIX**
  >
  > Add a named hold/release fixture around the responder’s real `onSpectateRequest`, wait for “entered,” launch resume, then release. Record request counts and assert the collision outcome occurred before the retry rather than inspecting private state.

- 🟠 **[TO7] The new phase event is not covered across the runtime boundary.** Production adds [`onResumePhase`](https://github.com/peer3to/state-channels-plus/blob/9f8967e4f8a773befd5baf3dc5cf5f7499713b0c/src/P2pEventHooks.ts#L27-L31), but the harness [`EventSpies` type](https://github.com/peer3to/state-channels-plus/blob/9f8967e4f8a773befd5baf3dc5cf5f7499713b0c/test/harness/core/types.ts#L150-L179) has no matching member. No test can assert that `reconnecting` then `resyncing` reaches the consumer, or that early failure suppresses the second phase.

  > **Fix TO7-FIX**
  >
  > Derive the spy signatures from `P2pEventHooks`, wire `onResumePhase` through the harness, and assert event order and payload through `P2pInstance.on` for success, transport failure, and hydration failure.

- 🟡 **[TY8] `localWasAhead` and public `hydrated` semantics have no assertion.** Production returns [`hydrated: !outcome.localWasAhead`](https://github.com/peer3to/state-channels-plus/blob/9f8967e4f8a773befd5baf3dc5cf5f7499713b0c/src/P2PManager.ts#L489-L490), but the tests assert only status or blacklist. Equal/ahead storage, actual state application, and the meaning of `hydrated` are unverified.

  > **Fix TY8-FIX**
  >
  > Cover behind, equal, and ahead durable storage, assert whether active state was applied, and assert every public result field. Rename the public field if it means “peer sync applied” rather than “storage hydrate succeeded.”

## ♻️ Code reuse

- 🟠 **[CO1] The test harness adds a weaker duplicate of `startSync`.** Existing [`startSync`](https://github.com/peer3to/state-channels-plus/blob/9f8967e4f8a773befd5baf3dc5cf5f7499713b0c/test/fixtures/customRpc/harnessControl/services/spectate/SpectateControlRpcMethods.ts#L50-L69) derives and validates the host channel ID; new [`triggerSync`](https://github.com/peer3to/state-channels-plus/blob/9f8967e4f8a773befd5baf3dc5cf5f7499713b0c/test/fixtures/customRpc/harnessControl/services/spectate/SpectateControlRpcMethods.ts#L91-L94) accepts an unchecked caller-supplied channel and starts the same operation.

  > **Fix CO1-FIX**
  >
  > Delete `triggerSync` and use `startSync`. Put deterministic hold/record behavior in the existing stub service rather than forking sync orchestration.

## 📏 AGENTS.md adherence

- 🟠 **[AO1] The new harness endpoint embeds a private accessor and forbidden structural cast in routable `RpcMethods`.** [`isSyncInFlight`](https://github.com/peer3to/state-channels-plus/blob/9f8967e4f8a773befd5baf3dc5cf5f7499713b0c/test/fixtures/customRpc/harnessControl/services/spectate/SpectateControlRpcMethods.ts#L96-L102) re-states a real SDK type with `as unknown as` and performs accessor logic inside the endpoint. The [repository RPC rule](https://github.com/peer3to/state-channels-plus/blob/9f8967e4f8a773befd5baf3dc5cf5f7499713b0c/AGENTS.md#L66-L78) requires accessors on the non-routable service and tests through public component behavior.

  > **Fix AO1-FIX**
  >
  > Remove the private-set probe and stage the race through the named hold fixture in TO6-FIX. If an accessor remains necessary, put typed access behind `SpectateControlService` and let the endpoint only delegate.

- 🟡 **[AY2] The new code adds essay-length comments that restate control flow.** The resume method’s multi-line narration and the [`failSync` policy block](https://github.com/peer3to/state-channels-plus/blob/9f8967e4f8a773befd5baf3dc5cf5f7499713b0c/src/rpc/services/spectate/SpectateService.ts#L1104-L1114) exceed the repository’s [“simple and to the point” comment rule](https://github.com/peer3to/state-channels-plus/blob/9f8967e4f8a773befd5baf3dc5cf5f7499713b0c/AGENTS.md#L119-L127) and already contain stale claims, such as treating only malformed bytes as unambiguous peer fault.

  > **Fix AY2-FIX**
  >
  > Keep short comments for non-obvious invariants only. Encode policy distinctions in named failure kinds and tests so comments do not have to defend or duplicate the implementation.

## 🧹 Miscellaneous (scope)

- 🟠 **[MO1] Consumer-visible failure reasons discard the actual peer outcome.** After any request failure, malformed payload, invalid proof, or exhausted candidate set, [`resyncWithPeerSwitch`](https://github.com/peer3to/state-channels-plus/blob/9f8967e4f8a773befd5baf3dc5cf5f7499713b0c/src/P2PManager.ts#L511-L515) returns `sync-timeout`; if rejoin or hydration consumes the deadline before an attempt, it returns `no-peers` even when peers are connected.

  > **Fix MO1-FIX**
  >
  > Track the terminal cause separately from attempt count and return a precise public reason. Reserve `no-peers` for an observed empty candidate set and `sync-timeout` for an expired deadline; add consumer assertions for each cause.

- 🟡 **[MY2] `hydrated` is false after successful hydration whenever local blocks are already current.** Hydration always completes before sync, but success inverts `localWasAhead`, so the field describes whether peer data was applied rather than whether durable hydration succeeded.

  > **Fix MY2-FIX**
  >
  > Rename it to `syncApplied` and expose `localWasAhead`, or return `hydrated: true` for every post-hydration restore. Document the chosen contract and test it at the public API.

## ❓ Open questions

- 🟢 **[OG1] No product decision is required to fix the findings above.**

  > **Fix OG1-FIX**
  >
  > No change required.

src/P2PManager.ts
momoshell: Good catch — fixed in `cb61b395`. Dropped the `channelId` param; `resumeFromBackground()` now reads it from `this.stateManager.getChannelId()`, and the per-channel map collapsed to a single `private resumeInFlight?: Promise<ResumeResult>` since this instance only ever tracks one channel.

momoshell:
Thanks for the thorough pass, all fixed:

1. **Missing tests** — added in `10c01dc5`: `E2E-ResumeFromBackground.test.ts` covers both bugs below, driven through the real `P2pInstance.resumeFromBackground()` consumer surface rather than an `execOnHost` escape hatch.
2. **Unreachable from consumer API** — fixed in `c5e83e67`: wired `resumeFromBackground` through the runtime port (worker protocol + host dispatch + client + `P2pInstance` method), same pattern as `quiesce`/`disconnectFromPeers`.
3. **Bug — resume gives up on a momentarily-busy peer** — fixed in `6f93394a`: an `in-flight-collision` no longer marks the peer "attempted"; it retries the same peer after the switch delay instead of burning a peer slot on a sync that was about to resync us anyway.
4. **Bug — undecodable payload goes unpunished on resume** — fixed in `161162e7`: decode failures get their own `malformed-payload` reason, and `failSync` blacklists for it unconditionally even in resume mode, since it's unambiguous peer fault rather than a legitimate on-chain race.

@lukastanisic99, your test cases made these trivial to verify, and folded them in almost verbatim (via the public API instead of `execOnHost`).

daiagi:
> your test cases made these trivial to verify

Glad you found them useful

PR #389, opened 2026-07-12
Agreement Manager unit tests

daiagi:
1. Unit test suite for `AgreementManager`
2.  Logged  a possible race condition bug: https://trello.com/c/Ij8sVdHd
3. Changes to `AgreementManager`

    - bug fix: `getStateProof` - take participants change set up until requestd block height.
    - change `tryBuildMilestone` from `public` to `private`
    - remove dead methods: `didParticipantPostOnChainLocal`, `getDoubleSignedBlock`
    - move methods `getForkDisputeConfirmations`, `getForkDisputes`, `getReduceData` so that all private methods will be at the bottom of the file

4. Changes to harness
    - bug fix on `submitInvalidStateTransitionBlock`, to allow it to submit the very first block on afork as invalid.
    - additional queries to support unit test suite for `AgreementManager`

5. Single owner for the dispute window
    - `EventSyncService.loadSynchronizedWindowCommitments` - chain read + missing dispute event recovery
    - reduce and spectate both enter through it
    - spectate takes the disputed flag from the chain too, not the local EVM

*

PR #393, opened 2026-07-14
unit test suite for ValidationService

daiagi:
Unit test suite for `ValidationService`


lukastanisic99:
rebase on top of #397  or whatever is the latest after fixes

daiagi:
> rebase on top of #397 or whatever is the latest after fixes

the current order is

#394  -> #393 -> #389 -> #397

lukastanisic99:
> > rebase on top of #397 or whatever is the latest after fixes
> 
> the current order is
> 
> #394 -> #393 -> #389 -> #397

@daiagi are these 3 PRs ready for review, since I don't see you requesting a review?

lukastanisic99:
1 red, the rest orange and yellow 

## 🧱 Fundamental

- 🟠 **[FO1] The probe drops the `QueuedBlockEntry` source attribution that production validation receives.** [`runBlockValidation`](https://github.com/peer3to/state-channels-plus/blob/3c09432d4ef21ff6809fe5ef36cdef86b3c3e076/test/fixtures/customRpc/harnessControl/services/stub/StubRpcMethods.ts#L1118-L1137) always calls `createEntry(block)` without a sender, so `sourcePeers` and `signatureSources` are empty. Hooks such as [`blockForkIsDisputed`](https://github.com/peer3to/state-channels-plus/blob/3c09432d4ef21ff6809fe5ef36cdef86b3c3e076/src/stateManager/validationStrategy/BlockValidationStrategy.ts#L211-L239) therefore run with different inputs from the gossip pipeline.

  > **Fix FO1-FIX**
  >
  > Add an encoded sender/source option to the endpoint and pass it to `createEntry`. Add source-aware cases that assert acknowledged disputed-fork suppliers are disconnected while an entry with an honest source is restored.

- 🟠 **[FO2] `buildAndEncodeBlock` does not make the signer the block author.** The helper signs [`block(overrides)`](https://github.com/peer3to/state-channels-plus/blob/3c09432d4ef21ff6809fe5ef36cdef86b3c3e076/test/factory.ts#L311-L317), but `block()` generates a random `transaction.header.participant` unless every caller repeats it. A future test can silently create a signer/author mismatch while using a helper named “authored.”

  > **Fix FO2-FIX**
  >
  > Resolve `await signer.getAddress()` and use it as the default `header.participant` before building the block. Keep an explicit override only for tests that deliberately craft a mismatch, and add a factory test that the recovered signer and block author match by default.

- 🟢 **[SG1] This PR adds test-only code and does not widen a production trust boundary.**

  > **Fix SG1-FIX**
  >
  > No change required.

## 🏁 Race conditions

- 🟠 **[RO1] The probe replaces live shared methods while unrelated session work can still call them.** [`runBlockValidation`](https://github.com/peer3to/state-channels-plus/blob/3c09432d4ef21ff6809fe5ef36cdef86b3c3e076/test/fixtures/customRpc/harnessControl/services/stub/StubRpcMethods.ts#L1161-L1191) patches the live dispute manager, P2P manager, and block queue manager until validation finishes. The new [concurrency test](https://github.com/peer3to/state-channels-plus/blob/3c09432d4ef21ff6809fe5ef36cdef86b3c3e076/test/unit/ValidationService.test.ts#L754-L821) deliberately advances the session during repeated probes, so a real queue restore, disconnect, or dispute can hit the record-only replacement.

  > **Fix RO1-FIX**
  >
  > Put the record-only patch/restore block behind a `StubService` mutex and do not overlap it with live state advancement. Test validation/storage interleavings with controlled holds around the real call boundary instead of repeatedly replacing shared methods during the workload.

## ⚡ Performance

- 🟢 **[PG1] The added work is confined to the test harness; no production hot path or round-trip changes.**

  > **Fix PG1-FIX**
  >
  > No change required.

- 🔴 **[TR1] The canonical gate fails `invalidStateProofGenesisLinkage.test.ts`.** [`initiatedWait`](https://github.com/peer3to/state-channels-plus/blob/3c09432d4ef21ff6809fe5ef36cdef86b3c3e076/test/e2e/disputeValidation/invalidStateProofGenesisLinkage.test.ts#L18-L23) times out after 15 seconds waiting for peer 1’s `onInitiatingDispute` event; see the local [run-458 failure log](../../../logs/run-458/error_invalidStateProofGenesisLinkage.test__E2E_dispute_validation_DisputeInvalidStateProof_genesis_linkage__unlinked_genesisStateSnapshotData_against_a_valid_genesis_dispute_submitter_slashed_honest_disputer_survives.ansi). A focused rerun failed the same way, and the same focused case also fails on base SHA `93498f913c1024e555541dc6e9ad0eb93c288476`, so PR #393 did not introduce it.

  > **Fix TR1-FIX**
  >
  > Replace the hardcoded 15-second wait with the height-0 timeout derived from [`protocolEventTimeoutMs`](https://github.com/peer3to/state-channels-plus/blob/3c09432d4ef21ff6809fe5ef36cdef86b3c3e076/test/harness/core/testTimeConfig.ts#L38-L48), then rerun the focused case and the full gate. Acceptance is one dispute initiation from peer 1 and 622 / 622 passing.

- 🟠 **[TO1] Unit: the on-chain active-participant fallback is untested.** [`getParticipantsUnionOrOnChain`](https://github.com/peer3to/state-channels-plus/blob/3c09432d4ef21ff6809fe5ef36cdef86b3c3e076/src/stateManager/ValidationService.ts#L538-L561) queries the contract only when the local union is empty; every new participant test starts with populated local state.

  > **Fix TO1-FIX**
  >
  > Teleport the peer to an empty local participant union while the contract still returns an active participant, validate that participant’s block through `runBlockValidation`, and assert it reaches the next guard instead of `blockAuthorIsNotParticipant`.

- 🟠 **[TO2] Unit: the on-chain pending-participant fallback is untested.** Pending participants are merged separately at [`getPendingParticipants`](https://github.com/peer3to/state-channels-plus/blob/3c09432d4ef21ff6809fe5ef36cdef86b3c3e076/src/stateManager/ValidationService.ts#L551-L561), but no case proves a pending participant is accepted when local snapshots are absent.

  > **Fix TO2-FIX**
  >
  > Create the same empty-local-state setup with the author only in the contract’s pending set, then assert validation does not fire `blockAuthorIsNotParticipant`.

- 🟠 **[TO3] Unit: the closed-channel guard remains skipped.** The suite leaves [`channel not opened → channelNotOpened`](https://github.com/peer3to/state-channels-plus/blob/3c09432d4ef21ff6809fe5ef36cdef86b3c3e076/test/unit/ValidationService.test.ts#L35-L38) pending even though [`isChannelOpen`](https://github.com/peer3to/state-channels-plus/blob/3c09432d4ef21ff6809fe5ef36cdef86b3c3e076/src/stateManager/ValidationService.ts#L227-L229) is public and its false branch drives a distinct restore/`NOT_READY` result.

  > **Fix TO3-FIX**
  >
  > Use a pre-open harness state with `ZeroHash`; assert `isChannelOpen` is false, then validate a crafted entry and assert `channelNotOpened`, `NOT_READY`, and one restore call.

- 🟠 **[TO4] Unit: previous-block calldata recovery that makes the timestamp valid is missing.** The recursive recovery path at [`tryRecoverBlockCalldataAndScheduleValidation`](https://github.com/peer3to/state-channels-plus/blob/3c09432d4ef21ff6809fe5ef36cdef86b3c3e076/src/stateManager/ValidationService.ts#L418-L468) is not exercised; the pending case instead names a `NOT_READY` result that this method never returns.

  > **Fix TO4-FIX**
  >
  > Stage matching calldata with a later on-chain timestamp for the previous block, submit a block that is invalid against the original timestamp but valid after recovery, and assert the recursive pass returns `SUCCESS` and caches the recovered timestamp.

- 🟠 **[TO5] Unit: previous-block calldata recovery that still leaves the timestamp invalid is missing.** The existing [non-first invalid timestamp case](https://github.com/peer3to/state-channels-plus/blob/3c09432d4ef21ff6809fe5ef36cdef86b3c3e076/test/unit/ValidationService.test.ts#L440-L479) covers “nothing recovered,” not “recovered better data and still invalid.”

  > **Fix TO5-FIX**
  >
  > Recover a previous on-chain timestamp that improves the relevant timestamp but still does not make the candidate valid; assert one recursive recheck, `objectiveInvalidTimestampDetected`, and an `InvalidTimestamp` proof.

- 🟠 **[TO6] Unit: the current block’s `ON_TIME` on-chain path is missing.** [`OnChainPostTiming.ON_TIME`](https://github.com/peer3to/state-channels-plus/blob/3c09432d4ef21ff6809fe5ef36cdef86b3c3e076/src/stateManager/ValidationService.ts#L471-L505) returns before the subjective clock check, but no public-surface unit case proves that behavior.

  > **Fix TO6-FIX**
  >
  > Validate a block with a cached on-chain timestamp exactly at the allowed post deadline and assert `SUCCESS`, including when the local wall clock is outside `agreementTime`.

- 🟠 **[TO7] Unit: the current block’s `TOO_LATE` on-chain path is missing.** The suite keeps [“posted on-chain too late”](https://github.com/peer3to/state-channels-plus/blob/3c09432d4ef21ff6809fe5ef36cdef86b3c3e076/test/unit/ValidationService.test.ts#L489-L491) skipped, so the one-second-over boundary and its fraud-proof hook are uncovered.

  > **Fix TO7-FIX**
  >
  > Validate the same block with an on-chain timestamp one second past the deadline and assert `objectiveInvalidTimestampDetected`, `DISPUTE`, and `InvalidTimestamp`.

- 🟠 **[TO8] Unit: recovery of the current block’s calldata is missing.** When no timestamp is cached, [`getOnChainPostTiming`](https://github.com/peer3to/state-channels-plus/blob/3c09432d4ef21ff6809fe5ef36cdef86b3c3e076/src/stateManager/ValidationService.ts#L576-L593) asks `EventSyncService`, reloads storage, and persists the result; none of the new cases drives that branch.

  > **Fix TO8-FIX**
  >
  > Stage recoverable current-block calldata with no initial cached timestamp, validate it, and assert recovery was queried, the timestamp was persisted, and the final result matches the recovered `ON_TIME` value.

- 🟠 **[TO9] Unit: active validation outside `agreementTime` remains skipped.** The pending [subjective-invalid case](https://github.com/peer3to/state-channels-plus/blob/3c09432d4ef21ff6809fe5ef36cdef86b3c3e076/test/unit/ValidationService.test.ts#L485-L487) is a component branch, not an E2E permutation.

  > **Fix TO9-FIX**
  >
  > Set a deterministic host clock, validate an objectively valid unposted block just outside `agreementTime`, and assert `subjectiveInvalidTimestampDetected` returns `NOT_ENOUGH_TIME`.

- 🟠 **[TO10] Unit: dispute validation’s subjective-time bypass is untested.** The active-only check at [`strategy instanceof BlockValidationStrategy`](https://github.com/peer3to/state-channels-plus/blob/3c09432d4ef21ff6809fe5ef36cdef86b3c3e076/src/stateManager/ValidationService.ts#L507-L533) means the same old, unposted block must succeed under `DisputeValidationStrategy`.

  > **Fix TO10-FIX**
  >
  > Reuse TO9’s block with `strategy: "dispute"` and assert `SUCCESS` with no subjective-invalid hook.

- 🟠 **[TO11] Unit: dispute validation’s disputed-fork exemption is untested.** The guard is explicitly skipped for `DisputeValidationStrategy` at [`validateBlockConfirmation`](https://github.com/peer3to/state-channels-plus/blob/3c09432d4ef21ff6809fe5ef36cdef86b3c3e076/src/stateManager/ValidationService.ts#L106-L116); the current [disputed-fork case](https://github.com/peer3to/state-channels-plus/blob/3c09432d4ef21ff6809fe5ef36cdef86b3c3e076/test/unit/ValidationService.test.ts#L639-L690) covers only the active strategy.

  > **Fix TO11-FIX**
  >
  > Run a linked candidate on a disputed fork with the dispute strategy and assert `blockForkIsDisputed` is not called and validation continues.

- 🟠 **[TO12] Unit: dispute validation’s future-height exemption is untested.** The second strategy-specific bypass at [`block.height > expectedNextHeight`](https://github.com/peer3to/state-channels-plus/blob/3c09432d4ef21ff6809fe5ef36cdef86b3c3e076/src/stateManager/ValidationService.ts#L118-L136) has no dispute-strategy case.

  > **Fix TO12-FIX**
  >
  > Teleport a complete dispute state-proof chain ahead of active storage, validate its next linked block, and assert `blockIsNotNextAndIsInTheFuture` is not called.

- 🟠 **[TO13] Unit: missing previous snapshot under dispute validation remains skipped.** [`validateBlockConfirmation`](https://github.com/peer3to/state-channels-plus/blob/3c09432d4ef21ff6809fe5ef36cdef86b3c3e076/src/stateManager/ValidationService.ts#L158-L173) has a dedicated error, but the suite combines it into a pending two-error placeholder.

  > **Fix TO13-FIX**
  >
  > Remove only the previous snapshot from a linked dispute fixture and assert the exact `Missing previous snapshot for dispute validation strategy` rejection.

- 🟠 **[TO14] Unit: missing previous state under dispute validation remains skipped.** The separate lookup and error at [`getStateMachineState`](https://github.com/peer3to/state-channels-plus/blob/3c09432d4ef21ff6809fe5ef36cdef86b3c3e076/src/stateManager/ValidationService.ts#L174-L188) needs its own fixture and assertion.

  > **Fix TO14-FIX**
  >
  > Keep the snapshot but remove its referenced state, then assert the exact `Missing previous state machine state for dispute validation strategy` rejection.

- 🟠 **[TO15] Unit: the dispute strategy’s wrong-leader result after `setState` is missing.** The only new dispute-strategy case covers [the successful leader](https://github.com/peer3to/state-channels-plus/blob/3c09432d4ef21ff6809fe5ef36cdef86b3c3e076/test/unit/ValidationService.test.ts#L705-L751), so it cannot catch ordering regressions between [`setState` and `getNextToWrite`](https://github.com/peer3to/state-channels-plus/blob/3c09432d4ef21ff6809fe5ef36cdef86b3c3e076/src/stateManager/ValidationService.ts#L190-L204).

  > **Fix TO15-FIX**
  >
  > Supply a linked block whose author is not the leader in the restored previous state and assert `invalidStateTransitionDetected` fires only after that state is installed.

- 🟠 **[TO16] Unit: `isDisputedFork` is not isolated with only the local marker true.** The method ORs [`didIDispute` with the contract query](https://github.com/peer3to/state-channels-plus/blob/3c09432d4ef21ff6809fe5ef36cdef86b3c3e076/src/stateManager/ValidationService.ts#L294-L304); the current real-dispute case can have both true and does not prove the local short-circuit.

  > **Fix TO16-FIX**
  >
  > Record a local dispute while the contract still reports false, call `isDisputedFork`, and assert true without an on-chain query.

- 🟠 **[TO17] Unit: `isDisputedFork` is not isolated with only the contract state true.** A peer that observes another peer’s on-chain dispute must return true before its local `didIDispute` marker is set.

  > **Fix TO17-FIX**
  >
  > Use an observer with no local dispute marker after another peer posts the dispute, then assert `isDisputedFork` returns true from the contract branch.

- 🟡 **[TY18] Unit: the concurrency case cannot reach a competing validation outcome.** It validates height `startHeight + 100` while advancing only 10 blocks, and its own comment says [`NOT_READY` is the only possible verdict`](https://github.com/peer3to/state-channels-plus/blob/3c09432d4ef21ff6809fe5ef36cdef86b3c3e076/test/unit/ValidationService.test.ts#L754-L819). This samples load; it does not test the future-to-next boundary or an interleaving.

  > **Fix TY18-FIX**
  >
  > Hold validation immediately after reading `expectedNextHeight`, advance storage across that candidate height, release the hold, and assert the allowed final result plus no false dispute or disconnect.

- 🟡 **[TY19] Unit: the verbatim replay case locks in an unreachable false accusation.** The test expects an honest stored block to become a double-sign dispute at [`ValidationService`](https://github.com/peer3to/state-channels-plus/blob/3c09432d4ef21ff6809fe5ef36cdef86b3c3e076/test/unit/ValidationService.test.ts#L335-L359), while production routes stored blocks through the merge path before validation.

  > **Fix TY19-FIX**
  >
  > Remove this case from the ValidationService contract. Keep the production behavior pinned at the queue boundary by asserting a stored verbatim duplicate merges or returns duplicate without reaching `doubleSignDetected`.

- 🟡 **[TY20] Unit/E2E classification: the on-chain timing boundary is a private-method unit probe living in E2E.** [`E2E-FirstBlockTimestampGrace`](https://github.com/peer3to/state-channels-plus/blob/3c09432d4ef21ff6809fe5ef36cdef86b3c3e076/test/e2e/E2E-FirstBlockTimestampGrace.test.ts#L173-L250) calls private `getOnChainPostTiming` directly. It does not exercise a system interaction.

  > **Fix TY20-FIX**
  >
  > Move the `ON_TIME`/`TOO_LATE` boundary matrix into `ValidationService.test.ts` through `validateBlockConfirmation`. Keep the E2E file’s real authoring and timeout workflows.

- 🟡 **[TY21] Unit/E2E classification: two out-of-scope E2E concerns are counted as pending unit tests.** The skipped [forged-signature case](https://github.com/peer3to/state-channels-plus/blob/3c09432d4ef21ff6809fe5ef36cdef86b3c3e076/test/unit/ValidationService.test.ts#L494-L497) belongs to queue authentication, and the skipped [real dispute-evidence case](https://github.com/peer3to/state-channels-plus/blob/3c09432d4ef21ff6809fe5ef36cdef86b3c3e076/test/unit/ValidationService.test.ts#L696-L703) belongs to the dispute integration workflow.

  > **Fix TY21-FIX**
  >
  > Remove both pending unit cases. Reference the existing queue forged-signature E2E and dispute fraud-proof E2E coverage in a short suite-boundary comment instead.

## ♻️ Code reuse

- 🟠 **[CO1] Dispute-strategy construction now has a fourth owner.** [`runBlockValidation`](https://github.com/peer3to/state-channels-plus/blob/3c09432d4ef21ff6809fe5ef36cdef86b3c3e076/test/fixtures/customRpc/harnessControl/services/stub/StubRpcMethods.ts#L1144-L1155) constructs the same strategy already assembled by three [`StubService` probes](https://github.com/peer3to/state-channels-plus/blob/3c09432d4ef21ff6809fe5ef36cdef86b3c3e076/test/fixtures/customRpc/harnessControl/services/stub/StubService.ts#L282-L377). Changes to constructor inputs or dispute setup can drift between them.

  > **Fix CO1-FIX**
  >
  > Add one `StubService.createDisputeValidationStrategy` owner and route all four probes through it.

## 📏 AGENTS.md adherence

- 🟠 **[AO1] `StubRpcMethods` contains the full host-side operation instead of only the endpoint.** The repository rule puts helpers, accessors, orchestration, and shared state on the non-routable service, but [`runBlockValidation`](https://github.com/peer3to/state-channels-plus/blob/3c09432d4ef21ff6809fe5ef36cdef86b3c3e076/test/fixtures/customRpc/harnessControl/services/stub/StubRpcMethods.ts#L1118-L1247) performs decoding, strategy creation, patching, instrumentation, validation, and projection in the routable class.

  > **Fix AO1-FIX**
  >
  > Leave a thin endpoint that delegates to `this.service.runBlockValidation(...)`; move the complete host operation and its synchronization to `StubService`.

- 🟡 **[AY2] The test file defines a reusable conversion helper.** [`solProofType`](https://github.com/peer3to/state-channels-plus/blob/3c09432d4ef21ff6809fe5ef36cdef86b3c3e076/test/unit/ValidationService.test.ts#L11-L16) is used across the suite, while `test/AGENTS.md` requires reusable test support to live in the factory or harness.

  > **Fix AY2-FIX**
  >
  > Move the conversion into `test/factory.ts` or compare against the typed Solidity value directly at each assertion.

## 🧹 Miscellaneous (scope)

- 🟢 **[MG1] The PR stays within test factories, harness control, and the new unit suite.**

  > **Fix MG1-FIX**
  >
  > No change required.

## ❓ Open questions

- 🟢 **[OG1] No product decision is needed; the remaining work is test completeness and harness structure.**

  > **Fix OG1-FIX**
  >
  > No change required.

daiagi:
F02: It's a bit of a stretch to call an detail of a `factory` method FUNDAMENTAL.
still, I've added the default signer it asked for, but not the test, I don't think we should go down the road of testing the test code.


other then that, ready for a re-review @lukastanisic99

lukastanisic99:
What's this polling in the last commit?

daiagi:
>What's this polling in the last commit?

trash, removed

lukastanisic99 (approved):

PR #394, opened 2026-07-16
unit test suite for DisputeManager

daiagi:


momoshell (commented):
What do you think about tightening up one flaky test before this lands? The suite itself is strong — I especially like using `verifyStateProof.staticCall` as the oracle for `constructDispute`, the behind-peer `isPartial` test with a synced control, and the concurrency sampler asserting slashes stay consistent with bundled proofs.

How I checked it: typecheck is clean on the head, and since the branch has no CI I ran the suite locally — 12 of 13 active tests pass; the one failure is the `killDispute` re-kill test, which failed differently on each run (details inline), so it looks flaky rather than wrong. `yarn format:check` also fails on this file (inline note).

One non-line note: this is stacked (`unit-test-validation-service` ← this ← `refactor-StateManager`), so merge order matters for the chain.

test/unit/DisputeManager.test.ts
```diff
@@ -0,0 +1,630 @@
+import { expect } from "chai";
+import { Codec, Type } from "@/utils";
+import { MathTestSession as TestSession } from "@test/harness";
+import * as factory from "@test/factory";
+import type { Address } from "@/types/types";
+
+
+describe("Unit: DisputeManager", function () {
```
momoshell, line 6: `yarn format:check` fails on this file — two doubled blank lines (after the imports and after the `getAuditingData` skip) plus two wraps Prettier wants differently. A quick `yarn format` before merge should do it, since husky won't have caught a file added this way.

test/unit/DisputeManager.test.ts
```diff
@@ -0,0 +1,630 @@
...
+                dispute.input.selfRemoval = false;
+            });
+            await h.event.waitForAllPeers("onDisputeKilled", 1, {
+                mode: "atLeast"
+            });
+
+            // re-kill from the stored proof: the spammer is already slashed, so
+            // the on-chain apply is idempotent -> killDispute completes with no
+            // throw and no second slash
+            const r = await h.execOnHost(h.getPeer(0), async (sm) => {
```
momoshell: This test was the only failure in my local runs, and it failed a different way each time:

1. In the full-suite run, this `execOnHost` block timed out — `RPC request 'scenario.exec' timed out after 2000ms`. The block does two on-chain reads plus the re-kill's on-chain apply, which doesn't fit the default 2s control-RPC budget under load. The concurrency test passes `{ timeoutMs: 20000 }` for exactly this reason — would you add the same here?
2. In isolation the test body passed, but the *after-each hook* failed on a detached transaction rejecting with `NONCE_EXPIRED` ("nonce has already been used") — the re-kill's multicall racing the original kill's detached tx. So even with the timeout fixed, the idempotence oracle may need the detached promise drained or its race errors handled (same `tryHandleEvmError` treatment the production paths get) before teardown.

Happy to share the logs if useful.

test/unit/DisputeManager.test.ts
```diff
@@ -0,0 +1,630 @@
...
+                    return { offenderSlashed: slashes.includes(args.offender) };
+                },
+                { offender: offender.address.toLowerCase() }
+            );
+
+            expect(r.offenderSlashed).to.equal(true);
+        });
+
+        // skipped: flaky
+        it.skip("fault-free writer-timeout on a pending-join fork → dispute posts with calldata", async function () {
```
momoshell: This skip documents a real product bug, not just test flakiness — a losing peer's `reduceAndFinalize` reverting `ErrorDisputeInboundMessageBlocksInvalid` isn't handled on the reduction path and ends as an unhandled detached rejection (~10% of runs, per the comment). What do you think about filing an issue for the unhandled-revert and linking it here, so the skip doesn't quietly rot? It's the same fire-and-forget error-handling theme as the partial-window race noted in #395.

daiagi: https://trello.com/c/MUwszX7B

test/unit/DisputeManager.test.ts
daiagi: fixed

lukastanisic99:
rebase on top of #397 or whatever is the latest after fixes

daiagi:
>rebase on top of https://github.com/peer3to/state-channels-plus/pull/397 or whatever is the latest after fixes

394 -> 393 -> 389 -> 397(yours)

lukastanisic99:
## 🏁 Race conditions

- 🔴 **[RR1] Test-harness — the re-kill test drains and discards every host failure.** [`await h.quiesceHosts()`](https://github.com/peer3to/state-channels-plus/blob/d56ea842207934fcdf9dc8f9a825929f647190b5/test/unit/DisputeManager.test.ts#L535-L545) ignores the returned `Error[]`, then calls optional `expectFirstDetachedError`, which [adds `NONCE_EXPIRED` to the global allowlist even when no error was observed](https://github.com/peer3to/state-channels-plus/blob/d56ea842207934fcdf9dc8f9a825929f647190b5/test/harness/session/TestSession.ts#L90-L110). The normal teardown instead [feeds every drained host error into the session failure channel](https://github.com/peer3to/state-channels-plus/blob/d56ea842207934fcdf9dc8f9a825929f647190b5/test/harness/session/registerTestSessionHooks.ts#L60-L76). This test can therefore pass while hiding a nonce failure or any unrelated detached rejection.

  > **Fix RR1-FIX**
  >
  > Quiesce before the manual re-kill and assert the returned errors are empty. Run the re-kill, quiesce again, and assert that result is empty too. Remove the optional detached-error allowance; if a nonce race is deliberately tested, inspect every returned error and reject anything except the exact expected failure.

- 🟠 **[RO2] Unit — the construction “concurrency” case never forces an interleaving.** Its [40/80 ms sleeps and before/after samples](https://github.com/peer3to/state-channels-plus/blob/d56ea842207934fcdf9dc8f9a825929f647190b5/test/unit/DisputeManager.test.ts#L630-L655) guarantee `min=0` and `max=all` even if no `constructDispute` call overlaps a proof insertion.

  > **Fix RO2-FIX**
  >
  > Add a harness hold at a real `constructDispute` async boundary, pause one call, insert a real fraud proof, release it, and assert the returned slash/proof snapshot. Remove both sleeps and the sampling loop.

- 🟠 **[RO3] Unit — the `dispute()` mutex is not tested under contention.** The only deduplication case [pre-marks the fork before calling `dispute`](https://github.com/peer3to/state-channels-plus/blob/d56ea842207934fcdf9dc8f9a825929f647190b5/test/unit/DisputeManager.test.ts#L306-L341); it does not verify that [two concurrent callers serialize and only one submits](https://github.com/peer3to/state-channels-plus/blob/d56ea842207934fcdf9dc8f9a825929f647190b5/src/disputeManager/DisputeManager.ts#L87-L103).

  > **Fix RO3-FIX**
  >
  > Hold the first upload, start a second `dispute(forkId)`, release the first call, and assert exactly one upload and one initiation hook. The waiter must acquire the mutex, observe `didIDispute=true`, and return.

- 🟠 **[RO4] Unit — duplicate `killDispute()` calls are not tested deterministically.** Production intentionally has [no mutex and relies on preflight plus EVM race handling](https://github.com/peer3to/state-channels-plus/blob/d56ea842207934fcdf9dc8f9a825929f647190b5/src/disputeManager/DisputeManager.ts#L213-L279), while the current re-kill case cannot distinguish the inactive-window return from a handled on-chain race.

  > **Fix RO4-FIX**
  >
  > Hold two calls while the kill window is live, release them together, and assert one slash, no detached errors, and a recorded handled outcome for the losing call.

## ⚡ Performance

- 🟢 **[PG1] The PR adds no production work.** Test cost is substantial but bounded by the existing harness and parallel runner.

  > **Fix PG1-FIX**
  >
  > No change required.

- 🟠 **[TO1] Unit — the four `dispute()` submission branches are not covered as component variations.** Production selects fraud-proof/no-proof × calldata/no-calldata across [`multicall`, `uploadDisputeWithCalldata`, and `uploadDispute`](https://github.com/peer3to/state-channels-plus/blob/d56ea842207934fcdf9dc8f9a825929f647190b5/src/disputeManager/DisputeManager.ts#L113-L160). The two live cases both stage a fraud proof, the no-proof/calldata case is skipped, and no-proof/no-calldata is absent.

  > **Fix TO1-FIX**
  >
  > Add four direct black-box unit cases with record-only contract probes. For each, assert the chosen method, exact arguments, transaction `wait()`, disputed-fork marker, and initiation hook.

- 🟠 **[TO2] Unit — `ErrorCantParticipateInDispute` is untested.** [`dispute()` consumes the custom error and clears the local marker](https://github.com/peer3to/state-channels-plus/blob/d56ea842207934fcdf9dc8f9a825929f647190b5/src/disputeManager/DisputeManager.ts#L168-L180).

  > **Fix TO2-FIX**
  >
  > Inject the real custom revert through the public call, assert `dispute()` resolves, and assert `didIDispute(forkId)` is false.

- 🟠 **[TO3] Unit — `RaceConditionDisputeTimeoutWindowCreatedTooEarly` is untested through `dispute()`.** Its [handler is a consumed no-op](https://github.com/peer3to/state-channels-plus/blob/d56ea842207934fcdf9dc8f9a825929f647190b5/src/disputeManager/DisputeManager.ts#L181-L186), which differs from the evidence-expired policy.

  > **Fix TO3-FIX**
  >
  > Cause that real custom revert, assert no rejection, and assert the disputed-fork marker is reset.

- 🟠 **[TO4] Unit — `RaceConditionDisputeEvidencePeriodExpired` is untested.** This handler [rethrows instead of consuming the error](https://github.com/peer3to/state-channels-plus/blob/d56ea842207934fcdf9dc8f9a825929f647190b5/src/disputeManager/DisputeManager.ts#L187-L195).

  > **Fix TO4-FIX**
  >
  > Cover a send-time revert and a `tx.wait()` revert. Assert rejection in both cases and pin the marker state on each side of the line that stores `true`.

- 🟠 **[TO5] Unit — the unknown `dispute()` failure policy is untested.** Unknown failures are [logged, swallowed, and followed by `storeDisputedFork(false)`](https://github.com/peer3to/state-channels-plus/blob/d56ea842207934fcdf9dc8f9a825929f647190b5/src/disputeManager/DisputeManager.ts#L198-L208).

  > **Fix TO5-FIX**
  >
  > Add unknown send/population and failed-`wait()` cases. Both must resolve without leaving the fork marked disputed.

- 🟠 **[TO6] Unit — `killDispute()` has no isolated active-window success case.** The current [re-kill workflow only checks unchanged slash count and no outward throw](https://github.com/peer3to/state-channels-plus/blob/d56ea842207934fcdf9dc8f9a825929f647190b5/test/unit/DisputeManager.test.ts#L469-L528), so it can pass without reaching `applyDisputeFraudProofs`.

  > **Fix TO6-FIX**
  >
  > Store one real dispute fraud proof, keep the window active, call `killDispute`, and assert that exact proof is submitted and the returned transaction is awaited.

- 🟠 **[TO7] Unit — both `killDispute()` precheck no-ops are missing.** [`windowExists=false` and `isExpired=true` return before submission](https://github.com/peer3to/state-channels-plus/blob/d56ea842207934fcdf9dc8f9a825929f647190b5/src/disputeManager/DisputeManager.ts#L226-L237), but neither boundary is asserted.

  > **Fix TO7-FIX**
  >
  > Add separate cases for a missing window and an expired existing window; both must make zero proof-submission calls.

- 🟠 **[TO8] Unit — the skipped missing-proof case states the wrong public behavior.** Its title says “throws,” but the [missing-proof throw is inside `killDispute()`’s catch-all](https://github.com/peer3to/state-channels-plus/blob/d56ea842207934fcdf9dc8f9a825929f647190b5/src/disputeManager/DisputeManager.ts#L213-L225) and is swallowed by the method.

  > **Fix TO8-FIX**
  >
  > Enable the case and assert the current contract: no proof submission and no outward rejection. If rejection is intended instead, move the invariant check outside the catch and update the test accordingly.

- 🟠 **[TO9] Unit — all four handled `killDispute()` races are missing.** No case distinguishes `RaceConditionDisputeKillPeriodExpired`, `RaceConditionOnChainSlashes`, `RaceConditionGenesisTimestampNotAvailable`, or `RaceConditionUnexpectedBlockCalldataPosted` in the [handler table](https://github.com/peer3to/state-channels-plus/blob/d56ea842207934fcdf9dc8f9a825929f647190b5/src/disputeManager/DisputeManager.ts#L253-L277).

  > **Fix TO9-FIX**
  >
  > Add one deterministic real-custom-revert case per handler and assert each is consumed without an unhandled or detached error.

- 🟠 **[TO10] Unit — unknown `killDispute()` submission failures are untested.** The method [logs and swallows errors outside its recognized race set](https://github.com/peer3to/state-channels-plus/blob/d56ea842207934fcdf9dc8f9a825929f647190b5/src/disputeManager/DisputeManager.ts#L280-L288).

  > **Fix TO10-FIX**
  >
  > Inject an unknown apply/wait failure and assert the public method resolves, no second submission is attempted, and teardown sees no detached rejection.

- 🟠 **[TO11] Unit — `constructDispute()` does not cover the zero-block/genesis boundary.** Before the first block, [`latestBlockHeight` becomes `-1`](https://github.com/peer3to/state-channels-plus/blob/d56ea842207934fcdf9dc8f9a825929f647190b5/src/disputeManager/DisputeManager.ts#L292-L309), which is relevant to a first-writer timeout.

  > **Fix TO11-FIX**
  >
  > Construct immediately after channel setup and assert a valid genesis-based dispute, empty proof, and correct timeout/input fields.

- 🟠 **[TO12] Unit — partial auditing data is not tested through `constructDispute()`.** The behind-peer case only calls `getAuditingData`; it never exercises [`constructDispute()`’s explicit partial-data rejection](https://github.com/peer3to/state-channels-plus/blob/d56ea842207934fcdf9dc8f9a825929f647190b5/src/disputeManager/DisputeManager.ts#L389-L395).

  > **Fix TO12-FIX**
  >
  > Invoke `constructDispute()` on a peer missing data referenced by its proof and assert `createDispute - isPartial auditingData`.

- 🟠 **[TO13] E2E — the pending-join writer-timeout workflow has no active regression.** The only case is [skipped because a losing reduction participant can throw `ErrorDisputeInboundMessageBlocksInvalid`](https://github.com/peer3to/state-channels-plus/blob/d56ea842207934fcdf9dc8f9a825929f647190b5/test/unit/DisputeManager.test.ts#L416-L460). The linked bug is downstream in reduction; it should not erase coverage of the DisputeManager upload branch.

  > **Fix TO13-FIX**
  >
  > Keep no-proof/calldata selection in a direct unit case. Add an E2E regression under `test/e2e` for timeout detection → calldata-backed commit → concurrent reduction, and enable it with zero detached errors once the reduction race is handled.

- 🟡 **[TY14] Unit — the three `constructDispute()` invalid/missing-state guards are not isolated.** There are no cases for [missing latest snapshot, missing state-machine state, or snapshot/state hash mismatch](https://github.com/peer3to/state-channels-plus/blob/d56ea842207934fcdf9dc8f9a825929f647190b5/src/disputeManager/DisputeManager.ts#L338-L363).

  > **Fix TY14-FIX**
  >
  > Add one black-box rejection case for each guard and assert its exact error.

- 🟡 **[TY15] Unit — slash filtering covers only one unslashed participant with one proof.** The implementation [intersects on-chain slashes with participants, then reads proofs only for the remainder](https://github.com/peer3to/state-channels-plus/blob/d56ea842207934fcdf9dc8f9a825929f647190b5/src/disputeManager/DisputeManager.ts#L365-L382).

  > **Fix TY15-FIX**
  >
  > Add separate cases for an out-of-set on-chain slash being removed, an already-slashed participant’s stored proof not being bundled, and a stored proof for a nonparticipant being ignored.

- 🟡 **[TY16] Unit — timeout tests assert only `participant`.** Both the empty and planted cases ignore the other fields [defaulted or copied into the dispute](https://github.com/peer3to/state-channels-plus/blob/d56ea842207934fcdf9dc8f9a825929f647190b5/src/disputeManager/DisputeManager.ts#L384-L387).

  > **Fix TY16-FIX**
  >
  > Assert all seven timeout fields for both absent and populated timeout states.

- 🟡 **[TY17] Unit — `selfRemoval` and inbound-head construction are untested at component level.** [`constructDispute()` copies force-exit state and the latest inbound hash/height, with zero defaults](https://github.com/peer3to/state-channels-plus/blob/d56ea842207934fcdf9dc8f9a825929f647190b5/src/disputeManager/DisputeManager.ts#L404-L421).

  > **Fix TY17-FIX**
  >
  > Add `selfRemoval=true`, empty inbound chain, and nonempty inbound chain cases; assert the exact resulting fields.

- 🟡 **[TY18] Unit — the generated dispute confirmation is never validated.** Production [signs the dispute, encodes it, and initializes peer signatures to `[]`](https://github.com/peer3to/state-channels-plus/blob/d56ea842207934fcdf9dc8f9a825929f647190b5/src/disputeManager/DisputeManager.ts#L471-L481).

  > **Fix TY18-FIX**
  >
  > Decode `encodedDispute`, compare it with the returned dispute, recover the signer, and assert `signatures=[]`.

- 🟡 **[TY19] Unit — output-snapshot computation failure is untested.** [`constructDispute()` logs and rethrows a failed `computeDisputeOutputSnapshotData.staticCall`](https://github.com/peer3to/state-channels-plus/blob/d56ea842207934fcdf9dc8f9a825929f647190b5/src/disputeManager/DisputeManager.ts#L423-L444).

  > **Fix TY19-FIX**
  >
  > Drive a real contract rejection at that boundary and assert the original error propagates.

- 🟡 **[TY20] Unit — the three independent partial-data causes are collapsed into one aggregate assertion.** `getAuditingData()` separately handles [missing milestone snapshots, missing latest snapshot, and missing finalized machine state](https://github.com/peer3to/state-channels-plus/blob/d56ea842207934fcdf9dc8f9a825929f647190b5/src/disputeManager/DisputeManager.ts#L512-L551).

  > **Fix TY20-FIX**
  >
  > Add one case per cause. Assert the genesis milestone placeholder, genesis latest-snapshot placeholder, and empty finalized state respectively, with `isPartial=true` in each case.

- 🟡 **[TY21] Unit — the successful milestone test checks only array length.** [`milestoneSnapshots` must preserve the proof’s milestone identity and order](https://github.com/peer3to/state-channels-plus/blob/d56ea842207934fcdf9dc8f9a825929f647190b5/src/disputeManager/DisputeManager.ts#L512-L520); wrong or reordered snapshots would still pass the current assertion.

  > **Fix TY21-FIX**
  >
  > Compare every returned snapshot with the snapshot referenced by the corresponding milestone.

- 🟡 **[TY22] Unit — no nonempty message-chain reconstruction is covered.** The method builds [bounded inbound and outbound ranges](https://github.com/peer3to/state-channels-plus/blob/d56ea842207934fcdf9dc8f9a825929f647190b5/src/disputeManager/DisputeManager.ts#L553-L584), including an optional dispute-supplied inbound upper hash.

  > **Fix TY22-FIX**
  >
  > Add inbound default-upper, explicit dispute-upper, and outbound genesis-to-latest cases; assert exact blocks and ordering.

- 🟡 **[TY23] Unit — unknown-fork/missing-genesis behavior remains skipped.** The public method has a concrete [missing-genesis error](https://github.com/peer3to/state-channels-plus/blob/d56ea842207934fcdf9dc8f9a825929f647190b5/src/disputeManager/DisputeManager.ts#L504-L510), while the suite dismisses it as defensive.

  > **Fix TY23-FIX**
  >
  > Enable the unknown-fork case and assert `getDisputeAuditingData - genesisStateSnapshot not found`.

## ♻️ Code reuse

- 🟡 **[CY1] Timeout staging duplicates an existing harness action.** The test calls the control RPC directly at [`DisputeManager.test.ts:156-163`](https://github.com/peer3to/state-channels-plus/blob/d56ea842207934fcdf9dc8f9a825929f647190b5/test/unit/DisputeManager.test.ts#L150-L163), while [`plantFreshTimeoutForNextWriter`](https://github.com/peer3to/state-channels-plus/blob/d56ea842207934fcdf9dc8f9a825929f647190b5/test/harness/actions/DisputeTamperingActions.ts#L407-L419) already owns that workflow.

  > **Fix CY1-FIX**
  >
  > Replace the inline query/RPC sequence with `h.tamper.plantFreshTimeoutForNextWriter(peer.index)`.

- 🟡 **[CY2] Invalid-transition proof staging is repeated inside the test file.** The same leader lookup, block construction, and `runBlockValidation` sequence appears in [the single-proof case](https://github.com/peer3to/state-channels-plus/blob/d56ea842207934fcdf9dc8f9a825929f647190b5/test/unit/DisputeManager.test.ts#L64-L87) and [the concurrency setup](https://github.com/peer3to/state-channels-plus/blob/d56ea842207934fcdf9dc8f9a825929f647190b5/test/unit/DisputeManager.test.ts#L557-L584).

  > **Fix CY2-FIX**
  >
  > Extract a discoverable harness action that stores a real invalid-transition proof for a chosen observer and offender, then reuse it in both tests.

## 📏 AGENTS.md adherence

- 🟠 **[AO1] Unit/E2E classification — three full system workflows are filed as unit tests.** The settled and pending-join cases [inject peer faults, wait for chain events, resolve a fork, and inspect on-chain slashes](https://github.com/peer3to/state-channels-plus/blob/d56ea842207934fcdf9dc8f9a825929f647190b5/test/unit/DisputeManager.test.ts#L343-L414); the re-kill case [posts a tampered dispute and waits for peer-wide kill events](https://github.com/peer3to/state-channels-plus/blob/d56ea842207934fcdf9dc8f9a825929f647190b5/test/unit/DisputeManager.test.ts#L469-L546). Those are E2E interactions, not component-level internal variations.

  > **Fix AO1-FIX**
  >
  > Move or merge the system workflows under `test/e2e`. Keep direct public-surface unit cases for branch permutations. Reuse the existing invalid-transition dispute/reduction and spam-dispute-kill E2Es instead of duplicating them.

- 🟡 **[AY2] The test adds unnecessary address casts.** [`offender.address as Address`](https://github.com/peer3to/state-channels-plus/blob/d56ea842207934fcdf9dc8f9a825929f647190b5/test/unit/DisputeManager.test.ts#L75-L82) and its concurrency copy restate a value already accepted by the factory input.

  > **Fix AY2-FIX**
  >
  > Pass `offender.address` directly and remove the unused `Address` type import.

## 🧹 Miscellaneous (scope)

- 🟠 **[MO1] The stack predates the final merged #397 fixes.** PR #394’s base descends from `3b7502e5`, while merged [PR #397](https://github.com/peer3to/state-channels-plus/pull/397) continued through `eb7c744b` with eight additional commits. The existing PR discussion already asks for a rebase onto the latest fixes.

  > **Fix MO1-FIX**
  >
  > Rebase the lowest open stack branch onto current `master`, then rebase #393 and #394 in order, update their PR bases, and rerun the focused and canonical gates.

- 🟡 **[MY2] The final-by-everyone case duplicates the healthy-fork test.** Both start the same three-peer/three-block state and assert [`postedAuditingData === false`](https://github.com/peer3to/state-channels-plus/blob/d56ea842207934fcdf9dc8f9a825929f647190b5/test/unit/DisputeManager.test.ts#L9-L55).

  > **Fix MY2-FIX**
  >
  > Delete the duplicate case or change it to a distinct finality boundary.

## ❓ Open questions

- 🟢 **[OG1] No unresolved product decision is needed for this review.** The required next steps are test corrections, coverage additions, and stack rebasing.

  > **Fix OG1-FIX**
  >
  > No change required.

test/unit/DisputeManager.test.ts
```diff
@@ -0,0 +1,1212 @@
...
+        });
+
+        // no test: `createDispute - isPartial auditingData` is unreachable from
+        // constructDispute. a peer's own proof only spans data it stored - a
+        // missing milestone snapshot already throws inside getStateProof
+        // ("Milestone built but corresponding snapshot not found") and the head
+        // snapshot is required by the "missing state snapshot" guard above it.
+        // getAuditingData only goes partial for a proof handed in by someone
+        // else, which is the audit path covered under `getAuditingData`.
+        it.skip("own proof missing referenced data → isPartial auditingData (unreachable)", function () {});
```
lukastanisic99, line 140: **Human note**

Commenting more to test my extension. The code does make this reachable by not having the `latestFinalizedStateStateMachineState` in storage `DisputeManager:544` so constructDispute would error `DisputeManager:394`

**AI-generated review**

🟠 **[TO12] The skipped test’s “unreachable” claim is false.** `getAuditingData()` independently marks the result partial when the finalized snapshot’s state-machine value is missing; that does not require a missing milestone or head snapshot. `constructDispute()` then reaches its explicit `createDispute - isPartial auditingData` rejection. The current suite never covers that public behavior.

> **Fix TO12-FIX**
>
> Build a real finalized session, remove only the finalized state-machine value through a typed harness action, call `constructDispute()`, and assert the exact partial-auditing-data error. Remove the skip and the unreachable explanation.

lukastanisic99 (commented):
**Human note**

Yes this is true or at leas I can't find it

**AI-generated review**

🟡 **[TY14] The three construction sanity guards still have no isolated coverage.** There is no case for a missing latest snapshot, missing latest state-machine state, or a snapshot/state hash mismatch in [`constructDispute()`](https://github.com/peer3to/state-channels-plus/blob/2d1bf3ec0c2c5b301c1763153f6303fab2345afa/src/disputeManager/DisputeManager.ts#L338-L363).

> **Fix TY14-FIX**
>
> Add one black-box rejection case per guard and assert each exact error.



---

**Human note**

In theory true

**AI-generated review**

🟡 **[TY15] Slash filtering covers only the ordinary unslashed-participant case.** The suite does not prove that an out-of-participant on-chain slash is removed, that a stored proof for an already-on-chain-slashed participant is not bundled, or that a stored proof for a nonparticipant is ignored by the filtering loop at [`DisputeManager.ts:365-382`](https://github.com/peer3to/state-channels-plus/blob/2d1bf3ec0c2c5b301c1763153f6303fab2345afa/src/disputeManager/DisputeManager.ts#L365-L382).

> **Fix TY15-FIX**
>
> Add one case for each set-membership boundary and assert both `onChainSlashes` and `fraudProofsToApply`.

lukastanisic99 (approved):

PR #395, opened 2026-07-16
Refactor-StateManager

daiagi:
#### Refactor StateManager

Break 2 new classes our from StateManager;

1. ForkReductionService
2. SnapshotPostingService

and write unit test suites for them


test/unit/ForkReductionService.test.ts
```diff
@@ -0,0 +1,334 @@
+import { expect } from "chai";
+import { MathTestSession as TestSession } from "@test/harness";
+import { hash as randomHash } from "../factory";
+import { waitFor } from "@test/utils/waitFor";
+
+describe("Unit: ForkReductionService", function () {
```
momoshell, line 6: Running this suite locally I got 7/8, with the failing test *changing* between runs.
Always the same signature though: `Nonce too high… can't be queued when automining`, thrown during harness contract deployment, i.e. before the code under test runs. 

So it's a deploy nonce race in the test infra, not this PR, but each of these tests boots a full multi-peer session, which gives the race a lot of chances to fire. Maybe worth a follow-up issue so future red runs don't get blamed on these suites?

src/stateManager/StateManager.ts
```diff
@@ -1861,537 +1465,18 @@ class StateManager<
...
-            this.logger.error(
-                "Error preparing update snapshot for the same fork",
-                {
-                    error:
-                        error instanceof Error ? error.message : String(error)
-                }
-            );
-            throw error;
-        }
+    public prepareUpdateSnapshotSameFork(forkId: ForkId) {
```
momoshell, line 1472: This delegator drops the optional `baseSnapshot` param.

The chaining onto the fork update's expected snapshot now only happens inside the service. Nothing in `src/` or `test/` passes it externally, and hiding that footgun is arguably an improvement, but since `StateManager` is exported from the library. 
Is the narrowing intentional? If yes, all good.

src/stateManager/ForkReductionService.ts
```diff
@@ -0,0 +1,463 @@
...
+    // is terminal. Cleared on fork transition (see StateManager's `forkId` setter).
+    private readonly killPeriodExpiryCache: Map<
+        string,
+        { isExpired: boolean; killPeriodEnd: number }
+    > = new Map();
+
+    private readonly logger: Logger;
+
+    constructor(
+        private readonly stateManager: StateManager,
```
momoshell, line 41: Both new services hold the whole `StateManager` and reach through it (~10 members), and `buildForkSnapshotCalldata` went public solely for them. Totally fine as a first mechanical step, but `DisputeValidationService` snapshots its deps into private fields instead. Would you consider doing the same (or injecting the narrower deps) in a follow-up?

src/stateManager/ForkReductionService.ts
```diff
@@ -0,0 +1,463 @@
...
+        logger: Logger
+    ) {
+        this.logger = logger.child({ component: "ForkReduction" });
```
momoshell, line 44: Heads-up: the child loggers rename components to `ForkReduction`/`SnapshotPosting`, so anyone filtering logs on the old `StateManager` component will see these flows move. Probably desirable, just flagging.

test/unit/ForkReductionService.test.ts
```diff
@@ -0,0 +1,334 @@
...
+        it.skip("no local disputes past kill period → initiates a local dispute, undefined", function () {});
+
+        // RACE, known red - skipped until fixed.
+        // a dispute commitment lands on-chain before onDisputeCommitted stores
+        // the struct. a reduction firing in that gap sees the commitment but no
+        // struct -> getForkDisputes throws "Missing Dispute in storage".
+        // reduceLocally doesn't catch it and tryReduce is fire-and-forget, so
+        // it ends as an unhandled rejection. wanted: return undefined and let
+        // the reschedule retry, like the other not-ready cases.
+        it.skip("partial dispute window (commitment on-chain, struct not yet stored) → reduceLocally discards, not throws", function () {});
```
momoshell: The ~90-line staged repro this replaces (from `AgreementManager.test.ts`) documented how to stage the race.

Commitment on-chain, struct dropped, `reduceLocally` by hand. Now the recipe only lives in git history. Maybe drop the commit hash or an issue link into this comment so it stays findable?

momoshell (commented):
What do you think about landing this mostly as-is? I went through it carefully and I'm happy with the extraction, inline notes below, none of them feel blocking to me.

How I checked it: I diffed each extracted method against the base branch with `this.stateManager.` normalized back to `this`. 
Every method comes out character-identical except Prettier re-wrapping, so no logic drift is hiding in the move. All external callers (`EventHandler`, `BlockQueueManager`, harness) go through the kept delegators. 

The suites themselves are genuinely nice.
Asserting promise identity for the single-flight case is the right oracle, reduce-to-genesis is a valuable edge case, and every skip documents why it can't be staged as a unit test.

src/stateManager/StateManager.ts
daiagi: intentional. baseSnapshot only feeds the internal chaining in postStateSnapshot, which now lives in the service and calls it directly -> nothing external passed it through StateManager. trivial to add back if something needs it.

test/unit/ForkReductionService.test.ts
daiagi: restored the test

src/stateManager/ForkReductionService.ts
daiagi: yes, this is part 1 of refactoring, there will be follow-ups

lukastanisic99:
I've already extracted some services/managers from stateManager in #397 feel free to continue

PR #396, opened 2026-07-16
fix(state-manager): handle dispute-storage race on reduction and snapshot-fork paths

sh3ll3x3c:
## Summary

A dispute commitment can land on-chain before our `onDisputeCommitted` handler
stores the dispute struct locally. Any code that reads the window in that gap
sees the on-chain commitment but no local struct, so `getForkDisputes` (and the
equivalent inline read) throws `"Missing Dispute in storage"`. This surfaced on
two paths:

- **Reduction path** (`reduceLocally` → `performLocalReduction` → `getForkDisputes`).
  `reduceLocally` did not catch it, and `tryReduce` is fire-and-forget, so the
  throw became an **unhandled rejection**.
- **Snapshot-fork path** (`prepareUpdateStateSnapshotFork`, reached via
  `postStateSnapshot` on the leaver exit path). It threw while building the
  window's disputes. The throw was caught and logged at the single caller, so no
  crash — but a transient race became a skipped, un-retried snapshot post.

### Changes

- `performLocalReduction` treats the missing struct as a transient "nothing to
  reduce yet" condition and **discards** (returns `undefined`); this fulfills the
  method's already-documented "no local dispute data → undefined" contract, and a
  later attempt converges once the struct lands.
- Removed the `tryReduce` catch that logged and re-threw — the source of the
  unhandled rejection.
- `prepareUpdateStateSnapshotFork` **stops traversal and waits for more data** on
  a missing struct, mirroring the existing no-commitments handling directly above
  it (partial reduction progress accumulated so far is preserved and posted).
- Introduced a typed `MissingDisputeInStorageError` so the discard decision is
  `instanceof`-checked instead of matched on an error-message string, and
  collapsed the duplicated literal onto one type.

## Test Plan

- New e2e suite `test/e2e/E2E-AgreementManager.test.ts` with two regressions that
  reproduce the race (no-op the sole struct writer on every peer, so the
  commitment lands on-chain but the struct never does), asserting:
  - `reduceLocally` discards rather than throwing.
  - `prepareUpdateStateSnapshotFork` stops traversal rather than throwing.
  - Both include genuineness guards (kill period expired, ≥1 on-chain commitment,
    struct absent) so the assertions aren't vacuous.
- Both tests pass; each was verified with a prove-it (reverting the fix makes the
  matching test fail with the exact `Missing Dispute in storage` error, then
  restored).
- `yarn tsc --noEmit -p tsconfig.json` and `-p tsconfig.browser.json`: clean.
- Regression: `E2E-DisputeManager` reduction tests pass (the one failing entry is
  a pre-existing WebSocket teardown flake in `MathPeerTestHarness.cleanup`,
  unrelated to these changes).

### DIFF
--- a/src/agreementManager/AgreementManager.ts
+++ b/src/agreementManager/AgreementManager.ts
@@ -24,6 +24,22 @@ import { ReduceData } from "@/types";
 import { LoggerUtils } from "@/utils/LoggerUtils";
 
 /**
+ * Thrown when a dispute commitment is present on-chain in the window but its
+ * struct has not been stored locally yet — the commitment landed before our
+ * onDisputeCommitted handler ran. Callers treat this as a transient "nothing to
+ * reduce yet" condition (discard and retry once the struct lands), so it is a
+ * typed error rather than a matched message string.
+ */
+export class MissingDisputeInStorageError extends Error {
+    constructor(public readonly commitment: string) {
+        super(
+            `Missing Dispute in storage for dispute commitment ${commitment}`
+        );
+        this.name = "MissingDisputeInStorageError";
+    }
+}
+
+/**
  * AgreementManager acts as a higher logic layer over storage
  * It interprets storage data and provides convenience methods
  */
@@ -527,9 +543,7 @@ class AgreementManager {
         for (const commitment of disputeCommitments) {
             const dispute = this.storage.disputes.getDispute(commitment);
             if (!dispute) {
-                throw new Error(
-                    `Missing Dispute in storage for dispute commitment ${commitment}`
-                );
+                throw new MissingDisputeInStorageError(commitment);
             }
 
             currentWindowDisputes.push(dispute);
--- a/src/stateManager/StateManager.ts
+++ b/src/stateManager/StateManager.ts
@@ -28,7 +28,9 @@ import {
 import { StateChannelManagerProxy } from "@typechain-types";
 
 // Core components
-import AgreementManager from "../agreementManager/AgreementManager";
+import AgreementManager, {
+    MissingDisputeInStorageError
+} from "../agreementManager/AgreementManager";
 import ADiamondStateMachine from "@/ADiamondStateMachine";
 import Clock from "@/Clock";
 import DisputeManager from "@/disputeManager";
@@ -509,21 +511,11 @@ class StateManager<
             );
         }
 
-        // Step 4: Perform reduction
-        try {
-            await this.performReduction(forkId);
-        } catch (error) {
-            if (
-                error instanceof Error &&
-                error.message.startsWith("Missing Dispute in storage")
-            ) {
-                this.logger.error(
-                    `Skipping reduction for fork ${forkId} because local dispute data is unavailable`,
-                    { error: error.message }
-                );
-            }
-            throw error;
-        }
+        // Step 4: Perform reduction. The "dispute struct not yet stored" race is
+        // handled at the source (reduceLocally discards), so performReduction
+        // returns without reducing rather than throwing an unhandled rejection
+        // out of this fire-and-forget path.
+        await this.performReduction(forkId);
     }
 
     /**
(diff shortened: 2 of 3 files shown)


src/agreementManager/AgreementManager.ts
```diff
@@ -23,6 +23,22 @@ import { StateChannelManagerProxy } from "@typechain-types/index";
...
 import { LoggerUtils } from "@/utils/LoggerUtils";
 
+/**
+ * Thrown when a dispute commitment is present on-chain in the window but its
+ * struct has not been stored locally yet — the commitment landed before our
+ * onDisputeCommitted handler ran. Callers treat this as a transient "nothing to
+ * reduce yet" condition (discard and retry once the struct lands), so it is a
+ * typed error rather than a matched message string.
+ */
+export class MissingDisputeInStorageError extends Error {
```
daiagi, line 33: not sure why we need to define a custom error class, and it will be first one in the code-base except for CustomEvmError

src/stateManager/StateManager.ts
```diff
@@ -568,11 +560,27 @@ class StateManager<
...
+                this.channelId,
+                forkId,
+                this.stateChannelManagerContract
+            );
+        } catch (error) {
+            // Race: a dispute commitment is already on-chain (in the window) but
+            // our onDisputeCommitted handler hasn't stored the struct yet.
+            // Nothing to reduce until it lands — discard; a later reduction
+            // attempt converges once the struct arrives.
+            if (error instanceof MissingDisputeInStorageError) {
```
daiagi, line 575: can use `if(!disputes)` instead of relying on the custom error class

test/e2e/E2E-AgreementManager.test.ts
```diff
@@ -0,0 +1,232 @@
...
+
+/**
+ * Drive four peers to a committed, kill-period-expired dispute in which no peer
+ * ever stored the dispute struct locally: no-op storeDisputeConfirmation (the
+ * only writer of the struct map) on every peer before the dispute, so the
+ * commitment lands on-chain but the struct never does — and, since no struct
+ * means no reduction, every peer stays on the disputed fork (deterministic).
+ * Returns an honest observer to exercise the window readers on.
+ */
+async function setupPartialDisputeWindow() {
```
daiagi, line 28: this test is quite a departure from our common test style

this function can be simplified:


```ts
const EVIDENCE_TIME = 6;
const MALICIOUS = 0;
const OBSERVER = 1;

/**
 * Drive four peers to a committed, kill-period-expired dispute in which no peer
 * ever stored the dispute struct locally. Returns an honest observer to exercise
 * the window readers on.
 */
async function setupPartialDisputeWindow() {

    // short evidence window so the on-chain kill period expires quickly
    await h.scenario.preDisputeSetup({
        peerCount: 4,
        timeConfig: { evidenceTime: EVIDENCE_TIME }
    });

    // reproduce the race on every peer: no-op the only writer of the dispute-
    // struct map. the on-chain commitment still lands, but no peer ever stores
    // the struct -> no peer reduces and there is no reduced fork to adopt
    // (keeps the scenario deterministic)
    await Promise.all(
        h.peers.map((peer) =>
            h.control(peer).stub.stubStoreDisputeConfirmation().request()
        )
    );

    // the malicious peer authors an invalid block -> the honest peers dispute it
    // and commit on-chain (their struct writes are no-ops)
    await h.byzantine.submitInvalidStateTransitionBlock(MALICIOUS);
    await h.assert.dispute.initiatedAndCommitedWait();

    // the kill period ends evidenceTime after the last evidence submission, and
    // nothing fires when it does -> wait it out. e2e mines on a 2s interval, so
    // block-time tracks wall-clock. once expired with commitments in the window,
    // further evidence reverts (DisputeManagerFacet) -> it stays expired.
    await sleep((EVIDENCE_TIME + 4) * 1000);

    return { h, observer: h.getPeer(OBSERVER) };
}
```


Note the use of 
- `stubStoreDisputeConfirmation` that needs to be created
- `initiatedAndCommitedWait` which waits for events to be registered, and DOES NOT poll
- asserts, rather then manually throw

test/e2e/E2E-AgreementManager.test.ts
```diff
@@ -0,0 +1,232 @@
...
+    }
+    if (!killExpired) {
+        throw new Error("kill period did not expire within timeout");
+    }
+
+    return { h, observer };
+}
+
+describe("E2E: Agreement Manager", function () {
+    it("partial dispute window (commitment on-chain, struct not yet stored) → reduceLocally discards, not throws", async function () {
```
daiagi, line 90: 1. probably  the `this.timeout(90000) is not needed, the AI keeps adding that.
2. we prefer succint readable tests, and push the imperative details to harness helpers
this is how this test could look like


```ts
    it("partial dispute window (commitment on-chain, struct not yet stored) → reduceLocally discards, not throws", async function () {
        this.timeout(90000);
        const { h, observer } = await setupPartialDisputeWindow();

        // the reproduction must be genuine, or the assertions below are vacuous
        const window = await h
            .control(observer)
            .query.getDisputeWindowStatus()
            .request();
        expect(window.commitmentCount, "on-chain dispute commitments").to.be.greaterThan(0);
        expect(window.anyStructPresent, "dispute struct present locally").to.equal(false);

        // reduceLocally reaches getForkDisputes, which finds a commitment with no
        // struct -> it must discard (return undefined), not throw
        const result = await h.execOnHost(observer, async (sm) => {
            try {
                const reduction = await sm.reduceLocally(sm.forkId);
                return { discarded: reduction === undefined, error: "" };
            } catch (error) {
                return {
                    discarded: false,
                    error: error instanceof Error ? error.message : String(error)
                };
            }
        });

        expect(result.error, "reduceLocally threw instead of discarding").to.equal("");
        expect(result.discarded, "reduceLocally discarded").to.equal(true);
    });

```


Same goes for rest of the tests in this file

daiagi (changes requested):

lukastanisic99:
@sh3ll3x3c can you see if my #397 PR resolved all of this? I think I handled all of these race conditions, re-syncing with missing events, etc..

sh3ll3x3c:
> @sh3ll3x3c can you see if my #397 PR resolved all of this? I think I handled all of these race conditions, re-syncing with missing events, etc..

It does, I'm closing this one.

PR #397, opened 2026-07-17
GREEN

lukastanisic99:
Hopefully no more flakes and everything green. 
If needed I'll write a proper PR description later, a lot of things were fixed, changed and things that were missing were implemented. I'm interested now to see, will tests be green on other machines too. 

1) Compile contracts:
`yarn compile`
2) Run parallel tests
`yarn test:parallel` 

You can run `yarn test:parallel -h` to see which flags you can provide

If a test fails it is NOT GOOD - we don't want flakes! 

All the failure test logs are persisted automatically under ./logs/run-N/* where N is auto incremented, so you don't have to worry or think about persisting the logs. I've updated my skills and pushed them to our SKILLS repository, so if a tests fails, you can use my $explain-log skill to fix it. It has a high hit rate, but like everything with AI, you can't trust it blindly.


contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol
```diff
@@ -514,7 +611,7 @@ contract DisputeFraudProofFacet is StateChannelCommon {
         (bool ok, uint256 maxValidTimestamp) =
             Math.tryAdd(previousTimestamp, getP2pTime() + getAgreementTime() + getChainFallbackTime());
```
daiagi: missing `firstBlockGrace`

src/stateManager/reduction/ReductionManager.ts
```diff
@@ -0,0 +1,296 @@
...
+                    if (this.stateManager.forkId !== forkId) {
+                        this.logger.debug(
+                            "Reduction completion skipped because the fork already changed",
+                            {
+                                forkId,
+                                currentForkId: this.stateManager.forkId,
+                                reducedForkId
+                            }
+                        );
+                        return false;
```
daiagi, line 204: this will get the promise to never settle and will lock the mutex
should resolve the promise instead of just return false

`failCompletion(completion, new Error("fork moved"))`

src/stateManager/validationStrategy/SpectatingValidationStrategy.ts
```diff
@@ -58,11 +60,11 @@ export default class SpectatingValidationStrategy extends AValidationStrategy {
     public async authenticateBlockFailed(
         _block: BlockConfirmationStruct
     ): Promise<BlockValidationResult> {
-        this.disconnect();
+        this.abort();
```
daiagi: seems like an easy DoS attack vector
any peer (not just channel participants) could create some junk block and cause a spectator to abort.

and what about `PENDING_PARTICIPANT` stauts?

src/stateManager/EventSyncService.ts
```diff
@@ -0,0 +1,530 @@
...
+            failed: false
+        };
+        state.pending += 1;
+        state.complete = false;
+        states.set(log.blockNumber, state);
+
+        const promise = this.dispatchLog(log, scheduledChannelId)
+            .catch((error) => {
+                state.failed = true;
+                this.logger.error("Contract event pipeline failed", {
```
daiagi, line 132: not sure it's reachable, but storing failed promises here is dangerous

on any transient failure the rejected promise sticks in the map -> every replay through queryAndProcessDisputeEvents refetches it and fails again

safer to drop failing promises from eventPromises / eventBlockNumbers in the .catch

contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol
lukastanisic99: yes, will add

src/stateManager/reduction/ReductionManager.ts
lukastanisic99: Will take a look at this one more clearly, it's not the mutex, but maybe the promise, but the design was:
1) we only initiate reduce if `stateManager.forkId == forkId` (we're currently at the correct fork) and the only way to change the forkId is to reduce and the first thing to reduce a forkId will resolve the promise, so if later during execution we observe that the fork has changed (different than when we started), we know we already reduced and that the promise is resolved. 

The only question now is, does the fork `really only change` on reduce, since if that's not true, the above assumptions break; 

There's some tunnel vision that the AI uses to skim through the code and deduct based on what it sees there, without understanding the higher level design -> what I'll probably do next is use AI to create a spec for everything and update my plan, implementation and review skills to update and keep track of the spec as we go and to point out any contradictions. Being verbose and pointing the AI in the right direction is all you need - having verbose logs is one of the best design decisions, a human would never go though them nor would it be possible considering the amount and AI does it nicely with a high hit rate since they are verbose and follow a pattern. We have to utilize that more.

src/stateManager/validationStrategy/SpectatingValidationStrategy.ts
lukastanisic99: Ye there is some merit here too - will consider in the morning

src/stateManager/reduction/ReductionManager.ts
daiagi: i take this comment back, it is unreachable "issue".

scripts/e2e-parallel/taskDiscovery.js
```diff
@@ -43,62 +43,37 @@ function collectDescribeTitlesFromIt(itCall) {
...
-                    }
-                });
+            const secondArg = args[1];
+            const isFunction =
+                secondArg.getKind() === SyntaxKind.ArrowFunction ||
+                secondArg.getKind() === SyntaxKind.FunctionExpression;
+            if (!isFunction) return;
+
+            const testName = getStringLiteralValue(args[0]);
+            if (!testName) return;
```
momoshell: `getStringLiteralValue` returns `null` for a template literal with substitutions (e.g. `` it(`rep ${rep}/${REPS}: ...`) ``), and line 68's `if (!testName) return;` drops that test from discovery entirely. It's never scheduled, never run, never counted. 

`E2E-Fuzz-Dispute-MVP.test.ts` has exactly one `it`, so the whole file silently contributes 0 tasks; the aggregate `tasks.length === 0` guard in `test-e2e-parallel.js` doesn't catch this since other files still produce plenty of tasks. 
Since this PR makes `yarn test:parallel` the canonical gate, this means a regression in the fuzz-dispute suite (or `E2E-RuntimeTransportModes`, `E2E-Spectate`'s height-pinning case, etc.) would report all-green. 

Would it be worth falling back to running the whole file when a title can't be statically resolved, or at least failing loudly when a discovered file yields 0 tasks?

src/evm/signer/HostNonceManager.ts
```diff
@@ -0,0 +1,137 @@
...
+        const transactionHash = keccak256(encodedTransaction);
+        let observedTransaction: TransactionResponse | null;
+        let pendingNonce: number;
+        try {
+            [observedTransaction, pendingNonce] = await Promise.all([
+                provider.getTransaction(transactionHash),
+                this.signer.getNonce("pending")
+            ]);
+        } catch (reconciliationError) {
+            this.nonceStateIndeterminate = true;
```
momoshell, line 118: `nonceStateIndeterminate` is set `true` here when the broadcast-failure reconciliation reads (`getTransaction`/`getNonce`) themselves throw, and it's never reset anywhere in the class.
Every subsequent `sendTransaction` call hard-asserts `!nonceStateIndeterminate` and throws. 

A transient blip on the reconciliation calls (e.g. a WebSocket hiccup) permanently wedges every future real-chain send for that peer — joins, top-ups, dispute submissions, until the process restarts. Given `ClientChainSigner` deliberately passes `timeoutMs: null` elsewhere because these ops can't be cancelled, this class seems to want the same care.

Could `sendTransaction` re-query `getNonce("pending")` on the next call instead of latching permanently?

src/evm/signer/ClientChainSigner.ts
```diff
@@ -0,0 +1,106 @@
...
+        );
+        // TODO: Revisit recovery for a port that dies while the host broadcast
+        // outcome is unknown. A timeout cannot cancel an in-progress send.
+        const serializedResponse =
+            await this.requester.request<SerializedTransactionResponse>(
+                {
+                    type: "chainSignerSendTransaction",
+                    serializedTransaction
+                },
+                { timeoutMs: null }
```
momoshell, line 74: Have a question here, and good call passing `timeoutMs: null` here since a timeout can't cancel an in-progress host send. 

The sibling `ClientP2pSigner.sendTransaction` (and its `connectToChannel`/`joinChannel`/`topUpBalance`) don't get the same treatment. 
They call `this.client.request(...)` with no `timeoutMs`, so they fall back to `P2pRuntimeClient`'s 30s default (that method isn't touched by this PR, so not asking for a change here, just flagging while the pattern's fresh). 

If a join/transition legitimately takes longer than 30s, the client times out and drops the pending entry while the host keeps running to completion. The caller sees a spurious failure and a retry could double-submit. 

Should those p2p-signer paths use the same `timeoutMs: null` pattern?

src/stateManager/reduction/ReductionExecutor.ts
```diff
@@ -0,0 +1,427 @@
...
+            return;
+        }
+
+        if (disputes.length === 0) {
+            this.logger.warn(
+                `No disputes found while reducing disputed fork ${forkId}; initiating local dispute`
+            );
+            await this.stateManager.disputeManager.dispute(forkId);
+            return;
+        }
```
momoshell, line 193: Wandering also here if I'm correct; when `disputes.length === 0`, this calls `disputeManager.dispute(forkId)` and returns without settling the fork's reduction `completion` (in `ReductionManager`). 

Since `completion.promise` is only resolved by `completeWithGenesis`, whatever awaited `tryReduce` for this fork stays pending until some later event re-triggers completion.

Is there a guaranteed re-trigger path after the local dispute is submitted, or could this hang under `BlockQueueManager.runForkRecovery` if that re-trigger doesn't fire?

contracts/V1/StateChannelDiamondProxy/JoinChannelFacet.sol
```diff
@@ -10,36 +10,66 @@ contract JoinChannelFacet is StateChannelCommon {
...
+            UtilityFacet(utilityFacetAddress).isAddressInArray(thresholdParticipants, jc.participant);
+        if (isTopUp) {
+            require(isExistingParticipant, ErrorTopUpBalanceParticipantNotFound());
+        } else {
+            require(!isExistingParticipant, ErrorJoinChannelParticipantAlreadyExists());
+            require(
+                !StateChannelManagerProxy(address(this)).isForkDisputed(channelId, expectedForkId),
+                RaceConditionForceInboundJoinForkDisputed()
+            );
+        }
```
momoshell, line 65: Also wandering here, the `joinChannel` enforces `!isForkDisputed` for a fresh join, but `topUpBalance` (the `isTopUp` branch) skips that guard entirely.

Oonly `isExistingParticipant` is checked. `expectedForkId` still has to match the current snapshot's fork, but if that fork is later reduced away, could the just-deposited `totalDeposits` end up stranded on the abandoned fork? I

s depositing to a disputed fork intentional for existing participants, or should this guard apply to both paths?

src/stateManager/reduction/ReductionManager.ts
```diff
@@ -0,0 +1,296 @@
...
+                    completion.resolve(result);
+                    return true;
+                },
+                { taskName: `ReductionManager.complete-${forkId}` }
+            );
+        } catch (error) {
+            this.failCompletion(completion, error);
+            await completion.promise;
+            return false;
+        }
```
momoshell, line 225: Just a nit in the `completeWithGenesis` catch block, `failCompletion` rejects `completion.promise`, so the following `await completion.promise` re-throws and `return false` on the next line is unreachable.

The method rejects rather than ever returning `false` on a fatal install error. Harmless (the caller already handles the throw), but worth dropping the dead line so it doesn't read as intentional non-throwing signaling.

src/cache/SignerRecoveryCache.ts
```diff
@@ -0,0 +1,44 @@
...
+ * any message+signature (blocks, join/open/transaction/dispute), not just blocks.
+ *
+ * Local-only derived data: never serialized or transmitted. One instance per
+ * worker thread (module singleton). Bounded by SIGNER_RECOVERY_CACHE_MAX with
+ * insertion-order (FIFO) eviction — an evicted entry is simply recovered again.
+ */
+const cache = new Map<string, Address>();
+
+function keyOf(message: Uint8Array | string, signature: Signature): string {
+    return hexlify(message) + signature;
```
momoshell: A nit here also for the `keyOf` does `hexlify(message) + signature`, but `verifyMessage` treats a `string` message as UTF-8 text and a `Uint8Array` as raw bytes. Which are differently recovered signers, while `hexlify("0xab..")` and `hexlify(Uint8Array[0xab..])` produce the same cache key. 

Currently latent since every caller passes `Uint8Array`, but the doc comment advertises this cache for "any message+signature." Worth narrowing the param to `Uint8Array` to close off a future string caller getting a wrong cached signer?

src/types/time.ts
```diff
@@ -1,8 +1,17 @@
...
 };
+
+// Height-0 evidenceTime grace; must mirror the on-chain firstBlockGrace
+// (DisputeFraudProofFacet._handleTimeoutTooEarly / FraudProofFacet._hasInvalidTimestamp)
+export function firstBlockGrace(
+    timeConfig: TimeConfig,
+    blockHeight: BlockHeight
+): Timestamp {
+    return blockHeight === 0 ? timeConfig.evidenceTime : 0;
+}
```
momoshell, line 17: Nit in the `firstBlockGrace`'s comment where it says it must mirror `DisputeFraudProofFacet._handleTimeoutTooEarly`/`FraudProofFacet._hasInvalidTimestamp` on-chain.

Per the project's shared-validator convention, a check like this should be called via typechain rather than re-implemented in TS, to rule out drift. The logic here is trivial (`height===0 ? evidenceTime : 0`) so risk is low, but should this forward to the contract instead per that rule?

test/AGENTS.md
```diff
@@ -181,10 +181,12 @@ in the parallel summary is the detector.
...
   is fully typed — a removed/renamed field is a compile error, your free
   checklist).
-- Unit/integration: `yarn test`. E2E inline: `yarn test:e2e`.
+- Canonical full gate: `yarn test:parallel`. Add `--e2e-only` to run only E2E
+  tests, or `--grep <regexp>` for the narrowest relevant task.
+- Legacy in-process unit/integration: `yarn test`. E2E inline: `yarn test:e2e`.
 - E2E in worker mode: `yarn test:e2e:worker` (per-file process isolation +
   internal X/N progress; needs the hardhat node — `yarn infra:hardhat-node`).
-- E2E parallel: `yarn test:e2e:parallel` — each run logs to a fresh
+- Parallel runner: `yarn test:parallel` — each run logs to a fresh
```
momoshell, line 189: And one more nit in the cross-reference two lines down ("See root `AGENTS.md` (\"E2E parallel run logs\")") points to a section title this same PR renamed to "Canonical test command and parallel run logs" in root `AGENTS.md`. 

The referenced heading no longer exists. Small doc fixup while this section's already being touched.

momoshell (commented):
Nice work on the extraction of `StateManager`'s monolith into `ReductionManager`/`SnapshotUpdateService`/`EventSyncService`. 
The dispute fraud-proof and P2P runtime code is solid overall, no access-control, reentrancy, or storage-layout issues found in the production contracts.

One thing I'd want resolved before merge: `yarn test:parallel` (which this PR promotes to the canonical gate) silently drops any test with a dynamically-titled `it()`, including the entire `E2E-Fuzz-Dispute-MVP` suite.
That way the gate can report green while real coverage never ran. A few other correctness questions inline, all narrow and easy to check.

scripts/e2e-parallel/taskDiscovery.js
momoshell: Ran the actual gate to check: `yarn compile && yarn test:parallel` and 562/562 passed clean, zero retries, ~11.2min wall. No flakiness in what ran. But I confirmed this bug empirically by comparing `discoverTasks()`'s output against the real test counts:

| File | Real tests | Scheduled |
|---|---|---|
| `E2E-Fuzz-Dispute-MVP.test.ts` | 1 | **0** |
| `E2E-RuntimeTransportModes.test.ts` | 5 (1 static + a 4-way worker/inline × dedicated-vm/inline-vm matrix) | **1** — the whole matrix never runs |
| `E2E-Spectate.test.ts` | 18 | 17 — height-pinning case dropped |

So "562 passing" isn't the full suite, the fuzz-dispute suite and the worker/VM-thread compatibility matrix (arguably the two areas most likely to be flaky) never get scheduled at all, silently. That's a stronger case for fixing this before relying on `test:parallel` as the canonical gate.

lukastanisic99:
@momoshell addressed most of your comments - I think I just skipped this one (https://github.com/peer3to/state-channels-plus/pull/397#discussion_r3615913110) since it's ultra simple and not worth invoking the EVM for, even though the comment is valid and in general we should do it

daiagi (approved):

sh3ll3x3c (approved):
lgtm! already using it as a base for future work

PR #398, opened 2026-07-20
fix first-block calldata grace, spectate junk-block abort, event-sync retry

daiagi:
review fixes for #397.

1. first-block calldata grace - `_validateTimeoutCalldataPostedProof` was missing the height-0 grace the other timeout paths already have
2. spectate junk-block abort - a non-participant sending an unauthenticated block could force a spectator offline; now it drops the sender and keeps running (E2E-SpectatingAbortDoS)

### DIFF
--- a/test/e2e/E2E-SpectatingAbortDoS.test.ts
+++ b/test/e2e/E2E-SpectatingAbortDoS.test.ts
@@ -3,8 +3,13 @@ import { expect } from "chai";
 import { Status } from "@/types";
 import { MathTestSession as TestSession } from "@test/harness";
 
+// A non-participant that sends a spectator a block it must reject should be
+// dropped + blacklisted, never able to take the spectator offline. Covers both
+// rejection paths: unauthenticated junk (rejected synchronously in ingest) and
+// an authenticated outsider-authored block (queued, then rejected when
+// executeQueuedEntry runs onBlockConfirmation).
 describe("E2E: spectating strategy junk-block handling", function () {
-    it("blacklists the sender but keeps a SYNCED spectator running on a junk block from a non-participant", async function () {
+    it("cuts the sender of an unauthenticated junk block and keeps a SYNCED spectator running", async function () {
         const h = TestSession.getHarness();
         await h.lifecycle.start(3, 1);
         const forkId = h.activeForkId!;
@@ -22,9 +27,6 @@ describe("E2E: spectating strategy junk-block handling", function () {
                 timeoutMessage: "attacker never connected to victim"
             }
         );
-        expect(
-            await h.control(victim).query.getStatus().request()
-        ).to.equal(Status.SYNCED);
 
         const { encodedBlockConfirmation } =
             await h.byzantine.craftJunkBlockConfirmation(0, forkId);
@@ -36,32 +38,22 @@ describe("E2E: spectating strategy junk-block handling", function () {
             )
             .request();
 
-        await h.disconnectionBarrier.waitFor(
-            async () =>
-                await h
-                    .control(victim)
-                    .query.isBlacklisted(attacker.address)
-                    .request(),
-            {
-                timeoutMs: h.event.protocolEventTimeoutMs(1),
-                timeoutMessage: "victim never blacklisted attacker"
-            }
-        );
-        expect(
-            await h.control(victim).query.getStatus().request()
-        ).to.equal(Status.SYNCED);
+        await h.assert.rpc.peerBlacklistedAndDisconnected({
+            observer: victim,
+            target: attacker
+        });
     });
 
-    it("keeps a PENDING_PARTICIPANT running on a junk block from a non-participant", async function () {
+    it("cuts the sender of an unauthenticated junk block and keeps a PENDING_PARTICIPANT running", async function () {
         const h = TestSession.getHarness();
         await h.lifecycle.start(3, 1);
         const forkId = h.activeForkId!;
 
         const victim = await h.join.addSpectatorWait();
         await h.join.joinChannelWait({ joiner: victim });
-        expect(
-            await h.control(victim).query.getStatus().request()
-        ).to.equal(Status.PENDING_PARTICIPANT);
+        expect(await h.control(victim).query.getStatus().request()).to.equal(
+            Status.PENDING_PARTICIPANT
+        );
 
         const attacker = await h.join.addSpectator();
         await h.connectionBarrier.waitFor(
@@ -86,43 +78,54 @@ describe("E2E: spectating strategy junk-block handling", function () {
             )
             .request();
 
-        await h.disconnectionBarrier.waitFor(
-            async () =>
-                await h
-                    .control(victim)
-                    .query.isBlacklisted(attacker.address)
-                    .request(),
-            {
-                timeoutMs: h.event.protocolEventTimeoutMs(1),
-                timeoutMessage: "victim never blacklisted attacker"
-            }
-        );
-        expect(
-            await h.control(victim).query.getStatus().request()
-        ).to.equal(Status.PENDING_PARTICIPANT);
+        await h.assert.rpc.peerBlacklistedAndDisconnected({
+            observer: victim,
+            target: attacker
+        });
     });
 
-    it("does not abort a SYNCED spectator validating an authenticated block from a non-participant author", async function () {
+    it("cuts the sender of an authenticated outsider-authored block over the live queue and keeps a SYNCED spectator running", async function () {
         const h = TestSession.getHarness();
         await h.lifecycle.start(3, 1);
         const forkId = h.activeForkId!;
 
         const victim = await h.join.addSpectatorWait();
-        expect(
-            await h.control(victim).query.getStatus().request()
-        ).to.equal(Status.SYNCED);
+        const attacker = await h.join.addSpectator();
+        await h.connectionBarrier.waitFor(
+            async () =>
+                await h
+                    .control(attacker)
+                    .query.isConnectedTo(victim.address)
+                    .request(),
+            {
+                timeoutMs: h.event.protocolEventTimeoutMs(1),
+                timeoutMessage: "attacker never connected to victim"
+            }
+        );
 
+        // the attacker is connected to the channel's p2p network but is not a
+        // channel participant. it authors + signs a well-formed next block with
+        // its own key: authentication passes, so it is queued and validated fresh
+        // -> reaches blockAuthorIsNotParticipant on the live queue, not the inline
+        // authenticate-failed path the junk cases hit. this is the vector that
+        // used to abort the spectator.
         const { encodedBlockConfirmation } =
-            await h.byzantine.craftOutsiderAuthoredBlockConfirmation(0, forkId);
-        const { aborted, result } = await h
-            .control(victim)
-            .stub.probeSpectateBlockNoAbort(encodedBlockConfirmation)
+            await h.byzantine.craftOutsiderAuthoredBlockConfirmation(
+                0,
+                forkId,
+                attacker.signer
+            );
+        await h
+            .control(attacker)
+            .byzantine.sendBlockConfirmation(
+                encodedBlockConfirmation,
+                victim.address
+            )
             .request();
 
-        expect(result).to.equal("DISCONNECT");
-        expect(aborted).to.equal(false);
-        expect(
-            await h.control(victim).query.getStatus().request()
-        ).to.equal(Status.SYNCED);
+        await h.assert.rpc.peerBlacklistedAndDisconnected({
+            observer: victim,
+            target: attacker
+        });
     });
 });
--- a/test/e2e/E2E-SpectatingAbortDoS.test.ts
+++ b/test/e2e/E2E-SpectatingAbortDoS.test.ts
@@ -1,0 +1,128 @@
+import { expect } from "chai";
+
+import { Status } from "@/types";
+import { MathTestSession as TestSession } from "@test/harness";
+
+describe("E2E: spectating strategy junk-block handling", function () {
+    it("blacklists the sender but keeps a SYNCED spectator running on a junk block from a non-participant", async function () {
+        const h = TestSession.getHarness();
+        await h.lifecycle.start(3, 1);
+        const forkId = h.activeForkId!;
+
+        const victim = await h.join.addSpectatorWait();
+        const attacker = await h.join.addSpectator();
+        await h.connectionBarrier.waitFor(
+            async () =>
+                await h
+                    .control(attacker)
+                    .query.isConnectedTo(victim.address)
+                    .request(),
+            {
+                timeoutMs: h.event.protocolEventTimeoutMs(1),
+                timeoutMessage: "attacker never connected to victim"
+            }
+        );
+        expect(
+            await h.control(victim).query.getStatus().request()
+        ).to.equal(Status.SYNCED);
+
+        const { encodedBlockConfirmation } =
+            await h.byzantine.craftJunkBlockConfirmation(0, forkId);
+        await h
+            .control(attacker)
+            .byzantine.sendBlockConfirmation(
+                encodedBlockConfirmation,
+                victim.address
+            )
+            .request();
+
+        await h.disconnectionBarrier.waitFor(
+            async () =>
+                await h
+                    .control(victim)
+                    .query.isBlacklisted(attacker.address)
+                    .request(),
+            {
+                timeoutMs: h.event.protocolEventTimeoutMs(1),
+                timeoutMessage: "victim never blacklisted attacker"
+            }
+        );
+        expect(
+            await h.control(victim).query.getStatus().request()
+        ).to.equal(Status.SYNCED);
+    });
+
+    it("keeps a PENDING_PARTICIPANT running on a junk block from a non-participant", async function () {
+        const h = TestSession.getHarness();
+        await h.lifecycle.start(3, 1);
+        const forkId = h.activeForkId!;
+
+        const victim = await h.join.addSpectatorWait();
+        await h.join.joinChannelWait({ joiner: victim });
+        expect(
+            await h.control(victim).query.getStatus().request()
+        ).to.equal(Status.PENDING_PARTICIPANT);
+
+        const attacker = await h.join.addSpectator();
+        await h.connectionBarrier.waitFor(
+            async () =>
+                await h
+                    .control(attacker)
+                    .query.isConnectedTo(victim.address)
+                    .request(),
+            {
+                timeoutMs: h.event.protocolEventTimeoutMs(1),
+                timeoutMessage: "attacker never connected to victim"
+            }
+        );
+
+        const { encodedBlockConfirmation } =
+            await h.byzantine.craftJunkBlockConfirmation(0, forkId);
+        await h
+            .control(attacker)
+            .byzantine.sendBlockConfirmation(
+                encodedBlockConfirmation,
+                victim.address
+            )
+            .request();
+
+        await h.disconnectionBarrier.waitFor(
+            async () =>
+                await h
+                    .control(victim)
+                    .query.isBlacklisted(attacker.address)
+                    .request(),
+            {
+                timeoutMs: h.event.protocolEventTimeoutMs(1),
+                timeoutMessage: "victim never blacklisted attacker"
+            }
+        );
+        expect(
+            await h.control(victim).query.getStatus().request()
+        ).to.equal(Status.PENDING_PARTICIPANT);
+    });
+
+    it("does not abort a SYNCED spectator validating an authenticated block from a non-participant author", async function () {
+        const h = TestSession.getHarness();
+        await h.lifecycle.start(3, 1);
+        const forkId = h.activeForkId!;
+
+        const victim = await h.join.addSpectatorWait();
+        expect(
+            await h.control(victim).query.getStatus().request()
+        ).to.equal(Status.SYNCED);
+
+        const { encodedBlockConfirmation } =
+            await h.byzantine.craftOutsiderAuthoredBlockConfirmation(0, forkId);
+        const { aborted, result } = await h
+            .control(victim)
+            .stub.probeSpectateBlockNoAbort(encodedBlockConfirmation)
+            .request();
+
+        expect(result).to.equal("DISCONNECT");
+        expect(aborted).to.equal(false);
+        expect(
+            await h.control(victim).query.getStatus().request()
+        ).to.equal(Status.SYNCED);
+    });
+});
--- a/test/e2e/E2E-SpectatingAbortDoS.test.ts
+++ b/test/e2e/E2E-SpectatingAbortDoS.test.ts
@@ -1,17 +1,26 @@
 import { expect } from "chai";
 
 import { Status } from "@/types";
-import { MathTestSession as TestSession } from "@test/harness";
+import {
+    MathTestSession as TestSession,
+    MIN_TEST_TIME_CONFIG
+} from "@test/harness";
 
 // A non-participant that sends a spectator a block it must reject should be
 // dropped + blacklisted, never able to take the spectator offline. Covers both
 // rejection paths: unauthenticated junk (rejected synchronously in ingest) and
 // an authenticated outsider-authored block (queued, then rejected when
 // executeQueuedEntry runs onBlockConfirmation).
+
+const LIVE_FORK_TIME = {
+    ...MIN_TEST_TIME_CONFIG,
+    chainFallbackTime: 12
+};
+
 describe("E2E: spectating strategy junk-block handling", function () {
     it("cuts the sender of an unauthenticated junk block and keeps a SYNCED spectator running", async function () {
         const h = TestSession.getHarness();
-        await h.lifecycle.start(3, 1);
+        await h.lifecycle.start(3, 1, { timeConfig: LIVE_FORK_TIME });
         const forkId = h.activeForkId!;
 
         const victim = await h.join.addSpectatorWait();
@@ -40,13 +49,14 @@ describe("E2E: spectating strategy junk-block handling", function () {
 
         await h.assert.rpc.peerBlacklistedAndDisconnected({
             observer: victim,
-            target: attacker
+            target: attacker,
+            expectedStatus: Status.SYNCED
         });
     });
 
     it("cuts the sender of an unauthenticated junk block and keeps a PENDING_PARTICIPANT running", async function () {
         const h = TestSession.getHarness();
-        await h.lifecycle.start(3, 1);
+        await h.lifecycle.start(3, 1, { timeConfig: LIVE_FORK_TIME });
         const forkId = h.activeForkId!;
 
         const victim = await h.join.addSpectatorWait();
@@ -80,13 +90,14 @@ describe("E2E: spectating strategy junk-block handling", function () {
 
         await h.assert.rpc.peerBlacklistedAndDisconnected({
             observer: victim,
-            target: attacker
+            target: attacker,
+            expectedStatus: Status.PENDING_PARTICIPANT
         });
     });
 
     it("cuts the sender of an authenticated outsider-authored block over the live queue and keeps a SYNCED spectator running", async function () {
         const h = TestSession.getHarness();
-        await h.lifecycle.start(3, 1);
+        await h.lifecycle.start(3, 1, { timeConfig: LIVE_FORK_TIME });
         const forkId = h.activeForkId!;
 
         const victim = await h.join.addSpectatorWait();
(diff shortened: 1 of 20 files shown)


test/e2e/E2E-SpectatingAbortDoS.test.ts
```diff
@@ -0,0 +1,97 @@
...
+        const victim = await h.join.addSpectatorWait();
+        const attacker = await h.join.addSpectator();
+        await waitFor(
+            async () =>
+                await h
+                    .control(attacker)
+                    .query.isConnectedTo(victim.address)
+                    .request(),
+            15000,
+            100
```
lukastanisic99: I forgot we're still using polling in some places - 100ms poll is imo too aggressive 

Image
Btw default 5ms is suicide - I don't know my review skill didn't flag this

daiagi: https://trello.com/c/wMnxDVC4

and removed from this test

lukastanisic99:
@daiagi take a look do you like the bellow format


**Bottom line:** rework — the gate is clean but two of the three stated fixes don't close the hole they claim to. Verified: `yarn compile` exit 0; `yarn tsc --noEmit` passes on **both** `tsconfig.json` and `tsconfig.browser.json`; `yarn test:parallel` **564 passing / 0 failing** (wall 798 s, `./logs/run-431`, zero error logs). Two 🔴 block, and neither is a test failure — they're holes the green suite doesn't cover. Fix 1 (calldata grace) is correct but untested; fix 2 (spectate abort) removes 1 of 11 `abort()` calls and leaves a *more reachable* one; fix 3 (event-sync retry) restores retryability but leaves the permanent watermark stall.

## 🧱 Fundamental

- 🔴 **[FR1] The spectate fix removes one `abort()` out of eleven, and the one left behind is easier to reach than the one removed.** The diff deletes `this.abort()` from [`authenticateBlockFailed`](https://github.com/peer3to/state-channels-plus/blob/25f39a1ee31ae5322061c6f59ac7b0698f2649fa/src/stateManager/validationStrategy/SpectatingValidationStrategy.ts#L60-L64) only; ten remain. [`blockAuthorIsNotParticipant:115`](https://github.com/peer3to/state-channels-plus/blob/25f39a1ee31ae5322061c6f59ac7b0698f2649fa/src/stateManager/validationStrategy/SpectatingValidationStrategy.ts#L115) is reached with a **strictly weaker** attack than the one just fixed: [`ValidationService:82-91`](https://github.com/peer3to/state-channels-plus/blob/25f39a1ee31ae5322061c6f59ac7b0698f2649fa/src/stateManager/ValidationService.ts#L82-L91) checks author membership *before* any linkage, height, leader or timestamp check, so an outsider signs a well-formed block with their own key — authentication now passes — and the victim still calls `abort()` → full `dispose()` ([`StateManager:254-265`](https://github.com/peer3to/state-channels-plus/blob/25f39a1ee31ae5322061c6f59ac7b0698f2649fa/src/stateManager/StateManager.ts#L254-L265)). The PR body's claim that a non-participant "could force a spectator offline; now it drops the sender and keeps running" holds only for the *unauthenticated* case.

  > **Fix FR1-FIX**
  >
  > Sweep the strategy rather than one hook: for every peer-attributable deviation return `DISCONNECT`/`DISPUTE` and blacklist `entry.sourcePeers ∪ block.author`, mirroring [`BlockValidationStrategy:62-66`](https://github.com/peer3to/state-channels-plus/blob/25f39a1ee31ae5322061c6f59ac7b0698f2649fa/src/stateManager/validationStrategy/BlockValidationStrategy.ts#L62-L66); reserve `abort()` for locally-fatal state. Acceptance: the TO2 test (outsider-authored block) fails before the sweep and passes after.

- 🔴 **[FR2] The event-sync fix restores retryability but leaves the block permanently unpublishable — the damaging half.** The two added `delete`s at [`EventSyncService:128-129`](https://github.com/peer3to/state-channels-plus/blob/25f39a1ee31ae5322061c6f59ac7b0698f2649fa/src/stateManager/EventSyncService.ts#L128-L129) correctly stop a replay from getting the cached rejection. But [`state.failed = true`](https://github.com/peer3to/state-channels-plus/blob/25f39a1ee31ae5322061c6f59ac7b0698f2649fa/src/stateManager/EventSyncService.ts#L130) is still never reset, and a replay re-fetches that same state object at [`:117`](https://github.com/peer3to/state-channels-plus/blob/25f39a1ee31ae5322061c6f59ac7b0698f2649fa/src/stateManager/EventSyncService.ts#L117), so `complete = pending === 0 && !state.failed` at [`:141`](https://github.com/peer3to/state-channels-plus/blob/25f39a1ee31ae5322061c6f59ac7b0698f2649fa/src/stateManager/EventSyncService.ts#L141) can never become true again. `publishCompletedBlocks` therefore still stops at that block forever and the processed-block watermark never advances — **even after a successful retry**. Because `state` is shared by every log in the block, one failed handler also poisons its siblings.

  > **Fix FR2-FIX**
  >
  > Replace the boolean with `failedEventKeys: Set<EventKey>` on the block state: add the key in the `.catch`, remove it when that key is rescheduled, and compute `complete = pending === 0 && failedEventKeys.size === 0`. Note this also requires rewriting [`EventSyncService.test.ts`](https://github.com/peer3to/state-channels-plus/blob/25f39a1ee31ae5322061c6f59ac7b0698f2649fa/test/stateManager/EventSyncService.test.ts) (owned by #397), whose `cursorAfter === cursorBefore` assertion currently codifies the stall as correct and will block the fix. Acceptance: fail a handler, let it settle, reschedule the same log with the handler restored — the handler runs a second time **and** `getLatestProcessedBlock` advances past that block.

- 🟢 **[FG3] The calldata-grace fix is correct.** [`DisputeFraudProofFacet:610`](https://github.com/peer3to/state-channels-plus/blob/25f39a1ee31ae5322061c6f59ac7b0698f2649fa/contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol#L610) now derives `firstBlockGrace = hasBlock ? 0 : getEvidenceTime()` and folds it into the `Math.tryAdd`, matching the sibling at [`:498`](https://github.com/peer3to/state-channels-plus/blob/25f39a1ee31ae5322061c6f59ac7b0698f2649fa/contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol#L498). `hasBlock` was already in scope, the `ok` guard is preserved, and it agrees with TS [`firstBlockGrace`](https://github.com/peer3to/state-channels-plus/blob/25f39a1ee31ae5322061c6f59ac7b0698f2649fa/src/types/time.ts#L12-L17) — I confirmed `!hasBlock` ⟺ timeout height 0 via [`:418`](https://github.com/peer3to/state-channels-plus/blob/25f39a1ee31ae5322061c6f59ac7b0698f2649fa/contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol#L418).

  > **Fix FG3-FIX**
  >
  > No change required; pin it with a test (TO1).

- 🔴 **[SR1] A handshaken non-participant can still permanently kill a spectating or joining node — the fixed hook was not the reachable one.** `StateTransitionService` gates only on handshake completion, and [`isBlockConfirmationAuthentic`](https://github.com/peer3to/state-channels-plus/blob/25f39a1ee31ae5322061c6f59ac7b0698f2649fa/src/stateManager/StateManager.ts#L695-L701) verifies the signature, not membership. A throwaway wallet signs a block carrying the victim's `channelId`/`forkId` (both readable via spectate/sync); authentication passes, FR1's path fires, the node disposes. Every non-`PARTICIPATING` status uses this strategy ([`getStrategyByStatus`](https://github.com/peer3to/state-channels-plus/blob/25f39a1ee31ae5322061c6f59ac7b0698f2649fa/src/stateManager/StateManager.ts#L2503-L2508)), so `PENDING_PARTICIPANT` nodes that have already committed their join on-chain are in scope. Ten hooks, including the ones a Byzantine *participant* reaches with a validly-signed bad block (`invalidStateTransitionDetected`, `wrongGenesisDetected`, `doubleSignDetected`), remain terminal.

  > **Fix SR1-FIX**
  >
  > Apply FR1-FIX. Acceptance: a handshaken non-participant sends a self-signed block on the victim's channel and fork; the victim blacklists the sender, keeps its status, stays `isDisposed === false`, and still processes the next honest block.

- 🟢 **[SG2] The fix introduces no new remote surface.** The three source changes are a subtraction (`abort()`), two map `delete`s on a local cache, and one arithmetic term — nothing new is reachable from a peer, and the harness additions are test-only.

  > **Fix SG2-FIX**
  >
  > No change required.

## 🏁 Race conditions

- 🟢 **[RG1] No new race, and the `delete`-then-`throw` ordering is safe.** The `.catch` deletes the cache entries before rethrowing, and the `.finally` that decrements `state.pending` runs after — so a concurrent `scheduleLog` for the same key either joins the still-cached promise or builds a fresh one, never observes a half-torn entry. FR2 is a liveness bug, not a race.

  > **Fix RG1-FIX**
  >
  > No change required.

## ⚡ Performance

- 🟠 **[PO1] `waitFor`'s 5 ms default poll interval is a live hazard at 32 call sites — this PR removes its own instance but not the default.** [`test/utils/waitFor.ts`](https://github.com/peer3to/state-channels-plus/blob/25f39a1ee31ae5322061c6f59ac7b0698f2649fa/test/utils/waitFor.ts) declares `pollIntervalMs: number = 5`. I counted call sites: **32 rely on that default**, 15 pass an explicit interval. Most poll an async control-RPC (`.request()`), so each waiting test issues a request every 5 ms; under `test:parallel` that is real contention, and this run's `event-loop delay: peak 741 ms` shows the headroom is not infinite. Not introduced by this PR — the diff *removes* the polling from its own test — but it is the general form of the problem raised in the PR thread, and it is why a green suite can still be a starved one.

  > **Fix PO1-FIX**
  >
  > Raise the default to something protocol-scaled (derive from `protocolEventTimeoutMs`, or a flat 100–250 ms) and migrate the 32 default-relying sites; prefer `connectionBarrier`/`disconnectionBarrier` where an event already exists to wait on. Acceptance: `grep` shows no `waitFor(` call relying on a sub-50 ms interval, and the gate's peak event-loop delay drops.

- 🟠 **[TO1] The one fully-correct fix is the one with no test.** Nothing in this PR pins the grace change. The two existing cases — [`DisputeVerificationFacet.t.sol:278`](https://github.com/peer3to/state-channels-plus/blob/25f39a1ee31ae5322061c6f59ac7b0698f2649fa/test/V1/StateChannelDiamondProxy/DisputeVerificationFacet.t.sol#L278) (valid proof) and [`:285`](https://github.com/peer3to/state-channels-plus/blob/25f39a1ee31ae5322061c6f59ac7b0698f2649fa/test/V1/StateChannelDiamondProxy/DisputeVerificationFacet.t.sol#L285) (wrong origin fork) — never touch the height-0 boundary, so deleting line 610 again leaves the suite green. The sibling path *is* pinned, at [`FraudProofFacet.t.sol:95`](https://github.com/peer3to/state-channels-plus/blob/25f39a1ee31ae5322061c6f59ac7b0698f2649fa/test/V1/StateChannelDiamondProxy/FraudProofFacet.t.sol#L95) (`test_hasInvalidTimestamp_firstBlockGraceBoundary`: valid at `prev + evidenceTime + p2pTime`, fraud at `+1`), with a no-grace-at-height-1 companion at [`:102`](https://github.com/peer3to/state-channels-plus/blob/25f39a1ee31ae5322061c6f59ac7b0698f2649fa/test/V1/StateChannelDiamondProxy/FraudProofFacet.t.sol#L102). That is the model to copy.

  > **Fix TO1-FIX**
  >
  > Add a Foundry case asserting `validateTimeoutCalldataPostedProof` accepts a height-0 calldata post at exactly `prev + evidenceTime + p2p + agreement + chainFallback` and rejects it at `+1`, plus a height-1 case proving no grace is granted.

- 🟠 **[TO2] The new e2e test exercises only the hook that was fixed, so it cannot detect FR1.** [`craftJunkBlockConfirmation:238`](https://github.com/peer3to/state-channels-plus/blob/25f39a1ee31ae5322061c6f59ac7b0698f2649fa/test/harness/actions/ByzantineActions.ts#L238) re-signs a real block with a throwaway key, so *authentication* fails — routing to `authenticateBlockFailed`. Reaching `blockAuthorIsNotParticipant` needs the opposite: a block the outsider signs **validly**, so authentication passes and only membership fails. Neither case does that, nor do they touch the other nine aborting hooks.

  > **Fix TO2-FIX**
  >
  > Add `craftOutsiderAuthoredBlock` (build a fresh block body authored by a throwaway key, signed correctly by that key — `craftBogusForkBlockZero` is the shape to start from) and assert both victims survive; add one Byzantine-*participant* case (invalid state transition) for the hooks a real member reaches.

- 🟢 **[TG3] The full canonical gate is green: 564 passing, 0 failing.** `yarn compile` exit 0, both typechecks pass, `yarn test:parallel` wall 798 s into `./logs/run-431` with **zero** `error*` logs (the runner cleans the run dir on a fully green run, which is why it is now empty). This also **retires round 1's 🔴 test finding**: the two failures seen earlier came from `yarn test`, the legacy in-process tier, where multi-peer harness tests contend for one event loop. Under the canonical gate — the authoritative one per AGENTS.md — `EventSyncService` passes. Round 1's suspicion that those were starvation artifacts is now confirmed rather than assumed.

  > **Fix TG3-FIX**
  >
  > No change required; keep `test:parallel` as the gate of record and treat `yarn test` results as advisory.

- 🟢 **[TG4] The PR head's version of the test is a real improvement over the local draft, and it closes the review thread.** `25f39a1e` replaces hardcoded `waitFor(…, 15000, 100)` polling with [`connectionBarrier`](https://github.com/peer3to/state-channels-plus/blob/25f39a1ee31ae5322061c6f59ac7b0698f2649fa/test/fixtures/PeerTestHarness.ts#L115)/[`disconnectionBarrier`](https://github.com/peer3to/state-channels-plus/blob/25f39a1ee31ae5322061c6f59ac7b0698f2649fa/test/fixtures/PeerTestHarness.ts#L118) `.waitFor(…, { timeoutMs: h.event.protocolEventTimeoutMs(1), timeoutMessage: … })`. These are `EventBarrier`s signalled from the harness's `onConnection`/`onDisconnection` hooks ([`:602`](https://github.com/peer3to/state-channels-plus/blob/25f39a1ee31ae5322061c6f59ac7b0698f2649fa/test/fixtures/PeerTestHarness.ts#L602), [`:609`](https://github.com/peer3to/state-channels-plus/blob/25f39a1ee31ae5322061c6f59ac7b0698f2649fa/test/fixtures/PeerTestHarness.ts#L609)), so waiting on `disconnectionBarrier` for a blacklist is event-driven and semantically apt, not a poll; the timeout is derived from the time config, matching the convention at [`AssertRPCActions.ts:18`](https://github.com/peer3to/state-channels-plus/blob/25f39a1ee31ae5322061c6f59ac7b0698f2649fa/test/harness/actions/assert/AssertRPCActions.ts#L18) and the intent of commit `7da25e4f`. Both cases also assert the positive outcome (attacker blacklisted) *and* the negative one (status unchanged, which does catch `abort()` since it sets `Status.OPENED`), and covering `PENDING_PARTICIPANT` separately is exactly the victim class people forget.

  > **Fix TG4-FIX**
  >
  > No change required; extend per TO2. Pull `25f39a1e` locally so the working tree stops lagging the PR.

## ♻️ Code reuse

- 🟠 **[CO1] The fix copy-pastes the grace expression instead of extracting it — the same shape that caused the bug.** `uint256 firstBlockGrace = hasBlock ? 0 : getEvidenceTime();` now appears verbatim at [`:498`](https://github.com/peer3to/state-channels-plus/blob/25f39a1ee31ae5322061c6f59ac7b0698f2649fa/contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol#L498) **and** [`:610`](https://github.com/peer3to/state-channels-plus/blob/25f39a1ee31ae5322061c6f59ac7b0698f2649fa/contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol#L610), and the full deadline sum is likewise duplicated across both. The original defect existed precisely because these two handlers maintain the same rule separately; restoring parity by adding a second copy fixes today's divergence without removing tomorrow's. AGENTS.md is explicit: "never copy-paste the behavior… keep one owner for each operation."

  > **Fix CO1-FIX**
  >
  > Extract one `internal` helper — `_maxValidTimeoutTimestamp(previousTimestamp, hasBlock) → (bool ok, uint256 max)` — owning both the grace term and the `p2p + agreement + chainFallback` sum, and call it from both handlers. Acceptance: `grep -c "getEvidenceTime()" DisputeFraudProofFacet.sol` returns 1, and TO1's boundary test still passes.

- 🟢 **[CG2] The new harness helper does not duplicate its sibling, and it encodes correctly.** [`craftJunkBlockConfirmation:238`](https://github.com/peer3to/state-channels-plus/blob/25f39a1ee31ae5322061c6f59ac7b0698f2649fa/test/harness/actions/ByzantineActions.ts#L238) re-signs an existing block, whereas [`craftBogusForkBlockZero:204`](https://github.com/peer3to/state-channels-plus/blob/25f39a1ee31ae5322061c6f59ac7b0698f2649fa/test/harness/actions/ByzantineActions.ts#L204) builds a fresh one via the factory — different purposes, and the only shared part is a trivial struct literal. It correctly returns a `Codec.encode`d `encodedBlockConfirmation` because that value crosses the control-RPC port, per the AGENTS.md encoding rule, while the sibling returns a raw struct because it does not.

  > **Fix CG2-FIX**
  >
  > No change required.

## 📏 AGENTS.md adherence

- 🟢 **[AG1] Clean.** No `console.*`, no casts, no class-layout issues. The harness helper lives on the action class with a short, accurate docstring and uses `Codec.encode`/`Codec.decode` at the port boundary. The Solidity change stays inside the owning facet — no new `public` surface on `StateChannelCommon`, no TS re-implementation. The e2e test uses the real RPC path with no mocks and factory/harness-built data. CO1 is the one rule this PR bends.

  > **Fix AG1-FIX**
  >
  > No change required beyond CO1-FIX.

## 🧹 Miscellaneous (scope)

- 🟡 **[MY1] The PR body will read as "fixed" for all three comments.** It states the spectator "keeps running" and that event-sync "can reprocess it" — both true only in the narrow sense FR1 and FR2 describe. Whoever resolves daiagi's comments on #397 from this description will close two holes that are still open.

  > **Fix MY1-FIX**
  >
  > Either land FR1-FIX and FR2-FIX here, or amend the body to scope each claim ("closes the unauthenticated-block path; the authenticated-outsider path remains — tracked in X") and leave the #397 comments open.

## ❓ Open questions

- 🔴 **[OR1] Land this as a complete fix, or as an explicitly-scoped increment?** As written it will read as closing daiagi's comments 3 and 4 while the spectator DoS and the watermark stall are both live. My recommendation: finish FR1 and FR2 here — a partial fix that closes the issue is worse than no fix, because the remaining hole gets marked resolved and stops being looked at.

  > **Fix OR1-FIX**
  >
  > Land FR1-FIX + FR2-FIX in this PR. If that is too much scope, retitle to name the narrow paths fixed and post the residual attack and stall on #397 so the comments stay open.

- 🟠 **[OO2] Should the `waitFor` default be fixed repo-wide now, or tracked?** PO1 affects 32 sites and plausibly contributes to the starvation-shaped failures that made #397 hard to call green.

  > **Fix OO2-FIX**
  >
  > Track it as its own PR rather than growing this one — but do it before the next flake investigation, since it changes what a "green" run means.

daiagi:
@lukastanisic99  ready for re-review

lukastanisic99:

- 🟠 **[SO1] An authenticated outsider is rejected but not disconnected on the real queued path.** Authenticated blocks are attributed and queued at [`queueBlock`](https://github.com/peer3to/state-channels-plus/blob/88a27783ad7f70dd917642f87f68612953e16867/src/stateManager/BlockQueueManager.ts#L147-L153). The spectator hook promises “drop + blacklist” but only returns `DISCONNECT` ([`blockAuthorIsNotParticipant`](https://github.com/peer3to/state-channels-plus/blob/88a27783ad7f70dd917642f87f68612953e16867/src/stateManager/validationStrategy/SpectatingValidationStrategy.ts#L126-L132)); later, [`executeQueuedEntry`](https://github.com/peer3to/state-channels-plus/blob/88a27783ad7f70dd917642f87f68612953e16867/src/stateManager/BlockQueueManager.ts#L518-L545) merely logs `false`. Unlike the stored-block merge path, it never disconnects `entry.sourcePeers`, so the authenticated outsider remains connected and can keep feeding fresh rejected blocks.

  > **Fix SO1-FIX**
  >
  > Pass the attributed `QueuedBlockEntry` into this deviation hook and perform the disconnect/blacklist inside every applicable strategy implementation, as required by the strategy boundary. Add a real RPC E2E assertion that the outsider is blacklisted while the spectator remains `SYNCED`.

## ⚡ Performance

- 🟠 **[PO1] One unretried early block causes unbounded event-sync state retention.** [`publishCompletedBlocks`](https://github.com/peer3to/state-channels-plus/blob/88a27783ad7f70dd917642f87f68612953e16867/src/stateManager/EventSyncService.ts#L471-L495) stops at the first incomplete block, so FR1 leaves later block states and promises accumulating for the lifetime of the process.

  > **Fix PO1-FIX**
  >
  > Implement FR1-FIX, then stress a long stream behind one transient failure and assert the watermark catches up and the tracked block/event maps return to their steady-state bound.

- 🟠 **[TO3] The authenticated-outsider test bypasses attribution and the live queue.** The test calls [`probeSpectateBlockNoAbort`](https://github.com/peer3to/state-channels-plus/blob/88a27783ad7f70dd917642f87f68612953e16867/test/e2e/E2E-SpectatingAbortDoS.test.ts#L105-L126), whose fixture creates a [sourceless entry and directly invokes validation](https://github.com/peer3to/state-channels-plus/blob/88a27783ad7f70dd917642f87f68612953e16867/test/fixtures/customRpc/harnessControl/services/stub/StubService.ts#L320-L343). It cannot catch SO1 because there is no sender to disconnect and `executeQueuedEntry` never runs.

  > **Fix TO3-FIX**
  >
  > Send the authenticated outsider block over the real state-transition RPC, wait for the receiver’s real queue path, and assert sender blacklisting/disconnection plus unchanged spectator status.

- 🟠 **[TO4] The first-block contract change has a Foundry boundary test but no peer-observable E2E.** The new tests correctly cover the exact edge and edge+1 ([`DisputeVerificationFacet.t.sol`](https://github.com/peer3to/state-channels-plus/blob/88a27783ad7f70dd917642f87f68612953e16867/test/V1/StateChannelDiamondProxy/DisputeVerificationFacet.t.sol#L301-L317)), but the existing `TimeoutCalldataPosted` E2E deliberately establishes height 1 to avoid first-block grace ([`timeout.test.ts`](https://github.com/peer3to/state-channels-plus/blob/88a27783ad7f70dd917642f87f68612953e16867/test/e2e/disputeValidation/disputeInputFields/timeout.test.ts#L279-L310)). The changed on-chain/off-chain dispute behavior at height 0 is therefore unverified end to end.

  > **Fix TO4-FIX**
  >
  > Add a height-0 flow in which calldata is posted after the normal deadline but inside evidence grace, a timeout dispute is submitted, `TimeoutCalldataPosted` is constructed and applied, and the honest first author is not slashed.

## ♻️ Code reuse

- 🟠 **[CO1] The same timeout boundary calculation now has two live owners in one facet.** [`_validateTimeoutDispute`](https://github.com/peer3to/state-channels-plus/blob/88a27783ad7f70dd917642f87f68612953e16867/contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol#L498-L503) and [`_validateTimeoutCalldataPostedProof`](https://github.com/peer3to/state-channels-plus/blob/88a27783ad7f70dd917642f87f68612953e16867/contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol#L610-L615) independently assemble the identical first-block grace plus P2P/agreement/fallback duration. This is exactly the invariant whose two paths drifted.

  > **Fix CO1-FIX**
  >
  > Extract one internal helper on `DisputeFraudProofFacet` that computes the checked timeout boundary from `previousTimestamp` and `hasBlock`; call it from both validators and keep boundary tests for both consumers.

## 📏 AGENTS.md adherence

- 🟠 **[AO1] The retry fixture violates the real-collaborator and trigger-source rules.** [`test/AGENTS.md`](https://github.com/peer3to/state-channels-plus/blob/88a27783ad7f70dd917642f87f68612953e16867/test/AGENTS.md#L123-L147) requires unit tests to keep real collaborators wired and fixtures to trigger source code rather than reimplement asserted behavior. [`probeRetriedEventSyncLog`](https://github.com/peer3to/state-channels-plus/blob/88a27783ad7f70dd917642f87f68612953e16867/test/fixtures/customRpc/harnessControl/services/stub/StubService.ts#L272-L310) replaces the entire handler and supplies the missing retry from the fixture.

  > **Fix AO1-FIX**
  >
  > Implement TO2-FIX using narrowly scoped fault injection that still executes the real event handler and production retry driver.

- 🟠 **[AO2] The authenticated-outsider test discards the pipeline’s required source attribution.** Root [`AGENTS.md`](https://github.com/peer3to/state-channels-plus/blob/88a27783ad7f70dd917642f87f68612953e16867/AGENTS.md#L80-L105) defines `QueuedBlockEntry` and its sources as the validation unit, while [`probeSpectateBlockNoAbort`](https://github.com/peer3to/state-channels-plus/blob/88a27783ad7f70dd917642f87f68612953e16867/test/fixtures/customRpc/harnessControl/services/stub/StubService.ts#L320-L340) intentionally creates a sourceless entry. That makes the asserted disconnect result incomplete at the trust boundary.

  > **Fix AO2-FIX**
  >
  > Implement TO3-FIX through the live RPC/queue path and assert the real side effect against the attributed source peer.

---------------------

Further I had a failing test due to the event barrier timeout being hardcoded to 15s while the reduction had to wait 15s - all the timeouts have a protocol aware default, so no need to hardcode. 

The fix is 2 lines of code, so I'll include it here as a screenshot so I don't create a PR for them, so please include these 2 lines:
image
image


-----------------------
@daiagi 

create a Trello card for this if you don't want to do it here:
In general log replay should be fully idempotent and we should not have 'partial execution' -> if a log throws it should be fatal and bubble up - the onEventX has to account for any race condition, so we're not 'rescheduling' failed logs -> any throw is fatal.

## 🧱 Fundamental

- 🔴 **[FR1] Failed subscription events become retryable, but nothing in production schedules the retry.** [`scheduleLog`](https://github.com/peer3to/state-channels-plus/blob/88a27783ad7f70dd917642f87f68612953e16867/src/stateManager/EventSyncService.ts#L114-L157) deletes the rejected promise and leaves the failed event key behind, while the live [`StateChannelEventListener`](https://github.com/peer3to/state-channels-plus/blob/88a27783ad7f70dd917642f87f68612953e16867/src/StateChannelEventListener.ts#L34-L47) schedules a log only when the provider emits it. A provider does not re-emit a log because an asynchronous application handler later rejected. The only query-driven rescheduling paths are specialized calldata recovery and dispute recovery ([calldata](https://github.com/peer3to/state-channels-plus/blob/88a27783ad7f70dd917642f87f68612953e16867/src/stateManager/EventSyncService.ts#L220-L240), [disputes](https://github.com/peer3to/state-channels-plus/blob/88a27783ad7f70dd917642f87f68612953e16867/src/stateManager/EventSyncService.ts#L306-L340)); ordinary events such as `ChannelOpened` and `StateSnapshotUpdated` still wedge the watermark forever.

  > **Fix FR1-FIX**
  >
  > Make `EventSyncService` own the retry lifecycle: retain the failed `Log`, schedule disposal-aware retries with capped exponential backoff, and keep retrying until the event succeeds or the service is explicitly stopped. Prove this through the real subscription/retry driver; the test must inject one transient failure and never call `scheduleLog` a second time itself.

- 🔴 **[FR2] Replaying the whole event handler is unsafe after a partial success.** A `StateSnapshotUpdated` retry re-enters [`onStateSnapshotUpdated`](https://github.com/peer3to/state-channels-plus/blob/88a27783ad7f70dd917642f87f68612953e16867/src/eventHandlers/EventHandler.ts#L93-L106) after [`LocalDiamond._acceptEvent`](https://github.com/peer3to/state-channels-plus/blob/88a27783ad7f70dd917642f87f68612953e16867/contracts/V1/StateChannelDiamondProxy/LocalDiamond.sol#L264-L280) may already have committed its coordinate. The Solidity call then no-ops, but TypeScript continues with downstream processing. Dispute events are riskier: [`onDisputeCommitted`](https://github.com/peer3to/state-channels-plus/blob/88a27783ad7f70dd917642f87f68612953e16867/src/eventHandlers/EventHandler.ts#L285-L312) forgets its in-flight guard after rejection and can reach on-chain [`killDispute` / `dispute`](https://github.com/peer3to/state-channels-plus/blob/88a27783ad7f70dd917642f87f68612953e16867/src/eventHandlers/EventHandler.ts#L557-L582) again. A transaction can succeed at the node and still surface a transport error, so retrying from the beginning can duplicate notifications, queue/state mutations, or paid chain actions.

  > **Fix FR2-FIX**
  >
  > Define event-coordinate idempotency for the complete handler, not just the LocalDiamond prefix. Persist completion per effect/phase (including dispute events, which currently receive no coordinate), resume only incomplete phases, and make on-chain actions reconcile transaction/on-chain state before resubmission. Add a failure injected after a real first effect and prove one retry completes without repeating that effect.

lukastanisic99:
Ping me when I should review

daiagi:
>Ping me when I should review

@lukastanisic99  ready

lukastanisic99 (approved):
@daiagi Approving and pls add the following to Trello:

- 🟠 **[TO4] Height-0 `TimeoutCalldataPosted` still lacks peer-observable fraud-proof coverage.** The Foundry edge and edge+1 tests directly call the validator ([`DisputeVerificationFacet.t.sol`](https://github.com/peer3to/state-channels-plus/blob/3a0cfcf86cc4c5c07665e17191e1c4c4ce7f25b6/test/V1/StateChannelDiamondProxy/DisputeVerificationFacet.t.sol#L292-L317)). The existing fraud-proof E2E deliberately advances beyond height 0 ([`timeout.test.ts`](https://github.com/peer3to/state-channels-plus/blob/3a0cfcf86cc4c5c07665e17191e1c4c4ce7f25b6/test/e2e/disputeValidation/disputeInputFields/timeout.test.ts#L279-L346)), so construction, submission/application, dispute killing, and honest first-author survival inside the new grace boundary remain unverified.

  > **Fix TO4-FIX**
  >
  > Add a height-0 flow that posts calldata after the ordinary deadline but inside evidence grace, applies `TimeoutCalldataPosted`, kills the invalid timeout dispute, and proves the honest first author is not slashed.

## ♻️ Code reuse

- 🟡 **[CO2] The same `sourcePeers ∪ author` attribution operation now has multiple owners.** It is rebuilt in both validation strategies for outsider author and missing genesis, while [`BlockQueueManager.requestSync`](https://github.com/peer3to/state-channels-plus/blob/3a0cfcf86cc4c5c07665e17191e1c4c4ce7f25b6/src/stateManager/BlockQueueManager.ts#L558-L568) independently assembles the same set. This is the attribution invariant the PR is trying to make consistent.

  > **Fix CO2-FIX**
  >
  > Give `QueuedBlockEntry` or one queue helper a single `sourcePeersAndAuthor` operation and reuse it at all punishment/sync sites.

daiagi:
>pls add the following to Trello

done

PR #399, opened 2026-07-20
Fix/Remove-polling (`waitFor`)

daiagi:
Follow-up to PR #398's [PO1] comment (waitFor's 5ms default poll interval is a hazard under test:parallel load). Converts the call sites that had a real event to wait on, and raises the shared default for the rest.

1. Converted to event-driven waits (7 test files + 1 new harness helper AssertSyncActions.peerBlacklistedWait): E2E-JoinChannelRaceConditions, E2E-MaliciousUpdateSnapshot, E2E-RuntimeContractEvents, E2E-CustomRpcRequestResponse, E2E-Spectate, E2E-SpectateStaleProofGuard, E2E-BlockQueueManager, E2E-FraudProofsBlockConfirmation. Each replaces a waitFor(condition, ms) poll with eventCountsBarrier/an existing harness wait (onAbort, onConnection, forkChangedWait, peerBlacklistedWait, or a listener-resolved promise), re-checking on real signals instead of a fixed interval.

2. Left as polling, deliberately — where there's no event to hook into:

E2E-RuntimeTransportModes / E2E-CustomRpcRequestResponse's waitForNode — polling a hardhat node before it's even listening; no event source can exist yet.
E2E-PingService — bespoke demo RPC manifest with no wired event infra; not worth inventing plumbing for an example fixture.
E2E-Spectate's Guard Protection (x2) and Concurrent-sync-dedup (x1) — bespoke stub counters (capturedInitHandshakeTransport, stubCountSpectateRequests) with no corresponding domain event.
E2E-BlockQueueManager's isBlockQueued eviction-at-timeout waits (x2) and heldSnapshotEventCount (x2) — the eviction is a bare scheduled timer with no event, and the held-event-count stub intentionally intercepts the real handler, so by design nothing fires normally while it's active.
AssertDisputeActions.reductionCompletedWait (E2E-FinalDispute, E2E-ReductionManager) — the signal path (onDisputeReducedResultCommitted) wasn't clearly wired to eventCountsBarrier; stopped short rather than guess on a helper shared by multiple suites.
One file (E2E-FraudProofsBlockConfirmation) had an explicit comment claiming event-based waiting is racy there — traced the real mechanism (stale spy-call-history match on blockHash, specific to ingestBlockConfirmationWait's own internal wait) and confirmed the converted follow-up waits read live state instead, so they aren't exposed to that race. Verified 5/5 individually + 5/5 under  parallel load.

 3. Raised waitFor's default pollIntervalMs 5→100ms for the ~32 remaining default-reliant call sites left untouched.

### DIFF
--- a/test/e2e/E2E-BlockQueueManager.test.ts
+++ b/test/e2e/E2E-BlockQueueManager.test.ts
@@ -123,14 +123,11 @@ describe("E2E: BlockQueueManager", function () {
             )
             .request();
 
-        await waitFor(
-            async () =>
-                await h
-                    .control(observer)
-                    .query.isBlacklisted(sender.address)
-                    .request(),
-            5000
-        );
+        await h.assert.sync.peerBlacklistedWait({
+            peerIndex: observer.index,
+            targetAddress: sender.address,
+            timeoutMs: 5000
+        });
         expect(
             await h
                 .control(observer)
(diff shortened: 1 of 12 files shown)


lukastanisic99 (approved):

lukastanisic99:
would still use a higher number than 100 for the poll like 500ms but ok

Issue #391, opened 2026-07-12
E2E flaky / chronic failure tracker


daiagi:
- **chronic** = failed both sample runs (21)
- **flaky** = failed exactly one of two runs (9)

### Dispute fraud-proof / state-proof
- [ ] `case1_inboundDivergence` — Case 1.1 → DisputeInvalidStateProof
- [ ] `case1_inboundDivergence` — Case 1.3 → DisputeInvalidStateProof
- [ ] `case1_inboundDivergence` — Case 1.4 → DisputeInvalidStateProof
- [ ] `case1_inboundDivergence` — Case 1.5 → DisputeInvalidStateProof
- [ ] `case5_milestoneBlockContent` — transactionCnt += 5 → DisputeInvalidBlockInStateProofApplyFraudProof
- [ ] `invalidStateProofAuditing` — [calldata posted] latestFinalizedStateStateMachineState = random → DisputeInvalidStateProof; proof author slashed
- [ ] `balanceInvariant` — locally-stored snapshot corrupted (sum ≠ totalBalance) → DisputeInvalidBalanceInvariant
- [ ] `timeout` — forged TimeoutTooEarly against a legitimate timeout dispute → proof author slashed
- [ ] `timeout` — timeout.blockHeight = block whose calldata is on-chain; isForced=true → TimeoutCalldataPosted
- [ ] `onChainSlashes` — onChainSlashes contains address not in latestStateSnapshot participants → InvalidDisputeReason

### Join / lifecycle / fork resolution
- [ ] `E2E-JoinChannelRaceConditions` — forceInboundJoin on disputed fork reverts
- [ ] `E2E-JoinChannelRaceConditions` — join on disputed fork reverts
- [ ] `E2E-JoinChannelRaceConditions` — pending joiner participates after dispute reduction
- [ ] `E2E-ForceJoinDispute` — force-join dispute after N turns of non-inclusion, then resolve with joiner PARTICIPATING

### Snapshot / spectate
- [ ] `E2E-StateSnapshots` — remove malicious participant after fork, then post updated snapshot on reduced fork — 2 independent snapshot updates
- [ ] `E2E-StateSnapshots` — creationTimestamp != 0 → on-chain snapshot updates but disputeWindowMap NOT cleared
- [ ] `E2E-MaliciousUpdateSnapshot` — colluded inflated stateMachineState balance → updateStateSnapshotSameFork succeeds, spectator aborts on balance invariant
[...]

PR #400, opened 2026-07-21
(test infra) parallel scheduler adapts to the machine

daiagi:
`test:parallel` sized itself off the account pool and total ram, not the machine:
concurrency capped at 40 (400 accounts / 10 per test) whatever the core count,
memory gated on our own rss vs total ram. on a box whose ram is half held by
browser/editor -> oversubscribe -> swap -> tests miss their wall-clock barriers.
green on roomier machines, so it read as flaky tests. #391

- gate admission on real os-available memory (`MemAvailable`), not just our rss
- `--workers` defaults from cpu cores instead of the account pool
- assign to an idle slot, not `slots[(seq - 1) % slotCount]`
- never boot more slots than the concurrency cap can use
- retry a failed task once when its event-loop peak shows starvation
- new `--mem-floor-gb`

my machine, full e2e suite: 14-26 failing -> 197/198, 3.8x speedup

### DIFF
--- a/scripts/e2e-parallel/adaptiveScheduling.js
+++ b/scripts/e2e-parallel/adaptiveScheduling.js
@@ -1,0 +1,81 @@
+/* eslint-disable no-console */
+// Host probes and scheduling policy: available memory, core-scaled concurrency,
+// slot picking, per-tick admission.
+const fs = require("fs");
+const os = require("os");
+const { HW_THREADS_PER_TEST } = require("./constants");
+
+const GIB = 1024 ** 3;
+
+// OS-available memory in GiB, counting every process on the box, not just ours.
+// Linux: MemAvailable (free + reclaimable cache). Elsewhere: os.freemem().
+function availableMemGb() {
+    if (process.platform === "linux") {
+        try {
+            const meminfo = fs.readFileSync("/proc/meminfo", "utf8");
+            const m = meminfo.match(/^MemAvailable:\s+(\d+)\s*kB/m);
+            if (m) return Number(m[1]) / 1024 / 1024;
+        } catch {
+            // fall through to os.freemem()
+        }
+    }
+    return os.freemem() / GIB;
+}
+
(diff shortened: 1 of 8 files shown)


lukastanisic99:
image
Now the concurrency is TOO low

daiagi:
>Now the concurrency is TOO low 

care to elaborate what you mean?
do you mean you don't like the binding of the slots to the workers or you don't like the default value?

any default value will be 
a.good for you and too high for me
b. good for me and too low for you

 It's just a default, you don't have to use it

lukastanisic99:
> > Now the concurrency is TOO low
> 
> care to elaborate what you mean? do you mean you don't like the binding of the slots to the workers or you don't like the default value?
> 
> any default value will be a.good for you and too high for me b. good for me and too low for you
> 
> It's just a default, you don't have to use it

My machine comfortably runs 13 concurrent tests and the new scheduler bounded it to a cap of 3 and signaled 'holding' while my cpu was 20% utilized. 

So if you're changing the scheduling logic to take into account more precisely the hardware of the machine, than it has to be more 'precise' e.g. `smarter` - both you and me can always set a hard cap with a flag, but then there is no point in even using a scheduling algorithm

daiagi:
fixed: does binary search to find the concurrency that keeps event-loop delay under 250ms.
no flag needed - on your machine I expect it to settle at 13 after ~8 finished tests (~2min
ramp), then holds.
use the `--workers` argument to skip this search and use the value you specify


Please try and let me know, run it without the `--workers` argument and let me know which concurrency value it settles on
for me it settles on 5, which is a bit higher then what I've manually found to work well on my machine (I usually use 3 or 4)

lukastanisic99:
image
after 2min
image


Concurrency 1.

Why don't you just add an option to override the default configuration and persist it locally, so you can set your own cpu/memory limit, worker count, scheduling interval - so next time you run the script it just uses your default configuration?

PR #401, opened 2026-07-23
fix(state-manager): handle stale reduceAndFinalize revert on the reduction path

sh3ll3x3c:
## Summary

On the dispute **reduction** path, `reduceAndFinalize` reverts with `ErrorDisputeInboundMessageBlocksInvalid` when the submitted reduction inputs have gone stale against the chain. That revert was unclassified, so it escaped **both** reduction-submit catch sites — the `simulateSubmission` staticCall and the detached `submitDetached` — as an **unhandled detached promise rejection**.

The revert is *not* a pure check over the submitter's own calldata. `reduce()` is a `view` over live storage, and the target hash is the channel's inbound head walked back to the dispute-window expiry (`lastEvidenceSubmissionTimestamp + evidenceTime`) — an expiry that moves whenever late evidence lands. So the usual cause is the chain moving between `compute()` and submission. `reduceAndFinalize` also only reaches this check while nothing is committed, so a non-zero committed result means another reducer won inside that window.

`classifyReductionRace` now distinguishes the two, and each call site applies its own policy:

- **committed result == our candidate** → `already-reduced`; we lost the race with a correct candidate, so install and converge.
- **anything else** → the candidate is stale:
  - **simulation path** (nothing installed yet) → reschedule a recompute against the moved head, bounded by 3 consecutive discards, after which the completion fails and aborts. Queued callers stand down until the backoff elapses instead of spending the budget on chain state that has not moved.
  - **detached path** (`complete()` already installed and settled, and the fork has moved) → **fail closed**. Nothing re-enters the executor for that source fork, so a candidate the chain never committed cannot be reconciled and the peer must not stay live on it.

## Test Plan

- `yarn tsc --noEmit` on `tsconfig.json` and `tsconfig.browser.json` — clean. Prettier and `git diff --check` clean.
- `yarn test:parallel` — full canonical gate.
- Reduction coverage in `test/e2e/E2E-ReductionManager.test.ts` and `test/unit/ReductionExecutor.test.ts`:
  - lost race with a matching committed result → install and converge;
  - stale with nothing committed → no install, rescheduled attempt completes the reduction;
  - stale through the whole retry budget → abort, budget released, no silent loop;
  - concurrently queued triggers → backoff held, run stays inside the budget;
  - stale on the detached submit → peer goes `OPENED` with the failure surfaced, not left `PARTICIPATING`.
- **Mutation check**: removing the reschedule makes the stale-with-nothing-committed test fail; restoring it passes.

### Coverage caveats

- The revert is **fault-injected** at the runner boundary (`call` for the simulation, `sendTransaction` for the submit). There is **no** E2E that produces it through a genuine on-chain race — a second actor moving the reduction inputs between a passing static call and a mined transaction. That staging needs a hold point between simulate and submit plus a controlled late-evidence write; it is not built here.
- The third classifier state — a **non-zero committed result that differs from our candidate** — is not staged. Reduction is deterministic, so producing it requires a deliberately divergent reducer, which the harness cannot currently stage. It is covered by the same `stale-candidate` branch as the zero case, but not independently asserted.

### DIFF
--- a/src/stateManager/reduction/ReductionExecutor.ts
+++ b/src/stateManager/reduction/ReductionExecutor.ts
@@ -52,8 +52,9 @@ const REDUCTION_RACE_ERRORS = [
     "RaceConditionReductionExpectationDoesntMatch",
     // Not a race per se: our reduce inputs went stale against the chain, so
     // reduceAndFinalize reverts the inbound-block check. Classified here so
-    // classifyReductionRace can gate it (swallow vs discard) instead of letting
-    // it escape the detached reduction path as an unhandled rejection.
+    // classifyReductionRace can tell converging on another reducer's result
+    // from a stale candidate, instead of letting the revert escape the detached
+    // reduction path as an unhandled rejection.
     "ErrorDisputeInboundMessageBlocksInvalid"
 ] as const satisfies readonly RaceConditionErrorName[];
 type ReductionRaceErrorName = (typeof REDUCTION_RACE_ERRORS)[number];
@@ -195,6 +196,15 @@ export default class ReductionExecutor {
                 attempts: pendingRetry.attempts,
                 notBefore: pendingRetry.notBefore
             });
+            // Re-arm rather than trusting the timer that scheduled this retry.
+            // Second-granularity time means the scheduled attempt can itself
+            // arrive a tick early and stand down here; without re-arming, that
+            // attempt would be the last one and the fork would strand.
+            this.stateManager.reductionManager.schedule(
+                forkId,
+                pendingRetry.notBefore,
+                true
+            );
             return;
         }
 
--- a/src/stateManager/reduction/ReductionExecutor.ts
+++ b/src/stateManager/reduction/ReductionExecutor.ts
@@ -430,6 +430,21 @@ export default class ReductionExecutor {
                         disputes,
                         "detached"
                     );
+                    // complete() already installed this candidate and settled
+                    // the completion, and the fork has moved on — so no later
+                    // attempt re-enters the executor for this source fork and
+                    // nothing can reconcile the install. A candidate the chain
+                    // did not commit therefore has to fail closed here.
+                    if (status === "stale-candidate") {
+                        this.logger.error(
+                            "Reduction submit rejected a candidate that is already installed locally - aborting",
+                            { forkId, candidateForkId }
+                        );
+                        this.stateManager.abort();
+                        throw new Error(
+                            `Installed reduction candidate ${candidateForkId} for fork ${forkId} was rejected on submit`
+                        );
+                    }
                     if (status) return;
                 }
                 throw tryDecodeCustomError(error) ?? error;
@@ -462,30 +477,12 @@ export default class ReductionExecutor {
                 forkId
             );
 
-        // ErrorDisputeInboundMessageBlocksInvalid compares our submitted
-        // inboundMessageBlocks against the target hash reduce() derives live
-        // from chain storage: the channel's inbound head walked back to the
-        // dispute-window expiry (lastEvidenceSubmissionTimestamp +
-        // evidenceTime). _verifyInboundMessageBlocks itself is pure, but that
-        // target is not — the expiry moves whenever late evidence lands. So the
-        // usual cause is the chain moving between compute() and this submission,
-        // and a recompute against the new head converges.
-        //
-        // reduceAndFinalize only reaches this check while nothing is committed
-        // (a committed result short-circuits earlier), so the committed value we
-        // read back is non-zero only when another reducer won inside this
-        // window. Matching our candidate means we merely lost the race with a
-        // correct candidate — safe to install and converge. Otherwise the
-        // candidate is stale (or nothing is committed yet): discard it rather
-        // than install a value that would later trip the reduced-result mismatch
-        // -> abort — but never rethrow into the detached path either.
-        //
-        // "Discard without installing" only holds on the simulation path.
-        // complete() installs the candidate before submitDetached runs, so on
-        // the detached path the stale genesis is already installed and the
-        // discard degrades to swallow-and-eventual-abort (the reduced-result
-        // mismatch in completeWithGenesis). Logged distinctly so the two are
-        // told apart.
+        // The revert means our inboundMessageBlocks no longer chain to the
+        // target hash reduce() derives from live chain storage, so our candidate
+        // is stale. Only a committed result equal to our candidate means we
+        // simply lost the race and can converge; anything else is a stale
+        // candidate, and each call site decides its own policy (retry before
+        // installing, fail closed once installed).
         if (errorName === "ErrorDisputeInboundMessageBlocksInvalid") {
             if (
                 String(reducedResult.reducedForkId) === String(candidateForkId)
--- a/src/stateManager/reduction/ReductionExecutor.ts
+++ b/src/stateManager/reduction/ReductionExecutor.ts
@@ -35,7 +35,12 @@ type ReductionSubmissionStatus = "submit" | "already-reduced" | "superseded";
 const REDUCTION_RACE_ERRORS = [
     "RaceConditionDisputeAlreadyReduced",
     "RaceConditionBlockHeightTooOld",
-    "RaceConditionReductionExpectationDoesntMatch"
+    "RaceConditionReductionExpectationDoesntMatch",
+    // Not a race per se: the losing peer's own reduce inputs went stale, so its
+    // reduceAndFinalize reverts the calldata-pure inbound-block check. Classified
+    // here so classifyReductionRace can gate it (swallow vs discard) instead of
+    // letting it escape the detached reduction path as an unhandled rejection.
+    "ErrorDisputeInboundMessageBlocksInvalid"
 ] as const satisfies readonly RaceConditionErrorName[];
 type ReductionRaceErrorName = (typeof REDUCTION_RACE_ERRORS)[number];
 
@@ -200,6 +205,7 @@ export default class ReductionExecutor {
         const submission = await this.prepareSubmission(candidate);
         const submissionStatus = await this.simulateSubmission(
             forkId,
+            candidate.reducedForkId,
             disputes,
             submission
         );
@@ -233,7 +239,12 @@ export default class ReductionExecutor {
                 }
             );
         if (installed && submissionStatus === "submit") {
-            this.submitDetached(forkId, disputes, submission);
+            this.submitDetached(
+                forkId,
+                candidate.reducedForkId,
+                disputes,
+                submission
+            );
         }
     }
 
@@ -318,6 +329,7 @@ export default class ReductionExecutor {
 
     private async simulateSubmission(
         forkId: ForkId,
+        candidateForkId: ForkId,
         disputes: DisputeStruct[],
         submission: { calldata: string[] }
     ): Promise<ReductionSubmissionStatus> {
@@ -332,6 +344,7 @@ export default class ReductionExecutor {
             const status = await this.classifyReductionRace(
                 custom?.name,
                 forkId,
+                candidateForkId,
                 disputes
             );
             if (status) return status;
@@ -341,6 +354,7 @@ export default class ReductionExecutor {
 
     private submitDetached(
         forkId: ForkId,
+        candidateForkId: ForkId,
         disputes: DisputeStruct[],
         submission: { calldata: string[] }
     ): void {
@@ -379,6 +393,7 @@ export default class ReductionExecutor {
                     const status = await this.classifyReductionRace(
                         raceErrorName,
                         forkId,
+                        candidateForkId,
                         disputes
                     );
                     if (status) return;
@@ -391,6 +406,7 @@ export default class ReductionExecutor {
     private async classifyReductionRace(
         errorName: string | undefined,
         forkId: ForkId,
+        candidateForkId: ForkId,
         disputes: DisputeStruct[]
     ): Promise<ReductionSubmissionStatus | undefined> {
         if (!isReductionRaceErrorName(errorName)) return undefined;
@@ -410,6 +426,43 @@ export default class ReductionExecutor {
                 this.stateManager.channelId,
                 forkId
             );
+
+        // ErrorDisputeInboundMessageBlocksInvalid is a pure check over THIS
+        // peer's own calldata (its reduce-derived output doesn't chain from its
+        // local snapshot), so it means our local reduction diverged from chain
+        // — not merely that another reducer won. Compare the committed result to
+        // our own candidate: only when they match did we simply lose the race
+        // with a correct candidate (safe to install and converge). Otherwise the
+        // candidate is divergent (or nothing is committed yet) — discard it
+        // rather than install a value that would later trip the reduced-result
+        // mismatch -> abort — but never rethrow into the detached path either.
+        if (errorName === "ErrorDisputeInboundMessageBlocksInvalid") {
+            if (
+                String(reducedResult.reducedForkId) === String(candidateForkId)
+            ) {
+                this.logger.info(
+                    "Ordinary reduction superseded by another reducer",
+                    {
+                        forkId,
+                        reducedForkId: reducedResult.reducedForkId,
+                        reducer: reducedResult.reducer
+                    }
+                );
+                return "already-reduced";
+            }
+            // Logged at warn so a genuine, persistent local divergence stays
+            // diagnosable instead of vanishing as a benign race.
+            this.logger.warn(
+                "Reduction inbound-blocks invalid; local candidate diverged from chain - discarding",
+                {
+                    forkId,
+                    candidateForkId,
+                    committedForkId: reducedResult.reducedForkId
+                }
+            );
+            return "superseded";
+        }
+
         const finalDispute = disputes.find(
             (dispute) =>
                 String(dispute.outputSnapshotDataHash) ===
--- a/src/stateManager/reduction/ReductionExecutor.ts
+++ b/src/stateManager/reduction/ReductionExecutor.ts
@@ -31,7 +31,14 @@ type LocalReductionCandidate = ReductionComputation & {
 };
 
 type ReductionCacheKey = string;
-type ReductionSubmissionStatus = "submit" | "already-reduced" | "superseded";
+type ReductionSubmissionStatus =
+    | "submit"
+    | "already-reduced"
+    | "superseded"
+    // Our own candidate went stale against the chain. Unlike "superseded" (a
+    // final dispute won and drives the fork itself) nothing else will settle
+    // this fork, so the attempt must be retried against the moved head.
+    | "stale-candidate";
 // Which call site classified the revert. "simulation" still owns the decision
 // to install; "detached" already installed the candidate before submitting.
 type ReductionSubmissionPath = "simulation" | "detached";
@@ -54,10 +61,20 @@ function isReductionRaceErrorName(
 }
 
 export default class ReductionExecutor {
+    // A stale candidate is retried against the moved chain head, but a
+    // divergence that survives this many recomputes is not transient — stop
+    // retrying and let the attempt fail loudly instead of looping in silence.
+    private static readonly MAX_STALE_CANDIDATE_RETRIES = 3;
     // Every ordinary trigger enters through tryReduce, including timers and
     // reschedules. Serialize the attempt itself, then release this mutex before
     // callers resume waiting on ReductionManager's shared completion promise.
     private readonly attemptMutex: Mutex;
+    // Consecutive stale-candidate discards per (channel, fork). Cleared once the
+    // fork's reduction installs, so only an unbroken run counts.
+    private readonly staleCandidateRetries = new Map<
+        ReductionCacheKey,
+        number
+    >();
     // Memoized `isKillPeriodExpired` per (channel, fork). A junk-fork flood then
     // costs O(1) chain reads per window instead of one per block: a "not expired
     // until killPeriodEnd" answer is reused until that time; an "expired" answer
@@ -80,6 +97,7 @@ export default class ReductionExecutor {
     }
 
     public dispose(): void {
+        this.staleCandidateRetries.clear();
         this.killPeriodExpiryCache.clear();
     }
 
@@ -108,7 +126,7 @@ export default class ReductionExecutor {
     public async isKillPeriodExpiredCached(
         forkId: ForkId
     ): Promise<KillPeriodObservation> {
-        const key = `${this.stateManager.channelId}:${forkId}`;
+        const key = this.getReductionCacheKey(forkId);
         const cached = this.killPeriodExpiryCache.get(key);
         if (
             cached &&
@@ -179,7 +197,7 @@ export default class ReductionExecutor {
             killPeriodEnd: Number(freshExpiry.killPeriodEnd)
         };
         this.killPeriodExpiryCache.set(
-            `${this.stateManager.channelId}:${forkId}`,
+            this.getReductionCacheKey(forkId),
             freshObservation
         );
         if (!freshObservation.windowExists) return;
@@ -212,8 +230,19 @@ export default class ReductionExecutor {
             disputes,
             submission
         );
+        if (submissionStatus === "stale-candidate") {
+            // Nothing was installed and no other actor settles this fork, so
+            // leaving now would strand the completion promise. The cause is the
+            // chain moving under us, so recompute against the new head after
+            // letting it settle. A divergence that outlives the retries is a
+            // real local fault: rethrow so the completion fails and aborts
+            // rather than looping forever.
+            this.retryStaleCandidate(forkId, candidate.reducedForkId);
+            return;
+        }
         if (submissionStatus === "superseded") return;
 
+        this.staleCandidateRetries.delete(this.getReductionCacheKey(forkId));
         await this.complete(
             forkId,
             candidate,
@@ -484,7 +513,7 @@ export default class ReductionExecutor {
                     path
                 }
             );
-            return "superseded";
+            return "stale-candidate";
         }
 
         const finalDispute = disputes.find(
@@ -500,5 +529,39 @@ export default class ReductionExecutor {
             disputer: finalDispute.input.disputer
         });
         return "superseded";
+    }
+
+    // Reschedule a recompute against the moved chain head, waiting for the write
+    // that moved it to settle. Throws once the run of consecutive discards is
+    // exhausted so ReductionManager fails the completion and aborts.
+    private retryStaleCandidate(forkId: ForkId, candidateForkId: ForkId): void {
+        const key = this.getReductionCacheKey(forkId);
+        const attempt = (this.staleCandidateRetries.get(key) ?? 0) + 1;
+        if (attempt > ReductionExecutor.MAX_STALE_CANDIDATE_RETRIES) {
+            this.staleCandidateRetries.delete(key);
+            throw new Error(
+                `Reduction candidate ${candidateForkId} for fork ${forkId} stayed stale across ${ReductionExecutor.MAX_STALE_CANDIDATE_RETRIES} recomputes`
+            );
+        }
+        this.staleCandidateRetries.set(key, attempt);
+        const delay = Math.max(
+            1,
+            this.stateManager.timeConfig.chainFallbackTime
+        );
+        this.logger.info("Rescheduling reduction after a stale candidate", {
+            forkId,
+            candidateForkId,
+            attempt,
+            delay
+        });
+        this.stateManager.reductionManager.schedule(
+            forkId,
+            Clock.getTimeInSeconds() + delay,
+            true
+        );
+    }
+
+    private getReductionCacheKey(forkId: ForkId): ReductionCacheKey {
+        return `${this.stateManager.channelId}:${forkId}`;
     }
 }
--- a/src/stateManager/reduction/ReductionExecutor.ts
+++ b/src/stateManager/reduction/ReductionExecutor.ts
@@ -31,6 +31,10 @@ type LocalReductionCandidate = ReductionComputation & {
 };
 
 type ReductionCacheKey = string;
+type StaleCandidateRetry = {
+    attempts: number;
+    notBefore: Timestamp;
+};
 type ReductionSubmissionStatus =
     | "submit"
     | "already-reduced"
@@ -69,11 +73,14 @@ export default class ReductionExecutor {
     // reschedules. Serialize the attempt itself, then release this mutex before
     // callers resume waiting on ReductionManager's shared completion promise.
     private readonly attemptMutex: Mutex;
-    // Consecutive stale-candidate discards per (channel, fork). Cleared once the
-    // fork's reduction installs, so only an unbroken run counts.
+    // Consecutive stale-candidate discards per (channel, fork), with the time
+    // the scheduled retry is due. Callers already queued on the mutex would
+    // otherwise each burn an attempt the moment the first one reschedules, so
+    // they stand down until `notBefore`. Cleared once the fork's reduction
+    // installs, so only an unbroken run counts.
     private readonly staleCandidateRetries = new Map<
         ReductionCacheKey,
-        number
+        StaleCandidateRetry
     >();
     // Memoized `isKillPeriodExpired` per (channel, fork). A junk-fork flood then
     // costs O(1) chain reads per window instead of one per block: a "not expired
@@ -166,6 +173,21 @@ export default class ReductionExecutor {
 
     private async tryReduceLocked(forkId: ForkId): Promise<void> {
         if (forkId !== this.stateManager.forkId) return;
+
+        // A stale-candidate retry is already scheduled and its backoff has not
+        // elapsed. Callers that were queued behind the mutex stand down rather
+        // than spending the remaining attempts on the same unmoved chain state.
+        const pendingRetry = this.staleCandidateRetries.get(
+            this.getReductionCacheKey(forkId)
+        );
+        if (pendingRetry && Clock.getTimeInSeconds() < pendingRetry.notBefore) {
+            this.logger.debug("Reduction retry not due yet; standing down", {
+                forkId,
+                attempts: pendingRetry.attempts,
+                notBefore: pendingRetry.notBefore
+            });
+            return;
+        }
 
         const { windowExists, isExpired, killPeriodEnd } =
             await this.isKillPeriodExpiredCached(forkId);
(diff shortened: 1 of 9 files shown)


lukastanisic99:
## 🧱 Fundamental

- 🟠 **[FO2] The discard branch never retries, but the failure it handles is usually transient — a fresh recompute would succeed.** The new comment claims the revert is ["a pure check over THIS peer's own calldata"](https://github.com/peer3to/state-channels-plus/blob/27f2877c4e708012852a8da2b714aa0df2a1e53c/src/stateManager/reduction/ReductionExecutor.ts#L430-L438), but [`reduce()` is a `view` that reads live chain storage](https://github.com/peer3to/state-channels-plus/blob/27f2877c4e708012852a8da2b714aa0df2a1e53c/contracts/V1/StateChannelDiamondProxy/DisputeVerificationFacet.sol#L61) — the verified target hash comes from the channel's on-chain inbound head [walked back to the dispute-window expiry](https://github.com/peer3to/state-channels-plus/blob/27f2877c4e708012852a8da2b714aa0df2a1e53c/contracts/V1/StateChannelDiamondProxy/DisputeVerificationFacet.sol#L106-L116), and that expiry moves whenever late (replacement) evidence bumps `lastEvidenceSubmissionTimestamp`. So the natural cause is chain movement between `compute()` and the submit simulation, and recomputing against the new head converges. Today the divergent/nothing-committed case returns `"superseded"` and [`tryReduceLocked` just returns](https://github.com/peer3to/state-channels-plus/blob/27f2877c4e708012852a8da2b714aa0df2a1e53c/src/stateManager/reduction/ReductionExecutor.ts#L212) — no reschedule, no install, and the fork-completion promise stays unsettled. Convergence then depends on some other trigger (another reducer's commit event, a later chain event); if this peer is the only live reducer, the fork strands with only a warn. Pre-PR it stranded too (with a crash log), so this is not a regression — but the fix is cheap.

  > **Fix FO2-FIX**
  >
  > On the discard branch (or at the `"superseded"` early-return in `tryReduceLocked` for this error name), call `this.stateManager.reductionManager.schedule(forkId, Clock.getTimeInSeconds() + a-small-backoff, true)` so the next attempt recomputes against the moved chain head. Also correct the comment: the check is pure *given its inputs*, but the target hash is chain-state-derived, which is exactly why a retry can succeed. Acceptance: a test that injects the error with **no** committed result and asserts a rescheduled attempt completes the reduction.

## 🏁 Race conditions

- 🟡 **[RY2] On the detached real-tx path, "discard without installing" is impossible — the install already happened.** [`complete()` installs the candidate before `submitDetached` runs](https://github.com/peer3to/state-channels-plus/blob/27f2877c4e708012852a8da2b714aa0df2a1e53c/src/stateManager/reduction/ReductionExecutor.ts#L241-L248), so if the real tx hits this error divergently (reachable only because chain state — e.g. late evidence — can move between the passing staticCall and the tx), the [swallow in the `.catch`](https://github.com/peer3to/state-channels-plus/blob/27f2877c4e708012852a8da2b714aa0df2a1e53c/src/stateManager/reduction/ReductionExecutor.ts#L392-L400) leaves a divergent genesis installed until the [reduced-result mismatch → `abort`](https://github.com/peer3to/state-channels-plus/blob/27f2877c4e708012852a8da2b714aa0df2a1e53c/src/stateManager/reduction/ReductionManager.ts#L227-L239) fires — the very outcome the classification comment says it prevents. The window is tiny and the pre-PR behavior (unhandled rejection *plus* the same installed state) was strictly worse, but the comment overpromises for this path.

  > **Fix RY2-FIX**
  >
  > Add one sentence to the block comment in `classifyReductionRace` stating that on the detached path the candidate is already installed, so the divergent case there degrades to swallow-and-eventual-abort; optionally log at `warn` with a distinct message when the divergent branch is reached from `submitDetached`. Acceptance: comment matches both call sites' actual semantics.
  
  - 🟠 **[TO2] The divergent/discard branch has zero coverage — both new tests exercise only the match branch.** The PR body admits this. The [e2e test](https://github.com/peer3to/state-channels-plus/blob/27f2877c4e708012852a8da2b714aa0df2a1e53c/test/e2e/E2E-ReductionManager.test.ts#L113-L163) and the [unit test](https://github.com/peer3to/state-channels-plus/blob/27f2877c4e708012852a8da2b714aa0df2a1e53c/test/unit/DisputeManager.test.ts#L11) both stage a winner commit first, so `committed == candidate` always holds and the [`"superseded"` discard return](https://github.com/peer3to/state-channels-plus/blob/27f2877c4e708012852a8da2b714aa0df2a1e53c/src/stateManager/reduction/ReductionExecutor.ts#L452-L463) — the branch that decides *not* to install — is never executed by any test. That's the half of the fix protecting against installing a wrong genesis.

  > **Fix TO2-FIX**
  >
  > Add a sibling e2e/unit case that injects `ErrorDisputeInboundMessageBlocksInvalid` with **no** winner released (committed result still `ZeroHash`), and assert: no install (`query.getCompletedReductionForkId` returns `null`), the warn log fires, `quiesceHosts()` returns no errors, and `getFirstDetachedError()` is `undefined`. This also becomes the acceptance test for FO2-FIX's reschedule.

- 🟡 **[TY3] The detached `.catch` path is not independently faulted (PR admits) and the mutation-check is a PR-body claim I did not re-run.** Both paths do route through the same `classifyReductionRace`, so simulation-side coverage is meaningful; the mutation-check ("remove the classification → both tests fail red") is plausible but unverified in this review.

  > **Fix TY3-FIX**
  >
  > Extend `stubNextReductionSimulationError`'s stub (or a sibling `stubNextReductionSubmitError`) to fault `runner.sendTransaction`/the tx path instead of `runner.call`, and run the same scenario through `submitDetached`. Acceptance: the detached `.catch` classify site executes under test.
  
## ♻️ Code reuse

- 🟡 **[CY2] The "winner commits → wait for its `onDisputeReducedResultCommitted` → inject error → release loser" staging sequence is duplicated inline in both new tests.** [e2e L117–L149](https://github.com/peer3to/state-channels-plus/blob/27f2877c4e708012852a8da2b714aa0df2a1e53c/test/e2e/E2E-ReductionManager.test.ts#L117-L149) and [unit L57–L95](https://github.com/peer3to/state-channels-plus/blob/27f2877c4e708012852a8da2b714aa0df2a1e53c/test/unit/DisputeManager.test.ts#L57-L95) — `test/AGENTS.md` says staging used by more than one test belongs in the harness (client-side combos on `rpcStub`).

  > **Fix CY2-FIX**
  >
  > Extract an `rpcStub` combo (e.g. `loseReductionRaceWithSimulationError(loserIndex, winnerIndex, errorName)`) next to the existing `releaseReductionWithSimulationError`, and call it from both tests. Acceptance: neither test file contains the inline sequence.
  
  ## 📏 AGENTS.md adherence
  - 🟡 **[AY2] The classification comment states something the contract contradicts.** "[A] pure check over THIS peer's own calldata" ([L430](https://github.com/peer3to/state-channels-plus/blob/27f2877c4e708012852a8da2b714aa0df2a1e53c/src/stateManager/reduction/ReductionExecutor.ts#L430-L438), same wording at [L39–L43](https://github.com/peer3to/state-channels-plus/blob/27f2877c4e708012852a8da2b714aa0df2a1e53c/src/stateManager/reduction/ReductionExecutor.ts#L39-L43)) — [`_verifyInboundMessageBlocks`](https://github.com/peer3to/state-channels-plus/blob/27f2877c4e708012852a8da2b714aa0df2a1e53c/contracts/V1/StateChannelDiamondProxy/StateChannelCommon.sol#L460-L479) is pure, but its target hash is chain-storage-derived (see FO2). The conclusion drawn ("our local reduction diverged from chain") is still correct; the premise as written will mislead the next reader about transiency and the detached path.

  > **Fix AY2-FIX**
  >
  > Reword both comments per FO2-FIX: the check compares our calldata against the chain's inbound head as of the (movable) dispute-window expiry. Acceptance: comments no longer claim calldata-purity.

## 🧹 Miscellaneous (scope)

- 🟠 **[MO1] Guaranteed add/add merge conflict with PR #394, which also creates `test/unit/DisputeManager.test.ts`.** #394's copy contains the original flaky test this PR ports, [still present as `it.skip` at its L422](https://github.com/peer3to/state-channels-plus/blob/d56ea842207934fcdf9dc8f9a825929f647190b5/test/unit/DisputeManager.test.ts#L422) — when both land, the files conflict and the repro exists twice (one skipped, one live, with different titles).

  > **Fix MO1-FIX**
  >
  > Decide the owner now: either rebase #394 onto this branch and delete its `it.skip` copy, or land #394 first and turn this PR's file-add into an insertion into that suite. Acceptance: after both merge, exactly one `DisputeManager.test.ts` with the ported test live and no `it.skip` twin.

- 🟡 **[MY2] An unrelated docs commit rides along.** [`27f2877c`](https://github.com/peer3to/state-channels-plus/commit/27f2877c4e708012852a8da2b714aa0df2a1e53c) imports `test/AGENTS.md` into `CLAUDE.md` — harmless and useful, but it's not part of the fix and will conflict if any sibling branch does the same.

  > **Fix MY2-FIX**
  >
  > Leave it if merging soon; otherwise cherry-pick it onto `general-test-fixes-2` so every stacked branch inherits it once.

## ❓ Open questions

- 🟠 **[OO1] Should the divergent/nothing-committed discard schedule a retry?** Recommended: yes (FO2-FIX) — the natural cause is transient chain movement, a recompute converges, and the retry is bounded because each attempt recomputes from the current head. If you instead decide the existing event triggers are guaranteed to re-fire `tryReduce` in every staleness scenario, document that invariant at the discard site.

  > **Fix OO1-FIX**
  >
  > Decide retry vs. documented-trigger-invariant. If retry: implement FO2-FIX. If not: add a comment naming the event(s) guaranteed to re-trigger reduction after a discard.

lukastanisic99:
Changed the target/base to `dispute` since I merged the rest

sh3ll3x3c:
Thanks — all of it held up under checking. Rebased onto the latest `dispute` (`3722634b`) and addressed every finding, one commit each.

| Finding | Commit | What changed |
|---|---|---|
| **AY2** | `7c90da7e` | Comments no longer claim calldata-purity |
| **RY2** | `e183713b` | Call site threaded into `classifyReductionRace`; detached case logged distinctly |
| **FO2 / OO1** | `398a6b4a` | Discard reschedules a recompute against the moved head |
| **MO1** | `e74f737f` | `test/unit/DisputeManager.test.ts` → `test/unit/ReductionExecutor.test.ts` |
| **CY2** | `4c60152f` | `loseReductionRaceWithSimulationError` extracted into `rpcStubActions` |
| **TO2** | `0110289f` | New e2e case with nothing committed |
| **TY3** | `bc6db275` | New `stubNextReductionSubmitError` faulting `runner.sendTransaction` |
| **MY2** | — | No change; see below |

## On FO2 — you were right, and it's a bit worse than the writeup

Confirmed `reduce()` is a `view` over live storage and the target hash is the inbound head walked back to `lastEvidenceSubmissionTimestamp + evidenceTime`, so the revert really is chain movement rather than a lost race.

One thing that sharpens it: `reduceAndFinalize` early-returns whenever `reducedResult.forkId != 0` (`DisputeVerificationFacet.sol:250-256`), so `ErrorDisputeInboundMessageBlocksInvalid` is only reachable while **nothing is committed**. The follow-up `getReducedResult` can only read a non-zero value if another reducer commits inside the gap between our revert and that read. So the "match → already-reduced" branch is a narrow coincidence and the **discard branch is the production path** — which is exactly the branch that had no coverage and no retry.

Also worth noting the pre-PR behaviour on the simulation path was `throw` → `failCompletion` → `abort()`. So as it stood this PR turned fail-loud into fail-silent, not "stranded either way".

## Two deviations from the proposed fixes

**1. The retry is bounded.** FO2-FIX as written is an unbounded reschedule, and `fcb5e307` ("make failed event-sync logs fatal, no retry") just moved the event-sync pipeline the other way. So: at most 3 consecutive stale-candidate discards, then throw → `failCompletion` → `abort()`. Transient case retries, a real divergence fails loudly, neither strands silently. Happy to drop the cap if you'd rather match FO2-FIX exactly.

**2. MO1 resolved by renaming rather than coordinating.** Both options you listed depend on merge order. The test covers `ReductionExecutor`'s submission classification, not dispute construction, so naming it for its subject removes the add/add collision whichever PR lands first. #394's `it.skip` at its L422 covers something genuinely different (dispute posts with calldata) and is still #394's to remove — flagging so it isn't assumed handled here.

**MY2** needs no action: `dispute`'s `CLAUDE.md` still lacks the import, so it rebased clean and the predicted conflict can't occur.

## Verification

- `tsc --noEmit` on `tsconfig.json` and `tsconfig.browser.json`, prettier clean
- `yarn test:parallel --e2e-only --grep ReductionManager` — 7 passing, incl. both new cases
- `yarn test:parallel --grep ReductionExecutor` — 1 passing (moved unit test)
- Mutation check: removing the `retryStaleCandidate` call makes the new discard test fail; restoring it passes — so it covers the fix rather than passing incidentally

One leftover: the original commit `fe136253` still carries the pre-review claim that the revert is "a pure check over the peer's own calldata". The code comments are corrected in `7c90da7e`; I left the commit message alone rather than rewrite reviewed history. Say the word if you'd prefer it amended.

lukastanisic99:
@sh3ll3x3c 

## 🧱 Fundamental

- 🔴 **[FR1] A stale detached submission is swallowed after the candidate has already become local canonical state.** [`complete()` installs and settles the reduction before calling `submitDetached`](https://github.com/peer3to/state-channels-plus/blob/bc6db2756c01f39d58137ff1ccf6fbe288348404/src/stateManager/reduction/ReductionExecutor.ts#L255-L280). If the transaction then gets `ErrorDisputeInboundMessageBlocksInvalid` with zero or a different committed result, [`classifyReductionRace` returns `"stale-candidate"`](https://github.com/peer3to/state-channels-plus/blob/bc6db2756c01f39d58137ff1ccf6fbe288348404/src/stateManager/reduction/ReductionExecutor.ts#L489-L516) and the detached catch [silently returns for every truthy status](https://github.com/peer3to/state-channels-plus/blob/bc6db2756c01f39d58137ff1ccf6fbe288348404/src/stateManager/reduction/ReductionExecutor.ts#L410-L435). The retry is only handled on the pre-install simulation path ([L233–L241](https://github.com/peer3to/state-channels-plus/blob/bc6db2756c01f39d58137ff1ccf6fbe288348404/src/stateManager/reduction/ReductionExecutor.ts#L233-L241)). The claimed “eventual abort” cannot happen: the completion was already resolved with the installed fork, and later events for the old source fork hit [`tryReduce`'s old-fork return](https://github.com/peer3to/state-channels-plus/blob/bc6db2756c01f39d58137ff1ccf6fbe288348404/src/stateManager/reduction/ReductionManager.ts#L119-L130), never the mismatch check in [`completeWithGenesis`](https://github.com/peer3to/state-channels-plus/blob/bc6db2756c01f39d58137ff1ccf6fbe288348404/src/stateManager/reduction/ReductionManager.ts#L227-L239). With no winner, there is not even a later reduced-result event. The peer stays `PARTICIPATING` on a fork that was never committed on-chain. This is the underlying RY2 issue from timeline row 3, not a comment-only discrepancy.

  > **Fix FR1-FIX**
  >
  > Do not treat a detached inbound-blocks error as success unless the chain's committed `reducedForkId` exactly equals the installed candidate. For zero or a different result, abort/fail immediately because the local install cannot be rolled back; the stronger design is to defer local installation until the transaction receipt or a matching committed result. Acceptance: the peer never remains `PARTICIPATING` with a locally completed reduction that has no matching on-chain result.

- 🟠 **[SO1] Another chain participant can drive FR1 through ordinary late evidence timing.** Solidity derives the reduction target from live storage, including [`lastEvidenceSubmissionTimestamp + evidenceTime` and the inbound head at that expiry](https://github.com/peer3to/state-channels-plus/blob/bc6db2756c01f39d58137ff1ccf6fbe288348404/contracts/V1/StateChannelDiamondProxy/DisputeVerificationFacet.sol#L61-L115). A participant that moves those inputs after the static call but before the transaction executes can make the detached transaction fail its [inbound-block verification](https://github.com/peer3to/state-channels-plus/blob/bc6db2756c01f39d58137ff1ccf6fbe288348404/contracts/V1/StateChannelDiamondProxy/DisputeVerificationFacet.sol#L305-L313). The local peer must not remain live on the rejected state after adversarially controllable chain timing.

  > **Fix SO1-FIX**
  >
  > Apply FR1-FIX and add a real-chain race test where a second actor changes the relevant chain state between simulation and mined submission; assert the submitting peer either converges to the committed result or aborts.

## 🏁 Race conditions

- 🟠 **[RO1] Already-queued `tryReduce` callers bypass the retry backoff and can consume all three stale retries immediately.** Each caller checks `completion.settled` before waiting for the executor ([`ReductionManager.tryReduce`](https://github.com/peer3to/state-channels-plus/blob/bc6db2756c01f39d58137ff1ccf6fbe288348404/src/stateManager/reduction/ReductionManager.ts#L119-L148)); the executor mutex only serializes them ([L156–L165](https://github.com/peer3to/state-channels-plus/blob/bc6db2756c01f39d58137ff1ccf6fbe288348404/src/stateManager/reduction/ReductionExecutor.ts#L156-L165)). After the first attempt schedules a delayed retry, callers already queued behind the mutex enter immediately because the completion is still unsettled. Timer, event, and queue-recovery entry points can all call the same operation ([timer L98–L105](https://github.com/peer3to/state-channels-plus/blob/bc6db2756c01f39d58137ff1ccf6fbe288348404/src/stateManager/reduction/ReductionManager.ts#L98-L105), [event L155–L163](https://github.com/peer3to/state-channels-plus/blob/bc6db2756c01f39d58137ff1ccf6fbe288348404/src/eventHandlers/EventHandler.ts#L155-L163), [reduced event L715–L731](https://github.com/peer3to/state-channels-plus/blob/bc6db2756c01f39d58137ff1ccf6fbe288348404/src/eventHandlers/EventHandler.ts#L715-L731), [queue recovery L206–L229](https://github.com/peer3to/state-channels-plus/blob/bc6db2756c01f39d58137ff1ccf6fbe288348404/src/stateManager/BlockQueueManager.ts#L206-L229)). A burst can therefore turn a transient chain race into an immediate abort instead of waiting `chainFallbackTime`.

  > **Fix RO1-FIX**
  >
  > Coalesce attempts per fork or store a `retryNotBefore` deadline checked inside the mutex. Queued callers before that deadline must join the pending completion without incrementing the stale counter. Add a deterministic concurrent-trigger test that proves the attempts remain separated by the configured delay.

## ⚡ Performance

- 🟡 **[PY1] Stale retry counters accumulate for superseded forks until the whole executor is disposed.** The map is [cleared globally only in `dispose`](https://github.com/peer3to/state-channels-plus/blob/bc6db2756c01f39d58137ff1ccf6fbe288348404/src/stateManager/reduction/ReductionExecutor.ts#L99-L102), and an individual key is deleted only on retry exhaustion or a normal non-stale completion ([L233–L245](https://github.com/peer3to/state-channels-plus/blob/bc6db2756c01f39d58137ff1ccf6fbe288348404/src/stateManager/reduction/ReductionExecutor.ts#L233-L245), [L537–L546](https://github.com/peer3to/state-channels-plus/blob/bc6db2756c01f39d58137ff1ccf6fbe288348404/src/stateManager/reduction/ReductionExecutor.ts#L537-L546)). If a fork changes after one stale attempt, later work returns on the old-fork guard and leaves that `(channel, fork)` key behind.

  > **Fix PY1-FIX**
  >
  > Delete the retry entry from the canonical fork-completion/failure/supersession cleanup path, not only from the executor's happy path. Add a probe that makes one attempt stale, changes forks, and verifies the old counter is gone.

- 🟢 **[TG1] All executed verification is green.** `yarn compile` passed; `yarn tsc --noEmit -p tsconfig.json` and `yarn tsc --noEmit -p tsconfig.browser.json` passed; `git diff --check` passed. Focused `ReductionExecutor` passed 1/1 (`logs/run-472`) and focused E2E `ReductionManager` passed 7/7 (`logs/run-473`). The canonical gate passed 591/591 in 883.52 s (`logs/run-474`), with no worker-process failures or failing-test logs.

  > **Fix TG1-FIX**
  >
  > No change required; keep these checks green after the blockers are fixed.

- 🔴 **[TR1] The detached-path regression test codifies FR1 as the expected success state.** The test injects the error before any transaction reaches the chain, then asserts [local reduction completion, `PARTICIPATING`, and zero host/detached errors](https://github.com/peer3to/state-channels-plus/blob/bc6db2756c01f39d58137ff1ccf6fbe288348404/test/e2e/E2E-ReductionManager.test.ts#L204-L226). That proves only that the catch swallowed the rejection. It also proves the local candidate remained installed even though no matching transaction committed it.

  > **Fix TR1-FIX**
  >
  > Replace the success assertions with the safe invariant from FR1-FIX: either a matching on-chain reduced result and local participation, or `OPENED`/failed completion after a zero or different result. Never accept `PARTICIPATING` solely because the detached error disappeared.

- 🟠 **[TO2] No E2E test produces the stale detached revert through the real contract/provider boundary.** [`stubNextReductionSubmitError` throws from `runner.sendTransaction` before forwarding](https://github.com/peer3to/state-channels-plus/blob/bc6db2756c01f39d58137ff1ccf6fbe288348404/test/fixtures/customRpc/harnessControl/services/stub/StubRpcMethods.ts#L1084-L1113), so it does not test a mined revert, `tx.wait()`, receipt recovery, or the live chain-state movement that causes this error.

  > **Fix TO2-FIX**
  >
  > Add a deterministic E2E race that lets the static call pass, moves the contract's reduction inputs from a second peer, sends the real transaction, and checks the final local/on-chain invariant plus detached-error handling.

- 🟠 **[TO3] The bounded retry policy has no boundary, reset, or concurrency coverage.** The production path permits [three stale recomputes and throws on the fourth](https://github.com/peer3to/state-channels-plus/blob/bc6db2756c01f39d58137ff1ccf6fbe288348404/src/stateManager/reduction/ReductionExecutor.ts#L534-L561), but the only retry test injects [one failure followed by a successful attempt](https://github.com/peer3to/state-channels-plus/blob/bc6db2756c01f39d58137ff1ccf6fbe288348404/test/e2e/E2E-ReductionManager.test.ts#L158-L202). It cannot catch an off-by-one cap, a counter that survives success, or RO1's queued-trigger bypass.

  > **Fix TO3-FIX**
  >
  > Add black-box unit cases for attempts 1–3, the fourth-attempt abort, reset after success, fork cleanup, and multiple concurrent triggers respecting the backoff.

- 🟠 **[TO4] The non-zero committed result that differs from the candidate is untested.** The current match case stages a winner with the same deterministic candidate ([E2E L113–L156](https://github.com/peer3to/state-channels-plus/blob/bc6db2756c01f39d58137ff1ccf6fbe288348404/test/e2e/E2E-ReductionManager.test.ts#L113-L156)), while the discard/retry case leaves the result at zero ([L158–L202](https://github.com/peer3to/state-channels-plus/blob/bc6db2756c01f39d58137ff1ccf6fbe288348404/test/e2e/E2E-ReductionManager.test.ts#L158-L202)). The third classifier outcome can therefore change without a focused failure.

  > **Fix TO4-FIX**
  >
  > Add a black-box unit case with a non-zero committed fork different from the candidate and assert the exact fail/retry policy chosen for that state.

## ♻️ Code reuse

- 🟠 **[CO1] The simulation and submit fault injectors duplicate the same one-shot runner patch lifecycle.** [`stubNextReductionSimulationError` plus restore](https://github.com/peer3to/state-channels-plus/blob/bc6db2756c01f39d58137ff1ccf6fbe288348404/test/fixtures/customRpc/harnessControl/services/stub/StubRpcMethods.ts#L1034-L1075) and [`stubNextReductionSubmitError` plus restore](https://github.com/peer3to/state-channels-plus/blob/bc6db2756c01f39d58137ff1ccf6fbe288348404/test/fixtures/customRpc/harnessControl/services/stub/StubRpcMethods.ts#L1078-L1126) separately capture, filter by selector, self-restore, and synthesize the same custom error. The next change to matching or restoration can drift between them.

  > **Fix CO1-FIX**
  >
  > Put one typed one-shot runner-patch helper on `StubService`, parameterized by `call` versus `sendTransaction` and its `StubKey`. Keep the routable `StubRpcMethods` functions as thin endpoints and reuse the helper for install and restore.

## 📏 AGENTS.md adherence

- 🟠 **[AO1] The `src/` behavior change does not have the required exhaustive black-box unit matrix.** The added unit suite contains [one correct-candidate lost-race scenario](https://github.com/peer3to/state-channels-plus/blob/bc6db2756c01f39d58137ff1ccf6fbe288348404/test/unit/ReductionExecutor.test.ts#L5-L109). It omits the retry boundary, exhaustion, recovery/reset, non-zero mismatch, detached stale state, and concurrent interleavings required by the repository's “Testing changes to `src/`” rule.

  > **Fix AO1-FIX**
  >
  > Add the black-box unit matrix in TO3-FIX and TO4-FIX, including FR1's detached-state invariant.

- 🟠 **[AO2] The E2E layer does not cross the materially affected external boundary.** Both new error paths are injected by patching the ethers runner; neither makes the contract reject a real reduction after chain state changes. That conflicts with the repository requirement that E2E cover affected system interactions and with `test/AGENTS.md`'s rule that fixtures trigger source behavior rather than reproduce the asserted side effect.

  > **Fix AO2-FIX**
  >
  > Add TO2-FIX's real-contract race and use fault injection only for internal classifier permutations that are impractical to stage on-chain.

- 🟡 **[AY3] The classifier has a long strategy essay whose key detached-path conclusion is false.** The [24-line block](https://github.com/peer3to/state-channels-plus/blob/bc6db2756c01f39d58137ff1ccf6fbe288348404/src/stateManager/reduction/ReductionExecutor.ts#L465-L488) says a detached stale install degrades to “swallow-and-eventual-abort,” but FR1 shows no code path reaches that abort. It also violates the repository rule to keep comments simple and to the point.

  > **Fix AY3-FIX**
  >
  > Fix the behavior first, then replace the essay with a short comment stating the invariant enforced at that call site.

## 🧹 Miscellaneous (scope)

- 🟡 **[MY2] The unrelated Claude guidance commit remains in the PR.** [`CLAUDE.md` imports `test/AGENTS.md`](https://github.com/peer3to/state-channels-plus/blob/bc6db2756c01f39d58137ff1ccf6fbe288348404/CLAUDE.md#L8-L13). It is useful, but it is not part of stale reduction handling and still expands this fix's merge surface.

  > **Fix MY2-FIX**
  >
  > Move the docs-only change to the base or a separate PR, or explicitly accept the scope expansion before landing.

- 🟡 **[MY3] The PR body describes the old branch and test state.** It still says the PR is stacked on `general-test-fixes-2`, calls the reduction target a pure check over the peer's calldata, and says the divergent and detached branches are untested. GitHub now reports `dispute` as the base, and this head added both injected cases.

  > **Fix MY3-FIX**
  >
  > Update the [PR description](https://github.com/peer3to/state-channels-plus/pull/401) to match the current base, chain-derived target, retry behavior, and actual coverage. Keep the real-contract caveat from TO2.

## ❓ Open questions

- 🟢 **[OG1] No design question needs to block the concrete fixes above.** The required invariant is straightforward: local participation must correspond to a matching committed reduction, or the peer must fail closed.

  > **Fix OG1-FIX**
  >
  > No decision required.

sh3ll3x3c:
**FR1 is right, and it's the finding I created last round.** RY2 said the detached path leaves a stale candidate installed; I responded by writing a comment claiming it "degrades to swallow-and-eventual-abort" instead of fixing it — and that claim is false. `unsafeSetLatestState` sets `this.forkId` to the reduced fork, so `tryReduce(sourceForkId)` early-returns at the old-fork guard and `completeWithGenesis`'s mismatch never re-runs. There is no later abort. The peer stayed `PARTICIPATING` on a fork the chain never committed, and TR1 is correct that my test asserted exactly that as success.

Fixed properly this time: the detached path fails closed. Rebased onto `dispute`, one commit per finding.

| Finding | Commit | What changed |
|---|---|---|
| **FR1 / SO1** | `087df091` | Detached submit aborts unless the committed result equals the installed candidate |
| **RO1** | `89fd0ff0` | Retry deadline recorded; queued callers stand down instead of burning the budget |
| **PY1** | `fc940e0d` | Retry entry dropped when the fork moves on |
| **TR1** | `59690dc7` | Test asserts `OPENED` + surfaced failure, not the swallow |
| **AY3** | `087df091` | 24-line essay → 6 lines stating the invariant |
| **CO1** | `bd8c4f76` | One `StubService` owner for the one-shot multicall fault; endpoints are thin |
| **TO3 / AO1** | `e5c7fa93` | Budget exhaustion, budget release, backoff vs concurrent triggers |
| **MY2** | — | Docs commit dropped; the diff is source + tests only |
| **MY3** | — | PR body rewritten |

Two of your findings sharpened things I'd have missed, so thank you for pushing on RY2 twice.

## Two findings deferred, not silently dropped

- **TO2 / AO2** — you're right that nothing produces this revert through the real contract boundary. Staging it needs a hold point between the passing static call and the mined transaction, plus a controlled late-evidence write from a second peer to move the inbound head. That's new harness machinery and I'd rather not ship a flaky approximation. Recorded as a caveat in the PR body.
- **TO4** — a non-zero committed result differing from our candidate needs a deliberately divergent reducer. Reduction is deterministic, so it can't arise naturally and the harness can't currently stage one. It shares the `stale-candidate` branch with the zero case but is not independently asserted.

If either should block the merge rather than land as a caveat, say so and I'll build the staging.

## Self-review caught two bugs in the fixes themselves

- The RO1 stand-down could **strand the fork**: second-granularity time lets the scheduled retry arrive a tick early and stand down on its own deadline, leaving nothing armed. It now re-arms (`f3070c2a`).
- The retry probe rebuilt the executor's cache-key format. A change there would make it read nothing while the new assertions passed vacuously — it now reaches the owner's key builder (`52460d99`).

## Verification

- `yarn tsc --noEmit` on both configs, prettier, `git diff --check` — clean
- `yarn test:parallel` — **593 passing, 0 failing** (680 s) at `f3070c2a`
- `--grep Reduction` re-run at final HEAD `52460d99` (the probe commit is harness-only) — **15 passing**
- Mutation check: removing the reschedule fails the stale-with-nothing-committed test; restoring it passes

sh3ll3x3c:
@lukastanisic99

sh3ll3x3c:
@lukastanisic99 — I went back and tried to actually reproduce the scenario behind **TO2 / AO2** on a local Hardhat devnet instead of leaving it as a caveat. Two probes, both negative, and the second one changes what this PR is.

## Probe 1 — stale local inbound view

Held `InboundMessagesProcessed` on one peer (new stub mirroring the existing `stubHold*Events` idiom), landed a join so the chain head moved, then drove that peer to reduce.

Staging worked — the peer's inbound head diverged from the other three. But it never reached the reduce path. It died earlier, at `DisputeValidationService.ts:584`:

```
Verify Dispute Output - sanity check - is data linked - failed
```

Dispute construction has its own linkage guard that fires first. `ErrorDisputeInboundMessageBlocksInvalid` appeared **0 times in 45,510 log lines**; the peer stayed `PARTICIPATING` and the retry budget was never touched.

## Probe 2 — genuine on-chain race

Built the hold point I claimed we lacked: park the reduction *after* its `multicall` staticCall returns successfully, land an inbound message block, then release so the real transaction mines against moved state. No fault injection anywhere — any revert would come from the contract.

The join transaction itself reverted during gas estimation with `0x807898a4`, which decodes to:

```
RaceConditionForceInboundJoinForkDisputed()
```

`JoinChannelFacet.sol:60-64` blocks a new participant from joining while the fork is disputed. Joins are the only caller of `_appendInboundMessages` (`StateChannelManagerProxy.sol:329`), which is the only writer of `channelBalances.latestInboundMessageBlockHash`. The inbound head cannot move during a dispute.

## The revert is unreachable, not just untested

Three independent locks on `reduce()`'s inbound-head derivation, all holding during a reduction:

1. **Inbound head frozen** — `RaceConditionForceInboundJoinForkDisputed` blocks the only writer. Confirmed on chain above.
2. **Window expiry frozen** — new evidence requires `block.timestamp < creationTimestamp + evidenceTime`, reduction requires `block.timestamp >= lastEvidenceSubmissionTimestamp + evidenceTime`, and `lastEvidenceSubmissionTimestamp >= creationTimestamp` (`DisputeUtils.sol:90-98`). Mutually exclusive — late evidence cannot land once reduction is legal.
3. **Belt and braces** — even if a block did land, `_appendInboundMessages` stamps `block.timestamp` (`StateChannelCommon.sol:250`), which is already past `killPeriodEnd`, so `reduce()`'s walk-back at `DisputeVerificationFacet.sol:109-112` skips it.

Note also that `areDisputesCommitted` compares the commitment set by exact length (`DisputeUtils.sol:113-116`), so even a hypothetical late evidence submission would revert `ErrorDisputeCommitmentNotAvailable` at `DisputeVerificationFacet.sol:263` — before `reduce()` runs at all.

## What this means for the PR

Being straight about it: **this PR hardens a revert that nothing can currently trigger.**

- The PR body's causal story — "the target hash moves whenever late evidence lands, so the usual cause is the chain moving between `compute()` and submission" — is **wrong**. I'll rewrite it.
- **TO2 / AO2 and TO4 are resolved rather than deferred.** There's no missing harness machinery; the scenarios are structurally impossible. TO4's divergent-committed-result case needs a non-deterministic reducer, which reduction being deterministic rules out.
- The classification itself is still defensible as defense-in-depth: an unclassified revert escapes both catch sites as an unhandled detached rejection, and a future contract change (e.g. relaxing the join guard, or top-ups appending inbound blocks) would make it reachable.

Open question for you, since it's a judgement call: the retry machinery — 3-attempt budget, `notBefore` backoff, the re-arm, the fail-closed abort on the detached path — is a lot of surface for a path nothing can enter. Options as I see them:

1. Keep as-is (defensive, but carries untriggerable code).
2. Keep the classification, drop the retry/backoff, treat the simulation path as fail-closed too.
3. Drop `ErrorDisputeInboundMessageBlocksInvalid` from `REDUCTION_RACE_ERRORS` entirely and let it surface as a hard error if it ever fires.

I lean toward 2 — it keeps the revert from escaping as an unhandled rejection without pretending we can recover from a state the chain won't produce. Happy to go with your call.

Both probes were scratch and are reverted; the branch is unchanged pending this decision.

sh3ll3x3c:
@lukastanisic99

sh3ll3x3c:
**Correction to my comment above** ([#issuecomment-5084808388](https://github.com/peer3to/state-channels-plus/pull/401#issuecomment-5084808388)) — the conclusion was wrong.

I claimed `ErrorDisputeInboundMessageBlocksInvalid` was structurally
unreachable, and that TO2/AO2 and TO4 were resolved rather than deferred. That
does not survive the evidence: a full parallel run on 2026-07-22 produced the
revert with no fault injection, rejecting two honest peers and aborting three of
four. My three "locks" argument was reasoning about the contract in isolation
without checking it against a log I already had.

The follow-up is stack #412 — #411 makes the error carry what it compared, #406
logs it. Whether the cause is a moved chain target or local divergence is still
unresolved; the point of that work is that the next occurrence will say which.

PR #402, opened 2026-07-23
(fix) departed member can abort a spectator via a stale membership snapshot

daiagi:
a participant that leaves the channel stays connected to the p2p network, and
state snapshots are keyed by hash and never pruned (`StateSnapshotStorage`) -> a
departed member can author a block whose `stateSnapshotHash` points at a snapshot
from while it was still a member.

`ValidationService`'s author gate unions the previous snapshot with the block's
declared resulting snapshot, so it re-admits the ex-member as a participant. the
block then reaches the leader check, fails `nextLeader !== author`, and
`SpectatingValidationStrategy` treats it as provable participant fraud ->
`abort()` -> the spectator goes offline (SYNCED -> OPENED).

fix binds the declared snapshot to the block's own coordinates (forkId + height),

### DIFF
--- a/src/models/StateSnapshot.ts
+++ b/src/models/StateSnapshot.ts
@@ -24,6 +24,26 @@ export default class StateSnapshot {
         return StateSnapshot.from(snapshot);
     }
 
+    static empty(): StateSnapshot {
+        const emptyBalance = { amount: 0n, data: "0x" };
+        return StateSnapshot.from({
+            forkId: ethers.ZeroHash,
+            blockHeight: 0n,
+            timestamp: 0n,
+            snapshotData: {
+                originForkId: ethers.ZeroHash,
+                stateMachineStateHash: ethers.ZeroHash,
+                participants: [],
+                latestInboundMessageBlockHash: ethers.ZeroHash,
+                latestInboundMessageBlockHeight: 0n,
+                latestOutboundMessageBlockHash: ethers.ZeroHash,
+                latestOutboundMessageBlockHeight: 0n,
+                totalDeposits: emptyBalance,
+                totalWithdrawals: emptyBalance
+            }
+        });
+    }
+
     toStruct(): StateSnapshotStruct {
         return this.snapshot;
     }
--- a/contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol
+++ b/contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol
@@ -8,6 +8,7 @@ import "./Errors.sol";
 import "../types/DisputeFraudProofTypes.sol";
 import "./utils/DisputeUtils.sol";
 import "./utils/GeneralUtils.sol";
+import "./utils/BlockUtils.sol";
 import "./UtilityFacet.sol";
 import {Math} from "@openzeppelin/contracts/utils/math/Math.sol";
 
--- a/contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol
+++ b/contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol
@@ -176,14 +176,9 @@ contract DisputeFraudProofFacet is StateChannelCommon {
             }
         }
 
-        if (
-            UtilityFacet(utilityFacetAddress).isAddressInArray(
-                proof.previousStateSnapshot.snapshotData.participants, signer
-            )
-                || UtilityFacet(utilityFacetAddress).isAddressInArray(
-                    proof.resultingStateSnapshot.snapshotData.participants, signer
-                )
-        ) return _invalid();
+        if (_isBlockAuthorParticipant(invalidBlock, proof.previousStateSnapshot, proof.resultingStateSnapshot)) {
+            return _invalid();
+        }
         return _valid(dispute.input.disputer);
     }
 
--- a/contracts/V1/StateChannelDiamondProxy/LocalDiamond.sol
+++ b/contracts/V1/StateChannelDiamondProxy/LocalDiamond.sol
@@ -7,6 +7,7 @@ import "../types/MessageTypeHashes.sol";
 import "../StateChannelManagerEvents.sol";
 import "./utils/DisputeUtils.sol";
 import "./utils/BlockUtils.sol";
+import "./utils/GeneralUtils.sol";
 import "hardhat/console.sol";
 
 /**
@@ -363,6 +364,14 @@ contract LocalDiamond is StateChannelManagerProxy {
         return DisputeVerificationFacet(disputeVerificationFacetAddress).checkDisputeAuditingDataCommitment(
             dispute, disputeAuditingData
         );
+    }
+
+    function isBlockAuthorParticipant(
+        Block memory _block,
+        StateSnapshot memory previousSnapshot,
+        StateSnapshot memory resultingSnapshot
+    ) public pure returns (bool) {
+        return _isBlockAuthorParticipant(_block, previousSnapshot, resultingSnapshot);
     }
 
     // function isCorrectAuditingData(Dispute memory dispute, DisputeAuditingData memory disputeAuditingData)
--- a/contracts/V1/StateChannelDiamondProxy/utils/BlockUtils.sol
+++ b/contracts/V1/StateChannelDiamondProxy/utils/BlockUtils.sol
@@ -2,6 +2,7 @@ pragma solidity ^0.8.8;
 
 import "../../types/DisputeTypes.sol";
 import "../Errors.sol";
+import "./GeneralUtils.sol";
 
 function _getBlockHeight(Block memory _block) pure returns (uint256) {
     return _block.transaction.header.transactionCnt;
@@ -33,4 +34,23 @@ function _areBlocksSameChannel(Block memory _block1, Block memory _block2) pure 
 
 function _doesBlockCommitToSnapshot(Block memory _block, StateSnapshot memory snapshot) pure returns (bool) {
     return _block.stateSnapshotHash == keccak256(abi.encode(snapshot));
+}
+
+function _isBlockAuthorParticipant(
+    Block memory _block,
+    StateSnapshot memory previousSnapshot,
+    StateSnapshot memory resultingSnapshot
+) pure returns (bool) {
+    address author = _getBlockAuthor(_block);
+    if (_isAddressInArray(previousSnapshot.snapshotData.participants, author)) {
+        return true;
+    }
+
+    if (
+        resultingSnapshot.blockHeight == _getBlockHeight(_block) && resultingSnapshot.forkId == _getBlockFork(_block)
+            && _isAddressInArray(resultingSnapshot.snapshotData.participants, author)
+    ) {
+        return true;
+    }
+    return false;
 }
--- a/contracts/V1/StateChannelDiamondProxy/utils/GeneralUtils.sol
+++ b/contracts/V1/StateChannelDiamondProxy/utils/GeneralUtils.sol
@@ -2,7 +2,6 @@ pragma solidity ^0.8.8;
 
 import "../../types/DisputeTypes.sol";
 import "../Errors.sol";
-import "./BlockUtils.sol";
 
 function _delegatecall(address target, bytes memory data) returns (bytes memory) {
     (bool success, bytes memory result) = target.delegatecall(data);
@@ -21,25 +20,6 @@ function _delegatecall(address target, bytes memory data) returns (bytes memory)
 function _isAddressInArray(address[] memory array, address adr) pure returns (bool) {
     for (uint256 i = 0; i < array.length; i++) {
         if (array[i] == adr) return true;
-    }
-    return false;
-}
-
-function _isBlockAuthorParticipant(
-    Block memory _block,
-    StateSnapshot memory previousSnapshot,
-    StateSnapshot memory resultingSnapshot
-) pure returns (bool) {
-    address author = _getBlockAuthor(_block);
-    if (_isAddressInArray(previousSnapshot.snapshotData.participants, author)) {
-        return true;
-    }
-
-    if (
-        resultingSnapshot.blockHeight == _getBlockHeight(_block) && resultingSnapshot.forkId == _getBlockFork(_block)
-            && _isAddressInArray(resultingSnapshot.snapshotData.participants, author)
-    ) {
-        return true;
     }
     return false;
 }
--- a/contracts/V1/StateChannelDiamondProxy/utils/GeneralUtils.sol
+++ b/contracts/V1/StateChannelDiamondProxy/utils/GeneralUtils.sol
@@ -24,3 +24,22 @@ function _isAddressInArray(address[] memory array, address adr) pure returns (bo
     }
     return false;
 }
+
+function _isBlockAuthorParticipant(
+    Block memory _block,
+    StateSnapshot memory previousSnapshot,
+    StateSnapshot memory resultingSnapshot
+) pure returns (bool) {
+    address author = _getBlockAuthor(_block);
+    if (_isAddressInArray(previousSnapshot.snapshotData.participants, author)) {
+        return true;
+    }
+
+    if (
+        resultingSnapshot.blockHeight == _getBlockHeight(_block) && resultingSnapshot.forkId == _getBlockFork(_block)
+            && _isAddressInArray(resultingSnapshot.snapshotData.participants, author)
+    ) {
+        return true;
+    }
+    return false;
+}
--- a/src/stateManager/ValidationService.ts
+++ b/src/stateManager/ValidationService.ts
@@ -72,14 +72,8 @@ export default class ValidationService {
             return await strategy.channelNotOpened(entry);
         }
 
-        //  Get participants
-        const participants = await this.getParticipantsUnionOrOnChain(
-            block,
-            block.channelId
-        );
-
         // Author is a participant
-        if (!participants.has(block.author)) {
+        if (!(await this.isBlockAuthorParticipant(block, block.channelId))) {
             this.logger.warn(
                 "validateBlockConfirmation - author is not participant",
                 {
@@ -535,19 +529,16 @@ export default class ValidationService {
 
     // ────────────────────── Helpers ─────────────────────
 
-    private async getParticipantsUnionOrOnChain(
+    private async isBlockAuthorParticipant(
         block: Block,
         channelId: ChannelId
-    ): Promise<Set<Address>> {
-        let participants = new Set(
-            this.storage.getParticipantsUnion(
-                block.coordinates,
-                block.stateSnapshotHash
-            )
+    ): Promise<boolean> {
+        const previousSnapshot = this.storage.getPreviousStateSnapshot(
+            block.coordinates
         );
 
-        if (participants.size === 0) {
-            // get participants from chain
+        if (!previousSnapshot) {
+            // No local anchor to bind against - fall back to the on-chain union.
             const [participantsFromChain, pendingParticipants] =
                 await Promise.all([
                     this.stateChannelManagerContract.getParticipants(channelId),
@@ -555,13 +546,25 @@ export default class ValidationService {
                         channelId
                     )
                 ]);
-            participants = new Set([
+            return new Set<Address>([
                 ...participantsFromChain,
                 ...pendingParticipants
-            ]);
+            ]).has(block.author);
         }
 
-        return participants;
+        // The author counts if it is in the previous snapshot, or in the block's
+        // declared resulting snapshot bound to the block's coordinates.
+        const resultingSnapshot =
+            this.storage.stateSnapshots.getStateSnapshotByHash(
+                block.stateSnapshotHash
+            );
+        // No declared snapshot in storage -> an empty snapshot contributes no
+        // participants, so the check degrades to the previous snapshot alone.
+        return await this.diamondStateMachine.localDiamondContract.isBlockAuthorParticipant(
+            block.blockStruct,
+            previousSnapshot.toStruct(),
+            (resultingSnapshot ?? StateSnapshot.empty()).toStruct()
+        );
     }
 
     private async getOnChainPostTiming(
--- a/test/V1/StateChannelDiamondProxy/DisputeVerificationFacet.t.sol
+++ b/test/V1/StateChannelDiamondProxy/DisputeVerificationFacet.t.sol
@@ -374,6 +374,28 @@ contract DisputeVerificationFacetTest is DiamondHarness {
         assertEq(harness.handleBlockAuthorNotParticipant(abi.encode(proof), dispute), dispute.input.disputer);
     }
 
+    function test_disputeBlockAuthorNotParticipant_authorInWrongForkResultingSnapshot_valid() public {
+        // The author is in the block's declared resulting snapshot, and the
+        // height matches, but the snapshot's forkId does not (a departed
+        // member naming a stale snapshot from a different fork at the same
+        // height). The coordinate binding must not count it, so the author is
+        // not a participant and the proof is valid - matching the off-chain
+        // author gate so an honest auditor is never slashed.
+        DisputeExpiryGuardHarness harness = new DisputeExpiryGuardHarness();
+        (Dispute memory dispute, DisputeBlockAuthorNotParticipant memory proof, address signer) =
+            _disputeBlockAuthorNotParticipantProof();
+
+        proof.resultingStateSnapshot.snapshotData.participants[0] = signer;
+        // stale: the resulting snapshot belongs to a different fork, not this block's
+        proof.resultingStateSnapshot.forkId = keccak256("other-fork");
+        Block memory invalidBlock = abi.decode(dispute.input.stateProof.signedBlocks[0].encodedBlock, (Block));
+        invalidBlock.stateSnapshotHash = keccak256(abi.encode(proof.resultingStateSnapshot));
+        dispute.input.stateProof.signedBlocks[0].encodedBlock = abi.encode(invalidBlock);
+        dispute.input.stateProof.signedBlocks[0].signature = _sign(1, abi.encode(invalidBlock));
+
+        assertEq(harness.handleBlockAuthorNotParticipant(abi.encode(proof), dispute), dispute.input.disputer);
+    }
+
     function test_blockInvalidStateTransition_wrongTurnWithCorrectSnapshot_slashesSigner() public {
         address[] memory participants = new address[](2);
         participants[0] = vm.addr(1);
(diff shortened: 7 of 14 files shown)


src/models/StateSnapshot.ts
```diff
@@ -24,6 +24,26 @@ export default class StateSnapshot {
         return StateSnapshot.from(snapshot);
     }
 
+    static empty(): StateSnapshot {
```
lukastanisic99, line 27: So this was added only to be able to do: `resultingSnapshot ?? StateSnapshot.empty()).toStruct()` ? 

Why not just:
```typescript
if(!resultingSnapshot)return false;
localDiamondContract.isBlockAuthorParticipant(...)
```

lukastanisic99:
Overall no red or orange - just 2 yellow which I'll just point out:



- 🟡 **[TY2] The tests protect the height check but not the fork-ID check.** The Solidity regression makes the resulting snapshot stale by changing only [`blockHeight`](https://github.com/peer3to/state-channels-plus/blob/7f3604552637929a423e465a154f4cb872feed55/test/V1/StateChannelDiamondProxy/DisputeVerificationFacet.t.sol#L356-L374). The TypeScript probe does the same at [`nextHeight + 100`](https://github.com/peer3to/state-channels-plus/blob/7f3604552637929a423e465a154f4cb872feed55/test/fixtures/customRpc/harnessControl/services/stub/StubService.ts#L513-L527). Its unknown-fork cases take the no-local-anchor fallback and never test a resulting snapshot whose height matches but fork differs. Removing the fork comparison would leave every new test green.

  > **Fix TY2-FIX**
  >
  > Add Solidity and TypeScript cases where the resulting snapshot includes the author and matches the block height but has a different fork ID. Both must classify the author as a non-participant.

## 📏 AGENTS.md adherence

- 🟡 **[AY1] The shared pure validator is in the wrong utility file.** [`_isBlockAuthorParticipant`](https://github.com/peer3to/state-channels-plus/blob/7f3604552637929a423e465a154f4cb872feed55/contracts/V1/StateChannelDiamondProxy/utils/GeneralUtils.sol#L28-L45) is a pure validator shared with the off-chain TypeScript pipeline. Root `AGENTS.md` requires that shape to live in `utils/DisputeUtils.sol` or `utils/BlockUtils.sol`, plus a `LocalDiamond` forwarder. `GeneralUtils.sol` is outside the allowed placement.

  > **Fix AY1-FIX**
  >
  > Move `_isBlockAuthorParticipant` to `BlockUtils.sol`, since it operates on a block and its snapshot coordinates. Keep the existing `LocalDiamond` forwarder and fraud-proof call sites, then rerun compilation, both TypeScript builds, and the focused Solidity suite.

lukastanisic99:
Changed the base to `dispute` since I merged the rest

src/models/StateSnapshot.ts
daiagi: yes.
simply returning `false` is wrong, the author could be in the `previousSnapshot`

other option could be:

1.  checking that in TS, which would be not DRY, and the sol shared method is already very simple, if you do a precheck in ts might as well just write out the logic both in ts and sol.
2. pass `resultingSnapshot` as bytes, not as struct, and use `"0x"` sentinel, which  creates overhead of encode decode and i don't see how that's any nicer interface

lukastanisic99 (approved):
No red no orange, just some yellow for tests; @daiagi take a look if the tests cases are relevant, approving regardless

## 🧱 Fundamental

- 🟢 **[FG1] The production rule now matches the intended membership model.** The live validator resolves the previous snapshot by the block's coordinates, resolves the declared snapshot by the committed hash, and calls the shared predicate at [`ValidationService.ts`](https://github.com/peer3to/state-channels-plus/blob/61468153006bba56dc8b1f4bec96ac279f79a226/src/stateManager/ValidationService.ts#L532-L567). The predicate accepts the previous-snapshot author unconditionally and accepts a resulting-snapshot author only when both coordinates match at [`BlockUtils.sol`](https://github.com/peer3to/state-channels-plus/blob/61468153006bba56dc8b1f4bec96ac279f79a226/contracts/V1/StateChannelDiamondProxy/utils/BlockUtils.sol#L39-L55).

  > **Fix FG1-FIX**
  >
  > No change required.

- 🟢 **[SG1] The stale-snapshot privilege escalation is closed in both trust-boundary consumers.** An ex-member's historical snapshot no longer grants live author eligibility, and the fraud-proof handler binds both supplied snapshots to the signed block chain before applying the same predicate at [`DisputeFraudProofFacet.sol`](https://github.com/peer3to/state-channels-plus/blob/61468153006bba56dc8b1f4bec96ac279f79a226/contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol#L147-L183).

  > **Fix SG1-FIX**
  >
  > No change required.

## 🏁 Race conditions

- 🟢 **[RG1] The change adds no mutable check/use gap.** Snapshot lookup and the pure membership call run inside the existing [`onBlockConfirmation` mutex](https://github.com/peer3to/state-channels-plus/blob/61468153006bba56dc8b1f4bec96ac279f79a226/src/stateManager/StateManager.ts#L873-L890); the selected snapshots are hash-addressed and resolved before the contract call.

  > **Fix RG1-FIX**
  >
  > No change required.

## ⚡ Performance

- 🟢 **[PG1] The new cost is bounded and keeps one validator owner.** Anchored validation adds [one pure local-diamond `eth_call`](https://github.com/peer3to/state-channels-plus/blob/61468153006bba56dc8b1f4bec96ac279f79a226/src/stateManager/ValidationService.ts#L563-L567) per block instead of duplicating the predicate in TypeScript. The no-anchor path still performs the existing current-plus-pending participant lookup.

  > **Fix PG1-FIX**
  >
  > No change required.

- 🟡 **[TY1] Unit / Test-harness — the new `ValidationService` test bypasses the component's public surface.** The test calls [`probeBlockAuthorParticipant`](https://github.com/peer3to/state-channels-plus/blob/61468153006bba56dc8b1f4bec96ac279f79a226/test/stateManager/ValidationService.test.ts#L5-L14), whose host implementation exposes the private helper through an [`as unknown as` structural cast](https://github.com/peer3to/state-channels-plus/blob/61468153006bba56dc8b1f4bec96ac279f79a226/test/fixtures/customRpc/harnessControl/services/stub/StubService.ts#L426-L432). It never executes [`validateBlockConfirmation`](https://github.com/peer3to/state-channels-plus/blob/61468153006bba56dc8b1f4bec96ac279f79a226/src/stateManager/ValidationService.ts#L46-L85), so it stays green if the public path stops calling the helper or dispatches the wrong strategy result.

  > **Fix TY1-FIX**
  >
  > Stage real queued entries in the harness and drive `validateBlockConfirmation` through its public surface with a real strategy. Use record-only instrumentation only where the strategy's disconnect or dispute side effect would derail the unit session, and assert the public result plus the reached strategy boundary.

- 🟡 **[TY2] Unit — there is no matching-coordinate negative case.** The probe proves that a matching resulting snapshot containing the outsider returns true and that wrong-height or wrong-fork snapshots return false at [`StubService.ts`](https://github.com/peer3to/state-channels-plus/blob/61468153006bba56dc8b1f4bec96ac279f79a226/test/fixtures/customRpc/harnessControl/services/stub/StubService.ts#L505-L551). It never tests a snapshot whose fork and height match but whose participant set excludes the author; returning true for any coordinate-matched snapshot would leave the test green.

  > **Fix TY2-FIX**
  >
  > Add a public-surface unit case with a stored, coordinate-matched resulting snapshot that excludes the author. Expect `blockAuthorIsNotParticipant` and the strategy's rejection result.

- 🟡 **[TY3] Unit — the missing-resulting-snapshot path is not tested for a previous-snapshot member.** Production deliberately substitutes [`StateSnapshot.empty()`](https://github.com/peer3to/state-channels-plus/blob/61468153006bba56dc8b1f4bec96ac279f79a226/src/models/StateSnapshot.ts#L27-L45) when the committed result hash is absent, then relies on the previous snapshot at [`ValidationService.ts`](https://github.com/peer3to/state-channels-plus/blob/61468153006bba56dc8b1f4bec96ac279f79a226/src/stateManager/ValidationService.ts#L555-L567). Every anchored probe currently points to a stored snapshot, so this valid branch never runs.

  > **Fix TY3-FIX**
  >
  > Add a public-surface unit case whose previous snapshot contains the author and whose result hash is absent from storage. Expect validation to pass the author gate.

- 🟡 **[TY4] Unit — the missing-resulting-snapshot path is not tested for an outsider.** The same unexecuted sentinel branch must reject an author absent from the previous snapshot; otherwise a malformed or unavailable result snapshot could admit an outsider.

  > **Fix TY4-FIX**
  >
  > Add a public-surface unit case with a real previous snapshot that excludes the author and an unstored result hash. Expect `blockAuthorIsNotParticipant`.

- 🟡 **[TY5] Unit — pending-only membership in the no-anchor fallback is not covered.** The fallback unions current and pending on-chain participants at [`ValidationService.ts`](https://github.com/peer3to/state-channels-plus/blob/61468153006bba56dc8b1f4bec96ac279f79a226/src/stateManager/ValidationService.ts#L540-L552), but the probe checks only a current participant and an unrelated outsider at [`StubService.ts`](https://github.com/peer3to/state-channels-plus/blob/61468153006bba56dc8b1f4bec96ac279f79a226/test/fixtures/customRpc/harnessControl/services/stub/StubService.ts#L553-L577). Removing `getPendingParticipants` would leave all new tests green.

  > **Fix TY5-FIX**
  >
  > Put a real peer into pending-participant state, validate a block with no local previous anchor, and assert that a pending-only author passes the gate.

- 🟡 **[TY6] Unit — the new contract-read failure and recovery path has no defined test.** Anchored validation now awaits the local-diamond call at [`ValidationService.ts`](https://github.com/peer3to/state-channels-plus/blob/61468153006bba56dc8b1f4bec96ac279f79a226/src/stateManager/ValidationService.ts#L563-L567). A rejection escapes queued execution at [`BlockQueueManager.ts`](https://github.com/peer3to/state-channels-plus/blob/61468153006bba56dc8b1f4bec96ac279f79a226/src/stateManager/BlockQueueManager.ts#L520-L548) after the entry has been dequeued, with no restoration in that path.

  > **Fix TY6-FIX**
  >
  > Define the transient-read policy and test it through the public queue path. The safer behavior is to restore the entry without punishing its sources, then prove a later successful read validates the same entry exactly once.

- 🟡 **[TY7] E2E — stale membership is not exercised against an active participant.** The added workflow sends the crafted block only to a spectator and asserts `Status.SYNCED` at [`E2E-SpectatingAbortDoS.test.ts`](https://github.com/peer3to/state-channels-plus/blob/61468153006bba56dc8b1f4bec96ac279f79a226/test/e2e/E2E-SpectatingAbortDoS.test.ts#L206-L255). Active peers use the distinct [`BlockValidationStrategy.blockAuthorIsNotParticipant`](https://github.com/peer3to/state-channels-plus/blob/61468153006bba56dc8b1f4bec96ac279f79a226/src/stateManager/validationStrategy/BlockValidationStrategy.ts#L121-L127) boundary.

  > **Fix TY7-FIX**
  >
  > Reuse [`craftStaleMembershipBlockConfirmation`](https://github.com/peer3to/state-channels-plus/blob/61468153006bba56dc8b1f4bec96ac279f79a226/test/harness/actions/ByzantineActions.ts#L308-L358) against a participating receiver. Assert that the leaver and source are cut, the receiver stays `PARTICIPATING` and synchronized, and no false dispute starts.

- 🟡 **[TY8] E2E — the stale-membership dispute path stops at Forge units.** The Forge cases cover stale height and same-height/wrong-fork classification at [`DisputeVerificationFacet.t.sol`](https://github.com/peer3to/state-channels-plus/blob/61468153006bba56dc8b1f4bec96ac279f79a226/test/V1/StateChannelDiamondProxy/DisputeVerificationFacet.t.sol#L356-L397), while the existing dispute E2E rewrites only a normal outsider at [`case3_signedBlocksOnly.test.ts`](https://github.com/peer3to/state-channels-plus/blob/61468153006bba56dc8b1f4bec96ac279f79a226/test/e2e/disputeValidation/stateProof/case3_signedBlocksOnly.test.ts#L305-L356). That E2E would pass if coordinate binding disappeared.

  > **Fix TY8-FIX**
  >
  > Add one stale-membership dispute workflow: commit a departed author's stale resulting snapshot, replay the dispute, assert that only `DisputeBlockAuthorNotParticipant` is stored, then resolve it and prove the malicious dispute is killed on-chain. Wrong-fork does not need a second E2E because it crosses the same boundary and is covered at unit and Forge levels.

## ♻️ Code reuse

- 🟡 **[CY1] Test-harness — the probe duplicates the repository's snapshot factory.** [`StubService.ts`](https://github.com/peer3to/state-channels-plus/blob/61468153006bba56dc8b1f4bec96ac279f79a226/test/fixtures/customRpc/harnessControl/services/stub/StubService.ts#L451-L471) manually rebuilds the complete `StateSnapshot` structure even though [`factory.snapshotData` and `factory.stateSnapshot`](https://github.com/peer3to/state-channels-plus/blob/61468153006bba56dc8b1f4bec96ac279f79a226/test/factory.ts#L299-L365) already own valid test-domain construction.

  > **Fix CY1-FIX**
  >
  > Replace the local builder with `factory.stateSnapshot`, overriding only participants, fork, height, and any scenario-specific fields.

## 📏 AGENTS.md adherence

- 🟡 **[AY1] The unit test violates the black-box component rule.** The private-helper cast and direct probe in [TY1](https://github.com/peer3to/state-channels-plus/blob/61468153006bba56dc8b1f4bec96ac279f79a226/test/fixtures/customRpc/harnessControl/services/stub/StubService.ts#L426-L432) bypass the public `ValidationService` contract that the repository requires unit tests to exercise.

  > **Fix AY1-FIX**
  >
  > Apply TY1-FIX and remove the private-helper cast.

- 🟡 **[AY2] The harness staging violates the factory-built-domain-object rule.** The hand-written snapshot in [CY1](https://github.com/peer3to/state-channels-plus/blob/61468153006bba56dc8b1f4bec96ac279f79a226/test/fixtures/customRpc/harnessControl/services/stub/StubService.ts#L451-L471) duplicates a canonical factory instead of using it.

  > **Fix AY2-FIX**
  >
  > Apply CY1-FIX.

## 🧹 Miscellaneous (scope)

- 🟢 **[MG1] The production diff stays on the reported defect.** The Solidity predicate, local forwarder, live call site, sentinel snapshot, adversarial block builder, and tests all support the stale-membership fix; no unrelated public API change is mixed in.

  > **Fix MG1-FIX**
  >
  > No change required.

## ❓ Open questions

- 🟢 **[OG1] None.** The inline suggestion to return `false` whenever the resulting snapshot is missing is correctly rejected in [the existing discussion](https://github.com/peer3to/state-channels-plus/pull/402#discussion_r3639334814): an author present in the previous snapshot remains eligible. TY3 and TY4 are the missing tests for that decision.

  > **Fix OG1-FIX**
  >
  > No change required.

daiagi:
@lukastanisic99 
I've added more tests
Do you want to review again?

lukastanisic99:
@daiagi looks good - 1 orange and a few yellow - your call if you want to merge

## 🧱 Fundamental

- 🟢 **[FG1] The live and on-chain paths now apply the same coordinate-bound membership rule.** [`BlockUtils.sol`](https://github.com/peer3to/state-channels-plus/blob/b32c1690705752422a345cb998c8b800d9a98f51/contracts/V1/StateChannelDiamondProxy/utils/BlockUtils.sol#L39-L55) accepts an author from the anchored previous snapshot, or from the declared resulting snapshot only when both fork and height match the block. [`ValidationService`](https://github.com/peer3to/state-channels-plus/blob/b32c1690705752422a345cb998c8b800d9a98f51/src/stateManager/ValidationService.ts#L532-L567) calls that Solidity owner instead of reimplementing it.

  > **Fix FG1-FIX**
  >
  > No change required.

## 🔒 Security

- 🟢 **[SG1] The stale-membership privilege escalation is closed at both trust boundaries.** Live validation resolves the previous snapshot from canonical storage and the result snapshot from the block's committed hash. The fraud-proof handler independently binds the supplied previous and resulting snapshots to the signed block chain before calling the same predicate at [`DisputeFraudProofFacet.sol`](https://github.com/peer3to/state-channels-plus/blob/b32c1690705752422a345cb998c8b800d9a98f51/contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol#L147-L183).

  > **Fix SG1-FIX**
  >
  > No change required.

## 🏁 Race conditions

- 🟢 **[RG1] The change adds no mutable check/use gap.** The snapshots are hash-addressed, and live validation runs under the existing `StateManager.onBlockConfirmation` mutex before applying the state transition.

  > **Fix RG1-FIX**
  >
  > No change required.

## ⚡ Performance

- 🟢 **[PG1] The added steady-state work is bounded.** Anchored blocks add one pure local-diamond call. The no-anchor branch retains the existing current-plus-pending participant reads.

  > **Fix PG1-FIX**
  >
  > No change required.

## 🧪 Tests

- 🟡 **[TY1] The public-surface unit probe still discards the public result.** The latest commit correctly replaces the private-helper cast with a real queued entry and calls [`validateBlockConfirmation`](https://github.com/peer3to/state-channels-plus/blob/b32c1690705752422a345cb998c8b800d9a98f51/test/fixtures/customRpc/harnessControl/services/stub/StubService.ts#L403-L428). However, it ignores the returned `BlockValidationResult` and returns only the recorded hook name at [lines 429–436](https://github.com/peer3to/state-channels-plus/blob/b32c1690705752422a345cb998c8b800d9a98f51/test/fixtures/customRpc/harnessControl/services/stub/StubService.ts#L429-L436). A regression that reaches `blockAuthorIsNotParticipant` but fails to propagate its `DISCONNECT` result stays green.

  > **Fix TY1-FIX**
  >
  > Return both the reached boundary and `validateBlockConfirmation`'s result from `runAuthorGate`. Assert the exact rejection result for every negative case and the absence of the author boundary for positive cases.

- 🟡 **[TY6] Contract-read failure and queue recovery remain undefined and untested.** Anchored validation awaits [`localDiamondContract.isBlockAuthorParticipant`](https://github.com/peer3to/state-channels-plus/blob/b32c1690705752422a345cb998c8b800d9a98f51/src/stateManager/ValidationService.ts#L557-L567). If that read rejects, [`executeQueuedEntry`](https://github.com/peer3to/state-channels-plus/blob/b32c1690705752422a345cb998c8b800d9a98f51/src/stateManager/BlockQueueManager.ts#L520-L548) has already dequeued the entry and does not restore it. The new tests cover every membership decision but never a transient read failure followed by recovery.

  > **Fix TY6-FIX**
  >
  > Define the transient-read policy and test it through the public queue path. Restore the attributed entry without punishing its sources, then prove a later successful read validates that entry exactly once.

- 🟢 **[TG2] The latest commit closes the earlier membership matrix and integration gaps.** [`ValidationService.test.ts`](https://github.com/peer3to/state-channels-plus/blob/b32c1690705752422a345cb998c8b800d9a98f51/test/stateManager/ValidationService.test.ts#L8-L90) now covers previous membership, matching result membership, stale height, wrong fork, matching coordinates without the author, both missing-result branches, current fallback, unknown fallback, and a real pending-only participant. The active receiver E2E proves the ex-member is cut without a dispute at [`E2E-SpectatingAbortDoS.test.ts`](https://github.com/peer3to/state-channels-plus/blob/b32c1690705752422a345cb998c8b800d9a98f51/test/e2e/E2E-SpectatingAbortDoS.test.ts#L259-L325). The new dispute E2E proves the dedicated fraud proof, excludes overlapping fallback proofs, and checks the on-chain kill at [`E2E-StaleMembershipDispute.test.ts`](https://github.com/peer3to/state-channels-plus/blob/b32c1690705752422a345cb998c8b800d9a98f51/test/e2e/E2E-StaleMembershipDispute.test.ts#L9-L152).

  > **Fix TG2-FIX**
  >
  > No coverage change beyond TY1-FIX and TY6-FIX.

## ♻️ Code reuse

- 🟡 **[CY1] The stale-membership block builder duplicates the existing next-authored-block pipeline.** [`craftOutsiderAuthoredBlockConfirmation`](https://github.com/peer3to/state-channels-plus/blob/b32c1690705752422a345cb998c8b800d9a98f51/test/harness/actions/ByzantineActions.ts#L277-L305) and [`craftStaleMembershipBlockConfirmation`](https://github.com/peer3to/state-channels-plus/blob/b32c1690705752422a345cb998c8b800d9a98f51/test/harness/actions/ByzantineActions.ts#L313-L358) both fetch and decode the head, resolve the author, construct head+1, set the previous hash, sign, and encode. The latter differs only by loading and assigning the stale snapshot hash.

  > **Fix CY1-FIX**
  >
  > Extract one private next-authored-block builder with an optional result-snapshot-hash override. Keep the two public adversarial helpers as named scenario wrappers.

## 📏 AGENTS.md adherence

- 🟠 **[AO1] The new dispute E2E uses the forbidden source-shipped callback path.** [`E2E-StaleMembershipDispute.test.ts`](https://github.com/peer3to/state-channels-plus/blob/b32c1690705752422a345cb998c8b800d9a98f51/test/e2e/E2E-StaleMembershipDispute.test.ts#L62-L105) passes an inline function to `stubConstructDispute`. Because it is not a registered strategy, [`DisputeTamperingActions`](https://github.com/peer3to/state-channels-plus/blob/b32c1690705752422a345cb998c8b800d9a98f51/test/harness/actions/DisputeTamperingActions.ts#L368-L374) serializes `tamper.toString()`, and [`DisputeRpcMethods`](https://github.com/peer3to/state-channels-plus/blob/b32c1690705752422a345cb998c8b800d9a98f51/test/fixtures/customRpc/harnessControl/services/dispute/DisputeRpcMethods.ts#L149-L159) reconstructs it with `new Function`. [`test/AGENTS.md`](https://github.com/peer3to/state-channels-plus/blob/b32c1690705752422a345cb998c8b800d9a98f51/test/AGENTS.md#L63-L73) says function-valued inputs must be named strategies and permits source shipping only for `scenario.exec` / `execOnHost`.

  > **Fix AO1-FIX**
  >
  > Add a named host-side stale-membership dispute tamper with typed `leaverAddress` and `staleSnapshotHash` arguments, and invoke it by strategy name. Do not send callback source across the runtime port.

- 🟡 **[AY2] The new probe surface erases domain and strategy-boundary types.** [`RecordingValidationStrategy.ts`](https://github.com/peer3to/state-channels-plus/blob/b32c1690705752422a345cb998c8b800d9a98f51/test/fixtures/customRpc/harnessControl/services/stub/RecordingValidationStrategy.ts#L4-L25) converts arbitrary proxy properties to `string`; the service and ten endpoints carry `Promise<string>`; and the test hand-restates `"blockAuthorIsNotParticipant"`. A strategy-hook rename is therefore a runtime failure instead of a compile error. The pending-participant endpoint also accepts `string` and casts it back to the already imported `Address` at [`StubRpcMethods.ts`](https://github.com/peer3to/state-channels-plus/blob/b32c1690705752422a345cb998c8b800d9a98f51/test/fixtures/customRpc/harnessControl/services/stub/StubRpcMethods.ts#L688-L693).

  > **Fix AY2-FIX**
  >
  > Derive a function-key union from `AValidationStrategy`, carry that type through the recorder/service/RPC/test, and use a typed sentinel. Accept `Address` directly at the pending-participant endpoint. AO1-FIX should likewise keep its argument object typed instead of `Record<string, unknown>` plus casts.

## 🧹 Scope

- 🟢 **[MG1] The PR diff stays on the stale-membership defect and its verification.** The apparent Holepunch relay changes from a two-dot comparison are target-branch-only changes after the merge-base; they are not in GitHub's 14-file three-dot PR diff.

  > **Fix MG1-FIX**
  >
  > No change required.

## Previous-review adjudication

The prior local review was [`3-review.md`](./3-review.md) at `61468153006bba56dc8b1f4bec96ac279f79a226`. The only new commit is `b32c169` (`additional test cases`); it changes test and harness files only.

| Prior finding | Status at `b32c169` | Evidence |
|---|---|---|
| TY1 public-surface unit | **PARTIALLY ADDRESSED** | The public method now runs, but its returned result is discarded. |
| TY2 matching coordinates without author | **ADDRESSED** | `ValidationService.test.ts` lines 37–42. |
| TY3 missing result + previous member | **ADDRESSED** | Lines 44–49. |
| TY4 missing result + outsider | **ADDRESSED** | Lines 51–54. |
| TY5 pending-only no-anchor fallback | **ADDRESSED** | Lines 68–89 use a real pending participant. |
| TY6 contract-read failure/recovery | **OPEN** | No failure/recovery case was added. |
| TY7 active participant E2E | **ADDRESSED** | `E2E-SpectatingAbortDoS.test.ts` lines 259–325. |
| TY8 stale-membership dispute E2E | **ADDRESSED** | `E2E-StaleMembershipDispute.test.ts` lines 9–152. |
| CY1 canonical snapshot factory | **ADDRESSED** | `StubService.storeSnapshotAt` now uses `factory.stateSnapshot` and `factory.snapshotData`. |
| AY1 black-box public surface | **ADDRESSED** | No private-helper cast remains. |
| AY2 factory-built domain objects | **ADDRESSED** | Addressed with the prior CY1 fix; current AY2 is a new type-safety finding. |

## GitHub feedback timeline

| Date | Source | Classification | Disposition |
|---|---|---|---|
| 2026-07-23 | [Inline thread: return false when the resulting snapshot is missing?](https://github.com/peer3to/state-channels-plus/pull/402#discussion_r3639334814) | **DISPUTED** | Returning false would incorrectly reject an author in the previous snapshot. The reply explains that distinction, and the new positive/negative missing-result tests now pin both branches. |
| 2026-07-23 | [PR comment containing the earlier fork-ID and utility-placement findings](https://github.com/peer3to/state-channels-plus/pull/402#issuecomment-5060291249) | **SELF-ECHO** | This is pasted output from the earlier local review. Both underlying issues were fixed before the previous `3-review.md`. |
| 2026-07-23 | [Base branch changed to `dispute`](https://github.com/peer3to/state-channels-plus/pull/402#issuecomment-5060329225) | **SUPERSEDED** | Coordination only; the current review uses `dispute` at `3722634b25c249dd752e9f41db8c37336865a429`. |
| 2026-07-24 | [Approved review containing `3-review.md`](https://github.com/peer3to/state-channels-plus/pull/402#pullrequestreview-4771915493) | **SELF-ECHO** | Exact prior review content; used as the review cutoff, not independent feedback. |
| 2026-07-26 | [Author: “I've added more tests / review again?”](https://github.com/peer3to/state-channels-plus/pull/402#issuecomment-5083006013) | **SUPERSEDED** | Coordination request. Commit `b32c169` is reviewed and adjudicated above. |

GitHub feedback counts: **0 OPEN, 0 ADDRESSED, 1 DISPUTED, 0 OUTDATED, 2 SUPERSEDED, 2 SELF-ECHO**. GitHub thread-resolution metadata was unavailable, so classifications are based on the current head, commit chronology, and thread contents.

## ❓ Open questions

- 🟢 **[OG1] None.** The missing-result-snapshot behavior is now explicit: previous-snapshot membership still counts, while an outsider is rejected.

  > **Fix OG1-FIX**
  >
  > No change required.

Issue #353, opened 2026-06-22
A peer that voluntarily leaves crashes when it sees the post-departure snapshot


momoshell:
When a peer voluntarily removes itself (a valid self-removal dispute), it stays marked as a participant. Then the reduced-fork snapshot it never built lands on chain, and the snapshot handler treats it as an unknown snapshot and throws a fatal error:

```ts
if (!knownSnapshot) {
    const status = this.stateManager.getStatus();
    if (status === Status.SYNCED) { /* warn + return */ }
    if (status === Status.PENDING_PARTICIPANT || status === Status.PARTICIPATING) {
        // ... "unknown snapshot while participant/pending, fatal"
        throw new Error(`onStateSnapshotUpdated: unknown snapshot ${...} while status=${status}`);
    }
}
```

The code that would recognise "I've actually left the channel" and step down to a spectator lives a bit further down in the same handler — but the throw above fires first, so it's never reached. The result: a peer that correctly left the channel crashes the moment its departure is finalized on chain, which is exactly when you least want a fatal.

**How to fix:** before throwing on an unknown snapshot, check whether this peer is still in the new snapshot's participant set. If it isn't, treat it as a clean departure (step down / resync) instead of a fatal. Reuse the participant cross-check that already exists lower in the handler, keep it scoped to peers that were participating, and make sure it returns early so it doesn't then try to follow a fork it can't reconstruct (see the reduced-fork transition gap). This also clears the benign "unknown snapshot" noise the tests currently have to swallow.

Issue #352, opened 2026-06-22
Posting a snapshot that both switches and updates the fork throws (or, if forced, corrupts withdrawals)


momoshell:
Some snapshot updates have to do two things in one transaction: move the on-chain snapshot to a new (reduced) fork, and apply a same-fork update on top of it. The code builds both halves up front:

```ts
const forkData = await this.prepareUpdateStateSnapshotFork();
const sameForkData = await this.prepareUpdateSnapshotSameFork(forkId);
```

The catch: the same-fork half is computed against the snapshot that's on chain *right now* (still the old fork), not against the state the fork half will produce. So whenever the local fork is ahead, it trips a guard and throws:

```ts
if (currentOnChainSnapshot.forkID !== latestSnapshot.forkID) {
    throw new Error(`Fork mismatch: current fork ..., new fork ...`);
}
```

It only succeeds when something else happens to move the on-chain snapshot to the new fork first — so it's timing-dependent and flaky.

Worth flagging: just making that guard return "nothing" instead of throwing is **not** a fix. We tried it — it silently drops the same-fork half's withdrawals and posts a snapshot with the wrong balance (a withdrawal delta came out at half its real value). The throw is the lesser evil and should stay until this is done properly.

**How to fix:** compute the same-fork half against the state the fork half will leave behind (the new fork's genesis), not the current on-chain snapshot. In practice that means rebasing three things onto the post-fork state — the fork-equality guard, the "is this snapshot newer" milestone filter, and the outbound-message range — with a fallback to the current snapshot when there's no fork half. A one-line guard tweak won't cut it.

_Repro (pre-rewrite test branch): "post snapshot on the reduced fork — multicall"._

Issue #351, opened 2026-06-22
A reduced fork stalls one signature short when its first block reaches the chain early


momoshell:
After a fork is reduced, the reduced fork's first block can land on chain (a peer re-queues it on a small clock skew) before everyone has counter-signed it off chain. When that happens, the peer whose turn is next quietly declines to sign it:

```ts
public async shouldSignBlock(block: Block): Promise<boolean> {
    if (this.p2pManager.isBlacklisted(block.author)) return false;
    if (this.status !== Status.PARTICIPATING) return false;
    if (block.onChainTimestamp !== undefined) {
        const nextToWrite = await this.diamondStateMachine.getNextToWrite();
        if (nextToWrite === this.signerAddress) {
            return false;          // ← skips a block it still needs to sign
        }
    }
    return true;
}
```

This is the only place a peer ever adds its own signature — no other path back-fills it, and the on-chain calldata path runs through the same check. Finalization still needs every participant's signature regardless of whether the block is already on chain, so the block sits forever one signature short and the fork never finalizes. It's deterministic (the next writer is fixed by the new fork's ordering), not a timing flake — it stays stuck no matter how long you wait.

**How to fix:** the skip should mean "don't re-sign a block I've already signed / that's already final" — not "don't sign because it's on chain and it's my turn next". Gate it on whether this peer's signature is already present rather than on the on-chain timestamp; then a still-missing signature always gets contributed, with no double-sign risk (a double-sign is two different blocks at the same height, not re-attaching your own signature to the same one).

_Repro (pre-rewrite test branch): spectate "traverse forks (dispute → reduced fork)", which deadlocks one signature short._

PR #403, opened 2026-07-25
chore: retire stale TODOs, remove dead code, and ship unhandled browser errors

sh3ll3x3c:
## Summary

A TODO-accounting pass over the repo: retire stale TODO comments, reword drifted
ones, remove the dead code they pointed at, and point deferred-test TODOs at their
tracked issues (#351, #352, #353, #391).

One behavioral change: `LogUploader` now captures an unhandled (window) error into
the log store **before** its early-returns, so browser crash context ships with the
uploaded logs. The capture is re-entrancy-safe (`Logger.error` calls back into
`uploadLogs`) and exception-safe (a hostile error graph that fails to encode must
not wedge later uploads). Covered by new unit tests.

Dead code removed — all provably unreachable or uncalled:
- `AStateMachine._clearOutboundMessages`: an unreachable `while (pop())` loop after
  `delete _outboundMessages` (`delete` already zeroes a dynamic array's length),
  plus its debug `console.log` and the now-unused `hardhat/console.sol` import.
- `DisputeVerificationFacet._isCorrectGenesis`: zero callers repo-wide.
- Four commented-out `disputeManager.dispute(...)` sketches referencing a field the
  class does not have.

Rebased onto `dispute` after #397 merged; adopted dispute's `block`→`entry`
validation-strategy refactor at the one overlapping call site.

## Test plan

- `yarn tsc --noEmit -p tsconfig.json` and `-p tsconfig.browser.json` — clean
- `yarn compile` — clean
- `yarn hardhat test --no-compile test/utils/LogUploader.test.ts` — 4 passing. The
  exception-safety guard is mutation-checked: moving the `finally` restore out of the
  guard makes the hostile-error test fail, proving the guard is load-bearing.
- eslint — no new errors versus the `dispute` baseline

### DIFF
--- a/contracts/V1/AStateMachine.sol
+++ b/contracts/V1/AStateMachine.sol
@@ -2,7 +2,6 @@ pragma solidity ^0.8.8;
 
 import "./types/DataTypes.sol";
 import "./types/MessageTypeHashes.sol";
-import "hardhat/console.sol";
 
 abstract contract AStateMachine {
     Transaction _tx; // This should be used instead of msg.sender at least for now
@@ -103,11 +102,6 @@ abstract contract AStateMachine {
 
     function _clearOutboundMessages() internal {
         delete _outboundMessages;
-        // TODO - pop loop since for some reason sometimes delete doesn't set length to 0
-        while (_outboundMessages.length > 0) {
-            _outboundMessages.pop();
-        }
-        console.log("Clearing outbound messages. Count after clear:", _outboundMessages.length);
     }
 
     function setState(bytes memory encodedState) external _nonReentrant {
@@ -157,7 +151,6 @@ abstract contract AStateMachine {
             }
         }
         Message[] memory recordedMessages = new Message[](_outboundMessages.length);
-        console.log("Recorded outbound messages count:", _outboundMessages.length);
         for (uint256 i = 0; i < _outboundMessages.length; i++) {
             recordedMessages[i] = _outboundMessages[i];
         }
--- a/contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol
+++ b/contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol
@@ -280,7 +280,7 @@ contract DisputeFraudProofFacet is StateChannelCommon {
     {
         DisputeInvalidStateProof memory proof = abi.decode(encodedFraudProof, (DisputeInvalidStateProof));
 
-        // TODO extract this into its own fraud proof and test it
+        // TODO extract genesis-fork linking, signed-block linking, and latest-state checks into fraud proofs
         if (!dispute.postedAuditingData) {
             // fraud prover supplies the genesis reference -> it must be linked to the fork
             if (!_isGenesisSnapshotDataLinkedToFork(dispute.input.forkId, proof.auditingData.genesisStateSnapshotData))
--- a/contracts/V1/StateChannelDiamondProxy/DisputeVerificationFacet.sol
+++ b/contracts/V1/StateChannelDiamondProxy/DisputeVerificationFacet.sol
@@ -419,15 +419,6 @@ contract DisputeVerificationFacet is StateChannelCommon {
 
     // =============================== State Proofs Verification  ===============================
 
-    function _isCorrectGenesis(Dispute memory dispute) internal view returns (bool) {
-        (bool hasBlock, Block memory latestBlock) = _getLatestBlock(dispute.input.stateProof);
-        if (!hasBlock) {
-            //no blocks in state proof - must be genesis
-            return true; // TODO This will be a dispute fraud proof, since the dispute struct doesn't have enough information to deduct this
-        }
-        return _areDisputeAndBlockSameFork(dispute, latestBlock);
-    }
-
     function _verifyDisputeOutboundMessageBlocks(DisputeAuditingData memory disputeAuditingData)
         internal
         view
--- a/contracts/V1/StateChannelDiamondProxy/StateProofFacet.sol
+++ b/contracts/V1/StateChannelDiamondProxy/StateProofFacet.sol
@@ -246,6 +246,7 @@ contract StateProofFacet is StateChannelCommon {
         address[] memory expectedParticipants,
         MilestoneProof memory milestone
     ) internal view returns (bool isFinal, bytes32 finalizedSnapshotHash) {
+        // TODO - need a gas limit on verifyMilestone and on verifyStateProof, so large proofs that can't be verified won't be spammed
         address[] memory thresholdSet = new address[](expectedParticipants.length);
         uint256 thresholdCount = 0;
         bytes memory previousEncodedBlock;
@@ -292,7 +293,6 @@ contract StateProofFacet is StateChannelCommon {
             }
             bool isParticipant = UtilityFacet(utilityFacetAddress).isAddressInArray(expectedParticipants, adr);
             if (isParticipant) {
-                // TODO - need a gas limit on verifyMilestone and on verifyStateProof, so large proofs that can't be verified won't be spamed
                 thresholdCount =
                     _tryInsertAddressInThresholdSet(adr, thresholdSet, thresholdCount, expectedParticipants);
             }
@@ -307,7 +307,6 @@ contract StateProofFacet is StateChannelCommon {
                 }
                 isParticipant = UtilityFacet(utilityFacetAddress).isAddressInArray(expectedParticipants, adr);
                 if (isParticipant) {
-                    // TODO - need a gas limit on verifyMilestone and on verifyStateProof, so large proofs that can't be verified won't be spamed
                     thresholdCount =
                         _tryInsertAddressInThresholdSet(adr, thresholdSet, thresholdCount, expectedParticipants);
                 }
--- a/docs/tests/todo.md
+++ b/docs/tests/todo.md
@@ -5,29 +5,32 @@ PR #384 fixes). Each needs a real dispute + on-chain reduction staged determinis
 dispute-heavy class that flakes on the sandbox (the harness stalls under load and tests fail for
 environmental reasons, not logic). They're better authored/run on a machine that can sustain the load.
 
-Source of the behaviours under test: `temp/pr/1-block-queue-crdt-redesign/implementation-1.md`.
-
 ---
 
-## 1. A junk-block flood costs one recovery attempt, not one per block
+## 1. An N-block junk flood is gated to one kill-period read while unexpired, one recovery after expiry
 
-**Scenario.** Our own current fork is under an active dispute whose kill period has **not** yet expired.
-A byzantine peer then floods us with many block confirmations, each on a different bogus fork.
+**Scenario.** Our own current fork is under an active dispute. A byzantine peer floods us with many block
+confirmations, each on a different bogus fork — first while the dispute's kill period has **not** yet
+expired, then again after it expires.
 
 **Expected.** Our node must:
 
-- schedule only **one** background local-reduction attempt for our current fork (not one per incoming
-  block), and
-- make only **one** on-chain "is the kill period expired?" read per kill-period window (subsequent
-  incoming junk blocks reuse the cached answer, they do not each hit the chain).
+- **While unexpired:** schedule **zero** background local-reduction attempts across all N incoming blocks
+  (`maybeScheduleForkRecovery` returns after caching the suppression window when `isExpired` is false),
+  and make only **one** on-chain "is the kill period expired?" read (subsequent junk reuses the cached
+  answer, it does not each hit the chain).
+- **After expiry:** a further junk burst schedules exactly **one** coalesced local-reduction attempt for
+  the fork.
 
 **Why it matters.** Without the per-fork recovery gate + the kill-period cache, a flood would schedule a
-task and an on-chain read per junk block — an amplification vector. The gate and cache are in the code;
-this test is the end-to-end proof.
+task and an on-chain read per junk block — an amplification vector. The existing single-block test
+already proves one local reduction for one wrong-fork block; this test must prove the N-block gate (no
+recovery while suppressed), the cached kill-period read, and the single coalesced recovery after expiry.
 
-**How to stage.** Put the node's current fork into dispute (kill period not yet expired), stub the
+**How to stage.** Put the node's current fork into dispute with the kill period not yet expired; stub the
 recovery reduction to a no-op that counts calls, and count on-chain kill-period reads. Ingest N distinct
-bogus-fork blocks. Assert: scheduled-recovery count == 1 and kill-period-read count == 1.
+bogus-fork blocks and assert scheduled-recovery count == 0 and kill-period-read count == 1. Then advance
+past kill-period expiry, ingest another junk burst, and assert exactly one coalesced recovery is scheduled.
 
 ---
 
@@ -78,12 +81,14 @@ author is never blacklisted, the node stays connected to it, and the block is pr
 
 ## Preferred minimum before merge (host-side, not full dispute e2e)
 
-Test #1 above (junk-flood → one recovery attempt / one kill-period read) is the DoS invariant this
-redesign introduces. A lighter, more deterministic **host-side** integration check (via `execOnHost`,
-in the style of the existing `E2E-BlockQueueManager` "never validates…" test) would pin it without the
-full dispute pipeline: stub `isForkDisputed` → true for the current fork, count `isKillPeriodExpired`
-chain reads and scheduled `runForkRecovery` tasks while ingesting N distinct bogus-fork blocks, and
-assert both counts == 1. Same idea can pin the B6 kill-period cache (N kicks → 1 chain read).
+Test #1 above (junk-flood gated to one kill-period read while unexpired, one recovery after expiry) is
+the DoS invariant this redesign introduces. A lighter, more deterministic **host-side** integration check
+(via `execOnHost`, in the style of the existing `E2E-BlockQueueManager` "never validates…" test) would
+pin it without the full dispute pipeline: stub `isForkDisputed` → true for the current fork, count
+`isKillPeriodExpired` chain reads and scheduled `runForkRecovery` tasks while ingesting N distinct
+bogus-fork blocks. While the kill period is unexpired, assert one chain read and zero recoveries; after
+advancing past expiry, ingest another burst and assert one coalesced recovery. Same idea can pin the B6
+kill-period cache (N kicks → 1 chain read).
 
 ## Follow-up: CRDT attribution ordering (not shipped)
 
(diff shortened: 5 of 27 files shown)


lukastanisic99:
## 🧱 Fundamental

- 🟠 **[FO1] Browser errors without an `Error` object are dropped.** [`onWindowError`](https://github.com/peer3to/state-channels-plus/blob/3a4cc17471c7ebcd29c943ea91599a57c875ea42/src/utils/logging/browser/BrowserLogUploader.ts#L10-L14) does nothing when `ErrorEvent.error` is absent. That is the normal shape for opaque cross-origin script failures, so the PR still misses a class of global browser errors it claims to ship.

  > **Fix FO1-FIX**
  >
  > Normalize every `ErrorEvent`: use `e.error` when it is an `Error`, otherwise build a safe error record from `message`, `filename`, `lineno`, and `colno`. Prove both event shapes through the browser listener.

- 🟠 **[FO2] A rejection reason can throw before the guarded uploader runs.** [`String(e.reason)`](https://github.com/peer3to/state-channels-plus/blob/3a4cc17471c7ebcd29c943ea91599a57c875ea42/src/utils/logging/browser/BrowserLogUploader.ts#L17-L22) executes in the global event callback, outside `uploadLogs`' `try`. An object with a throwing `toString` makes the crash handler itself throw and loses the original rejection.

  > **Fix FO2-FIX**
  >
  > Move unknown-reason normalization into a non-throwing helper that falls back to a fixed message when conversion fails. Dispatch a rejection whose `toString` throws and assert one safe crash record is uploaded.

- 🟡 **[FY3] The logger used for a global crash depends on whichever child was created last.** Every [`BrowserLogger` constructor calls `setLogger(this)`](https://github.com/peer3to/state-channels-plus/blob/3a4cc17471c7ebcd29c943ea91599a57c875ea42/src/utils/logging/browser/BrowserLogger.ts#L25-L39), while [`createChild` reuses the same uploader](https://github.com/peer3to/state-channels-plus/blob/3a4cc17471c7ebcd29c943ea91599a57c875ea42/src/utils/logging/browser/BrowserLogger.ts#L41-L51). Child creation therefore overwrites the logger that records later window errors, giving them an unrelated and order-dependent component context.

  > **Fix FY3-FIX**
  >
  > Bind the uploader once to the root logger, or give it an immutable crash-capture context instead of replacing its logger from child constructors. Create two children and verify a later window error keeps the root crash context.

- 🔴 **[SR1] A rejected network request can upload credentials and request bodies inside its raw error graph.** [`uploadLogs` stores the original error object](https://github.com/peer3to/state-channels-plus/blob/3a4cc17471c7ebcd29c943ea91599a57c875ea42/src/utils/logging/LogUploader.ts#L79-L90), and [`encodeLogEntry` copies every own `Error` property](https://github.com/peer3to/state-channels-plus/blob/3a4cc17471c7ebcd29c943ea91599a57c875ea42/src/utils/logging/logEncoder.ts#L24-L38). An `AxiosError`, for example, owns `config`, `request`, and `response`; `config.headers.Authorization`, cookies, and `data` can therefore be compressed and sent to the log endpoint. The existing [`sanitizeAxiosErrorForLogging`](https://github.com/peer3to/state-channels-plus/blob/3a4cc17471c7ebcd29c943ea91599a57c875ea42/src/utils/logging/axiosErrorUtils.ts#L14-L20) runs only on failures raised by the upload request itself and does not sanitize the captured application error.

  > **Fix SR1-FIX**
  >
  > Convert captured errors to an allowlisted DTO before logging: retain only `name`, `message`, `stack`, an approved `code`, and explicitly safe scalar fields. Never traverse `config`, headers, request, response, cause, or arbitrary own properties. Upload an Axios-like error containing bearer tokens, cookies, and a request body, then decode the payload and prove none of those secrets appear.

## 🏁 Race conditions

- 🟠 **[RO1] The in-flight guard ends before the expensive and asynchronous part of the upload.** [`uploadInitiated` is cleared immediately after jitter](https://github.com/peer3to/state-channels-plus/blob/3a4cc17471c7ebcd29c943ea91599a57c875ea42/src/utils/logging/LogUploader.ts#L99-L107), before snapshotting, encoding, compression, and the retried HTTP request. An error during those phases starts another full upload of the retained buffer, so a crash burst can overlap compression and duplicate the same logs across requests.

  > **Fix RO1-FIX**
  >
  > Keep one in-flight promise through compression, retries, and completion. Mark arrivals after the current snapshot as dirty and serialize exactly one follow-up upload; clear the lock in `finally`. Pause the HTTP request, inject several errors, and assert one active request plus one follow-up containing the late records.

## ⚡ Performance

- 🟢 **[PG1] The cleanup adds no independent steady-state cost beyond the overlapping-upload race above.**

  > **Fix PG1-FIX**
  >
  > No change required.

- 🟠 **[TO2] There is no test proving captured errors are safe to upload.** The happy-path test expects the entire [`Error` in uploaded metadata](https://github.com/peer3to/state-channels-plus/blob/3a4cc17471c7ebcd29c943ea91599a57c875ea42/test/utils/LogUploader.test.ts#L53-L70), which locks in SR1 instead of protecting the trust boundary.

  > **Fix TO2-FIX**
  >
  > Replace that assertion with the allowlisted error shape from SR1-FIX and add secret-bearing nested properties whose values must be absent from the decoded upload.

- 🟠 **[TO3] No test exercises the browser event and HTTP boundaries together.** All four cases instantiate [`NodeLogUploader` directly](https://github.com/peer3to/state-channels-plus/blob/3a4cc17471c7ebcd29c943ea91599a57c875ea42/test/utils/LogUploader.test.ts#L12-L34), so the new `window` listeners, event normalization, and actual request payload are never covered as a browser workflow.

  > **Fix TO3-FIX**
  >
  > Add a browser E2E with a local HTTP receiver. Dispatch `ErrorEvent` and `PromiseRejectionEvent`, receive and decode the request, and assert the normalized crash record reaches the server.

- 🟠 **[TO4] The burst test proves storage, not delivery.** [`records an unhandled error that arrives while an upload is in flight`](https://github.com/peer3to/state-channels-plus/blob/3a4cc17471c7ebcd29c943ea91599a57c875ea42/test/utils/LogUploader.test.ts#L83-L107) checks the mutable `LogStore` after completion but never decodes the sole POST. It passes even if the late error missed that upload entirely.

  > **Fix TO4-FIX**
  >
  > Decode the posted payload and assert the late error is present. If the intended contract is next-upload delivery, assert the first payload omits it and exactly one serialized follow-up contains it.

- 🟠 **[TO5] The missing-`error` ErrorEvent path has no unit coverage.** That untested branch is the data-loss case in FO1.

  > **Fix TO5-FIX**
  >
  > Dispatch an `ErrorEvent` with only message, filename, line, and column fields through the listener and assert a normalized uploaded record.

- 🟠 **[TO6] Non-`Error` rejection normalization has no failure-path coverage.** No test supplies a primitive rejection or an object whose string conversion throws, so FO2 remains invisible.

  > **Fix TO6-FIX**
  >
  > Cover a string rejection, a plain object, and a throwing `toString`; assert each handler returns without throwing and uploads one bounded, safe record.

- 🟠 **[TO7] The concurrency test never reaches the network-phase boundary.** Its second call arrives only during [`jitter`](https://github.com/peer3to/state-channels-plus/blob/3a4cc17471c7ebcd29c943ea91599a57c875ea42/test/utils/LogUploader.test.ts#L86-L94), while the production guard is still true. It cannot catch RO1, where the guard is false and the first HTTP request is still pending.

  > **Fix TO7-FIX**
  >
  > Hold the first HTTP request unresolved, inject another error after jitter, and assert the single-flight/follow-up contract from RO1-FIX.

- 🟡 **[TY8] Child-logger crash attribution is untested.** The suite never creates children around a shared uploader, so FY3's last-child overwrite is not observed.

  > **Fix TY8-FIX**
  >
  > Create multiple child loggers, trigger the global listener, decode the result, and assert the chosen root crash context is stable regardless of child creation order.

## ♻️ Code reuse

- 🟠 **[CO1] Browser and Node own different unhandled-error capture flows.** The browser adapter calls [`uploadLogs(error)`](https://github.com/peer3to/state-channels-plus/blob/3a4cc17471c7ebcd29c943ea91599a57c875ea42/src/utils/logging/browser/BrowserLogUploader.ts#L10-L22), while the Node adapter separately calls [`logger.error`](https://github.com/peer3to/state-channels-plus/blob/3a4cc17471c7ebcd29c943ea91599a57c875ea42/src/utils/logging/node/NodeLogUploader.ts#L29-L50). Redaction, normalization, re-entrancy, and upload scheduling can now drift between two crash paths.

  > **Fix CO1-FIX**
  >
  > Put `captureUnhandled(reason, source)` on `LogUploader` as the single owner of safe normalization, storage, and scheduling. Keep platform adapters limited to translating their event shape and delegating; run the same contract cases against both adapters.

## 📏 AGENTS.md adherence

- 🟠 **[AO1] The new unit suite mocks both HTTP and time randomness.** [`sinon.stub(axios, "post")` and `sinon.stub(Math, "random")`](https://github.com/peer3to/state-channels-plus/blob/3a4cc17471c7ebcd29c943ea91599a57c875ea42/test/utils/LogUploader.test.ts#L42-L56) violate `test/AGENTS.md`'s no-mocks rule and avoid the actual upload boundary.

  > **Fix AO1-FIX**
  >
  > Use a local HTTP receiver and inject a deterministic jitter/clock through the real uploader fixture. Assert the received request rather than stubbing Axios.

- 🟡 **[AY2] Reusable setup and decoding helpers live in the test file.** [`createUploader` and `uploadedLogsFrom`](https://github.com/peer3to/state-channels-plus/blob/3a4cc17471c7ebcd29c943ea91599a57c875ea42/test/utils/LogUploader.test.ts#L10-L40) are fixture code, contrary to `test/AGENTS.md`'s rule that test files contain tests rather than helpers.

  > **Fix AY2-FIX**
  >
  > Move uploader setup, the local receiver, and payload decoding into a dedicated logging fixture under `test/fixtures`; leave this file with scenario setup and assertions.

## 🧹 Miscellaneous (scope)

- 🟠 **[MO1] The first deferred-test recipe expects recovery while explicitly staging the no-recovery branch.** [`docs/tests/todo.md`](https://github.com/peer3to/state-channels-plus/blob/3a4cc17471c7ebcd29c943ea91599a57c875ea42/docs/tests/todo.md#L10-L28) says the kill period has not expired and then expects one scheduled recovery. Production returns without scheduling when [`isExpired` is false](https://github.com/peer3to/state-channels-plus/blob/3a4cc17471c7ebcd29c943ea91599a57c875ea42/src/stateManager/BlockQueueManager.ts#L170-L184), so the documented test cannot pass as written.

  > **Fix MO1-FIX**
  >
  > Split the recipe at the boundary: while unexpired, assert one cached chain read and zero recoveries; after advancing past expiry, ingest another junk burst and assert one coalesced recovery.

- 🟡 **[MY2] The timeout TODO points at an issue that explicitly excludes that test cluster.** [`E2E-Timeouts.test.ts`](https://github.com/peer3to/state-channels-plus/blob/3a4cc17471c7ebcd29c943ea91599a57c875ea42/test/e2e/E2E-Timeouts.test.ts#L35-L50) assigns its `RaceConditionBlockCalldataTimestampTooLate` note to #391, but [#391's latest report](https://github.com/peer3to/state-channels-plus/issues/391#issuecomment-5019374883) classifies all seven `E2E-Timeouts` failures as a one-off anomaly and says not to add them to the tracker.

  > **Fix MY2-FIX**
  >
  > Remove the TODO if the anomaly is closed, or open a dedicated issue with this exact test, revert, and log evidence, then point the comment there.

## ❓ Open questions

- 🟢 **[OG1] No product decision is required to fix the findings above.**

  > **Fix OG1-FIX**
  >
  > No change required.

sh3ll3x3c:
@lukastanisic99

lukastanisic99:
@sh3ll3x3c baci pogled - nema nista critical samo neke todos za tests koliko vidim, pa sam odluci da li ces da merge 
## 🧱 Fundamental

- 🟠 **[FO1] Browser errors without an `Error` object are still dropped.** [`onWindowError`](https://github.com/peer3to/state-channels-plus/blob/f49d295d3e1d7746ed46599e4e05093d5c9a3e95/src/utils/logging/browser/BrowserLogUploader.ts#L10-L14) only captures the event when `e.error` is truthy. Opaque cross-origin script failures normally expose only `message`, `filename`, `lineno`, and `colno`, so this crash path still loses the event.

  > **Fix FO1-FIX**
  >
  > Always pass an error record to `captureUnhandled`: use `e.error` when present and otherwise normalize the safe scalar `ErrorEvent` fields. Dispatch both event shapes through the browser listener and assert both reach the receiver.

- 🟡 **[FY3] The global crash logger still depends on whichever child logger was created last.** Every [`BrowserLogger` constructor calls `setLogger(this)`](https://github.com/peer3to/state-channels-plus/blob/f49d295d3e1d7746ed46599e4e05093d5c9a3e95/src/utils/logging/browser/BrowserLogger.ts#L25-L39), and [`createChild` passes the shared uploader into another constructor](https://github.com/peer3to/state-channels-plus/blob/f49d295d3e1d7746ed46599e4e05093d5c9a3e95/src/utils/logging/browser/BrowserLogger.ts#L41-L51). Later child creation therefore changes the component context assigned to window crashes.

  > **Fix FY3-FIX**
  >
  > Bind the uploader once to the root logger, or give it a fixed crash-capture context. Create children in different orders and prove a later global error keeps the root context.

- 🟢 **[SG1] The force-push fixes SR1: captured `Error` graphs are replaced with an allowlisted record before `JSON.stringify`, and the real HTTP-boundary test proves Axios credentials, cookies, and request data do not reach the wire.**

  > **Fix SG1-FIX**
  >
  > No change required.

## 🏁 Race conditions

- 🟠 **[RO1] The upload lock still ends before encoding and the HTTP request.** [`uploadInitiated` is cleared immediately after jitter](https://github.com/peer3to/state-channels-plus/blob/f49d295d3e1d7746ed46599e4e05093d5c9a3e95/src/utils/logging/LogUploader.ts#L110-L117), while the request does not start until [`axios.post`](https://github.com/peer3to/state-channels-plus/blob/f49d295d3e1d7746ed46599e4e05093d5c9a3e95/src/utils/logging/LogUploader.ts#L143-L157). An error during compression, retry delay, or a pending request starts a second upload of the retained buffer, causing concurrent compression and duplicate requests.

  > **Fix RO1-FIX**
  >
  > Hold one in-flight promise through encoding, retries, and request completion. Mark later arrivals dirty and serialize one follow-up upload after the first finishes; clear the lock in `finally`. Pause the receiver's first response, capture several errors, and assert one active request plus one follow-up containing the late records.

## ⚡ Performance

- 🟢 **[PG1] No separate performance defect remains beyond RO1's duplicate compression and HTTP work.**

  > **Fix PG1-FIX**
  >
  > No change required.

- 🟠 **[TO3] No test crosses either browser event boundary.** All uploader cases build a [`NodeLogUploader` directly](https://github.com/peer3to/state-channels-plus/blob/f49d295d3e1d7746ed46599e4e05093d5c9a3e95/test/fixtures/logging/LogUploader.fixture.ts#L101-L126) and call `captureUnhandled` themselves. The changed `window.error` and `window.unhandledrejection` listeners, event normalization, listener attachment, and listener teardown remain unverified.

  > **Fix TO3-FIX**
  >
  > Add browser integration coverage that attaches the real `BrowserLogUploader`, dispatches `ErrorEvent` and `PromiseRejectionEvent`, receives and decodes the POST, then destroys the uploader and proves the listeners no longer fire.

- 🟠 **[TO5] The missing-`error` `ErrorEvent` path has no coverage.** The existing tests never invoke [`onWindowError`](https://github.com/peer3to/state-channels-plus/blob/f49d295d3e1d7746ed46599e4e05093d5c9a3e95/src/utils/logging/browser/BrowserLogUploader.ts#L10-L15), so FO1 remains invisible.

  > **Fix TO5-FIX**
  >
  > Dispatch an `ErrorEvent` with only message, filename, line, and column fields and assert the normalized record reaches the HTTP receiver.

- 🟠 **[TO7] The concurrency test still stops at the jitter boundary.** [`delivers a captured error that arrives while an upload is in flight`](https://github.com/peer3to/state-channels-plus/blob/f49d295d3e1d7746ed46599e4e05093d5c9a3e95/test/utils/LogUploader.test.ts#L69-L87) injects the second error during the 50 ms sleep, while the lock is true. It cannot detect RO1, which starts after that sleep while the first HTTP request remains pending.

  > **Fix TO7-FIX**
  >
  > Extend the receiver fixture with a response hold, wait until the first request arrives, capture another error before releasing it, and assert the serialized one-active-request plus one-follow-up contract from RO1-FIX.

- 🟠 **[TO9] The changed Node process listeners are not tested through their public boundary.** [`NodeLogUploader` now delegates both process events to `captureUnhandled`](https://github.com/peer3to/state-channels-plus/blob/f49d295d3e1d7746ed46599e4e05093d5c9a3e95/src/utils/logging/node/NodeLogUploader.ts#L29-L41), but direct `captureUnhandled` calls do not prove event names, attachment, or teardown.

  > **Fix TO9-FIX**
  >
  > Attach the real Node uploader to an isolated process-like event source, emit both supported events, verify their uploads at the receiver, destroy the uploader, and prove both listeners were removed.

- 🟡 **[TY8] Child-logger crash attribution remains untested.** The fixture creates only one [`NodeLogger`](https://github.com/peer3to/state-channels-plus/blob/f49d295d3e1d7746ed46599e4e05093d5c9a3e95/test/fixtures/logging/LogUploader.fixture.ts#L117-L126), so FY3's last-child overwrite is not observed.

  > **Fix TY8-FIX**
  >
  > Create multiple children around one uploader, trigger the platform event, decode the uploaded entry, and assert the root crash context is stable regardless of child creation order.

## ♻️ Code reuse

- 🟢 **[CG1] The force-push fixes CO1: browser and Node listeners now delegate normalization, storage, and scheduling to the single `captureUnhandled` owner.**

  > **Fix CG1-FIX**
  >
  > No change required.

## 📏 AGENTS.md adherence

- 🟡 **[AY2] Reusable encoder setup still lives in the test file.** [`SECRETS`, `secretAxiosError`, `encodeMeta`, and `expectNoSecrets`](https://github.com/peer3to/state-channels-plus/blob/f49d295d3e1d7746ed46599e4e05093d5c9a3e95/test/utils/logEncoder.test.ts#L6-L44) stage nearly every case. `test/AGENTS.md` says reusable staging belongs under `test/fixtures`, leaving test files with scenarios and assertions.

  > **Fix AY2-FIX**
  >
  > Move the secret-bearing error builder, entry encoder, and shared assertion into the logging fixture, then keep `logEncoder.test.ts` to the case-specific setup and expectations.

## 🧹 Miscellaneous (scope)

- 🟢 **[MG1] The force-push corrects the kill-period recipe and explicitly removes #391 from the timeout TODO; the dead-code and remaining TODO edits match their stated scope.**

  > **Fix MG1-FIX**
  >
  > No change required.

## ❓ Open questions

- 🟢 **[OG1] No product decision is required to fix the remaining findings.**

  > **Fix OG1-FIX**
  >
  > No change required.

lukastanisic99 (approved):

PR #406, opened 2026-07-29
fix(reduction): log the inbound failure from the revert instead of a later chain read

sh3ll3x3c:
> **Stacked on top of the contract change.** This PR's diff is the TypeScript
> layer only. Review the contract layer separately in the PR below it.

## Summary

A reduction submission that fails with an error we do not classify as a race is
rethrown with only the decoded error name — from `simulateSubmission` as a
thrown error, and from `submitDetached` as a detached rejection. The error name
alone does not say what the contract compared.

Now that the revert carries that comparison (#411), this logs it. The decoding
is **generic**: an `ErrorDescription`'s args are an ethers `Result` whose field
names come from the error's own ABI, so one helper covers every custom error
without naming a single field.

```ts
LoggerUtils.getCustomEvmErrorMetadata(customError)
// → { errorName, args: { submittedSnapshotInboundHash, expectedTargetInboundHash, … } }
```

`ErrorDisputeInboundMessageBlocksInvalid` therefore logs its hashes, break index
and failure reason, and every other custom error in the codebase gets the same
treatment for free. Numeric args stay `bigint` — the log pipeline already
stringifies them on the way out.

Also adds `previousBlockHash` to `LoggerUtils.getMessageBlockMetadata`. The
on-chain walk compares that field against the running head, so without it the
log cannot show where a linkage check broke. Three existing call sites benefit.

## Review changes

Three commits, one per review comment:

| commit | change |
|---|---|
| `refactor(reduction): read disputes off the reduction candidate` | The candidate already carries its disputes, so `simulateSubmission`, `submitDetached` and `complete` no longer take them as a second parameter. |
| `feat(logging): log any custom EVM error generically from its ABI args` | The one helper above. Drops the per-error decoder, the `INBOUND_FAILURE_*` reason table, and the `cause` verdict derived from them. |
| `refactor(reduction): name the submission-failure log fields for what they hold` | `custom` → `customError`, `path` → `submissionPath`, `inbound` → `inboundChain`. |

The `cause` verdict (`chain-target-moved` / `local-chain-diverged`) is **gone**
by design: case-by-case handling belongs where an error drives a decision, not
in logging. Nothing is lost — the generic args carry
`expectedTargetInboundHash` and `getReductionInboundMetadata` logs
`computedTargetInboundHash`, so both hashes still land side by side and the
comparison is one glance away.

## Test Plan

- `yarn compile` — clean
- `yarn tsc --noEmit` on `tsconfig.json` and `tsconfig.browser.json` — clean
- `yarn test:parallel --grep "ReductionExecutor|LoggerUtils"` — **8 passing**

New tests for the generic helper: named args resolved from the real ABI, bigint
args preserved rather than narrowed, an error with no args, and `null` /
`undefined` inputs.

`test/factory.ts` gains `customEvmError(errorName, args)`, which ABI-encodes a
real revert and decodes it through the SDK's own `tryDecodeCustomError` — so a
rename in `Errors.sol` breaks these tests instead of silently feeding them an
`undefined` field. Its `ethers.Interface` is built on first use, not at module
scope: this file is loaded by the SDK worker thread and `GeneratedArtifacts` is
~2 MB.

## Known gaps

No end-to-end test drives a real submission failure through to the logged
metadata. The only fault seam, `stubNextReductionSimulationError`, patches
`runner.call` on the `multicall` selector and accepts only the three
*classified* race errors, so forcing an unclassified failure needs new stub
machinery — more new code than the diagnostics it would cover.

`ValidationService.test.ts` has two or three timestamp cases that fail
intermittently on this branch **and** on its base at the same rate. Pre-existing
flakiness under load, not introduced here.

## Background

This replaces an earlier approach on this branch that re-read `reduce()` after
the failure to see whether the target had moved. That read ran *after* the
submission had already failed, so it described a later chain state: anything
that moved in between would be reported as the cause even when it was
unrelated, and the detached path waited for mining first, widening the window.
Review feedback was that the missing information belonged in the contract
rather than in a TypeScript re-read, which is what the PR below this one does.
The re-read and the deadline helper it needed are gone.


src/stateManager/reduction/ReductionExecutor.ts
```diff
@@ -372,11 +416,80 @@ export default class ReductionExecutor {
...
+     * never linking, and those two call for opposite responses.
+     */
+    private async logUnclassifiedSubmissionFailure(
+        errorName: string | undefined,
+        forkId: ForkId,
+        candidate: LocalReductionCandidate,
+        disputes: DisputeStruct[],
+        path: ReductionSubmissionPath
+    ): Promise<void> {
+        const chainTarget = await this.readChainInboundTarget(disputes);
```
lukastanisic99: **Human note**

Yes I agree with the AI poitining out that we're not logging the actuall run (which failed) - we're interested in the actual failed run. (Don't look at the fix that the AI suggested - it's shit) 

Another thing I want to point out - in typescript we're logging along the way, so we have infromation to deduct - the information that was missing was in soliditity (e.g. DisputeVerificationFacet.sol:312 ). Most of our custom errors don't include any `args` so we just look at the error name, but that may not be enough. In the one I linked we should include the hashes, so it's clear agianst which state the chain compared and deducted the error. That was the root why we couldn't deduct and didn't know what was going on since it seems that in the failing case the state on-chain moved (somehow) even though by design it should not be possible - if we had that logged we would know. So most of the usefull stuff around this and in general is populating our custom solidity Errors to include usefull args that give them more meaning.

**AI-generated review**

🔴 **[FR1] The comparison reads a later chain state, so it can report the wrong cause.** [`logUnclassifiedSubmissionFailure()`](https://github.com/peer3to/state-channels-plus/blob/578390519b772e0aa8d0636d544744a4d0d5c3fe/src/stateManager/reduction/ReductionExecutor.ts#L439-L465) starts a new latest-state `reduce.staticCall` only after the submission call has failed. If the failure at block N came from locally broken inbound linkage and another reduction or snapshot transaction changes the effective chain state before this read, the later target can differ even though that movement did not cause the failure. The detached path has a larger version of the same window because it waits for mining and error recovery first.

> **Fix FR1-FIX**
>
> Capture the block used by simulation and run both static calls against that explicit `blockTag`. For a mined detached failure, use the failed receipt's block number when available and report `unavailable` when no failure block can be established. Add a deterministic test that changes the effective reduction target after the caught failure but before the diagnostic read and prove the comparison still describes the failure block.

lukastanisic99:
There's also a merge conflict

sh3ll3x3c:
## This is now two stacked PRs

Following the review feedback, this work split into a stack. If you looked at
this PR before, its diff has changed — it is now the TypeScript layer only.

**Stack #412**

| | PR | what it changes | files |
|---|---|---|---|
| bottom | #411 | Solidity: the error carries what the check compared | 4 |
| top | this PR (#406) | TypeScript: decode that and log it | 5 |

Review them separately — each PR's diff shows only its own layer. #411 merges
into `dispute` first, then this one.

## Why it split

This PR started as logging only. The original approach logged our local view and
then re-read `reduce()` after a failed submission to see whether the chain's
target had moved.

That re-read was wrong, for the reason raised in review: it runs *after* the
failure, so it describes a later chain state. Anything that moved in between
gets reported as the cause even when it was unrelated, and the detached path
waits for mining first, which widens the window further. A diagnostic that can
name the wrong cause is worse than no diagnostic.

The review also made the more important point: the information was never
obtainable in TypeScript. The contract holds the three hashes it compares and
throws all of them away, because `ErrorDisputeInboundMessageBlocksInvalid` took
no arguments. Every off-chain attempt was a worse proxy for data that only
exists inside the revert.

So the fix moved into Solidity — and that is a contract change with a breaking
ABI impact, which does not belong hidden inside something titled "log the
inbound-chain evidence". Hence the split: #411 is the contract change and can be
reviewed on its own terms, this PR is the consumer.

## What changed since the first review

- The post-failure `reduce()` re-read is **gone**, along with the deadline
  helper it needed.
- The revert now carries the compared hashes, the break index, and a
  `failureReason` distinguishing a broken hash link from a broken height
  sequence — those stop at the same index and were otherwise
  indistinguishable (#411).
- `cause` is now derived from two values that both come from the failed call
  itself, so there is no window for an unrelated transaction to move the target.
- Rebased onto current `dispute`; the earlier merge conflict is gone.

src/stateManager/reduction/ReductionExecutor.ts
sh3ll3x3c: @lukastanisic99 — this is done, and it landed in #411 rather than here.

You were right on both counts. The re-read described a later chain state than
the one that failed, so it could name a cause that had nothing to do with the
failure — I deleted it rather than patching it with a `blockTag`.

And your actual point was the one that mattered: the information was never
obtainable in TypeScript. `ErrorDisputeInboundMessageBlocksInvalid` took no
arguments, so the contract compared three hashes and discarded all of them.
#411 makes it carry them, plus a `failureReason` — a broken hash link and a
broken height sequence stop at the same block index and were otherwise
indistinguishable, which I only found because the first attempt at this shipped
that ambiguity.

Not done: the general sweep. The other ~40 errors in `Errors.sol` still take no
arguments. Happy to open an issue for that if you want it prioritised.

Also still open — the same check runs through `_isDataLinkedToDisputeInput`,
whose two callers turn a failure into a bare `false`. Those can't revert, so it
needs an event on the fraud-proof path and a return-shape change on the
`staticCall` path. Documented as a known gap in #411 rather than widened into it.

src/stateManager/reduction/ReductionExecutor.ts
```diff
@@ -307,6 +368,7 @@ export default class ReductionExecutor {
 
     private async simulateSubmission(
         forkId: ForkId,
+        candidate: LocalReductionCandidate,
```
lukastanisic99, line 309: Since you're passing the `candidate` now, you don't need to pass `disputes` (they are already contained in the candidate), so use them from there - same in other places like `submitDetached`

src/stateManager/reduction/ReductionExecutor.ts
```diff
@@ -32,6 +33,65 @@ type LocalReductionCandidate = ReductionComputation & {
...
+    runningInboundHash: string;
+    breakIndex: number;
+    submittedBlockCount: number;
+    // "hash-link" and "height-sequence" both break at breakIndex and are
+    // otherwise indistinguishable; "unknown" means the contract sent a code
+    // this build does not know.
+    failureReason: (typeof INBOUND_FAILURE_REASONS)[number] | "unknown";
+};
+
+export function tryReadInboundMessageBlocksFailure(
```
lukastanisic99: We should do 1 geenral logger utill that covers all custom errors instead of doing this case by casem since custom.errorDescription.args always contain the correct property names at runtime (e.g. if you log the `args` object for this at runtime you'll see `{submittedSnapshotInboundHash:"",...}` withouth ever explicitly mentioning `submittedSnapshotInboundHash` or other propreties). 

Just check at runtime is `args typeof ethers.Result` and is so our EthersResultProxy contains a method to convert all ethers.results to objects since deailing with ethers results is anoying. 

The only place where we'll handle this on a case by case basis is if we're taking action on these errors, where for every error we decide what to do next so in some cases this can help us form our decision.

src/stateManager/reduction/ReductionExecutor.ts
```diff
@@ -32,6 +33,65 @@ type LocalReductionCandidate = ReductionComputation & {
...
+ * during that call, and `submittedTarget` is what we derived locally before
+ * submitting. A difference means the chain's target moved out from under the
+ * candidate; equality means the target was right and our own supplied blocks
+ * did not link.
+ */
+export function compareSubmittedInboundTarget(
+    failure: InboundMessageBlocksFailure,
+    submittedTarget: string
+) {
+    return failure.expectedTargetInboundHash === submittedTarget
```
lukastanisic99: **Human note**

Yes this is dead code + not sure why AGENTS.md is not honored on the naming `submittedTarget: string` is not descriptive enough for me to know it's a `hash` - I just deduct it from the check bellow by seeing it `==` some hash - that's why I added the rule to AGENTS.md to make the veriable names and types descriptive so whoever reads the code doesn't have to 'think' `what is this`

**AI-generated review**

🔴 **[FR2] `chain-target-moved` compares two copies of the same submitted value.** Solidity puts `reducedOutput.latestInboundMessageBlockHash` into `expectedTargetInboundHash` in the revert at [`DisputeVerificationFacet.sol:314-320`](https://github.com/peer3to/state-channels-plus/blob/22f441f7db2fcb0a7936d13e22d7b51be65e1fa1/contracts/V1/StateChannelDiamondProxy/DisputeVerificationFacet.sol#L314-L320). `submittedTarget` comes from the same candidate field at [`ReductionExecutor.ts:469-471`](https://github.com/peer3to/state-channels-plus/blob/22f441f7db2fcb0a7936d13e22d7b51be65e1fa1/src/stateManager/reduction/ReductionExecutor.ts#L469-L471). A real revert therefore always takes the equality branch. The test creates an impossible mismatch by hand, so the log cannot establish that the chain target moved as claimed.

> **Fix FR2-FIX**
>
> Do not emit a confident movement cause from these two fields. Either log only the raw failure evidence and `failureReason`, or extend the Solidity error with two independently derived values that can prove movement. Cover the classifier through a real contract revert instead of constructing an impossible decoded error in TypeScript.

lukastanisic99 (commented):
**Human note**

Yes this one is part of my previous comment of we don't want a case by case solution for logging - we have a solution to reuse solidity generated types in typescript, but in reality as with my previous comment you can just log the `args` it will list [0..2] instead of the human readable error reason, but since we don't look at the logs anymore mannually the AI will deduct what it is + it's verbose in solidity. You can reuse our soldity to enum generator, but would probably be an overkill and hard to do in the general case where you just see a `uint` at runtime and don't know what it is.

**AI-generated review**

🟠 **[CO2] Failure-reason codes now have separate Solidity, production TypeScript, and test owners.** Solidity defines the numbers at [`Errors.sol:59-65`](https://github.com/peer3to/state-channels-plus/blob/22f441f7db2fcb0a7936d13e22d7b51be65e1fa1/contracts/V1/StateChannelDiamondProxy/Errors.sol#L59-L65), while [`ReductionExecutor.ts:39-45`](https://github.com/peer3to/state-channels-plus/blob/22f441f7db2fcb0a7936d13e22d7b51be65e1fa1/src/stateManager/reduction/ReductionExecutor.ts#L39-L45) and the unit test independently assume `0/1/2`. Changing a Solidity value can silently give logs the wrong label while both language-specific tests remain green.

> **Fix CO2-FIX**
>
> Generate the TypeScript mapping from the Solidity source, or keep the raw numeric reason in TypeScript. Add a cross-layer case that triggers each real Solidity branch and feeds its revert through the production decoder.

src/stateManager/reduction/ReductionExecutor.ts
sh3ll3x3c: Done in `21ca9d9a`. `simulateSubmission`, `submitDetached` and `complete` now
read `candidate.disputes` — `complete` had the same redundancy you flagged here.

`prepareLocalCandidate` and `classifyReductionRace` keep the parameter, since
neither takes a candidate.

src/stateManager/reduction/ReductionExecutor.ts
sh3ll3x3c: Done in `972e22f3`, exactly as described.

```ts
LoggerUtils.getCustomEvmErrorMetadata(customError)
// → { errorName, args: { submittedSnapshotInboundHash, expectedTargetInboundHash, … } }
```

It checks `isEthersResult` at runtime and converts with the
`Codec.convertEthersResultToObject` already used by `EthersResultProxy`, so no
error names a field anywhere in TypeScript and every custom error in the
codebase gets the same treatment. Unnamed solidity params fall out positional.

Deleted with it: `tryReadInboundMessageBlocksFailure`, the `INBOUND_FAILURE_*`
reason table, and the `cause` verdict they fed. Case-by-case decoding now only
exists where an error drives a decision, which is the race classifier.

Nothing is lost from the log — the args carry `expectedTargetInboundHash` and
`getReductionInboundMetadata` still logs `computedTargetInboundHash`, so both
hashes land side by side.

src/stateManager/reduction/ReductionExecutor.ts
sh3ll3x3c: Naming done in `932e7354`: `custom` → `customError`, `path` → `submissionPath`,
log key `inbound` → `inboundChain`. `submittedTarget` needed no rename in the
end — the generic helper in `972e22f3` deleted it outright.

On **CO2**: agreed, and the duplication is gone. The reason table no longer
exists in TypeScript; the raw `uint8` from the revert lands in the log as-is.

One correction on **FR2**, since the code it describes is gone but the claim
would otherwise stand as the record. It says `expectedTargetInboundHash` and
`submittedTarget` are two copies of the same submitted value, so a real revert
always takes the equality branch. That is not the case — we submit
`reduceAndFinalize(disputes, …)` and never send `reducedOutput`. The contract
recomputes it itself with `reduce(disputes)` at `DisputeVerificationFacet.sol:262`
and puts *that* into the revert, so the revert's hash was chain-derived and ours
was local. The comparison was between two independent values.

Removing it is still right, but for your reason (verdicts are not a logging
concern), not that one.

lukastanisic99 (approved):

PR #407, opened 2026-07-29
use protocol-derived harness timeouts and pin mocha

lukastanisic99:
- replace fixed harness waits with protocol-calculated deadlines
- keep Hardhat and tests on one Mocha runtime


test/harness/actions/assert/AssertSyncActions.ts
```diff
@@ -19,7 +19,7 @@ export class AssertSyncActions<
         const {
             expectedStateMachineStateHash,
             peerIndices,
-            timeout = 10000,
+            timeout = this.harness.event.protocolEventTimeoutMs(1),
```
daiagi: I'd change the signature:

```ts
export function protocolEventTimeoutMs(
    timeConfig: TimeConfig,
    options?: { withFirstBlockGrace?: boolean }
): number
```

with default value `withFirstBlockGrace:false`

to make the callsites meaning more clear

lukastanisic99: Sure, but why the added timeConfig in the args?

daiagi (approved):

PR #411, opened 2026-08-01
feat(contracts): report what the inbound-blocks check compared when it rejects

sh3ll3x3c:
## Summary

`ErrorDisputeInboundMessageBlocksInvalid` carried no arguments, so off-chain all
we had was the error name. Two opposite causes produce it:

- the chain's reduction target moved between `compute()` and submission — our
  candidate went stale through no fault of our own; or
- our own supplied blocks never linked — our local view diverged.

They call for opposite responses, and the name alone cannot tell them apart.

This makes the check report what it compared:

```solidity
error ErrorDisputeInboundMessageBlocksInvalid(
    bytes32 submittedSnapshotInboundHash,  // where the submitted snapshot said the chain starts
    bytes32 expectedTargetInboundHash,     // where reduce() said it ends
    bytes32 runningInboundHash,            // how far the walk actually got
    uint256 breakIndex,                    // the block that failed, or submittedBlockCount
    uint256 submittedBlockCount,
    uint8   failureReason                  // INBOUND_FAILURE_*
);
```

`failureReason` earns its place: a broken hash link and a broken height sequence
stop at the *same* index, and in the height case the running head and the
block's `previousBlockHash` still match — so without it the log shows a
consistent-looking chain that broke for no visible reason.

It is a `uint8` with named constants rather than an enum on purpose.
`scripts/generate-enums.ts` numbers the generated TS enums by discovery order
(`(i+1)*100`), and `src/types/sol-enums.ts` is tracked — a new enum in a file
sorting before `ProofTypes.sol` would have silently renumbered `FraudProofType`
and `DisputeFraudProofType`.

No block heights are added to the payload: the submitter already holds the
blocks, and the off-chain log records each block's height and
`previousBlockHash`, so the heights around `breakIndex` are already available.
That also keeps the added bytecode small — see the size note below.

There is now a single `_verifyInboundMessageBlocks`, which always returns the
full report `(bool isValid, bytes32 runningInboundMessageBlockHash, uint256
breakIndex, uint8 failureReason)`. `reduceOutputToSnapshotData` forwards the
whole report into `ErrorDisputeInboundMessageBlocksInvalid`, while
`_isDataLinkedToDisputeInput` destructures only `isValid` and discards the rest.
The bool-only wrapper that used to sit in front of the walk is gone; it cost
nothing to remove, as the optimizer had already inlined it (bytecode is
byte-identical).

## Test Plan

- `yarn compile` — clean
- `yarn tsc --noEmit` on `tsconfig.json` and `tsconfig.browser.json` — clean
- `forge test --match-path test/V1/StateChannelDiamondProxy/DisputeVerificationFacet.t.sol`
  — **32 passing** (was 25)
- `forge test` (full) — **52 passing**

Seven new forge cases, via an `InboundVerificationHarness` exposing the internal
check:

- valid chain matching the target
- first block not anchored to the snapshot head → `breakIndex 0`, hash-link
- mid-chain link break → `breakIndex 1`, running head = hash of block 0
- **skipped height** → height-sequence reason, and asserts the hash link at the
  break is *intact* — pinning exactly the ambiguity `failureReason` removes
- all blocks linked but wrong target → `breakIndex == submittedBlockCount`
- empty block array boundary
- plus one driving the real revert through `reduceOutputToSnapshotData` with
  `vm.expectRevert(abi.encodeWithSelector(...))`, so the payload is proven
  end-to-end and not only at the helper

## Reviewer notes

**Breaking ABI change.** The selector moves from `0x4b021fb0` to `0x3a7249bd`.
An SDK built before this change, pointed at a chain running these contracts,
will fail `parseError` and log no error name at all — strictly less than today.
Contracts and SDK should ship together. Tests never catch this because they
deploy fresh contracts every run.

**Contract size.** `DisputeVerificationFacet` goes from 24783 to 24866 bytes
against the 24576 EIP-170 limit. It was **already over** before this change —
measured with a clean compile of the base contracts — so this is not a new
regression, but it moves the wrong way on a facet that already cannot deploy to
mainnet. `StateChannelManagerProxy`, `DisputeFraudProofFacet` and `LocalDiamond`
are over the limit on `dispute` too.

**Known gap.** Only the reduction path gets this evidence. The same check also
runs through `_isDataLinkedToDisputeInput`, whose two callers
(`DisputeVerificationFacet.isDisputeOutputCorrect`,
`DisputeFraudProofFacet._handleDisputeInvalidOutputState`) still turn a failure
into a bare `false`. Those must keep returning false rather than reverting, so
surfacing detail there needs a different mechanism — an event for the
fraud-proof path, and a return-shape change for the `staticCall` path. Left for
a follow-up rather than widened into this PR.

### DIFF
--- a/contracts/V1/StateChannelDiamondProxy/StateChannelCommon.sol
+++ b/contracts/V1/StateChannelDiamondProxy/StateChannelCommon.sol
@@ -450,20 +450,10 @@ contract StateChannelCommon is StateChannelManagerStorage, StateChannelManagerEv
             return false;
         }
 
-        return _verifyInboundMessageBlocks(
+        (bool isValid,,,) = _verifyInboundMessageBlocks(
             latestStateSnapshot.snapshotData.latestInboundMessageBlockHash,
             dispute.input.latestInboundMessageBlockHash,
             inboundMessageBlocks
-        );
-    }
-
-    function _verifyInboundMessageBlocks(
-        bytes32 previousInboundMessageBlockHash,
-        bytes32 latestInboundMessageBlockHash,
-        MessageBlock[] memory inboundMessageBlocks
-    ) internal pure returns (bool) {
-        (bool isValid,,,) = _walkInboundMessageBlocks(
-            previousInboundMessageBlockHash, latestInboundMessageBlockHash, inboundMessageBlocks
         );
         return isValid;
     }
@@ -475,7 +465,7 @@ contract StateChannelCommon is StateChannelManagerStorage, StateChannelManagerEv
     /// `inboundMessageBlocks.length` for a final-target mismatch;
     /// `failureReason` is one of the INBOUND_FAILURE_* constants and is only
     /// meaningful when `isValid` is false.
-    function _walkInboundMessageBlocks(
+    function _verifyInboundMessageBlocks(
         bytes32 previousInboundMessageBlockHash,
         bytes32 latestInboundMessageBlockHash,
         MessageBlock[] memory inboundMessageBlocks
--- a/contracts/V1/StateChannelDiamondProxy/StateChannelCommon.sol
+++ b/contracts/V1/StateChannelDiamondProxy/StateChannelCommon.sol
@@ -462,20 +462,47 @@ contract StateChannelCommon is StateChannelManagerStorage, StateChannelManagerEv
         bytes32 latestInboundMessageBlockHash,
         MessageBlock[] memory inboundMessageBlocks
     ) internal pure returns (bool) {
+        (bool isValid,,,) = _walkInboundMessageBlocks(
+            previousInboundMessageBlockHash, latestInboundMessageBlockHash, inboundMessageBlocks
+        );
+        return isValid;
+    }
+
+    /// Walks the inbound chain and reports where and why it stopped, so a
+    /// caller that reverts can say what was compared instead of only that it
+    /// failed. `runningInboundMessageBlockHash` is the head the walk had
+    /// reached; `breakIndex` is the block that failed, or
+    /// `inboundMessageBlocks.length` for a final-target mismatch;
+    /// `failureReason` is one of the INBOUND_FAILURE_* constants and is only
+    /// meaningful when `isValid` is false.
+    function _walkInboundMessageBlocks(
+        bytes32 previousInboundMessageBlockHash,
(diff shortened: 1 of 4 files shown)


contracts/V1/StateChannelDiamondProxy/StateChannelCommon.sol
```diff
@@ -462,20 +462,47 @@ contract StateChannelCommon is StateChannelManagerStorage, StateChannelManagerEv
...
+    }
+
+    /// Walks the inbound chain and reports where and why it stopped, so a
+    /// caller that reverts can say what was compared instead of only that it
+    /// failed. `runningInboundMessageBlockHash` is the head the walk had
+    /// reached; `breakIndex` is the block that failed, or
+    /// `inboundMessageBlocks.length` for a final-target mismatch;
+    /// `failureReason` is one of the INBOUND_FAILURE_* constants and is only
+    /// meaningful when `isValid` is false.
+    function _walkInboundMessageBlocks(
```
lukastanisic99: My only comment on this PR is that the current `_verifyInboundMessageBlocks` is no logger needed and `_walkInboundMessageBlocks` should always be used instead or just rename `_walkInboundMessageBlocks` to `_verifyInboundMessageBlocks` and delete the original

lukastanisic99 (approved):

contracts/V1/StateChannelDiamondProxy/StateChannelCommon.sol
sh3ll3x3c: Done in `277d15e4` — took the rename option. There is now a single
`_verifyInboundMessageBlocks` returning the full report; the bool wrapper is
deleted and `_isDataLinkedToDisputeInput` destructures just `isValid`.

Bytecode is byte-identical before and after, so the optimizer had already
inlined the wrapper — the extra name was the only thing it was costing us.

Related, and left alone deliberately: `DisputeValidationService.ts:611` is a
hand-written TypeScript copy of this same `_isDataLinkedToDisputeInput`,
including a private `verifyInboundMessageBlocks` at `:683` that re-implements
the walk. That is the on/off-chain duplication AGENTS.md warns about, and a
candidate cause for the reverts this stack exists to diagnose. Raised as its own
backlog card rather than widened into this PR.

PR #413, opened 2026-08-02
Event bus

lukastanisic99:
unify runtime events behind one EventBus bridged over the port

- one bus per realm (p2p hooks, contract events, handler mirrors) with a
  single busEvent wire message; typed ethers delivery via attachContractEvents
  on both sides; breaking API change documented in the README
- custom RPC roots gain a dispose() hook that runs before runtime teardown
- EventBarrier: register before first check, bounded deadline check and
  timeout diagnostics, settle guard against late timeout side effects
- treat a stale DisputeReducedResultCommitted event as consumed instead of
  calling the reducer with an empty dispute set
- Clock re-initializes on a fresh provider; worker signer uses the no-op
  event provider so typed subscriptions start no real-chain filters


test/stateManager/EventBus.test.ts
```diff
@@ -0,0 +1,754 @@
...
+                            () =>
+                                resolve({
+                                    firstListenerTurnCount,
+                                    secondListenerTurnCount,
+                                    unsubscribedListenerTurnCount,
+                                    blockFinalizedCount,
+                                    turnWriter,
+                                    turnAgreementTime
+                                }),
+                            300
```
daiagi: these 300ms timeouts don't work for me, I needed to increase it to 3000 in order to get the test to pass.
and we should not rely on magic numbers and specific timing to get test passing.

give this to you agent and it should resolve it:

```
remove every fixed sampling delay from test/stateManager/EventBus.test.ts and replace timing assumptions with ordering guarantees. three tools cover every case:

1. positives resolve on the asserted event, not after a delay
- onTurn/onBlockFinalized test: add a `maybeResolve()` that returns unless both counters are nonzero, call it from BOTH listeners, delete the `setTimeout(..., 300)`. reason the 300ms fails: onBlockFinalized needs all 4 peers' signatures to converge on peer 0 -> reliably slower than 300ms after the first onTurn on loaded/slower machines.

2. snapshot counters one microtask after the resolving dispatch
- wrap the resolve in `queueMicrotask(...)`. the bus emit is synchronous, so a microtask runs after the whole dispatch (all listeners + bridge tap) completed -> only whole dispatches are observed -> the two healthy onTurn counters compare equal, and a wrongly-still-subscribed listener would already have counted before the snapshot.
- the same tool replaces the 300ms in the eventHandler-both-realms test and the setP2pEventHooks test (there the publishing proxy runs the bus emit BEFORE the app target within one invocation; the microtask lands after both -> counts match).
- consumer-built-contract test: the typed ethers delivery is async, so gate `maybeResolve` on all three positives (bus Addition, bus Roster, typed `once` fired) and put one `setImmediate` before reading getSum and the negatives, so sibling mirror dispatch chains from the same bus emit have drained. these are ordering primitives, not durations.

3. cross-port negatives ("the poisoned event never crossed") use a fifo fence instead of sleeping
- the poisoned post fails synchronously at structured-clone time inside the bridge tap -> nothing was enqueued. after the poisoned attempt returns, post a second CLONEABLE event through the same producer path:
  `sm.p2pEventHooks.onConnection?.(args.fenceAddress as never, true)`
- the runtime port delivers successful posts in order, so once the fence arrives on the main thread, every earlier worker post has already been dispatched -> assert the negative at that point:
  `await fenceArrived; expect(deliveredAddresses).to.deep.equal([fenceAddress])`
- fence address comes from `randomAddress()` in test/factory (no hardcoded hex), passed into the exec body via the `args` parameter (exec bodies are closure-free). compare with plain `===`, no String()/toLowerCase - the value round-trips the port verbatim.
- applies to both clone-error tests. in the hook one `workerDeliveries` becomes 2 (poisoned local delivery + fence). in the redelivery one, also replace the `setTimeout(..., 100)` "let this dispatch finish" deferral with `queueMicrotask`.

leave the disposal test's timers alone: "nothing arrives after the port closed" has no later message to fence with - wall clock is inherent there.

acceptance: no numeric setTimeout left in the file outside the disposal test; suite green repeatedly
```

lukastanisic99:
@daiagi take a look now

daiagi (approved):

PR #415, opened 2026-08-04
(refactor) StateManager

daiagi:
refactor: split StateManager into subsystem services (2509 → ~450 lines)

- Split `StateManager` into subsystem services + unit test suites for each subsystem
- `tryMergeStoredBlockConfirmation`: use `strategy.goodNewSignaturesOnExistingBlock`
  instead of inline `instanceof`-strategy. Note: `onBlockFinalized` now fires after
  the merge re-gossip, not before
- `timeoutWaitTime` consolidated onto one owner
- `StateManager` get/set surface: `getStatus`/`getChannelId`/`getSignerAddress`/
  `getChannelStatus` deleted, status/channelId/forkId behind accessors
- public method that were used only in tests cahnged into private (bracket access in fixtures),
  incl. ReductionManager/SnapshotUpdateService outside the `StateManager` split
- Removed dead code: `StateApplicationService.setLatestState`, `isForkDisputedOnChain`
  wrapper (replaced with direct inline call to the StateChannelManager contract)
- Rebased over #408; its ValidationService edits re-applied into the moved
  `ingest/ValidationService.ts`
- Added a diagram: diagrams/blockPipelines.drawio


lukastanisic99:
@daiagi conflicts here too - I'll still review in the meantime

daiagi:
>conflicts here too - I'll still review in the meantime 

resolved

src/stateManager/ingest/StoredBlockMergeService.ts
```diff
@@ -0,0 +1,133 @@
...
+
+        const persisted = sm.storage.blocks.getBlock(block.hash);
+        if (persisted) {
+            P2pEventHooksUtils.maybeNotifyBlockFinalized({
+                block: persisted,
+                storage: sm.storage,
+                p2pEventHooks: sm.p2pEventHooks,
+                logger: this.logger
+            });
+        }
```
momoshell, line 129: One thing I noticed: `strategy.goodNewSignaturesOnExistingBlock(block)` now runs (and broadcasts, for the live/spectating strategies) before `maybeNotifyBlockFinalized` runs.

The old inline code notified first, then broadcast. Is that reorder intentional? It's probably fine since the broadcast is fire-and-forget, but since it's peer-observable, might be worth a quick comment here noting it's deliberate so it doesn't get "fixed" back by accident later.

test/harness/actions/TransitionActions.ts
```diff
@@ -393,4 +395,74 @@ export class TransitionActions<
...
     }
+    /**
+     * White-box: run `tryMergeStoredBlockConfirmation` on the host, under the
+     * peer's live block strategy or a fabricated dispute strategy. Returns the
+     * merge result and the persisted signature set.
+     */
+    async runStoredBlockMerge(options: {
+        peerIndex: number;
+        confirmation: { signedBlock: unknown; signatures: string[] };
+        strategy?: "active" | "dispute";
```
momoshell: Small one here.

 `runStoredBlockMerge`'s `strategy` option only takes `"active" | "dispute"`, so there's no way yet to drive the merge through `SpectatingValidationStrategy` from the harness. 
 
 Since this PR is what routes the merge through per-strategy `goodNewSignaturesOnExistingBlock`, could be nice to add `"spectating"` here too so `StoredBlockMergeService.test.ts` can cover that arm as well.
 
 No pressure though, just flagging it.

src/stateManager/validationStrategy/CalldataCommittedStrategy.ts
```diff
@@ -69,6 +69,8 @@ export default class CalldataCommittedStrategy extends AValidationStrategy {
     public async goodNewSignaturesOnExistingBlock(
         _block: Block
     ): Promise<BlockValidationResult> {
+        // Unreachable tripwire: the calldata event always carries
+        // signatures: [], so the merge never finds new signatures here.
```
momoshell, line 73: This throw is unreachable today because calldata events always carry `signatures: []`, but that's an invariant living over in `EventHandler.ts` rather than enforced here. Would it be worth a small test pinning "merging a calldata confirmation on an already-stored block returns `noNewSignaturesOnExistingBlock`, never throws".

Just so a future change to the event shape fails loudly here instead of surprising someone at runtime?

src/stateManager/ingest/StoredBlockMergeService.ts
daiagi: yes, intentional. the re-gossip is the part that differs between strategies (dispute withholds it, live/spectating send it) so it moved into the hook, and the notify stayed where it was. the broadcast is fire and forget anyway so the order doesn't really matter in practice. don't think it needs a comment in the code

test/harness/actions/TransitionActions.ts
daiagi: done

src/stateManager/validationStrategy/CalldataCommittedStrategy.ts
daiagi: done

lukastanisic99:
@daiagi should I review this one? Is it rebased? Adding the docs to this one?

lukastanisic99 (commented):
**Human note**

All valid cases

**AI-generated review**

🟠 **[TO6] The ingest suite still misses major public-boundary branches.** It does not cover stale versus unknown fork handling at ingest, authentication/general rejection, broken inbound state, failed transition, both unexpected-signer outcomes, live versus dispute success, or restoration after a post-transition throw.

> **Fix TO6-FIX**
>
> Drive public ingest with attributed queue entries and assert result, VM/queue restoration, punishment, persistence, and processed notification for every branch.



---

**Human note**

Most of my notes are test cases, so not sure if it's worth it, the agent can also generate them

**AI-generated review**

🟠 **[TO7] Participant-timeout tests miss the calldata and dispute-window races.** Existing-window handling, recovery retries, matching calldata delay, commitment-forced timeout, stand-down, and timeout evidence fields remain unproved.

> **Fix TO7-FIX**
>
> Add scheduled black-box cases for every recovery and commitment outcome and assert retry, stand-down, dispute choice, and evidence.



---

**Human note**

Nit, but true

**AI-generated review**

🟡 **[MY2] Test documentation still names files and method owners removed by this refactor.** The ownership map at [`E2E-DisputeManager.test.ts:6–14`](https://github.com/peer3to/state-channels-plus/blob/9224ad5ad427bad153963476b36b7d4c06ea0f08/test/e2e/E2E-DisputeManager.test.ts#L6-L14) points readers to `src/stateManager/DisputeValidationService.ts` and `src/stateManager/ValidationService.ts`, but those paths no longer exist; the implementations now live at [`stateManager/dispute/DisputeValidationService.ts`](https://github.com/peer3to/state-channels-plus/blob/9224ad5ad427bad153963476b36b7d4c06ea0f08/src/stateManager/dispute/DisputeValidationService.ts#L26) and [`stateManager/ingest/ValidationService.ts`](https://github.com/peer3to/state-channels-plus/blob/9224ad5ad427bad153963476b36b7d4c06ea0f08/src/stateManager/ingest/ValidationService.ts#L35). The lifecycle explanation at [`E2E-ParticipantLifecycle.test.ts:117–125`](https://github.com/peer3to/state-channels-plus/blob/9224ad5ad427bad153963476b36b7d4c06ea0f08/test/e2e/E2E-ParticipantLifecycle.test.ts#L117-L125) says the synchronous status transition happens in `StateManager.joinChannel()`, while the call now delegates through [`LocalP2pSigner.joinChannel`](https://github.com/peer3to/state-channels-plus/blob/9224ad5ad427bad153963476b36b7d4c06ea0f08/src/evm/signer/LocalP2pSigner.ts#L158-L167) to [`MembershipService.joinChannel`](https://github.com/peer3to/state-channels-plus/blob/9224ad5ad427bad153963476b36b7d4c06ea0f08/src/stateManager/membership/MembershipService.ts#L44-L59), which owns the pre-`await` status change. This does not change execution, but it makes the tests' architecture map and race explanation false exactly where the PR is redefining ownership.

> **Fix MY2-FIX**
>
> Update the ownership map to the `dispute/` and `ingest/` paths. Change the lifecycle comment to say that `MembershipService.joinChannel()` sets `PENDING_PARTICIPANT` before its first `await`, reached through `LocalP2pSigner.joinChannel()`.

src/stateManager/StateManager.ts
```diff
@@ -233,6 +208,39 @@ class StateManager<
             this.self,
             this.logger
         );
+        this.stateApplicationService = new StateApplicationService(
```
lukastanisic99, line 211: **Human note**

True

**AI-generated review**

🔴 **[FR1] Custom RPC constructors run before any extracted service exists.** `P2PManager` constructs the user RPC synchronously here, but the new services are assigned only at [lines 200–260](https://github.com/peer3to/state-channels-plus/blob/9224ad5ad427bad153963476b36b7d4c06ea0f08/src/stateManager/StateManager.ts#L200-L260). A custom RPC constructor reading a non-optional service receives `undefined` during startup.

> **Fix FR1-FIX**
>
> Initialize the services first, or delay custom RPC construction until `StateManager` is complete. Add a startup test whose custom RPC constructor reads a new service.

test/unit/BlockIngestService.test.ts
```diff
@@ -0,0 +1,424 @@
...
+                .control(observer)
+                .query.getBlockByHeight(forkId, 1)
+                .request();
+            expect(after!.onChainTimestamp).to.equal(onChainTimestamp);
+        });
+
+        // no test here: the stray-signature merge and its blacklisting is
+        // E2E-BlockQueueManager's "queued entry that becomes stored merges at
+        // queue timeout: strays stripped, supplier blacklisted".
+        it.skip("stray signatures on a stored block → stripped and supplier cut", function () {});
```
lukastanisic99, line 234: **Human note**

Ture

**AI-generated review**

🟠 **[TO1] The stored-block stray-signature component case is skipped.** The E2E queue workflow does not prove black-box stripping and exact supplier attribution inside `StoredBlockMergeService`.

> **Fix TO1-FIX**
>
> Replace the skip with an attributed queued entry and assert that only the stray signature supplier is disconnected.

test/unit/BlockProductionService.test.ts
```diff
@@ -0,0 +1,271 @@
...
+
+            expect(message).to.contain("Channel not open");
+
+            await Promise.all(restores.map((restore) => restore()));
+        });
+
+        // no test: a locally stored inbound block can only arrive from the
+        // on-chain event, which is chained by construction - defensive against
+        // a corrupted inbound store.
+        it.skip("pending inbound blocks do not form a chain → throws", function () {});
```
lukastanisic99, line 149: **Human note**

True

**AI-generated review**

🟠 **[TO2] The broken pending-inbound-chain guard has no running test.** This invalid-state branch can regress without changing the happy-path tests.

> **Fix TO2-FIX**
>
> Stage a malformed real inbound chain, call the public authoring path, and assert the error plus unchanged VM and height.

test/unit/BlockProductionService.test.ts
```diff
@@ -0,0 +1,271 @@
...
+            expect(block!.timestamp).to.be.greaterThan(clockBefore);
+            expect(block!.timestamp).to.be.at.most(clockAfter + 1);
+        });
+
+        // the was-in-the-past clamp (timestamp raised to previousTimestamp)
+        // needs the previous block's timestamp to exceed the author's local
+        // clock + 1. The authoring path caps every block at its author's
+        // now + 1 against a chain-synced clock, so no honest flow reaches it -
+        // it guards clock skew between peers the harness cannot stage.
+        it.skip("previous block ahead of the local clock → timestamp raised to it", function () {});
```
lukastanisic99, line 191: **Human note**

True, but some things here might change - regardless genuine test cases even though they're trivial

**AI-generated review**

🟠 **[TO3] The peer-clock-skew timestamp boundary is skipped.** Prompt and late-author cases do not cover raising the timestamp to a future previous block.

> **Fix TO3-FIX**
>
> Stage a previous block ahead of the author clock and assert that the new block uses the previous timestamp.

test/unit/MembershipService.test.ts
```diff
@@ -0,0 +1,170 @@
...
+
+            expect(await recorder.submissions()).to.deep.equal([]);
+            h.assert.dispute.noDisputes();
+            await recorder.restore();
+        });
+
+        // no test here: making it fire needs the joiner excluded for N turns,
+        // which is E2E-ForceJoinDispute's "should trigger force-join dispute
+        // after N turns of non-inclusion" staging.
+        it.skip("fires exactly at joinSubmissionHeight + participants + 1", function () {});
```
lukastanisic99, line 112: **Human note**

why it.skip - many tests are it.skip-ed ?

**AI-generated review**

🟠 **[TO4] The exact force-join firing boundary is skipped at component level.** [`MembershipService.maybeInitiateForceJoinDispute`](https://github.com/peer3to/state-channels-plus/blob/9224ad5ad427bad153963476b36b7d4c06ea0f08/src/stateManager/membership/MembershipService.ts#L147-L164) computes `N = participants.length + 1`, then fires only when `block.height === joinSubmissionHeight + N`. Both the `+ 1` and strict equality are boundary-sensitive: changing the formula to `participants.length` fires one block early, while missing the exact-height invocation means the dispute never fires on a later block. The existing [`E2E-ForceJoinDispute.test.ts`](https://github.com/peer3to/state-channels-plus/blob/9224ad5ad427bad153963476b36b7d4c06ea0f08/test/e2e/E2E-ForceJoinDispute.test.ts#L6-L51) starts with two participants, advances three blocks, and proves that a dispute eventually resolves with the joiner participating. It does not inspect the dispute recorder after block two, assert the submitted dispute's triggering height, or advance past the boundary while proving there is no second submission. Therefore an implementation that fires early can still satisfy the final E2E outcome. The only direct unit case covers the missing-submission-height no-op, while the exact-boundary case is explicitly [`it.skip`](https://github.com/peer3to/state-channels-plus/blob/9224ad5ad427bad153963476b36b7d4c06ea0f08/test/unit/MembershipService.test.ts#L75-L112).

> **Fix TO4-FIX**
>
> Create a pending join through the real join flow, suppress inclusion, and install the allowed record-only dispute probe. Commit blocks one at a time and assert zero submissions at `fireOnBlockHeight - 1`, exactly one carrying the active fork at `fireOnBlockHeight`, and no additional submission at `fireOnBlockHeight + 1`. Keep the existing full E2E for real on-chain dispute and reduction behavior.

test/unit/SnapshotAssemblyService.test.ts
```diff
@@ -0,0 +1,611 @@
...
+                .request();
+            const timestamp = await h
+                .control(writer)
+                .query.getClockTimeInSeconds()
+                .request();
+            const call = await h
+                .getPeer(writer.index)
+                .p2pInstance.p2pContractInstance.add.populateTransaction(1);
+
+            // plain-number header fields so the struct crosses the port raw
```
lukastanisic99, line 190: **Human note**

Ok if it's working, but in some cases I had to use Codec to serialize/deserialize

**AI-generated review**

🟠 **[AO1] New tests deliberately send raw TypeChain structs across the runtime port.** This transaction and later snapshots cross raw, and [`BlockIngestService.test.ts:253–265`](https://github.com/peer3to/state-channels-plus/blob/9224ad5ad427bad153963476b36b7d4c06ea0f08/test/unit/BlockIngestService.test.ts#L253-L265) does the same with `BlockConfirmationStruct`. Root AGENTS requires `Codec` for every ethers/typechain struct.

> **Fix AO1-FIX**
>
> Encode each struct before the port, decode host-side, and name every port value `encoded*`.

test/unit/BlockIngestService.test.ts
```diff
@@ -0,0 +1,424 @@
...
+                    );
+                    const model =
+                        sm.snapshotAssemblyService.getPreviousStateSnapshotOrThrow(
+                            { forkId: args.forkId, height: next }
+                        );
+                    // a genesis fork id IS the hash of its snapshot data;
+                    // the StateSnapshot class is reached through the live
+                    // instance (closures are closure-free)
+                    const otherForkId = model.snapshotDataHash;
+                    const SnapshotC = model.constructor as unknown as {
```
lukastanisic99, line 82: **Human note**

Nit

**AI-generated review**

🟡 **[AY3] The test restates model constructors through `as unknown as`.** The same pattern appears for `Block` at lines 138–141, which the repository type-safety rule bans.

> **Fix AY3-FIX**
>
> Move staging to a typed harness endpoint using the real `StateSnapshot` and `Block` classes.

test/harness/actions/TransitionActions.ts
```diff
@@ -393,4 +395,74 @@ export class TransitionActions<
...
         }
     }
+    /**
+     * White-box: run `tryMergeStoredBlockConfirmation` on the host, under the
+     * peer's live, spectating, calldata, or a fabricated dispute strategy.
+     * Returns the merge result and the persisted signature set.
+     */
+    async runStoredBlockMerge(options: {
+        peerIndex: number;
+        confirmation: { signedBlock: unknown; signatures: string[] };
```
lukastanisic99, line 405: **Human note**

Nit, but true

**AI-generated review**

🟡 **[AY4] The harness restates `BlockConfirmationStruct` and casts it back.** The real named type is already available.

> **Fix AY4-FIX**
>
> Type `confirmation` as `BlockConfirmationStruct` and remove the cast at lines 415–418.

lukastanisic99 (approved):
Overall, we can merge since there's nothing really critical and all of the test and fixes can be done in a followup + I guess the agent will generate 90% of them again when he makes the implementation report - however you prefer

PR #419, opened 2026-08-05
fix(workers): let workers drain on dispose instead of force-stopping

lukastanisic99:
- worker-side process.exit() / parent-side terminate() abort the whole process with "uv_loop_close() while having open handles" when close callbacks are still pending; workers now close their ports after replying to dispose and exit naturally, and the parent awaits that exit without a timeout (abandoning a live worker just moves the abort to process exit)
- the sdk worker closes every handle still referencing its loop (sockets, timers, servers) before closing the bootstrap port
- create Hyperswarm lazily on first join so debug-transport peers never own the native utp handles that tripped the same assert
- reject executor calls after disposal; cover the shutdown helper, post-dispose rejection, and threaded-peer teardown in unit + e2e

### DIFF
--- a/src/Holepunch.ts
+++ b/src/Holepunch.ts
@@ -13,22 +13,9 @@ class Holepunch {
     p2pManager: P2PManager;
     topics: Buffer[] = [];
     connectionCount = 0;
+    private readonly isBrowserRuntime: boolean;
     constructor(p2pManager: P2PManager) {
         this.p2pManager = p2pManager;
-        const setup = () => {
-            this.swarm.removeAllListeners(["connection"]); // since hyperwarm is injected into the runtime, creating a new Holepunch object still holds the same refrence to hyperwarm
-            this.swarm.on("connection", (socket: any, info: any) => {
-                this.p2pManager.logger.info("New holepunch peer connection", {
-                    connectionCount: ++this.connectionCount,
-                    transportType: "HOLEPUNCH"
-                });
-                this.p2pManager.logger.debug("Holepunch peer info", {
-                    peerInfo: info
-                });
-                new HolepunchTransport(socket, info, this.p2pManager);
-            });
-            this.rejoinTopics();
-        };
 
         // `window` is absent in a Web Worker, so RUN_SDK_IN_THREAD (Holepunch
         // running inside the SDK's browser worker) must also detect the worker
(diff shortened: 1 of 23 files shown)


sh3ll3x3c (approved):
Approving. I checked the CI failures this branch fixes, and found their root cause.

## Root cause of the red CI

`test/utils/HolepunchRelay.test.ts:30` calls `stubModule("hyperswarm", FakeHyperswarm)` at module top level. It overwrites `require.cache` and never restores it. `FakeHyperswarm` is an empty class — no `removeAllListeners`, no `on`, no `join`, no `destroy`.

CI runs all test files in one mocha process, and that file loads before `test/stateManager/*` and `test/rpc/*`. So every later test that builds a real `Holepunch` fails with `this.swarm.removeAllListeners is not a function`. The parallel runner forks one process per file, so it stays green locally.

Measured with one test and `--grep`, so the stub file loads but its own tests do not run:

| Branch | Files loaded | Result |
|---|---|---|
| `refactor/memory-array-shrink-helper` | target file alone | 1 passing |
| `refactor/memory-array-shrink-helper` | `HolepunchRelay.test.ts` + target | 1 failing — `removeAllListeners is not a function` |
| `worker-shutdown` | `HolepunchRelay.test.ts` + target | 1 passing |

This branch fixes it: the lazy swarm means the `Holepunch` constructor no longer touches the stubbed module.

It fixes the symptom, not the leak. The poisoned `require.cache` entry stays, so any test that reaches `Holepunch.join()` after that file loads hits `FakeHyperswarm` again — it has no `join` or `destroy` either. Restoring the cache entry in an `after()` hook in `HolepunchRelay.test.ts` would close it for good, here or as a follow-up.

## Two optional points, neither blocking

**1. `parentPort.once` in `onWorkerBootstrap`.** `once` removes the listener before it calls the handler, and Node unrefs and stops a `MessagePort` when its last `message` listener goes away. The handler then awaits `startP2pRuntimeHost`, whose first statement is `await createRuntimeChainContext(...)`, while the transferred runtime port still has no listener — `adaptPort` attaches one only inside `onMessage`, which runs after startup. So during bootstrap the thread stays alive only through whatever handle ethers happens to create. `closeWorkerBootstrapPort()` already ends with `port.close()`, so `.on` drains the same and keeps the margin.

**2. Wall-clock assertion in `E2E-WorkerShutdown.test.ts`.** `Date.now() - startedAt < 5000` can fail under the parallel runner as a plain assertion error rather than a starvation signature, which hides load as a real failure. It also does not cover the original bug — that was an abort at `uv_loop_close`, which elapsed time cannot detect. Asserting a clean worker exit would test it directly.

Minor, no action needed: `closeWorkerBootstrapPort` gives up silently after 10 passes and logs nothing, so the "visible hang that names the leaking teardown" described in `workerShutdown.ts` is not actually produced; `process._getActiveHandles()` is private Node API; and the new e2e test calls `harness.cleanup()` while the registered `afterEach` calls it again through `TestSession.clear()`.

I did not run the new tests in this PR, and did not run a clean typecheck of it.

PR #421, opened 2026-08-07
Distributed testing

lukastanisic99:
# How to run the distributed test pool

## Required setup

1. Check out the same revision of this branch on the orchestrator and every worker machine.
2. Install dependencies on every machine:

   ```bash
   yarn install --frozen-lockfile
   ```

3. Generate one long random pool secret:

   ```bash
   openssl rand -hex 32
   ```

4. Add that same secret to the ignored `.env` file on the orchestrator and every worker:

   ```dotenv
   SCP_TEST_POOL_SECRET=<the-same-random-secret-on-every-machine>
   ```

   The secret authenticates the pool and must only be shared with trusted machines. Workers execute code and dependency lifecycle scripts supplied by the authenticated orchestrator.

5. Ensure every machine can reach the public Hyperswarm DHT. No worker IP address or bootstrap server needs to be configured.

## Start each worker server

Run one persistent server per physical host. Give every host a unique, stable lowercase name:

```bash
yarn test:parallel:server --name worker-one
```

Example with explicit capacity controls:

```bash
yarn test:parallel:server \
  --name worker-one \
  --slots 1 \
  --workers 10 \
  --interval 500
```

- `--slots` controls the number of warm, isolated test-infrastructure slots. Use `0` for tests that do not need one.
- `--workers`/`-w` limits concurrent test processes and is clamped to the funded-account capacity.
- `--interval`/`-i` controls how often the scheduler admits another test.
- `--target-load` and `--mem-limit-gb` can further constrain admission.
- `--work-root <path>` moves the persistent workspace, pnpm store, and temporary leases from `temp/distributed-worker/` to another local disk.

The server remains available for later runs. Stop it with `Ctrl+C`/`SIGINT` or `SIGTERM`. Restart it after checking out a revision with a different distributed protocol version.

## Start the orchestrator

First run a smoke test:

```bash
yarn test:parallel:distributed \
  --test-pattern 'scripts/e2eParallelDistributedE2E.test.ts' \
  --discovery-timeout 60000
```

Then run the complete canonical suite:

```bash
yarn test:parallel:distributed
```

Useful variants:

```bash
# E2E tests only
yarn test:parallel:distributed --e2e-only

# Tests whose full Mocha title matches a regular expression
yarn test:parallel:distributed --grep 'pattern'

# Forward a required test environment variable (repeatable)
yarn test:parallel:distributed --forward-env VARIABLE_NAME
```

Do not forward `SCP_TEST_POOL_SECRET`; the runner deliberately excludes it. Results, failure logs, and worker infrastructure diagnostics are written under a fresh `logs/run-N/` directory on the orchestrator.

# Summary

This PR adds persistent, authenticated distributed execution to the parallel Mocha runner. An orchestrator discovers trusted worker servers over Hyperswarm, synchronizes the current source workspace, and schedules the normal test task set across available machines.

The implementation keeps the local `yarn test:parallel` path intact while adding `yarn test:parallel:server` and `yarn test:parallel:distributed` for multi-host runs.

## What changed

- Added authenticated worker/orchestrator discovery with role-specific topics, challenge proofs derived from `SCP_TEST_POOL_SECRET`, stable orchestrator identity, duplicate-connection convergence, and reverse-dial recovery.
- Added persistent worker workspaces and a shared pnpm store. Initial runs transfer the complete source set; later runs transfer only changed files and deletions while reusing dependencies and prepared build output.
- Preserved recursive `link:`/`file:` dependency layouts and prepared linked repositories in dependency order.
- Added host-level locking, bounded archives, checksum verification, path-containment checks, restricted child environments, and cleanup of complete per-lease runtime trees.
- Added per-host capacity controls for slots, process concurrency, CPU load, memory, and admission interval.
- Added queueing for concurrent orchestrators with live position, progress, and wait estimates.
- Made active runs resilient to worker loss: live assignments return to the shared queue, discovery remains active, and replacement workers can continue the run.
- Added speculative duplicate execution after the unassigned queue is empty. The first successful copy wins, while failed copies remain provisional when another copy is still running.
- Spool test output locally on workers without retaining a duplicate in memory. Successful tests return metadata only; only failures and infrastructure failures upload their logs to the orchestrator.
- Preserve worker colors across reconnects by worker ID/name and provide a 12-color worker palette.
- Added detailed discovery, connection, lease, workspace-preparation, scheduling, transfer, and cleanup diagnostics.
- Kept canonical logs and summaries on the orchestrator, including `error_*.ansi` failure logs and per-worker infrastructure logs.

## Worker storage lifecycle

The default worker root is `temp/distributed-worker/`:

```text
temp/distributed-worker/
├── pnpm-store/                 # persistent dependency cache
├── workspaces/<project-id>/    # persistent synchronized/prepared source
└── leases/lease-*/             # temporary runtime, infra, and output spools
```

Prepared workspaces and the pnpm store survive between runs. Lease directories are removed after completion, cancellation, failure, or orchestrator loss.

## Failure and recovery behavior

- A lost worker does not terminate the orchestrator while work remains; its assignments are returned to the queue.
- Servers remain available after lease cleanup and can accept the next queued or newly discovered orchestrator.
- Routine abandoned, duplicate, reverse-dial, and wrong-role handshakes do not flood terminals with authentication errors.
- Preparation, protocol, heartbeat, archive, spool, and discovery limits fail loudly instead of leaving a run hung.
- A successful test does not upload its spool. A failed test uploads and checksum-verifies its complete ordered stdout/stderr log before reporting the result.

## Verification

- `yarn tsc --noEmit -p tsconfig.json`
- `yarn tsc --noEmit -p tsconfig.browser.json`
- Focused `test/scripts/e2eParallel*.test.ts` suites
- Socket-backed distributed protocol and lifecycle tests
- Workspace cache, bundle/extraction, scheduler, coordinator, lease, logging, and preparation tests
- `git diff --check`

The local full-suite landing gate remains:

```bash
yarn test:parallel
```

The distributed full-suite gate is:

```bash
yarn test:parallel:distributed
```


daiagi (commented):
superseded — see the review below

daiagi (commented):
Used this infra to run the tests while developing the dispute-validation unit
tests. Results:

- smoke test: 12/12
- full gate, 5 workers, merged tree: 967 passing / 9 failing, 976 tasks, wall
  537s, 29.1x speedup
- 4 runs aborted during workspace preparation before one completed:
  `zlib: unexpected end of file` on server-3, server-8, server-7,
  `Source archive does not match its manifest` on server-1 (= D-01, fixed by #423)
- of the 9 failures, 4 shared revert selector `0xd15152cc`
  (`RaceConditionBlockCalldataTimestampTooLate`); load at the time was cpu avg
  86% / peak 100%, event-loop delay peak 9991ms, 6 tasks recovered from starvation

2 blockers, 7 majors below. D-01 excluded.

scripts/e2e-parallel/distributed/orchestrator.js
```diff
@@ -0,0 +1,736 @@
...
+            fs.appendFileSync(
+                logStore.infrastructurePath(worker.id, worker.label),
+                `${error.message}\n`
+            );
+            workerStatus(
+                worker,
+                `Preparation failed: ${message.header.message}`,
+                process.stderr
+            );
+            finishing = true;
```
daiagi: `PREPARATION_ERROR` sets `finishing = true` and calls `completedReject(error)`.
That rejects the promise `runDistributed` awaits at L702 -> the run ends while
other workers are mid-task.

`WORKER_ERROR` at L605-619 handles the same situation (worker can no longer
serve the run) by retiring only that worker.

Effect: run reliability is `1-(1-p)^N` on an N-host pool. 4 of my 5 aborted runs
were a single host failing preparation.

Deleting L601-603 alone hangs the run instead: `clearRediscoveryTimeout()` at
L418 runs on every new connection, so a worker that reconnects and re-fails
resets the deadline indefinitely. Needs a per-worker failure count:

```js
// declare next to ignoredWorkers (L215), same per-worker bookkeeping
const preparationFailures = new Map();

// replacing L601-603
const failures = (preparationFailures.get(worker.id) ?? 0) + 1;
preparationFailures.set(worker.id, failures);
if (failures >= 2) ignoredWorkers.add(worker.id);
if ([...workers.keys()].every((id) => ignoredWorkers.has(id))) {
    finishing = true;
    clearRediscoveryTimeout();
    completedReject(error);
}
```

`retireWorker` already requeues via `coordinator.disconnectWorker` and re-arms
rediscovery when the last worker goes. `server.js:533-544` already releases the
lease and cleans the runtime after `PREPARATION_ERROR`.

Note `ignoredWorkers` (L215) is the natural home for "stop using this host", but nothing currently reads it — L349-350 only use it to avoid repeating the protocol-mismatch warning. The post-auth path needs to actually skip workers in the set for this to take effect.

scripts/e2e-parallel/distributed/orchestrator.js
```diff
@@ -0,0 +1,736 @@
...
+                ),
+            options.discoveryTimeoutMs
+        );
+    }
+    for (const publicKey of options.peerPublicKeys || []) {
+        pool.swarm.joinPeer(publicKey);
+    }
+
+    const coordinator = new TaskCoordinator(options.tasks, {
+        speculative: true,
```
daiagi, line 300: `speculative: true` is unconditional, no flag to disable.

Sequence when a task fails on worker A while a duplicate runs on worker B:

1. A reports code 1 -> recorded as provisional failure, `onResult` not called
   (`taskCoordinator.js:228-249`)
2. B reports code 0 -> `finalize` runs
   `provisionalFailures.delete(assignment.taskId)` (`taskCoordinator.js:207`)
3. `failed[]` never sees it, run exits 0
4. A's log was uploaded (`server.js:54-56`) but `promoteAttemptLog` only runs for
   `disposition === "complete"` (L268), so `cleanupNonErrorLogs`
   (`logging.js:244`) deletes it

Result: a failure is reported as a pass and its log is removed. 967/9 is a lower
bound. `e2eParallelTaskCoordinator.test.ts:134-165` asserts this behaviour.

```suggestion
        speculative: options.speculative === true,
```

Two follow-ups:
- a dropped provisional failure should be recorded as a disagreement, its log
  promoted to `error_`, and printed
- `finalize` (`taskCoordinator.js:210-213`) drops sibling assignments but never
  signals the worker to kill the process -> it runs to completion, then its
  result is rejected as stale

scripts/test-e2e-parallel.js
```diff
@@ -162,17 +177,109 @@ async function main(options = {}) {
...
+                );
+            const transferRoot = path.join(logDir, "distributed-transfer");
+            const archivePath = path.join(transferRoot, "source.tgz");
+            try {
+                const manifest = await buildRuntimeBundle(
+                    process.cwd(),
+                    archivePath,
+                    console.log
+                );
+                const stats = await runDistributed({
```
daiagi, line 226: `--slots`, `-w/--workers`, `--target-load`, `-i/--interval`, `--mem-limit-gb`
are resolved at L117-132 and not passed to `runDistributed`. `argParser.js:301`
guards the reverse direction only.

`runHeader` prints at L148, before the branch. So
`yarn test:parallel:distributed --workers 4 --slots 4` prints `concurrencyCap=4`
while each host uses `serverArgParser.js:8-21` defaults: `workers` =
`MAX_SLOTS_FROM_POOL` = 40, `targetLoad` 0.8, `memLimitGb` 0.8 x totalmem.

Fix: forward them in `RUN_CONFIG` (L476-479) as an upper bound the server clamps
to at `server.js:704-720`, or reject them under `--distributed`. Move `runHeader`
into the local branch either way.

Separate, pre-existing: `ResourceGate.allows` (`resourceGate.js:128-136`) admits
one process per tick with no ramp-up model, and reserves headroom for one average
test while N are still starting.

scripts/e2e-parallel/distributed/server.js
```diff
@@ -0,0 +1,777 @@
...
+}
+
+function progressElapsedMs(connection, now = Date.now()) {
+    const startedAt =
+        connection.runStartedAt || connection.leaseStartedAt || now;
+    return Math.max(0, now - startedAt);
+}
+
+function shouldTransferAttemptLog(result) {
+    return result.code !== 0 || Boolean(result.infrastructureFailure);
```
daiagi, line 57: `shouldTransferAttemptLog` returns false for `code === 0`, worker runs with
`captureOutput: false` (`worker.js:228`), orchestrator passes `stdout: ""` to
`completeAttempt` (`orchestrator.js:538-543`).

`reduceAttempt` derives `starveCount`, `oomCount` and all timing metrics from
that string (`taskCoordinator.js:3-25`).

So for a test that logs the watchdog line and exits 0:
- local: `starveCount > 0` -> retry, second hit forces code 1 via
  `repeatedStarvation` (`taskCoordinator.js:120-136`)
- distributed: `countStarvation("")` = 0 -> pass, no retry, no record

Also means the summary timing table is populated only from failing tests in
distributed mode.

Fix: reduce on the worker and put `{oomCount, starveCount, timing}` on the
`ATTEMPT_RESULT` header. Keeps the no-upload-on-pass bandwidth saving.

package.json
```diff
@@ -66,12 +67,14 @@
     "buffer": "^6.0.3",
     "ethereum-waffle": "4.0.10",
     "fflate": "^0.8.2",
+    "fs-ext": "^2.1.1",
```
daiagi, line 70: `fs-ext` (L70), `tar` (L77), `@hyperswarm/dht` (L61) are in `dependencies`. None
is referenced under `src/`; consumers are
`scripts/e2e-parallel/distributed/{hostLock,runtimeBundle,runtimeExtractor,orchestratorIdentity}.js`
and `test/fixtures/distributed/testTransport.ts`.

`fs-ext` pulls `nan` and needs node-gyp. Its chain includes `nopt@10.0.1`,
`engines: ^22.22.2 || ^24.15.0 || >=26.0.0`. On node 22.22.0
`yarn install --frozen-lockfile` fails — that is step 2 of this PR's own
description. No `engines.node` floor is declared.

`test/scripts/e2eParallelWorkerLease.test.ts:176-194` calls `acquireHostLock`
directly, and `hostLock.js:18-27` requires `fs-ext` unconditionally -> the repo's
test suite fails on any host without a built binding, orchestrator or worker.

Fix: move all three to `devDependencies`. For the lock itself, `O_CREAT|O_EXCL`
plus owner pid and a `process.kill(pid, 0)` staleness check is equivalent here,
needs no native code, and works on Windows — `hostLock.js:13-17` currently
refuses Windows outright.

test/e2e/disputeValidation/stateProof/case3_signedBlocksOnly.test.ts
```diff
@@ -61,7 +61,7 @@ describe("E2E: dispute validation / stateProof / Case 3 (signedBlocks-only)", fu
     });
 
     describe("stateProof.signedBlocks[0].transaction.header.transactionCnt != 0", function () {
-        it("first signedBlock height is not 0 → DisputeInvalidStateProof", async function () {
+        it("first signedBlock height is not 0 → DisputeInvalidBlockStructure", async function () {
```
daiagi, line 64: Expected proof changes from `DisputeInvalidStateProof` to
`DisputeInvalidBlockStructure`.

This PR touches 0 files under `src/` and 0 under `contracts/`, so this is not a
behaviour change. It follows from `MathScenarioActions.ts:217-227`, which moves
the JOIN in `preDisputeSetupDisconnectedPeer` to after two transitions, changing
the fork's join point.

Two consequences:
- the file is under `stateProof/` and its describe block is
  `stateProof.signedBlocks[0].transaction.header.transactionCnt != 0`, but it now
  asserts a block-structure proof. Nothing else covers the state-proof path.
- `preDisputeSetupDisconnectedPeer` is also used by `case4_blockInjection.test.ts`
  and `disputeInputFields/latestStateSnapshotHash.test.ts`, which now build
  crafted proofs against a different topology.

scripts/e2e-parallel/distributed/orchestrator.js
```diff
@@ -0,0 +1,736 @@
...
+    );
+    return {
+        peakCpu: Math.max(0, ...stats.map((entry) => entry.peakCpu)),
+        avgCpu: cpuSamples
+            ? stats.reduce(
+                  (sum, entry) => sum + entry.avgCpu * entry.cpuSampleCount,
+                  0
+              ) / cpuSamples
+            : 0,
+        peakOccupiedGb: stats.reduce(
```
daiagi: `aggregateWorkerStats` uses four different reductions in one row:
`peakOccupiedGb` and `memBoundGb` are sums across hosts, `peakCpu` is a max,
`avgCpu` is a sample-weighted mean.

The reported "peak owned 36.1GB" is a sum of five peaks that did not co-occur;
"peak cpu 100%" does not identify the host.

Neither answers whether a single host was overcommitted, which is what [the capacity-flag comment](https://github.com/peer3to/state-channels-plus/pull/421#discussion_r3743005409)
raises. `completedByWorker` and per-worker `stats` on `leasedWorkers` are already
retained.

Fix: print a per-worker line (peak cpu, peak owned GB / bound, tests completed)
and label the aggregate as a sum of peaks.

scripts/e2e-parallel/distributed/artifactTransfer.js
```diff
@@ -0,0 +1,183 @@
...
+    }
+    const offered = new Set(manifest.files.map((entry) => entry.path));
+    if (
+        need.changed.some(
+            (entry) => typeof entry !== "string" || !offered.has(entry)
+        )
+    ) {
+        throw new Error("Worker requested a file outside the offered manifest");
+    }
+    const delta = await buildDeltaBundle(manifest, need.changed, archivePath);
```
daiagi: `buildDeltaBundle` derives `fileCount` / `expandedBytes` from `manifest.files`
rather than from the tar it wrote. `runtimeExtractor.js:60-65` validates the
received archive against those numbers, so the two can disagree independently of
the path collision #423 fixes.

Fix: compute them from the tar entries actually written.

Also in this file's subsystem: `buildRuntimeBundle`
(`test-e2e-parallel.js:214-218`) tars, gzips and sha256s the whole workspace;
`WORKSPACE_OFFER` sends only `wireManifest` + `files`, `BUNDLE_META` overwrites
the archive fields with the delta's (L70-72), and the first `sendBundle` deletes
the file (`runtimeBundle.js:251`). Nothing reads it.

scripts/e2e-parallel/distributed/workerLeaseManager.js
```diff
@@ -0,0 +1,133 @@
...
+
+    assertActive(connection) {
+        if (this.active !== connection)
+            throw new Error("Connection does not own the lease");
+    }
+
+    grantNext() {
+        const next = this.waiters.shift();
+        if (!next) return;
+        this.active = next;
```
daiagi, line 142: Two concurrent syncs into one workspace directory produce a merged tree. Measured
with two 361-file sources into the same destination: 421 files, both exit codes
0, no error output. Each run's deletion set was computed from a pre-write listing,
so neither removes the other's files.

`workRoot/workspaces/<id>/workspace` is keyed on workspace shape and shared across
leases, with no file lock and no temp-then-rename. The only serialization is the
in-memory `WorkerLeaseManager`, which does not survive a server restart and is
bypassed by `--allow-shared-host`.

Not addressed by #423 or by [the artifactTransfer comment](https://github.com/peer3to/state-channels-plus/pull/421#discussion_r3743005424) — both concern how bytes move, not
destination ownership.

Fix: `O_EXCL` lock on the workspace directory held across prepare + run, refuse
the lease on conflict.

package.json
lukastanisic99: @daiagi I don't have a quick fix for this, so can you add a trello cards? Here's a comment: 

The browser bundle does not import these modules: it resolves the separate `dist/browser` entry, while the distributed runner is published through Node-only subpath exports. However, ordinary dependencies are installed before export resolution, so `fs-ext`'s native `node-gyp` build can still prevent a browser-only consumer from installing the SDK. I agree that this packaging cost should be removed from the main SDK, but moving the modules directly to `devDependencies` would break the currently exported runner/server/worker. We should either stop publishing that infrastructure and make the modules development-only, or move the infrastructure into a separate package with its own runtime dependencies; making `fs-ext` optional is a smaller intermediate option. The cited `nopt@10` chain is not in this lockfile, and a stale PID file is not equivalent to an OS-held flock.

test/e2e/disputeValidation/stateProof/case3_signedBlocksOnly.test.ts
lukastanisic99: So this is correct and we're missing a test case now after the changes?

lukastanisic99:
@daiagi @sh3ll3x3c I've pushed changes to account for the review comments - both of you observed the prepare build failure, which was a mistake on my end (I didn't push 2 of my commits that fixed it yesterday) - Either way I merged #423 and rebased on top of that and updated the server workers to have this latest stuff - pls rereview

scripts/e2e-parallel/distributed/orchestrator.js
daiagi: quarantine's good, the :374 check does it.

`WORKER_ERROR` at :647 didn't get it though, still a plain `retireWorker`. so a host that dies at test worker boot reconnects, clears the deadline at :451, dies again, and round it goes. with one host in the pool that hangs instead of failing.

same call as :634:

```js
const failure = recordPreparationFailure(
    preparationFailures,
    quarantinedWorkers,
    worker.id
);
if (failure.quarantined) {
    workerStatus(
        worker,
        `Quarantined after ${failure.failures} failures`,
        process.stderr
    );
}
retireWorker(worker, "test worker reported a fatal error");
```

scripts/e2e-parallel/distributed/orchestrator.js
daiagi: coordinator takes the flag now but :287 still passes `true` hardcoded, so nothing's different at runtime.

`finalizeOrDefer` only promotes the provisional when the second result also fails (taskCoordinator.js:292). a passing duplicate goes straight to `finalize`, which drops it at :251 and never touches `failed`, so the test reports green and `cleanupNonErrorLogs` bins the log.

of everything i flagged this is the one i'd actually block on.

```suggestion
        speculative: options.speculative === true,
```

package.json
daiagi: yeah you're right. nopt@10 isn't in the lockfile, only nopt@3.x, and node-gyp isn't in there at all, so that engines thing doesn't apply here. and i missed the exports, devDeps would break `./test-parallel*` and `files: scripts/e2e-parallel`. pid file vs flock is fair too, flock dies with the process and pids get reused.

lazy require keeping the loader error (hostLock.js:29) is better than what was there anyway.

https://trello.com/c/ojo2QI3S

test/e2e/disputeValidation/stateProof/case3_signedBlocksOnly.test.ts
daiagi: correct, though less than i made it sound. DisputeInvalidStateProof is still covered in case1_inboundDivergence, invalidStateProofGenesisLinkage, latestStateSnapshotHash and E2E-DisputeManager.

what we lost is just this rule, `signedBlocks[0].transaction.header.transactionCnt != 0` getting rejected by the state proof. moving the JOIN (MathScenarioActions.ts:217-228) means that input hits the structural check first and comes back DisputeInvalidBlockStructure, so nothing asserts the state proof arm now.

wouldn't block the pr on it, trello + a test that gets there under the new topology.

scripts/e2e-parallel/distributed/workerLeaseManager.js
```diff
@@ -0,0 +1,133 @@
...
+        this.assertActive(connection);
+        this.state = "cleaning";
+        this.notifyWaiters();
+        try {
+            await cleanup();
+        } catch (error) {
+            this.onFault(error);
+        }
+        this.active = null;
+        this.state = "idle";
```
daiagi, line 83: cleanup throwing before `lock.release()` (leaseRuntime.js:55) is fine, you don't want to release a workspace that might still have writers in it. it's what happens after.

release() swallows it at :67 and goes idle + `grantNext()` anyway, and server.js:118 prints "worker remains available with a fresh lease directory". the lease dir is fresh, the workspace lock isn't, so the next lease on the same workspaceId (server.js:339) hits flock held by this same process. `cleaning` is memoised so it never retries either. host is dead for every run after that until it's restarted.

should stop granting on fault rather than re-advertise.

scripts/test-e2e-parallel.js
daiagi: rejected under `--distributed` now (argParser.js:325) and the header's skipped, works. side effect is a distributed run prints no header at all, so there's nothing showing what capacity it actually ran with.

scripts/e2e-parallel/distributed/server.js
daiagi: worker reduces the spool now and puts `reduced` on the result, validated coordinator side, local path untouched. good.

scripts/e2e-parallel/distributed/orchestrator.js
daiagi: per-worker line plus the "sum of worker peaks" label, good.

scripts/e2e-parallel/distributed/artifactTransfer.js
daiagi: counts come from the tar entries now. checked full/partial/empty deltas against the lockfile's tar and they match what's written.

scripts/e2e-parallel/distributed/workerLeaseManager.js
daiagi: lock's held across prepare+run and `--allow-shared-host` doesn't bypass it any more, workspaceId is validated hex too.

leftover is the fault path, separate comment: https://github.com/peer3to/state-channels-plus/pull/421#discussion_r3743817482

lukastanisic99:
@daiagi think I've addressed all critical ones

daiagi (approved):
ran the new script tests: 51 passing, 2 failing, both `e2eParallelWorkerLease` on the fs-ext binding my computer can't build (the trello card).

case3 state proof rule still needs its own follow-up, not a blocker here.

scripts/e2e-parallel/distributed/orchestrator.js
daiagi: budget is on WORKER_ERROR now, recordWorkerFailure + quarantine, same as the preparation branch. good.

scripts/e2e-parallel/distributed/orchestrator.js
daiagi: better than what i suggested. `provisionalFailures` is gone, the first result finalizes so nothing can delete a recorded failure, and a late losing duplicate lands as `late-failure` with its log promoted instead of binned.

scripts/e2e-parallel/distributed/workerLeaseManager.js
daiagi: `faulted` state does it, stops granting rather than re-advertising a host that still owns the flock, and the waiters get told it needs a restart.

PR #422, opened 2026-08-07
feat(discovery): channel discovery — lobby, ad protocol, intent-then-commit acquire

momoshell:
## What

Lets a peer find channels that already exist and join one, or meet a stranger in a shared lobby and open a new channel together. Negotiation stays 1v1 — two peers agree, everyone else joins the opened channel afterward.

Off by default: `LOBBY_ENABLED` is `false`, so nothing changes for existing consumers until it's switched on.

## Why it's built this way

**The lobby runs on its own swarm, not the channel stack.** This is the decision that sets the size of this PR, so it's the one worth challenging first.

Reusing the existing RPC/transport stack would have been much less code, and I didn't take it. On that stack, any peer completing the handshake lands in the channel broadcast set; a dispute sweep blacklists every connected peer; closing a channel disconnects them all; and closing a holepunch transport bans that peer at the DHT level. A lobby is full of strangers you haven't agreed to play with, so on that stack it would leak channel traffic to them and ban itself during normal operation.

Making reuse safe means changing those four behaviours in shared code that disputes and settlement depend on. I chose to leave that code alone and give the lobby its own swarm and a small signed protocol instead — no `ATransport`, no `P2PManager`. That buys isolation at the cost of roughly 1,500 lines that would otherwise not exist.

**The alternative is still on the table.** Narrowing those four behaviours instead would collapse most of `src/discovery/` into a service on the existing stack: materially less code here, in exchange for changes to the paths disputes and settlement run through. That's a real trade rather than a settled question, worth raising in review if the we would rather spend the risk than the lines.

**Acquisition races a cheap question, not real connections.** Hyperswarm gives one connection per peer, an instance holds one channelId and one negotiation slot, and dropping a loser would ban or blacklist it. So we ask candidates "will you have me?", commit to the first yes, and politely release the rest. The expensive part — connect, sync, on-chain join — stays sequential because it has to.

**Ads are hints, never authority.** They're unsigned: they only ever arrive over a connection whose identity the handshake already proved, so a signature would restate that. Nothing trusts an ad — the commit path re-verifies on chain, so a lying ad costs a wasted round-trip, never money. The signature field is reserved for when ads get relayed rather than delivered directly.

**Declining isn't an accusation.** Choosing your counterparty is the whole point of 1v1 negotiation, so "no thanks" now sends an abort instead of a blacklist. Objective fraud — bad signature, impersonation, mismatched terms — still blacklists.

**The ad store assumes an adversary.** Ads are validated on receipt, capped per peer and globally, dropped when their advertiser disconnects, and a peer can only ever evict its own entries — one spammer can't push everyone else's tables out.

## Also in here
- `setStakeAmount`: the stake was previously unreachable from outside, so consumers silently opened channels at a hardcoded default.
- Per-topic `Holepunch.leave` and a negotiation re-arm, so a failed attempt can fall back to the next candidate instead of being one-shot.
- `RelayerPool` extracted from `HolepunchRelay` so the lobby's relay reuses the existing failover/backoff instead of a second copy.

## Size

The diff looks large; most of it isn't logic.

| | Lines |
|---|---|
| Production code | ~4,240 |
| In-process lobby hub (test-only, lets the e2e run without a live DHT) | 452 |
| Tests | 8,955 |
| **Total** | **13,661** |

Two thirds of this is tests. The production code is one new directory (`src/discovery/`) plus wiring; nothing existing was restructured, and no Solidity changed.

The test weight is deliberate: this is peer-to-peer code where the failure modes are adversarial (a peer that lies in an ad, floods the store, declines forever, or disconnects mid-commit), and those cases are cheap to test and expensive to discover in production. The e2e suite earned that during development — it caught a case where the table creator's own buy-in wasn't being applied.

## Deliberately left as seams

Three things are stubbed rather than built, each in one place, so the next step doesn't need a redesign:

- **Racing several candidates.** `parallelism`/`maxWinners` are accepted and rejected with a typed `unsupported`. Callers already pass them, so enabling a real race later doesn't change any call site.
- **Per-peer admission.** Accept/decline runs through one policy object with a permissive default. Wiring an app-level "do I want to play this person?" decision means supplying that object — the decline path it feeds is already built and non-punitive.
- **Relayed ads.** Ads carry an unused `signature` field. Today they're only accepted directly from an authenticated peer, so a signature would add nothing; it's there so gossiping ads through third parties later is a field-fill rather than a format change.

## Where to look

New code is confined to `src/discovery/`; everything else is wiring.

| Path | What it is |
|---|---|
| `discovery/ChannelAd.ts`, `lobbyTopic.ts` | The ad wire type and how a lobby topic is derived. **Start here** — everything else assumes these. |
| `discovery/LobbyProtocol.ts` | Frame encoding and the challenge/response that authenticates a lobby peer. |
| `discovery/LobbyClient.ts` | Lobby membership: connect, authenticate, exchange ads, hold/release intents. The biggest file. |
| `discovery/LobbyAdStore.ts` | Ad state: TTL, caps, eviction rules, drop-on-disconnect. |
| `discovery/ChannelAcquisitionCoordinator.ts` | The acquire path: ask, commit, fall back, release losers. |
| `discovery/AdmissionPolicy.ts` | Whether to accept an inbound request, and why a decline isn't punitive. |
| `discovery/lobbySwarm/` | Node/browser transport seam, plus an in-process backend so the e2e runs without a real DHT. |
| `rpc/services/openChannelNegotiation/` | The decline-instead-of-blacklist change, `setStakeAmount`, and the re-arm that makes fallback possible. |
| `evm/P2pInstance.ts`, `evm/signer/*`, `evm/p2pRuntime/*` | The `discovery` facade and its trip across the runtime port. |
| `Holepunch.ts`, `HolepunchRelay.ts`, `transport/relay/RelayerPool.ts` | Per-topic leave, and the relay failover logic extracted for reuse. |
| `utils/config.ts`, `utils/Codec.ts`, `types/ethers.ts` | Config keys and wire-type registration. |
| `test/discovery/**`, `test/e2e/E2E-Lobby`, `E2E-ChannelAcquire` | Unit suites, and the two real-chain e2e suites. |

Reading it in that order follows the data: wire format → authentication → membership → ad state → acquisition.

## Tests
Full e2e suite green (219/219), including two new suites on a real chain: two peers meet in the lobby and open a channel at the agreed stake, a third discovers and joins it, a stale ad is rejected at commit with no funds moved, and a counter asserts zero wrongful blacklists or bans.

## CI note
`yarn build` is already broken on `dispute` by a stale call in `test/e2e/E2E-BlockQueueManager.test.ts:175` — type-only, predates this branch, and the e2e suite runs fine. No Solidity changed here.


lukastanisic99:
@momoshell my thoughts bellow 

--------------------------------------

# Discovery should be chain-first, with the lobby as a fallback

The overall direction is good: discovery and channel traffic should stay partitioned. I do not think the lobby should be the entry point for all discovery.

For existing channels, start from the chain. Query the available game or channel IDs, then try those channel partitions directly. We can work out filtering and ranking later. The first version only needs a usable list of IDs and a direct connection attempt for each one.

Create a fixed pool of separate SDK and P2P instances, up to a small concurrency ceiling such as five. Each instance runs setup once and owns one direct spectate attempt at a time. Assign each instance a channel ID, set that ID, and run the existing spectate and sync path.

A candidate succeeds only when spectating completes and the sync-completed hook or result confirms that the peer is usable. At that point, use that successful instance to start the join process. Spectating takes a short timeout parameter. If no completion signal arrives within that window, report the slot as failed, abort the cheap spectate attempt, and signal the same prepared SDK instance to try the next channel ID. Do not create and destroy a fresh SDK setup for every candidate, and do not wait forever for an unknown peer.

When every chain candidate has failed, timed out, or been rejected, then use the lobby. That is where it makes sense to find peers who are looking to open a new channel, or to negotiate when there is no suitable existing channel. The lobby should be optional fallback infrastructure, not a dependency for every peer that wants to join.

I also do not think this needs a separate lobby protocol and transport stack. The reason given was that reusing the custom RPC path would add lobby peers to the normal P2P manager and therefore send channel broadcasts to peers that are not in the channel. That is a coupling problem in the handshake lifecycle, not a reason to build a second protocol.

`InitHandshakeService` already has the right boundary. After the handshake proves the peer identity, register or update the peer in `ProfileManager`, but do not immediately add that transport to `P2PManager.openConnections`. Emit a `handshakeCompleted` P2P hook with the verified peer and transport. The caller can then choose what happens next:

- For a lobby connection, use the custom RPC service to negotiate discovery, but do not promote the transport into the channel peer set.
- For a direct connection to a chosen channel, promote and track the transport, then continue with the existing sync or channel-negotiation path.

This keeps the two cases separate without duplicating authentication, framing, transports, timeouts, and cleanup.

`ProfileManager` is still useful before promotion. It records the verified address-to-transport mapping, lets targeted RPC look up the authenticated peer, and keeps identity and blacklist state. It does not define the active connection set. `P2PManager.openConnections` is the set used for broadcasts, `getConnectedPeers`, and connection-wide cleanup. Registering a profile must not imply that the peer receives channel traffic.

The cleaner shape is to make `InitHandshakeService` fully lifecycle-neutral after verification. It should verify the peer, update `ProfileManager`, and emit one `handshakeCompleted` P2P hook with the peer identity and transport. It should not decide whether to sync, negotiate, or promote the transport itself.

The service that created the transport subscribes to that hook and owns the next step. The handshake, profile, and transport layers stay generic: they do not need to know whether the peer came from the lobby or a game. When the service selects a concrete game or channel, it adds the transport to `P2PManager` and starts sync or negotiation. That promotion is the only relevant distinction from the networking perspective.

Another reason to keep the existing flow and services is that we reuse the complete transport lifecycle. After the initial handshake, the shared stack can upgrade a Holepunch connection to WebRTC in the background. The lobby service does not need to know which transport is active or add a special case for the upgrade; `ProfileManager` preserves the verified identity and manages the replacement.

The same behavior then works for lobby, direct channel, and future connection flows. Discovery adds lobby-specific RPC endpoints and admission rules, not a second authentication protocol, framing layer, swarm wrapper, or transport lifecycle.

The broadcast concern should be addressed at promotion time. It is not a reason to build a separate protocol.

The network is already partitioned by what peers are trying to do. Peers in a channel connect to the other peers in that channel. Peers using the lobby connect to other lobby peers. A lobby peer is not automatically a peer in an active channel, and a channel participant does not need to join the lobby. If we keep those connection groups separate, channel broadcasts have no route to lobby peers.

That partitioning is also the scaling story. We should connect peers within the relevant channel partition or the lobby partition, rather than put every discoverable peer into one shared channel connection set. The handshake change makes that explicit: a completed lobby handshake does not create a normal channel connection; direct channel selection does.

The connection does not need a separate policy because it came from the lobby or a channel topic. Use one shared Hyperswarm so the same peer is deduplicated into one physical connection. The relevant owner is the `P2PManager` that holds the connection. `ProfileManager` tracks the peer identity and its transports; the active transport can be Holepunch or WebRTC.

`ProfileManager` must be the only owner of ban and unban policy. A completed Holepunch handshake exposes the peer-info through the transport, but neither `HolepunchTransport`, the lobby RPC service, nor transport-upgrade code may call `peerInfo.ban()` directly. They ask `ProfileManager` to apply the policy for that identity; it knows the active transport and invokes the required Holepunch, WebRTC, or future transport behavior. Use that policy in two explicit cases:

- When the profile is explicitly blacklisted, ban the Holepunch peer and never unban it.
- When the connection upgrades from Holepunch to WebRTC, temporarily ban the Holepunch peer so Hyperswarm does not reconnect the relay transport while the preferred WebRTC transport is healthy.

When the WebRTC transport closes, `ProfileManager` inspects the profile. If the peer is not explicitly blacklisted, it unbans the stored Holepunch peer-info so Holepunch may reconnect as the fallback transport. If the profile is blacklisted, it keeps the ban. A normal transport close must only close the socket; it must not silently choose either ban policy.

The lobby service should normally return a busy or declined response rather than ban a peer. If it ever needs temporary reconnection suppression, it must request it through `ProfileManager` and let that single owner decide when to clear it. All ban and unban calls therefore follow one lifecycle and cannot drift between lobby, channel, Holepunch, and WebRTC code.

This needs the latest Hyperswarm API. The repository currently uses Hyperswarm 4.3.5, whose installed API has no usable unban operation. The latest API supports `peerInfo.ban(true)` and `peerInfo.ban(false)`, which matches this upgrade-and-fallback lifecycle. Upgrade the dependency only with compatibility coverage for permanent blacklist, successful WebRTC upgrade, WebRTC failure fallback, and ordinary Holepunch close.

Keep one shared Hyperswarm for lobby and game or channel connections. This avoids duplicate physical connections to the same peer. The profile-owned policy, rather than swarm separation, decides whether that peer is currently allowed to reconnect and how that decision maps to each transport.

lukastanisic99:
@momoshell there are conflicts, I guess it's not ready for review?

momoshell:
Reworked against the review and rebased onto `dispute` (`7bad54a0`). Fourteen commits, meant to be read in order — the ordering is the argument.

## Chain-first discovery

`acquireChannel` with no explicit candidates now enumerates open channels from chain, probes them, and joins the first that proves usable; the lobby is consulted only after every chain candidate has failed. So an outsider can join what already exists without touching the lobby, and the lobby goes back to being where peers meet to open a *new* channel. Failures in either stage fall through rather than sinking the acquire. Explicit candidates skip the chain phase — which is what opening a new channel needs, since the chain can't answer that by definition.

## The lobby has no protocol of its own

Its swarm join, framing and challenge/response auth are gone. It's an opt-in `LobbyService` + `LobbyRpcMethods` pair wired like `OpenChannelNegotiationService`: `joinLobby()` is a `holepunch.join(topic)` on the shared swarm, and `HandshakeCompletedGuard` replaces the hand-rolled auth.

The first three commits are what made that safe, and they're the fix at the source you asked for. `InitHandshakeService` is lifecycle-neutral — it verifies, updates `ProfileManager`, emits `handshakeCompleted`, and decides nothing about syncing, negotiating or promoting. Promotion into `openConnections` is gated on channel participation, so a lobby peer is authenticated and reachable by targeted RPC but never receives channel broadcasts. `ProfileManager` is the only caller of `peerInfo.ban()`, with the cases you specified; a normal close no longer bans.

Domain logic carried over unchanged — ad envelope, topic derivation, TTL/cap eviction, admission policy, reservations. Only delivery changed. History was rewritten, so the branch never contains the deleted stack: reviewing commit-by-commit never means reading code that was later thrown away.

## Concurrent probing

Probing was serialized because checking a candidate went through `connectToChannel` + wait-for-`SYNCED`, which sets `channelId`. That path was never needed: spectate is request/response, so `SpectateService.probeChannel` validates a peer's proof **without applying it**. Applying is what mutates local EVM state; a probe is a round-trip plus a pure decode, so any number run at once on one instance. Arming happens once, later, at commit. It's non-punitive by design — a peer that can't serve a channel we merely asked about isn't misbehaving.

lukastanisic99:
## What I like

- I agree with the direction: channel discovery, and eventually a network where peers can collectively help with discovery, data availability, and other services, is where the system should go.

- Using `eth_getLogs` to find channel IDs is an acceptable simple first step. It gives us an intermediate path without first building indexer infrastructure.

- The PR is trying to make the system more capable once a peer wants to find or create a channel. That direction is right, even though I disagree with the scope and implementation in this change.

- I like the [`ProfileManager` handling of Holepunch ban policy and WebRTC upgrades](https://github.com/peer3to/state-channels-plus/blob/0fd362f81c7a5f3b306d587c42533c3926b06b24/src/ProfileManager.ts#L150-L195). The policy belongs to the peer identity and correctly survives a transport upgrade.

## Feedback and concerns

### Keep the interim implementation deliberately simple

I agree with the direction this is heading. The final system should let peers help other peers with discovery, indexing, gossip, data availability, and related services. That infrastructure needs to be designed carefully as one coherent system, with clear effects on the rest of the protocol.

This PR is not that final system. We do not need to overcomplicate the interim implementation. We should work together on the final system: indexers, peers helping other peers, watchtowers, and the other infrastructure that is still missing.

For now, the initial system should be dumb, good enough, and easy to swap out. Obtain channel IDs from recent logs, then use the existing full channel path. When the final infrastructure is designed, replace that one discovery piece with permissionless indexers and an open network of peers communicating with and helping one another.

### The probe misuses spectating and does not prove anything useful

The probe is the main design problem in this PR. A successful transport connection already gives the useful heartbeat signal: the peer is reachable now. If it times out, move on to another channel.

The probe does not establish that the peer has valid or current channel state. Spectating was designed to synchronize state with hard guarantees, so a peer can follow and replicate the state and decide whether to participate. It is also the data-availability path. A peer answering a lightweight probe can return something that makes the probe look successful, then fail when the real spectate and sync path starts. That extra round trip does not reduce uncertainty.

Do not run several weak probes through one partial acquisition flow. Once candidate channel IDs are known, run full spectating attempts for several candidates concurrently, each in its own full instance. A candidate either completes the real path or it does not. That gives a meaningful result instead of a separate probe stage that must be disproved later.

### `getLogs` is an interim discovery mechanism, not the final architecture

Scanning open and closed events is acceptable for the first iteration, but it should not become the final discovery system. Repeated log scans become more expensive over longer ranges, especially if every peer performs them. More importantly, historical logs do not give discovery the latest channel state.

The current approach does account for channel opening and closing, but it does not account for the intermediary state transitions. It therefore does not give the latest state that peers need to make a discovery decision: the current amount, participants, balances, and other terms. The initial open event often has only two peers, which is not enough information to make the decision.

Discovery should return current, filtered state. For example, it should be able to find channels with a particular player count, stake or amount range, and other current terms. That information has to be current when the peer is deciding where to connect.

The longer-term design should use permissionless indexers, among other infrastructure. Different operators can implement their own indexers and choose how to serve those queries. This first step is only a dumb stepping stone that gives us basic discovery until the current-state discovery system is designed.

This is compatible with the simple first step. Both sources feed the same decision: given the information available, should we attempt this channel or not? The initial log-based source gives a rough answer; the later filtered, current-state source gives a more precise one. In both cases, the system reuses the existing full spectating and join paths after it decides to attempt the channel. Once the full system is implemented, the swap should be out of the box because the rest of the system works the same; it only makes a more precise decision about whether to connect.

### Use a partition topic as the simple first lobby implementation

The first, dumb lobby implementation should be a topic. A topic is already a partition, so it can scope the range of peers that want to find one another. For example, players interested in a one-to-two-dollar range can join one topic, while players interested in a ten-to-one-hundred-dollar range join another. Once they connect, they negotiate their specific terms directly.

That is enough for the first implementation. The full marketplace can come later, once we have carefully designed the full protocol for how peers communicate, gossip, and help one another. I like that direction, but the intermediary implementation here adds marketplace machinery before that design is ready.

This also keeps peers connected for one purpose at a time. Peers already participating in a channel and replicating its state do not need lobby discovery. Lobby peers are short-lived: they enter the lobby, find a peer, negotiate, then become spectators or participants. A peer should be discovering or participating, rather than carrying several unrelated responsibilities at once.

### Reuse open-channel negotiation for the simple lobby

The simple lobby does not need a separate service and RPC surface. Its job is already mostly the job of open-channel negotiation: from a partition topic containing several peers, select and lock one peer, negotiate terms directly, and commit by opening a channel. The open-channel negotiation service is currently more hard-coded than it needs to be, but extending it to choose a channel ID covers most of this initial lobby flow.

That makes most of the added discovery directory unnecessary for the interim implementation. Keep the first step small and easy to replace: discover a peer through the partition topic, then use the extended existing negotiation flow to agree on terms and open the channel.

The richer lobby belongs to the later infrastructure design. Indexers, watchtowers, peers helping other peers, and peers with different roles should reuse a carefully designed common protocol with role-specific behavior. We should design that system together before adding its marketplace and lifecycle machinery. I like the direction, but this intermediary implementation is neither the final design nor a useful replacement for the smaller first step.

### Discovery needs an explicit peer status

The current [`Status` enum](https://github.com/peer3to/state-channels-plus/blob/0fd362f81c7a5f3b306d587c42533c3926b06b24/src/types/flags.ts#L11) has `NOT_OPENED`, `OPENED`, `SYNCED`, `PENDING_PARTICIPANT`, and `PARTICIPATING`. It is missing `DISCOVERING`: a peer that is looking for a channel ID or using the partition lobby to find someone with whom to negotiate.

That makes the post-handshake decision a simple function of status. After a connection completes the existing handshake path and is accepted, `DISCOVERING` invokes the lobby logic. Every other status follows the existing logic for negotiation, spectating, synchronization, or participation. The current handshake-completion path is the concrete place to add that branch: [`InitHandshakeService`](https://github.com/peer3to/state-channels-plus/blob/0fd362f81c7a5f3b306d587c42533c3926b06b24/src/rpc/services/initHandshake/InitHandshakeService.ts#L395).

### Keep Holepunch ban state on the peer profile

The Holepunch peer-info handle belongs on the [`PeerProfile`](https://github.com/peer3to/state-channels-plus/blob/0fd362f81c7a5f3b306d587c42533c3926b06b24/src/PeerProfile.ts#L15-L63), not on the general [`ATransport`](https://github.com/peer3to/state-channels-plus/blob/0fd362f81c7a5f3b306d587c42533c3926b06b24/src/transport/ATransport.ts#L58-L60) interface. There is one profile per address. A peer can begin on Holepunch or the relay, then upgrade to WebRTC, while the same profile continues to represent that identity. The profile is therefore the right place to retain the Holepunch peer-info handle and toggle the ban policy when needed.

Adding `getBannablePeerInfo()` to every transport, with a default no-op implementation and a Holepunch-only override, is unnecessary. The transport does not own that policy. The profile manager already owns the identity, the active transport, upgrade handling, and the ban decision; it should receive and retain the Holepunch information at profile creation or update without widening the general transport abstraction.

### Handshake completion should be a status decision, not a second membership lifecycle

The new [`P2PManager.onHandshakeCompleted`](https://github.com/peer3to/state-channels-plus/blob/0fd362f81c7a5f3b306d587c42533c3926b06b24/src/P2PManager.ts#L108-L267) overcomplicates what should be a small status-based decision. The long architectural comments and the separate `pendingChannelMembershipTransports` lifecycle make the manager infer channel membership from every completed handshake.

The simpler flow is already visible in the lifecycle model above. After a handshake completes and the connection is accepted, dispatch by local status. If the local status is `DISCOVERING`, invoke the lobby logic. For every other status, use the negotiation, spectating, synchronization, and participation behavior that already exists. There is no need for [`reevaluatePendingChannelMembership`](https://github.com/peer3to/state-channels-plus/blob/0fd362f81c7a5f3b306d587c42533c3926b06b24/src/P2PManager.ts#L276-L341) or a second deferred promotion state.

When a discovery negotiation opens a channel, change the local status, add the selected existing connection to `P2PManager`, leave the lobby topic, and subscribe to the channel topic. The same Hyperswarm instance can continue using the same peer connection. This is an explicit lifecycle transition, not something `P2PManager` needs to rediscover later from chain membership.

### Configuration and code boundaries are too broad

- Adding sixteen entries to the main configuration is bloat. These settings need to be separated into fixed protocol values, local implementation defaults, and the few real application choices.

- Many of the long comments explain architecture and history rather than a narrow local invariant. That material belongs in feature documentation (we have documentation now). The code should be simpler enough that it needs fewer long explanations.

- `ChannelAd` is a TypeScript-defined discovery type, not a Solidity type and never an on-chain value. Adding it to the shared Solidity-oriented codec mixes two ownership boundaries that do not need to be mixed. Discovery should own its own encoding and versioning, unless the project deliberately redefines the shared codec as the format for all cross-runtime protocol messages.

momoshell:
@lukastanisic99 — you were right on the substance, and this is too big to review as one thing. I'm splitting it.

Two PRs replace it:

- **https://github.com/peer3to/state-channels-plus/pull/457** — the handshake/promotion/ban-policy work you said you liked, with no discovery code in it at all. ~1,700 lines.
- **https://github.com/peer3to/state-channels-plus/pull/458** — discovery, stacked on the first and reworked against your review rather than carried over unchanged.

I'm closing this one; the history is worth more linked from those two than kept alive as a third.

## Where you were right

**The probe misuses spectating, and it's worse than a wasted round trip.** Chasing it turned up two things I'd missed. A probe goes through `onSpectateRequest`, which now promotes the requester into the responder's `openConnections` — so probing N chain candidates makes N peers add us to channels we never join, and nothing removes us. That's exactly the broadcast leak this whole change is supposed to prevent. And the responder blacklists when `generateSyncPayload` returns undefined, which happens for transient local reasons like a dispute window not being loaded yet. Since `blacklistProfile` now escalates to a DHT-level ban, probing a channel that happens to be mid-dispute permanently bans an honest prober. "Non-punitive" was only ever true on the prober's side.

**Chain-first still needed the lobby.** I said an outsider could join an existing channel without touching the lobby. `acquireChannel` calls `requireLobbyService()` before anything else, so it throws with no lobby wired, including on the pure chain path. That claim was wrong.

**The interim implementation is too big.** Accepted. The discovery half is being cut down rather than defended.

## What splitting turned up

Two defects in the promotion gate, both now fixed in **https://github.com/peer3to/state-channels-plus/pull/457**:

The immediate and deferred paths had drifted onto different definitions of membership — the immediate one asking the on-chain participant union, the deferred one asking dispute eligibility, which is false at genesis even for a founding participant. The immediate path carried a comment explaining why that stalls a channel at height zero; the deferred path was still doing it.

Worse, a deferred peer could stay deferred forever. Re-evaluation only triggered on our *own* status changing, but membership grows when someone *else* joins — which moves the joiner's status, not ours. A peer that defers a stranger and then sits idle never looks again. I tried `onSetState` and the chain-driven `onStateSnapshotUpdated` and neither reaches an idle peer; what works is the joiner announcing once its join transaction is mined, with the receiver still re-reading the union so the announcement is a hint rather than authority. There's a test that fails against every other trigger.

## What already changed in https://github.com/peer3to/state-channels-plus/pull/458

- **The probe is gone.** Chain-first walks candidates newest-first through the ordinary connect/sync/join path. A candidate either completes it or it doesn't.
- **Chain-first no longer requires a lobby**, so joining an existing channel really does work without one now.
- **The duplication is cleared**: one `getOptionalRpcService` instead of three structural casts, one `deriveChannelTopic` instead of three copies, the write-only `boundOpenChannelId` removed, and the leaked plan labels stripped from source and a test filename.

175 passing, 0 failing on the discovery suites.

## What I left open there, deliberately

- **The stake is still chosen up front.** Your "spectate, then sit, then pick a deposit" note doesn't fit an API that takes `amount` before anything has been discovered. It wants splitting into a discover step and a sit step. I'd rather agree that shape with you than guess at it, so `setStakeAmount` sits in the discovery PR where it can change.
- **Whether the lobby stays a topic-only partition**, as you suggested. It currently carries an ad store on top of the topic. I can cut that back, but that's a scope call rather than a defect.
- **Some unit tests stub collaborators** instead of teleporting a harness session, and a few still register `it()` in loops. Both are real violations; both are tracked in the PR.

## Two things unrelated to this PR

`yarn tsc -p tsconfig.json` is red on `dispute` itself — `test/scripts/e2eParallelProtocol.test.ts` imports `tar` without `@types/tar` installed. (This supersedes the stale `E2E-BlockQueueManager` note in the description above.)

The full gate has load-dependent failures on `dispute` independent of any of this. Same selection at the same concurrency gives 155 passing / 6 failing on both the merge base and this branch, with overlapping sets — `JoinChannelSignatureRequest` and two `E2E-Spectate` cases fail on both sides. Worth looking at separately.

PR #427, opened 2026-08-08
refactor(clock): extract the resync tolerance predicate and cover it with tests

sh3ll3x3c:
## Summary

`Clock.syncClock` decides whether to re-adjust the local clock by comparing the
local/chain timestamp difference against `averageBlockTime`. That comparison
carried an open question — `//TODO - think - should it be 2* or 1* or something
else?` — which had never been decided and had no test coverage.

This settles it at **one** interval, which is the existing behaviour. **No
behaviour change**: the predicate is byte-for-byte the same comparison, just
moved.

- Extracts the comparison into a pure exported
  `isBeyondBlockIntervalTolerance(differenceSeconds, averageBlockTimeSeconds)`,
  leaving `syncClock` a thin async wrapper, so the decision is unit testable
  with synthetic inputs.
- Records the reasoning next to the predicate as a **tolerance policy**: the
  comparison cannot tell real clock skew from ordinary "no block yet" lag, so
  the threshold is a budget for undetected skew rather than a classifier. One
  interval and not two because both avoid correcting when no block has simply
  landed, but 1x leaves less real skew uncorrected.
- Documents a real limit that the code does **not** handle, rather than
  papering over it: `averageBlockTime` is a trailing mean over the last ≤10 gaps
  and does not bound the gap still open after the latest block. An unusually
  late block therefore reads as a disagreement and pulls local time back to that
  block's timestamp. Narrowing that needs an asymmetric or gap-aware rule — a
  behaviour change, deliberately out of scope here and tracked separately.
- Adds 7 unit tests: both sides of the boundary, both signs, the exact-boundary
  case, zero difference, and a fractional interval.
- Drops a commented-out debug `console.log` block from `syncClock` (the repo
  bans `console.*`). Called out explicitly rather than removed silently.

Unchanged on purpose: the `blockCnt === 0` and `!this.averageBlockTime` early
returns keep their current semantics, so a zero average block time still adjusts
once without recursing.

Stacked on #424 so it can run on the distributed pool; no code dependency, and
GitHub will auto-retarget as the lower PRs land.

## Test Plan

- `yarn tsc --noEmit -p tsconfig.json` — clean
- `yarn tsc --noEmit -p tsconfig.browser.json` — clean
- Distributed pool, `--test-pattern 'Clock.test.ts'` — **11/11 passing**
- **Mutation check:** flipping the predicate's `>` to `>=` fails exactly the two
  boundary tests (`holds the clock when the disagreement is exactly one block
  interval`, `compares against a fractional block interval`) and passes the
  other 9 — confirming the new tests pin the boundary rather than just running.

External review run before raising: 3 rounds, 2 findings fixed, final round
clean.

### DIFF
--- a/src/Clock.ts
+++ b/src/Clock.ts
@@ -1,5 +1,18 @@
 import { ethers } from "ethers";
 
+// The latest block's timestamp trails chain time by up to one block interval,
+// so a skew within one averageBlockTime is explained by block production alone,
+// not by a wrong local clock. One interval is the tightest threshold that still
+// covers that natural lag: below it we would chase block quantization (and each
+// correction re-syncs recursively), above it we would tolerate real skew. Hence
+// 1x, not 2x.
+export function isClockSkewBeyondBlockInterval(
+    differenceSeconds: number,
+    averageBlockTimeSeconds: number
+): boolean {
+    return Math.abs(differenceSeconds) > averageBlockTimeSeconds;
+}
+
 class Clock {
     private static instance: Clock | undefined;
     private static initialization: Promise<void> | undefined;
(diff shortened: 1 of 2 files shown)


src/Clock.ts
```diff
@@ -96,17 +114,11 @@ class Clock {
...
-        // console.log(
-        //     `Average block time calculated: ${this.averageBlockTime}s over ${blockCnt} blocks`,
-        //     `past block timestamp: ${pastTimestamp}, latest block timestamp: ${latestTimestamp}`,
-        //     `current time: ${currentTime}, difference: ${difference}s`
-        // );
         if (!this.averageBlockTime) {
             this.clockAdjustmentSeconds += difference;
             return;
         }
-        //TODO - think - should it be 2* or 1* or something else?
```
lukastanisic99, line 108: What's the design decision around this? This was left as a `todo` so I/whoever can think about a holistic solution to this that makes sense - I don't see the criteria on which this decision was taken and just refactoring a function isn't really useful.

PR #434, opened 2026-08-15
start runtime monitors after each context is ready

lukastanisic99:
- await application and precompile readiness before monitoring
- report runtime startup separately and preserve starvation logs
- cover inline and worker lifecycle paths and refresh the specification

@daiagi @sh3ll3x3c @momoshell @evainfeld
Among implementing further what I need for poker, I'm also demonstrating how to use and maintain the docs. 
I've verified and reviewed all files changed and their 1:1 report mirrors. I think this is a good example of how things are kept up to date and how we can NOT have stale documentation, since this structure forces us (the agent) to follow a specific pattern and maintain specific document mirrors instead of 'having to figure out what to change from unstructured docs that don't follow a pattern`, well it would probably deduct what to change, but we (the engineer) as the verifier wouldn't probably know should something else also be updated. 

Like this we know where to look, and can see all the changes which are digestible in small chunks. All the test cases from this pass are assigned (no gaps) and all tests from this run have at least 1 test ID assigned (no gaps). 
This does not guarantee that nor me nor the agent didn't miss a test case, but later if we deduct that we have - adding and tracking it is O(1)


docs/spec/generated/pending-review.md
```diff
@@ -4,8 +4,8 @@
 > its content hash with `SPEC_REVIEWER="Name" node docs/spec/tools/review.js <file...>`.
 > Any later edit invalidates the record automatically (the file returns to pending as stale).
 
-- Verified (current): **3/516** (1%)
-- Pending (never reviewed): **513**
+- Verified (current): **13/515** (3%)
+- Pending (never reviewed): **502**
 - Stale (edited since review): **0**
```
daiagi, line 9: pending-review.md is stale: committed report says 13/515 verified / 0 stale, but rerunning `yarn spec:refresh` on this branch gives 9/515 / 4 stale. all 4 are this PR's own docs (WorkerContractExecutor + worker protocol mirrors and the test report) - edited after review.js recorded their hashes -> the report shows drifted content as verified.

lukastanisic99: ye I probably forgot to run spec:refresh - probably all this generated stuff should be .gitignored + the audit also (still need to look into the audit) - the other stuff should be good

src/evm/p2pRuntime/P2pRuntimeHost.ts
```diff
@@ -406,6 +397,27 @@ export async function startP2pRuntimeHost<
...
+                config.EVENT_LOOP_DELAY_ERROR_THRESHOLD_SECONDS > 0 &&
+                threadLabel
+            ) {
+                logger!.startPerformanceMonitoring({ threadLabel });
+            }
+            if (
+                typeof process !== "undefined" &&
+                typeof process.stdout?.write === "function"
+            ) {
+                process.stdout.write(
```
sh3ll3x3c, line 414: This emitter is not gated on `EVENT_LOOP_DELAY_ERROR_THRESHOLD_SECONDS > 0`, unlike the monitor block directly above it and every other `##E2E_TIMING##` writer. `src/evm/node/workerStartupTiming.ts` gates on it explicitly "so production worker creation stays silent". As written, any app embedding the SDK on Node gets runner markers in its stdout on every `p2pSetup`.

Two smaller things on the same lines:

- No `try/catch`. `workerStartupTiming.ts` wraps its write because "stdout may be closed during teardown". Here an EPIPE throws inside `buildRuntime`, so `deployComplete` rejects and the runtime is disposed.
- `process.stdout` is a node API at a browser-compiled path. `tsconfig.browser.json` excludes `src/evm/p2pRuntime/node`, not `src/evm/p2pRuntime`, so per AGENTS.md this belongs under a `node/` dir behind a `@platform/*` alias — and `workerStartupTiming.ts` already owns this exact operation.

Approving since the fix is small, but worth doing before this reaches an integrator.

sh3ll3x3c (approved):

PR #436, opened 2026-08-17
spec review: storage and security docs signed

daiagi:
read and signed the 13 assigned docs under specification/storage/ and specification/security/.

- fixed spec text where it didn't match verified code: block tip rules, posting-timestamp merge, force-exit lifecycle, range bounds, timeout refresh
- trust model: no watchtower in v1, partition -> channel
- findings register: 1 closed, 2 updated, 2 new; OQ-23 and OQ-43 updated
- test matrices extended to match, generated reports refreshed


docs/spec/specification/security/README.md
```diff
@@ -16,21 +16,21 @@ algorithms.
 - **Public inputs.** Every other system's stated assumptions, security considerations, and
   verification plans.
 - **Public outputs.** The authoritative trust model (`REQ-TRUST-*`), the data-availability cost and
-  griefing model (`REQ-DA-*`, `INV-DA-*`), topology and partition-size limits, watchtower
+  griefing model (`REQ-DA-*`, `INV-DA-*`), topology and channel-size limits, continuous-availability
```
lukastanisic99: what's this `continuous-availability` - I left watchtower, since other solutions are more complicated for V1 and we'll explore in v2

docs/spec/specification/security/trust-model.md
```diff
@@ -83,8 +83,8 @@ voids the guarantees that depend on it.
...
+| A8  | **Continuous availability for offline participants.**                                                                     | An offline participant loses contest rights unless a node holding its key stays online on its behalf; a keyless third party can only submit dispute fraud proofs (§7). No watchtower/delegate mechanism exists in version one.                                                                                                                                                                                                                                                                                                                        |
```
lukastanisic99: Same here - the watchtower is needed in V1 (we'll do it in the end) and in V2 we'll do a probabilistic system where your peers/friends are your hidden watch towers - you protect them, they protect you and an attacker doesn't know do they exists or not -> they have negative expected value if they try to 'attack'

docs/spec/specification/security/trust-model.md
```diff
@@ -137,35 +137,23 @@ calldata posts, snapshot updates) and contest invalid ones within the protocol w
...
-  which key material; a delegate that can submit contests but not steal funds is the target shape.
-- **Timeouts:** the delegate must act within the same on-chain windows as the participant
-  ([../protocol/disputes.md](../disputes/disputes.md), [../protocol/time.md](../protocol-model/time.md)).
-- **Failure:** what the participant's exposure is if the delegate fails, and how failure is
-  detected.
-
-**Required verification:** offline-participant tests (honest participant
-offline while a counterparty submits an invalid dispute/timeout; delegate contests in time) and
-collusion tests (remaining participants collude against the offline participant) — required downstream coverage.
+**<a id="req-trust-4-kw24nf"></a>`REQ-TRUST-4-KW24NF`.** Version one ships no watchtower. An honest participant that may go
```
lukastanisic99: In my opinion the easiest way to do a watchtower in V1, is to add 1 more `status` == WATCHTOWER and 1 more polymorphic strategy and have the watchtower just be 'another peer' without anyone knowing they're connected to the watchtower - they can not differentiate a spectator from a watchtower + this allows anyone to run a watchtower by just spinning up a peer - maybe even in V1 all honest peers run 1 watch tower and spectate N random games (if it's not too heavy, depending on their available resources) - in the beginning we'll run the watchtower for sure and spectate all games, but doing it like this is totally permissionless and anyone can do it and we don't need to implement anything extra for V2 + this will be useful for recovery

docs/spec/specification/security/trust-model.md
```diff
@@ -213,11 +201,11 @@ must identify prevention, detection, recovery, and residual exposure; an unowned
 
 **[`REQ-TRUST-2-X8GCZ7`](trust-model.md#req-trust-2-x8gcz7).** At least one available, honest RPC connection required; redundancy helps availability but does not remove the assumption.
 
-**[`REQ-TRUST-3-3YWEZR`](trust-model.md#req-trust-3-3ywezr).** At least one non-Byzantine participant per partition; all-Byzantine partitions have no safety.
+**[`REQ-TRUST-3-3YWEZR`](trust-model.md#req-trust-3-3ywezr).** At least one non-Byzantine participant per channel; all-Byzantine channels have no safety.
 
-**[`REQ-TRUST-4-KW24NF`](trust-model.md#req-trust-4-kw24nf).** Watchtower/delegate required for offline honest participants, with stated data, privacy, availability, authorization, timeout, and failure assumptions.
+**[`REQ-TRUST-4-KW24NF`](trust-model.md#req-trust-4-kw24nf).** Continuous availability required for offline honest participants: a node holding the participant's key stays online; keyless third parties are limited to fraud-proof kills; no delegate mechanism exists in version one.
```
lukastanisic99: -||-

docs/spec/specification/security/trust-model.md
```diff
@@ -153,19 +153,28 @@ watchtower design:
   delegate outage recreates the original problem.
 - **Authorization:** which actions the delegate may take on the participant's behalf, and with
   which key material; a delegate that can submit contests but not steal funds is the target shape.
+  Today dispute fraud proofs are permissionless (`applyDisputeFraudProofs` accepts valid proofs
+  from any sender), while opening a dispute or contesting a timeout requires the participant's own
+  key — how a keyless delegate gains contest authority is
+  [`OQ-43-HWRTNF`](../open-questions.md#oq-43-hwrtnf).
```
lukastanisic99, line 158: yes this is a good question - I've though about it a bit, but let it remain an open question till we do the watchtowers - in short:
The watchtower doesn't need to commit to a dispute, it just needs to kill any invalid dispute. This can lead to a state where no valid dispute is submitted and that's fine as long as no invalid ones survive and in the worst case the participant peer when they come back online (eventually) can submit the valid dispute which would allow it to settle and exit if no one timed it out - the game theory behind this will of course be part of the spec when this is implemented in the end

docs/spec/specification/security/trust-model.md
```diff
@@ -153,19 +153,28 @@ watchtower design:
...
 - **Timeouts:** the delegate must act within the same on-chain windows as the participant
   ([../protocol/disputes.md](../disputes/disputes.md), [../protocol/time.md](../protocol-model/time.md)).
 - **Failure:** what the participant's exposure is if the delegate fails, and how failure is
   detected.
 
+_Non-normative direction (review note):_ the intended version-one shape is a watchtower running as
```
lukastanisic99: Why would 'review notes' be part of the spec? You can add an open question stating that design, but the spec should always be 'normative'

docs/spec/specification/storage/README.md
```diff
@@ -37,8 +37,12 @@ strength, never observable semantics.
   durability contract the medium change must satisfy.
 - **Ordering and concurrency.** Module operations are atomic at the single-operation level; merge
   operations (signature sets, source attribution) are monotone and idempotent so concurrent and
-  duplicate delivery converge. Cross-module consistency at an operation boundary is the *caller's*
-  transaction, specified by [`REQ-STOR-2-TARP8S`](durability.md#req-stor-2-tarp8s).
+  duplicate delivery converge, and a known on-chain posting timestamp is never discarded by a merge
+  ([blocks.md](./blocks.md), [queue.md](./queue.md)). Cross-module consistency at an operation
+  boundary is the *caller's* transaction, specified by [`REQ-STOR-2-TARP8S`](durability.md#req-stor-2-tarp8s).
+- **Aliasing and isolation.** Stores copy their arguments on write and their results on read, so no
+  caller-held reference aliases store state. Iterator reads are the exception: they yield live
```
lukastanisic99: Is this true? If so we should explicitly force iterators to yield `copies` so they don't break the mental model. All mutations to storage should be `explicit` and pass the `serialization boundary` like they would if it was a normal DB under the hood - no shared reference should exist `inside` and `outside`

docs/spec/specification/security/trust-model.md
daiagi: removed

docs/spec/specification/storage/README.md
daiagi: it is true, I've corrected the spec and filed an open finding
- FIND-STORAGE-6
- REQ-STOR-6

daiagi:
all done @lukastanisic99 , ready for another rounf

lukastanisic99 (approved):

PR #437, opened 2026-08-18
make runtime type checks module-graph independent

lukastanisic99:
- replace instanceof checks for custom RPC services, transports, and ethers Result
- harden RPC dispatch, response delivery, frame sizing, and cleanup
- expand exact verification coverage and dependency traceability

-------
This started with getting rid of `instanceof` in a few places since the bundler had 2 dependency graphs (e.g. the main app bundle and the worker bundle that is loaded dynamically at runtime), so even though it's the same code, it's not the same constructor and the worker has to load the CustomRpc dynamically at runtime since the SDK doesn't know it. 

Other changes to src are fixes along the way, while providing coverage to files I've touched with the original changes. The ones I reviewed to a degree I marked with the review script, the rest I've updated, but didn't mark -> deferring their checking to the future, to not explode the PR too much + to get other stuff done


docs/spec/generated/pending-review.md
```diff
@@ -4,17 +4,26 @@
 > its content hash with `SPEC_REVIEWER="Name" node docs/spec/tools/review.js <file...>`.
 > Any later edit invalidates the record automatically (the file returns to pending as stale).
 
-- Verified (current): **13/515** (3%)
-- Pending (never reviewed): **502**
-- Stale (edited since review): **0**
+- Verified (current): **40/525** (8%)
+- Pending (never reviewed): **480**
+- Stale (edited since review): **5**
```
daiagi, line 9: same staleness as 434 but wider: committed pending-review.md says 40/525 / 5 stale, fresh `yarn spec:refresh` -> 30/525 / 15 stale. 10 docs with 2026-08-18 review records were edited after the hash was recorded, Codec and HandshakeCompletedGuard mirrors among them (Codec.ts.md hashes 945b3857 vs recorded 122036de in review-state.json).

lukastanisic99: Ye same comment as on 434 - probably best to .gitignore and have the CI run it later - for now this is not useful since a lot of things need to be filled/cleaned, but once it gets to a stable state it should be - we'll see

lukastanisic99: We'll probably also git ignore the audit, since the auditor runs his own audit instead of relying on the audit of the implementor - the implementor by doing the PR implicitly audited, reviewed and approved their implementation. 

So we'll strip:
/generated 
/audit
review and approve scripts

test/rpc/ARpcService.test.ts
```diff
@@ -0,0 +1,339 @@
+import { expect } from "chai";
+
+import { ARpcServiceFixture } from "@test/fixtures/ARpcServiceFixture";
+
+describe("ARpcService", function () {
```
sh3ll3x3c, line 5: Heads up: none of the new unit suites in this PR actually execute. `yarn test` runs files in `find` order, and a failing `afterEach` in `test/unit/EventBarrier.test.ts` aborts the whole mocha run, so every file after it is skipped — this one included.

Evidence: CI run 32085291466 (this branch) and run 31908667630 (base `await-ready`) both report exactly `127 passing / 11 pending / 4 failing`, and no new `describe` name appears in either log. About 4.6k lines of new test code moved the counts by zero.

The root cause is pre-existing on the base branch, so it isn't this PR's doing. Flagging it because the coverage credited in the verification layer isn't real evidence until the abort is fixed and these suites run.

src/rpc/ARpcService.ts
```diff
@@ -1,11 +1,36 @@
...
+function resolveRpcEndpoint(
+    rpcMethods: ARpcMethods,
+    methodName: string
+): RpcEndpoint | undefined {
+    if (methodName === "constructor") return undefined;
+
+    let owner: object | null = rpcMethods;
+    while (
+        owner &&
+        owner !== ARpcMethods.prototype &&
```
sh3ll3x3c, line 21: This stop condition has the same module-graph dependency the PR is removing elsewhere — `x !== SomeClass.prototype` is as graph-sensitive as `x instanceof SomeClass`. For a custom-RPC methods class loaded from the second bundle, `ARpcMethods.prototype` here is a different object than the one that class extends, so the walk never stops at the base and continues through its own properties.

Not exploitable today: `ARpcMethods.prototype` has only `constructor` (rejected on the first line of the function) and `remoteRpc` (an accessor, so `descriptor.value` is `undefined`). Both safe by accident.

It becomes a hole the moment anyone adds a plain method to `ARpcMethods`: a crafted `{service, method}` payload would reach it on worker-loaded custom RPCs while still being correctly rejected on SDK-native services. Whoever adds that method would have no reason to suspect they were opening an endpoint.

Suggest a brand marker instead of prototype identity — a `Symbol.for(...)`-keyed property on the base is stable across bundles — or build an explicit endpoint-name set at construction.

sh3ll3x3c (approved):

test/rpc/ARpcService.test.ts
lukastanisic99: @sh3ll3x3c  Is this based on the CI only or you're seeing failures locally too? We haven't updated the CI in ages - that also need to be done - you can connect it to run distributed tests (`yarn test:parallel:distributed`), just give it the shared secret when it's running that. Also tell your agent to look at AGENTS.md so only uses yarn `test:parallel` or `yarn test:parallel:distributed` (way faster) - if you're running the sequential mocha tests you'll be running them for hours

src/rpc/ARpcService.ts
lukastanisic99: You're right that `prototype identity` is graph-sensitive, but runRPC comes from the custom service’s own ARpcService copy. Its lexical ARpcMethods.prototype comes from the same graph as the custom methods class, so the walk stops correctly.
The hole would require deliberately mixing an ARpcService from graph A with an ARpcMethods instance from graph B. The normal worker-loaded custom RPC path does not do that, they are always paired - `the implementation report documented this :D `

PR #439, opened 2026-08-18
docs(spec): specify execution suspension on disputed forks

sh3ll3x3c:
## Summary

The specification never states what happens to off-chain execution while a dispute is open. The lifecycle diagram has an `Executing → Disputing` edge, but no document defines the liveness contract during a dispute — even though the implementation enforces one. This PR adds it as `REQ-LIF-7-0XZBDM` in `specification/settlement/lifecycle.md`:

- A committed dispute suspends off-chain execution on the disputed fork: queued blocks are discarded and newly received blocks are rejected, until execution resumes on the canonical successor fork (`REQ-LIF-4-SW8GVY`), with valid non-final work carried forward.
- A dispute against a non-current fork must not suspend the current fork.
- For this protocol version, the expected pause spans the three consecutive dispute periods (evidence, kill, reduce challenge) — approximately `3 × evidenceTime` — and the runtime surfaces the pause and its expected duration to the integrating application.
- Security considerations now name the griefing surface: any dispute-eligible participant can pause the fork for ~`3 × evidenceTime`, bounded by dispute costs, kill-period slashing, and the per-disputer throttle.

Also links the new requirement from the three implementation file reports that enforce it, adds a test-matrix row with six real permutations (currently unassigned — a tracked verification gap), and regenerates the inverse reports (`yarn spec:refresh`).

## Evidence (current implementation behavior)

- **Commitment-time purge:** `EventHandler.onDisputeCommitted` clears the disputed fork's block queue — `src/eventHandlers/EventHandler.ts:362` → `BlockQueueManager.clearFork` (`src/stateManager/BlockQueueManager.ts:361`) drops queued entries and cancels their timeouts ("Cleared queued blocks for disputed fork").
- **Standing gate:** `ValidationService.validateBlockConfirmation` diverts every block on a disputed fork to the strategy deviation `blockForkIsDisputed` instead of the normal pipeline — `src/stateManager/ValidationService.ts:99-107`. So production cannot resume on the disputed timeline; this is a continuous rule, not a one-off flush.
- **Pause duration surfaced:** on the first committed dispute for the fork, the runtime fires `onDisputeStarted(evidenceTime * 3)` — `src/eventHandlers/EventHandler.ts:370-373` — matching the evidence + kill + reduce-challenge sequence in lifecycle.md §8.

None of this was stated in `lifecycle.md`, `finality.md`, or `disputes.md` (searched for any halt/pause/production rule).

## Test plan

- `yarn spec:ids:fix` — 5214 definitions, 0 issues.
- `yarn spec:refresh` — regenerates the six inverse reports; includes `yarn spec:ids:check`, passing.
- `yarn spec:impact` — flags the touched reports' linked requirements for reverification as expected; the edits only append (new requirement, new linked rows), no existing rows changed.
- No repository tests cover the new permutations yet; the six `REQ-LIF-7-0XZBDM.T1.P1–P6` permutations are recorded as the tracked gap in the regenerated verification coverage.

Engineer sign-off (audit/approvals) intentionally untouched; the aggregate approval goes to `Reverification required` automatically.


docs/spec/specification/settlement/lifecycle.md
```diff
@@ -148,6 +148,18 @@ period and its opener slashed, but the window it opened still resolves to a cano
...
 
+**<a id="req-lif-7-0xzbdm"></a>`REQ-LIF-7-0XZBDM`.** A committed dispute suspends off-chain execution on the disputed
+fork. From the moment a participant observes a dispute committed against its current fork, it
+MUST stop advancing that fork: blocks already queued for the fork are discarded, and newly
+received blocks on the fork are rejected instead of validated. Execution resumes only on the
+canonical successor fork ([`REQ-LIF-4-SW8GVY`](lifecycle.md#req-lif-4-sw8gvy)), with valid
+non-final work carried forward — the pause costs time, not progress. A dispute against a fork
+that is not the participant's current fork MUST NOT suspend execution on the current fork.
+**For this protocol version:** the expected pause spans the three consecutive on-chain dispute
+periods — evidence, kill, and reduce challenge — i.e. approximately `3 × evidenceTime` (§8), and
```
lukastanisic99: Currently we run reduce on-chain so we finalize it instantly (we pay in gas), we can also pay in time (and commit to the reduce outcome and have it be challenged) and wait for the additional period to expire, but in reality we don't need to wait since all peers can instantly know if the reduction is good or not and can safely build on the winning reality (even if someone tries reducing to a wrong one) - the reduction challenge can happen totally async without impacting the users (the only impact would be if someone wanted to withdraw immediately - they'd have to wait for the challenge to expire)

docs/spec/specification/settlement/lifecycle.md
```diff
@@ -220,6 +232,12 @@ partial opening, concurrent execution and joins, stale snapshot attempts, unavai
 failed settlement, retry/recovery, and all terminal states. The diagram is not sufficient evidence: every edge
 and prohibited edge needs an observable oracle and must preserve value and canonical-history invariants.
 
+The execution pause of [`REQ-LIF-7-0XZBDM`](lifecycle.md#req-lif-7-0xzbdm) is itself a griefing surface: any
+dispute-eligible participant can suspend the fork's off-chain execution for roughly `3 × evidenceTime` by
+committing a dispute, valid or not. The cost of doing so (dispute fees, kill-period slashing of invalid
```
lukastanisic99: 2x evidenceTime regardless if the reduce is challengable or not, but that's the reason why a dispute has to have valid reason - if you just create a dispute to create a dispute without a valid reason, you get slashed and the dispute killed. So you can only do it once.

docs/spec/specification/settlement/lifecycle.md
sh3ll3x3c: Thanks — corrected in aa17185f. I had read the challenge period as gating resumption; it does not.

Traced it: `reduceAndFinalize` computes `reduce(disputes)` on-chain and commits the result with `reductionTimestamp = block.timestamp - getEvidenceTime()`, so `_isReduceChallengePeriodExpired` is already true when the result lands — instant finality, paid in gas. And `areDisputesCommitted` requires the full committed set (exact length plus ordered commitment match), so the outcome is not steerable by submitting a subset; every peer can derive the same winning fork and build on it right away.

`REQ-LIF-7` now states that the reduce challenge period does not extend the pause, and that a pending challenge period delays only on-chain snapshot adoption (an immediate withdrawal), never resumption of off-chain execution. Added the same clarification to §5 path 2, since that text reads as though adoption always waits.

docs/spec/specification/settlement/lifecycle.md
sh3ll3x3c: Confirmed and fixed — it is `2 × evidenceTime`, not 3.

The kill period ends at `lastEvidenceSubmissionTimestamp + evidenceTime`, and evidence can arrive up to `creationTimestamp + evidenceTime`, so the worst case is two periods from window creation. (`_isEvidencePeriodExpired` is marked "not used anywhere right now" — the kill period is the binding gate.) Reduction then finalizes instantly per your other comment, so nothing follows it.

Also rewrote the griefing paragraph the way you framed it: the bound is accountability, not rate — an unfounded dispute is killed and its opener slashed, so it can be inflicted once, at the cost of their stake. I had written it as a repeatable throttle-limited nuisance, which was wrong.

One thing worth a look, outside this PR: the runtime fires `onDisputeStarted(evidenceTime * 3)` (`src/eventHandlers/EventHandler.ts:370`), so integrating applications are told to expect three periods. If that is deliberate padding it is fine; if it is a leftover from the same three-period reading I had, it is telling apps to freeze the UI about 30s longer than the protocol needs. Happy to open a separate issue.

lukastanisic99 (approved):

PR #444, opened 2026-08-20
test(open): prove the open-terms trust boundary instead of enforcing it

NiKrause:
# Prove the open-terms trust boundary instead of enforcing it

Closes [`DEF-1-92NTAG`](docs/spec/audit/open-findings.md#def-1-92ntag) as **won't fix**. **No contract change** — the finding is not a defect, and this PR makes that decision falsifiable instead of leaving it as prose.

## What the register claimed

`open()` indexes `balances[i]` over `participants.length` without requiring the sequences to agree, and accepts `address(0)` participants. Measured against `dispute` with valid unanimous signatures: a short sequence panics (`0x32`), a surplus entry is silently discarded.

I originally shipped on-chain checks for this. On review the design call went the other way, and I think that is right — my own analysis had already concluded neither case is exploitable.

## Why no check is needed

**Unanimity is the trust boundary.** A malformed term needs the full signature set, so one honest listed participant keeps it below the threshold. An on-chain shape check would only repeat what the signature set already decides. The SDK cannot even construct such a term — `OpenChannelNegotiationService` builds `participants` as a fixed pair with matching balances.

**A fully colluding set affects only itself.** Per-participant data is consumed only for listed participants: `joinChannels` is built with exactly `participants.length` entries, genesis derives from `processedJoins`, `totalDeposits` comes from the deposits actually composed, and `channelBalances` is keyed by `channelId`. A surplus entry is never escrowed, so it creates no withdrawable credit, and [`INV-LIF-5-ENQB91`](docs/spec/specification/settlement/lifecycle.md#inv-lif-5-enqb91) still caps withdrawals at what was escrowed.

**The `address(0)` half was never reachable at all.** `verifyThresholdSigned` recovers via OpenZeppelin `ECDSA.recover`; the repo pins and installs **5.0.2**, where `recover` routes failures through `_throwError` (`ECDSA.sol:90-91`) and reverts rather than returning `address(0)` — older versions did return it, which is presumably where the concern came from.

So the residual cost is diagnostic, not custodial: a malformed term surfaces as a threshold failure or an out-of-range fault whose reported cause points away from the term itself. Not worth +87 bytes on a contract already ~6.2 KB over EIP-170.

## What ships instead

Six tests against the full `DiamondHarness` diamond, holding both properties to evidence so the decision can be falsified if either stops holding:

| Test | Property |
| --- | --- |
| `abstainingParticipantBlocksThreshold` | one honest participant keeps a malformed term below the threshold |
| `colluderCannotForgeTheAbstainer` | and the colluder cannot substitute its own signature for the missing one |
| `collusiveSurplusBalances_createNoDepositCredit` | unanimously signed surplus opens the channel but records only the listed balances |
| `wellFormedTerms_recordSameTotalsAsSurplusTerms` | control — without it, "the totals look right" is unfalsifiable |
| `collusiveShortBalances_leaveNoState` | the short direction reverts and strands nothing |
| `duplicateParticipants_reverts` | unchanged; that check *is* enforced on-chain |

## Test file renamed to `.t.sol`

The spec tooling recognizes Solidity tests only as `*.t.sol`:

```js
// docs/spec/tools/shared/test-inventory.js:11
const TEST_FILE_RE = /(?:\.(?:test|spec)\.[cm]?[jt]sx?|\.t\.sol)$/;
```

At `StateChannelManagerProxyOpen.test.sol` the file was invisible to every coverage report — including its pre-existing `test_open_duplicateParticipants_reverts`. Already true on `dispute`, not a regression here. All sibling Foundry suites already use `.t.sol`, and Foundry discovers by directory, so execution is unaffected.

## Specification

- The decision is recorded in [admission-and-funds.md](docs/spec/specification/enforcement/admission-and-funds.md) — an **assumption** (the correspondence is not validated, and why) plus a **security consideration** (what bounds the colluding case).
- `DEF-1-92NTAG` closed as won't-fix with its rationale, linking the decision and the evidence.
- The two permutations that previously existed only to *document the finding* (`…NTYR71.P6`, `.P11`) now carry the trust-model evidence, with `.P12`–`.P15` added for the remaining cases. IDs preserved, nothing renumbered.
- New verification report for the suite — the file had none.

Implementation coverage is unchanged at **200/243**: no requirement was added, because the decision is that no requirement is warranted.

## Verification

Rebased onto `dispute` at `7bad54a0`.

- `forge test` — 76 passed, 0 failed
- `yarn spec:refresh` — clean, check mode byte-identical; a second refresh is a no-op
- `yarn spec:impact --staged` — exit 0, no unmapped files
- Both malformed directions reproduced against unpatched `dispute` before any of this

On the previous revision of this branch (which carried the contract change), `yarn test:parallel:distributed` on the project swarm gave **1198 passing, 12 failing** in `wall 375.74s · 42.0x speedup`, with none of the 12 attributable to the change — 9 were `"after each"` teardown races where the test body passed, 3 were harness/P2P infrastructure. This revision removes the contract change entirely, so that evidence only becomes more favourable; happy to re-run the swarm gate on request.

## Note for the reviewer

`yarn spec:impact --base <merge-base>` reports one unmapped file: the **pre-rename** path `StateChannelManagerProxyOpen.test.sol`. `report-change-impact.js` collects the diff with `git diff --name-only` and no `-M`, so a rename surfaces its old path, which cannot be mapped because the file no longer exists. The new path is fully mapped. Worth fixing separately in the shared tooling.


contracts/V1/StateChannelDiamondProxy/StateChannelManagerProxy.sol
```diff
@@ -108,8 +108,20 @@ contract StateChannelManagerProxy is StateChannelManagerInterface, StateChannelC
         (bool isOpen,) = isChannelOpen(openChannelData.channelId);
         require(!isOpen, RaceConditionChannelAlreadyOpen());
 
-        // reject duplicate participants
+        // balances is indexed by participant when building the joins below: a short
+        // balances array panics (0x32), a long one silently drops entries every
+        // participant signed. Reject the shape before any of that.
+        require(
+            openChannelData.participants.length == openChannelData.balances.length,
```
lukastanisic99: This is not logically wrong, but for the security assumptions we have it's not needed. The main design decision is we try to not do unnecessary compute `on-chain`, since the chain is the `bottleneck` - it's the shared global state we can not scale + computation on-chain costs fees. If you take a look at our entire protocol, we try and use auditable commitments, where peers commit to something on-chain and other peers can audit and verify it. In the happy path no computation is done on-chain and if someone tries to be malicious then the chain does the computation and slashes someone. Economic game theory should keep all of this in order. 

So to come back here, while this check and on-chain enforcement is not wrong, it's not needed due to the assumption that at least 1 honest peer is in the channel and that peer audited the payload its signing, so if the payload is malformed a threshold can `NOT` be reached with an honest participant in the channel. 

What is important that regardless if peers are colluding or not, the chain actually keeps track of actual processed deposits and withdrawals, so no channel can withdraw more than the sum of their deposits - this is important and that's why all the deposit and withdraw flows happen through 2 functions. 

Later we'll go and optimize more places in our contracts to follow these patterns, since in a few places we actually still do computation in the happy path and that computation is often not needed.

NiKrause: Thanks — this landed before I saw the review. The contract change is gone entirely (`git diff dispute -- contracts/` is empty); what ships now is six tests holding the two properties your reasoning rests on, so the decision is falsifiable rather than prose:

- an abstaining participant keeps a malformed term below the threshold, and a colluder can't substitute its own signature for the missing one
- unanimously signed surplus balances open the channel but record exactly the totals of a well-formed open — with a control case, so "the totals look right" isn't unfalsifiable

On your last point — the withdrawal cap under collusion — I went looking for a gap and there isn't one: `E2E-MaliciousUpdateSnapshot.test.ts:20` ("colluded over-withdrawal → `CantWithdrawMoreThanDeposits`") already covers it, mapped to `INV-LIF-5-ENQB91.T1.P3`, enforced at `StateSnapshotFacet.sol:141`. So I left it alone rather than duplicating it.

`DEF-1-92NTAG` is closed as won't-fix with the rationale, and the decision is recorded in `admission-and-funds.md` under assumptions + security considerations.

PR #445, opened 2026-08-21
Test containers for distributed tests

lukastanisic99:
# Human

This is backwards compatible with the current orchestrator which everyone is using for testing. This introduces workers running all tests inside a docker container, instead of running bare metal on the host. The performance overhead is more or less non existent since docker containers are so efficient. The added benefit is that everyone has their own cached container on the worker, so they don't compete with everyone else on the file system and we don't have to rebuild and reinstall dependencies every time someone else uses the test infra between our runs. Mostly did this, so a bad test can't fuck up the host machine where I'd have to manually restart it. You can only fuck up your container.

It also introduces access management though key pairs, but it's optional and can be turned on later. 

There are again some test flakes, but they are the same pattern from my previous PR - apparently didn't fix everything - will see later what's a good solution for everything so these same bugs don't cause flakes all the time

# Agent

## Summary

This PR moves distributed test workers into isolated, per-orchestrator Docker environments instead of preparing and executing untrusted projects directly on the worker host.

Docker is the default backend. An explicit `unsafe-host` backend remains available for trusted, disposable development environments.

## Key changes

- Add a hardened Docker execution backend with:
  - non-root execution
  - read-only base filesystem
  - dropped capabilities and `no-new-privileges`
  - CPU, memory, PID, and disk limits
  - isolated networking with host/private-network egress restrictions
  - no host mounts, Docker socket, worker credentials, or host shell access

- Move source extraction, dependency installation, builds, test infrastructure, and test execution inside the isolated environment.

- Add a versioned, bounded, data-only host/guest protocol for:
  - source transfer
  - environment setup
  - task execution
  - status and diagnostics
  - failure artifact transfer
  - resource-limit reporting

- Cache one persistent environment per orchestrator identity:
  - source, dependencies, and build artifacts are reused across leases
  - stopped environments consume disk but no reserved CPU or memory
  - cache count and disk usage are bounded with idle LRU eviction
  - corrupt or dirty environments are rebuilt without affecting other identities

- Add configurable execution profiles covering scheduler settings, worker slots, CPU, memory, disk, and process limits. Oversized requests are rejected before starting the environment.

- Preserve the existing external distributed-test flow and gate new protocol extensions through advertised worker capabilities.

- Add transport-key authorization alongside the existing shared worker-set secret:
  - persistent orchestrator and admin allowlists
  - migration mode for unlisted shared-secret clients
  - remotely managed authorization entries
  - worker-owned audit logs for admission, execution, resource, and administration events

- Add operator-facing commands:
  - `yarn distributed:identity` prints the persistent orchestrator public key
  - `yarn distributed:admin workers` discovers workers
  - `yarn distributed:admin authorization-list`
  - `yarn distributed:admin authorization-add`
  - `yarn distributed:admin authorization-remove`
  - admin commands target all discovered workers by default or one worker with `--worker`

- Add Docker self-check, integration, and benchmark commands for validating real worker hosts.

## Reliability improvements

- Preparation uses an inactivity timeout rather than a fixed total deadline, with guest keepalives for silent long-running commands.
- Recoverable preparation failures retain the environment cache for retry.
- Cached containers update CPU, memory, and PID limits when later leases request a different profile.
- Environment listeners are detached after every lease, preventing duplicate frame handling on reused environments.
- Artifact transfers have bounded timeouts and preserve stdout/stderr chronology.
- Unexpected worker exits are classified as resource exhaustion or worker failures and affect only that environment.
- Docker containers use an init process so terminated task descendants are reaped correctly.
- Unsafe-host environments are identity-scoped and recoverable under the configured worker root.

## Verification

- Added focused coverage for isolated-runtime lifecycle, protocol validation, workspace caching, execution profiles, authorization, administration, auditing, egress policy, failure artifacts, resource limits, and crash recovery.
- Passed both Node and browser TypeScript builds.
- Passed compilation and specification impact checks.
- Passed real-Docker self-check and disposable integration testing on a Debian worker, covering:
  - filesystem and identity isolation
  - public and loopback networking
  - worker-host and private-network denial
  - CPU, memory, PID, and disk enforcement
  - cached-container reuse and profile updates
  - crash cleanup and environment replacement
- The latest distributed full run completed with two unrelated test-harness race flakes; both were diagnosed as stale/overly strict dispute-test assertions rather than Docker or worker infrastructure failures.


sh3ll3x3c:
@lukastanisic99 Two tests fail on this branch, independent of #447 which stacks on it.

- `e2eParallelProtocol.test.ts` — "extracts a chunked source transfer with a manifest larger than 512 KiB" — `Error: Timed out waiting for PREPARED` (10s timeout in the test)
- `e2eParallelAuthorizationAdministration.test.ts` — "uses the persistent admin identity to add an admin key over worker discovery" — `Error: Protocol stream is closed`

I checked out `test-containers` in a clean worktree, without my commit, and ran only these two on the pool. Both fail there too.

They are probably not load flakes. They also failed on a run at 45% average CPU. Three other tests that failed in the same full run passed on re-run; these two did not.

Flagging so reviewers of #447 do not read them as regressions from my change.

PR #447, opened 2026-08-21
feat(test-runner): run Foundry tests as parallel and distributed tasks

sh3ll3x3c:
## Summary

Runs the Foundry test suite through the same parallel and distributed runner as
the Mocha tests, so `forge` tests are part of the canonical gate instead of never
running in CI.

Foundry test contracts are discovered alongside the Mocha tests and scheduled as
ordinary tasks, one task per test contract (7 today). A contract counts as a test
contract when it declares a `test`, `invariant`, or `statefulFuzz` function,
including through a same-file base, so harness and helper contracts sharing a
file are left out.

### Why forge runs through a Hardhat task

A distributed worker executes tasks with the runner code from its **own
checkout** — `trustedRunnerManifest` copies `scripts/e2e-parallel/**` from the
worker host into the container. Only project sources are synced to a worker. So
a runner taught to spawn the `forge` binary directly would not work on any worker
until that worker was updated and restarted.

`hardhat.config.ts` is a synced project source, and the runner already spawns the
Hardhat CLI. So forge runs through a `forge-test` Hardhat task
(`tasks/forgeTest.ts`), which reaches every worker as an ordinary source file.
Forge tests run on the existing pool with no worker changes at all.

The task takes `--match-contract` and `--threads`, spawns forge with inherited
stdio so its output streams into the runner's log capture, and propagates forge's
exit code so the runner's pass/fail detection works. It does not depend on the
compile task, so no task recompiles.

Each task is pinned to one thread: `forge test` otherwise sizes its pool from the
logical core count, which inside a CPU-limited container is still the host's
count, and the runner already parallelizes across tasks.

Forge tasks need no Hardhat node, so they take neither a warm slot nor a funded
account partition, and a run made only of forge tasks skips provisioning the
Hardhat node and discovery server altogether.

### Packaging

Two changes make this work on distributed workers, which already have `forge` in
the runner image from #445:

- `lib/forge-std` was tracked but matched `.gitignore`, so the runtime bundle
  dropped it as a tracked-but-ignored path and workers had no `forge-std` to
  compile against. It now ships (+57 files, ~976 KB).
- The cached prepare script builds the forge artifacts, so parallel forge tasks
  never start concurrent `via_ir` compiles into a shared `out/`. Local runs build
  once before scheduling. `foundry.toml` is a contract-compile cache input;
  editing a forge `.sol` test selects the cached prepare, not a full Hardhat
  recompile.

### Flags

```shell
yarn test:parallel --forge-only
yarn test:parallel --no-forge
yarn test:parallel --forge-threads 2
```

`--e2e-only` selects the Mocha end-to-end tier and drops the forge tier with it.

## Test Plan

- `test/scripts/e2eParallelForgeTasks.test.ts` — unit coverage for discovery
  (prefixes, same-file inheritance, harness exclusion, comment and string
  handling, duplicate-name detection), the Hardhat-task argument form, agreement
  between discovery and the task's registered parameter names, the wire round
  trip, tier selection and flag conflicts, slot resolution, and attempt reduction
  of forge output.
- A parity test asserts the static discovery finds exactly the contracts
  `forge test --list --json` reports, so the hand-rolled Solidity parsing cannot
  drift from foundry. It skips when artifacts are cold or `forge` is unavailable.
- Discovery verified by dry run: `--e2e-only` yields 0 forge tasks; conflicting
  flags are rejected; a forge-only run resolves to 0 slots.
- **Forge tests on the distributed pool: 7/7 passing**, on unmodified workers.
- Full distributed run, Mocha and forge together: **1349 passing of 1351 tasks**
  (1344 Mocha + 7 forge), 370s wall, 50.7x speedup. Both failures are the
  pre-existing ones below; all 7 forge tasks passed.
- Two failures seen in full runs — `e2eParallelProtocol` "extracts a chunked
  source transfer with a manifest larger than 512 KiB" and
  `e2eParallelAuthorizationAdministration` "uses the persistent admin identity to
  add an admin key over worker discovery" — **fail identically on
  `test-containers` without this commit**, verified from a clean worktree of the
  base, so they are pre-existing and not caused here.
- `yarn spec:impact --base origin/test-containers`: 0 impacted requirements.

Stacked on #445, which supplies the Docker runner image that installs `forge`.


scripts/e2e-parallel/shared/forgeTaskDiscovery.js
```diff
@@ -0,0 +1,225 @@
...
+            bases: parseBaseNames(match[3]),
+            declaresTest: TEST_FUNCTION.test(blockBody(source, openIndex))
+        });
+    }
+    return declarations;
+}
+
+/**
+ * Whether a contract owns a test function, directly or through a base declared
+ * in the same file. Resolution stops at the file boundary — an imported base is
```
lukastanisic99, line 110: **Human note**

Yes

**AI-generated review**

🟠 **[FO3] `--forge-only` can finish green without running valid tests inherited from an imported base.** Foundry runs a concrete `ContractTest is SharedTests` even when all `test*` functions live in an imported file. This parser intentionally stops at the file boundary, so it omits that contract. The default mixed run may catch current-repository drift through the Mocha parity test, but `--forge-only` has no such oracle and silently skips the test.

> **Fix FO3-FIX**
>
> Resolve imported base contracts, or discover contracts from Forge's authoritative listing after one warm build. Add a two-file fixture whose concrete contract inherits every test function from an imported abstract base, and assert that `--forge-only` schedules it.

scripts/test-e2e-parallel.js
```diff
@@ -79,44 +129,74 @@ async function main(options = {}) {
...
     }
 
     // ---- resolve config ----
-    // Slots are provisioned unconditionally and offered to every task; no
-    // task is classified by directory or source. A test uses the shared node
-    // (via PROVIDER_URL) or ignores it and keeps hardhat's in-process network.
     const requestedSlotCount = cli.slots ?? DEFAULT_SLOTS;
-    const slotCount = Math.min(requestedSlotCount, MAX_SLOTS_FROM_POOL);
-    if (requestedSlotCount > slotCount) {
+    const slotCount = resolveSlotCount(
```
lukastanisic99, line 194: **Human note**

Nit

**AI-generated review**

🟠 **[PO1] Distributed `--forge-only` still provisions Hardhat infrastructure.** `resolveSlotCount` correctly computes zero slots, but that value is used only by the local path. Unless the caller explicitly passes `--slots`, this execution profile omits `slots`; the server restores its default of one and the worker starts one Hardhat node and discovery process before any Forge task. A Forge-only run can therefore fail on unrelated infrastructure and pays its CPU, memory, and startup cost.

> **Fix PO1-FIX**
>
> Merge `{ slots: 0 }` into the distributed execution profile whenever the discovered task list contains no chain task. Add a distributed assertion that the worker receives zero and never provisions node or discovery infrastructure.

lukastanisic99 (commented):
**Human note**

Sure

**AI-generated review**

🟠 **[TO2] Tier-selection tests stop before distributed preparation and profile resolution.** [`e2eParallelForgeTasks.test.ts:600-615`](https://github.com/peer3to/state-channels-plus/blob/f228420851e1e147b6fb67aa9fda2c65d58c8896/test/scripts/e2eParallelForgeTasks.test.ts#L600-L615) checks only the pure discovery-selection object, and [`e2eParallelForgeTasks.test.ts:709-716`](https://github.com/peer3to/state-channels-plus/blob/f228420851e1e147b6fb67aa9fda2c65d58c8896/test/scripts/e2eParallelForgeTasks.test.ts#L709-L716) checks only the local slot helper. Neither reaches the worker preparation/profile paths that cause FO2 and PO1.

> **Fix TO2-FIX**
>
> Exercise the distributed entry point with `--no-forge`, `--e2e-only`, and `--forge-only`. Assert the first two never run a Forge command, and the last sends `slots: 0` and starts no Hardhat infrastructure.



---

**Human note**

sure

**AI-generated review**

🟠 **[TO3] The local scheduler's Forge branch has no execution test.** The suite checks `requiresChainSlot` and `resolveSlotCount`, but does not run [`local/scheduler.js:132-148`](https://github.com/peer3to/state-channels-plus/blob/f228420851e1e147b6fb67aa9fda2c65d58c8896/scripts/e2e-parallel/local/scheduler.js#L130-L149). A regression that allocates an account, rotates a slot, or leaks provider/account variables into Forge can pass all new tests.

> **Fix TO3-FIX**
>
> Run one mixed local schedule through the real scheduler and assert the Forge child receives only the base environment, takes no partition, does not advance slot rotation, and releases all resources after success and failure.



---

**Human note**

sure

**AI-generated review**

🟠 **[TO4] The distributed worker's Forge branch is tested only through wire serialization.** No test reaches [`distributed/worker.js:225-275`](https://github.com/peer3to/state-channels-plus/blob/f228420851e1e147b6fb67aa9fda2c65d58c8896/scripts/e2e-parallel/distributed/worker.js#L225-L275), so the claim that remote Forge tasks take no slot/account and omit chain environment is not pinned.

> **Fix TO4-FIX**
>
> Start a worker with one mixed assignment stream and assert Forge gets no slot, partition, or provider variables, while the adjacent Hardhat task still gets all three. Cover Forge failure and cancellation too.



---

**Human note**

Sure

**AI-generated review**

🟠 **[TO5] Forge process failures are not tested through the Hardhat task.** The tests regex-read [`tasks/forgeTest.ts`](https://github.com/peer3to/state-channels-plus/blob/f228420851e1e147b6fb67aa9fda2c65d58c8896/tasks/forgeTest.ts#L23-L39) and test generic output reduction, but never execute the task with a missing binary, non-zero exit, or signal. They therefore do not prove those conditions become failed attempts with useful persisted output.

> **Fix TO5-FIX**
>
> Run the registered task in a temporary project with a controlled `PATH` and a real fixture executable for success, non-zero status, and signal; use an absent executable for spawn failure. Assert the parallel runner marks each failure and preserves the diagnostic in the task log.



---

**Human note**

Yes

**AI-generated review**

🟡 **[CY1] Forge protocol constants have several owners.** The binary name is duplicated in [`taskRunners.js`](https://github.com/peer3to/state-channels-plus/blob/f228420851e1e147b6fb67aa9fda2c65d58c8896/scripts/e2e-parallel/shared/taskRunners.js#L11-L14) and [`forgeTest.ts`](https://github.com/peer3to/state-channels-plus/blob/f228420851e1e147b6fb67aa9fda2c65d58c8896/tasks/forgeTest.ts#L15-L21); the thread default and `forge-test` task name are also duplicated across JS and TS. Comments say to keep them in sync, and tests regex-inspect source instead of using one owner.

> **Fix CY1-FIX**
>
> Put the task name, binary, and default threads in one shared module that both layers import. Test the shared values through behavior rather than source regexes.



---

**Human note**

yes

**AI-generated review**

🟠 **[OO1] What is the intended scope of `--test-pattern` now that Forge is a default tier?** It is documented as a consumer filename filter and existing consumers use `**/*.ts`, but main passes it only to Mocha even though Forge discovery also accepts a `testPattern` option. The README's small distributed smoke command now schedules its one Mocha file plus every Forge contract.

> **Fix OO1-FIX**
>
> Make `--test-pattern` filter the complete suite: both Mocha files and Forge test contracts. Add `--mocha-test-pattern` and `--forge-test-pattern` for tier-specific filtering. Pass each effective pattern to its corresponding discovery path, allow a tier filtered to zero tasks without failing the run, and cover the shared and tier-specific flags in distributed mode.

tasks/forgeTest.ts
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * Run forge in the project root and resolve its exit code. `stdio: "inherit"`
+ * hands forge this process's own stdout/stderr, so its output streams straight
+ * into the runner's log capture. A null exit code means a signal or a failed
+ * spawn — both failures.
+ */
+function runForge(args: string[], cwd: string) {
+    return new Promise<number>((resolve) => {
+        const child = spawn(FORGE_BIN, args, { cwd, stdio: "inherit" });
```
lukastanisic99, line 31: **Human note**

Yes

**AI-generated review**

🟠 **[PO2] Forge and solc memory is invisible to the scheduler's memory gate.** The runner tracks only the Hardhat child PID. This task then starts Forge as a grandchild, while `ResourceGate` samples only the exact tracked PIDs with `ps -p`. Its RSS is omitted, and repeated small-wrapper samples can lower the per-test estimate to the 0.25 GiB floor. Constrained workers can then admit many real Forge processes past `memBoundGb` and be killed by the host or cgroup.

> **Fix PO2-FIX**
>
> Account for the complete detached process group/process tree, or report the Forge child PID to the runner, for current RSS and per-task averages. Add a gate test with a tracked grandchild that allocates memory and prove admission uses that RSS.

test/scripts/e2eParallelForgeTasks.test.ts
```diff
@@ -0,0 +1,734 @@
...
+            expect(() => discoverForgeTasks(root)).to.throw(
+                /matches contract names project-wide/
+            );
+        } finally {
+            fs.rmSync(root, { recursive: true, force: true });
+        }
+    });
+
+    it("finds exactly the test contracts foundry itself lists for the repository", function () {
+        if (!forgeArtifactsAreWarm()) this.skip();
```
lukastanisic99, line 352: **Human note**

Sure

**AI-generated review**

🟠 **[TO6] The only parser-versus-Forge oracle can silently skip twice.** This is the one case capable of detecting a contract that the static parser omitted, but it passes as skipped when artifacts are cold or Forge is unavailable. It also validates only the current same-file layout, so FO3 remains uncovered.

> **Fix TO6-FIX**
>
> Make parity a required Forge-tier test after the runner's warm build, fail with a clear prerequisite error instead of skipping, and include the imported-base fixture from FO3.

test/scripts/e2eParallelForgeTasks.test.ts
```diff
@@ -0,0 +1,734 @@
...
+    });
+
+    it("reads a wire task without a runner as a hardhat task", function () {
+        const root = fs.mkdtempSync(path.join(os.tmpdir(), "legacy-wire-"));
+        try {
+            const task = fromWireTask(
+                {
+                    label: "test:logic.test.ts:runs",
+                    logName: "logic__runs",
+                    runner: undefined as unknown as string,
```
lukastanisic99, line 556: **Human note**

Nit

**AI-generated review**

🟡 **[AY2] This test bypasses the real wire type with the banned `as unknown as` pattern.** The scenario is valid—legacy payloads omit `runner`—but the cast restates an invalid shape instead of letting the boundary type describe backward compatibility.

> **Fix AY2-FIX**
>
> Make `runner` optional on the input wire type and normalize it at the boundary, then omit the property in this fixture with no cast.

README.md
```diff
@@ -86,6 +86,45 @@ Compile the contracts and run tests
 yarn testc
 ```
 
+### Foundry tests
```
lukastanisic99, line 89: **Human note**

True

**AI-generated review**

🟠 **[DO1] A fresh checkout cannot follow the documented default test setup.** The setup still says only `yarn`, but the default parallel gate now requires an external `forge` binary and the `lib/forge-std` git submodule. A normal clone without recursive submodules leaves that dependency empty, so the new mandatory prebuild fails.

> **Fix DO1-FIX**
>
> Document a pinned Foundry installation and `git submodule update --init --recursive`, or bootstrap and validate both prerequisites in a setup script. Update any checkout that adopts this gate to fetch submodules.

README.md
```diff
@@ -86,6 +86,45 @@ Compile the contracts and run tests
...
+when it declares a `test`, `invariant`, or `statefulFuzz` function, so harness
+and helper contracts sharing a file are left out.
+
+```shell
+yarn test:parallel --forge-only     # only the forge tier
+yarn test:parallel --no-forge       # only the Mocha tier
+yarn test:parallel --forge-threads 2
+```
+
+Each forge task runs on a single thread. `forge test` otherwise sizes its thread
```
lukastanisic99, line 102: **Human note**

Nit

**AI-generated review**

🟡 **[DY3] The Foundry thread documentation contradicts its example.** The example immediately above sets two threads, but this sentence says every task uses one.

> **Fix DY3-FIX**
>
> Say that each task uses one thread by default, and explain that `--forge-threads` overrides it.

scripts/e2e-parallel/shared/argParser.js
```diff
@@ -130,6 +140,31 @@ function parseCliArgs(argv) {
...
+        if (arg === "--forge-only") {
+            options.forgeOnly = true;
+            continue;
+        }
+        if (arg === "--no-forge") {
+            options.forge = false;
+            continue;
+        }
+        if (arg === "--forge-threads") {
+            const v = takeNumber(argv[i + 1], (s) => Number.parseInt(s, 10));
```
lukastanisic99, line 152: **Human note**

Nit

**AI-generated review**

🟡 **[MY1] `--forge-threads` accepts invalid values while claiming to require an integer.** `parseInt` turns `1.5` into `1` and `2junk` into `2`; the current tests reject only zero.

> **Fix MY1-FIX**
>
> Parse the complete value and require `Number.isInteger(value) && value > 0`. Add decimal, suffix, negative, missing, and non-numeric cases for both flag forms.

PR #448, opened 2026-08-22
fix(test-runner): drop the fs-ext host lock for a pure-JS pid lock

sh3ll3x3c:
## Summary

Fixes the dependency problem reported in review on #421 (`package.json` L61/L70/L77).

`fs-ext`, `tar` and `@hyperswarm/dht` were in `dependencies`, but none is referenced under `src/` — the consumers are `scripts/e2e-parallel/distributed/*` and `test/`. `fs-ext` additionally pulls `nan` and needs a node-gyp build, so a plain `yarn install --frozen-lockfile` failed on any host without a working native toolchain, and `hostLock.js` required it unconditionally — meaning the test suite could not run at all without a built binding. `acquireHostLock` also refused to run on Windows.

Changes:

- **`hostLock.js` no longer uses `flock(2)`.** The lock is now an atomic `link()` of a pid file: the owner writes its pid to a staging file and links it into place (so the lock file is never observed empty), and a lock whose recorded owner no longer answers `process.kill(pid, 0)` is reclaimed. That restores the release-on-owner-death behaviour `flock` gave for free, with no native code. The existing symlink rejection is kept.
- **Platform gates removed.** `link()` and `process.kill(pid, 0)` work everywhere, so `acquireHostLock`/`acquireOsFileLock` no longer throw on Windows and the two `this.skip()` guards in the lease tests are gone.
- **`fs-ext` removed outright** from `package.json` and `yarn.lock` (along with its now-orphaned `nan` entry). `tar` and `@hyperswarm/dht` move to `devDependencies`. `@hyperswarm/dht-relay` stays in `dependencies` — it is used by `src/HolepunchRelay.ts`.
- `pnpm.onlyBuiltDependencies` and `peer3TestDistribution.verifyNativeModules` are now empty lists. Both consumers already handle an empty list (`workspacePreparation.js:15`, `runtimeBundle.js:175`); the plumbing stays for the next native dependency.
- `engines.node` declared as `>=20`, matching CI.

Stacked on #447 (stack #446: #443 → #445 → #447 → #448). No code dependency on #447 — the commit cherry-picks cleanly onto it, which is also how I confirmed the two PRs touch no common files.

It is stacked rather than branched as a sibling because `test-containers` is an active branch that keeps moving; a single chain is one rebase instead of two. Going the other way (onto `dispute` or `fix-flakes`) is not free: `test-containers` deleted two lease tests and rewrote a third, so the commit conflicts in `e2eParallelWorkerLease.test.ts` on those bases.

## Test Plan

- `test/scripts/e2eParallelWorkerLease.test.ts` — the existing exclusivity/bypass case keeps its coverage (renamed, since the lock is no longer OS-held), plus three new cases: reclaiming a lock whose owning process is gone, reclaiming one whose owner cannot be parsed, and refusing a symlinked lock path.
- `yarn tsc --noEmit -p tsconfig.json` — clean for the touched files. (Four pre-existing errors remain in `DisputeFraudProofService.ts` / `DisputeValidationService.ts` / `Codec.ts` / `DisputeManager.test.ts` from stale typechain output; unrelated to this change and present on the base.)
- Behaviour verified directly against the module: exclusivity while held, `allowSharedHost` bypass, lock file removed on release, reacquire after release, reclaim of a dead-pid lock, reclaim of an unparseable lock, symlink rejection, independent paths not contending, and a second **process** blocked with the contention message.
- Full run left to the distributed pool.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
     "ethers": "6.13.7"
   },
   "devDependencies": {
+    "@hyperswarm/dht": "^6.0.1",
     "@nomicfoundation/hardhat-chai-matchers": "^2.0.0",
     "@nomicfoundation/hardhat-ethers": "^3.0.0",
     "@nomicfoundation/hardhat-foundry": "^1.1.3",
@@ -50,6 +51,7 @@
     "sinon": "^20.0.0",
     "solidity-coverage": "^0.8.1",
     "source-map-support": "^0.5.21",
+    "tar": "^7.4.3",
     "tsc-alias": "^1.8.16",
     "typechain": "^8.3.0",
     "vite": "^7.1.5",
@@ -58,7 +60,6 @@
   "dependencies": {
     "@ethereumjs/evm": "^3.1.0",
     "@ganache/console.log": "^0.4.2",
-    "@hyperswarm/dht": "^6.0.1",
     "@hyperswarm/dht-relay": "0.3.0",
     "@nomiclabs/hardhat-ethers": "2.2.3",
     "@nomiclabs/hardhat-waffle": "2.0.6",
@@ -67,14 +68,12 @@
     "buffer": "^6.0.3",
     "ethereum-waffle": "4.0.10",
     "fflate": "^0.8.2",
-    "fs-ext": "^2.1.1",
     "get-webrtc": "^1.0.1",
     "glob": "10.3.10",
     "hardhat": "^2.22.1",
     "hyperswarm": "4.3.5",
     "mocha": "10.7.3",
     "prompt": "1.3.0",
-    "tar": "^7.4.3",
     "ts-morph": "^27.0.2",
     "ts-node": "^10.9.2",
     "tsconfig-paths": "^4.2.0",
@@ -82,6 +81,7 @@
     "ws": "^8.18.0"
   },
   "engines": {
+    "node": ">=20",
     "yarn": ">=1.15"
   },
   "scripts": {
@@ -200,16 +200,12 @@
   },
   "author": "",
   "pnpm": {
-    "onlyBuiltDependencies": [
-      "fs-ext"
-    ]
+    "onlyBuiltDependencies": []
   },
   "peer3TestDistribution": {
     "prepareScript": "test:parallel:prepare",
     "cachedPrepareScript": "test:parallel:prepare:cached-contracts",
-    "verifyNativeModules": [
-      "fs-ext"
-    ],
+    "verifyNativeModules": [],
     "contractCompileInputs": [
       "contracts/",
       "hardhat.config.ts",
--- a/scripts/e2e-parallel/distributed/hostLock.js
+++ b/scripts/e2e-parallel/distributed/hostLock.js
@@ -10,11 +10,6 @@ const HOST_LOCK_PATH = path.join(HOST_LOCK_DIR, "server-v8.lock");
 
 function acquireHostLock(options = {}) {
     if (options.allowSharedHost) return { release() {} };
-    if (process.platform !== "darwin" && process.platform !== "linux") {
-        throw new Error(
-            "Host locking is unsupported; pass --allow-shared-host to bypass it"
-        );
-    }
     return acquireOsFileLock(
         options.lockPath || HOST_LOCK_PATH,
         "Another test:parallel:server owns this host",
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -50,6 +51,7 @@
     "sinon": "^20.0.0",
     "solidity-coverage": "^0.8.1",
     "source-map-support": "^0.5.21",
+    "tar": "^7.4.3",
```
lukastanisic99, line 54: **Human note**

yes

**AI-generated review**

🔴 **[FR1] The installed package no longer declares dependencies required by its public test-runner exports.** The package ships and exports [`./test-parallel`, `./test-parallel-server`, and `./test-parallel-worker`](https://github.com/peer3to/state-channels-plus/blob/7b6e2612251c6442a5197582aab6519efde2e8d0/package.json#L161-L188). The runner eagerly loads [`runtimeBundle.js`](https://github.com/peer3to/state-channels-plus/blob/7b6e2612251c6442a5197582aab6519efde2e8d0/scripts/test-e2e-parallel.js#L40-L47), which directly requires [`tar`](https://github.com/peer3to/state-channels-plus/blob/7b6e2612251c6442a5197582aab6519efde2e8d0/scripts/e2e-parallel/distributed/runtimeBundle.js#L1-L6); the distributed identity path directly requires [`@hyperswarm/dht`](https://github.com/peer3to/state-channels-plus/blob/7b6e2612251c6442a5197582aab6519efde2e8d0/scripts/e2e-parallel/distributed/persistentIdentity.js#L1-L5). Package consumers do not install this package's `devDependencies`. A flat install may hide the error through accidental transitive hoisting, but a strict install can fail as soon as one of these public entry points loads.

> **Fix FR1-FIX**
>
> Keep `tar` and `@hyperswarm/dht` in `dependencies`. Add a packed-package consumer test that installs production dependencies only under a strict package manager and loads/runs all three exported tools.

scripts/e2e-parallel/distributed/hostLock.js
```diff
@@ -10,72 +10,109 @@ const HOST_LOCK_PATH = path.join(HOST_LOCK_DIR, "server-v8.lock");
...
+        throw error;
+    } finally {
+        fs.rmSync(stagingPath, { force: true });
+    }
+}
+
+function clearStaleLock(lockPath) {
+    const ownerPid = readLockOwner(lockPath);
+    if (ownerPid === undefined) return true; // vanished under us; the slot is free
+    if (ownerPid !== null && isProcessAlive(ownerPid)) return false;
```
lukastanisic99, line 73: **Human note**

yes

**AI-generated review**

🟠 **[FO2] A dead owner's PID can be reused by an unrelated process and make the lock unreclaimable.** `kill(pid, 0)` proves only that some process currently has that number. If the operating system reuses the PID before the next server starts, this code treats the unrelated process as the old owner and rejects every acquisition until that process exits. `flock` released the old kernel lock when the owning process died, so this is a new failure mode.

> **Fix FO2-FIX**
>
> Do not use a PID as the complete owner identity. Use a maintained cross-platform lock primitive with atomic stale recovery and owner tokens, or store and validate a process-generation value plus a unique token. Add a test that simulates PID reuse and proves the dead owner's lock is reclaimable without deleting a current owner's lock.

scripts/e2e-parallel/distributed/hostLock.js
```diff
@@ -10,72 +10,109 @@ const HOST_LOCK_PATH = path.join(HOST_LOCK_DIR, "server-v8.lock");
...
+    fs.rmSync(lockPath, { force: true });
```
lukastanisic99, line 74: **Human note**

Same solution as for FO2

**AI-generated review**

🔴 **[RG1] Stale-lock recovery can delete a replacement lock and let two live servers acquire it.** The stale check and pathname deletion are not one atomic operation. Reproduced interleaving: process A and B both observe the same dead owner; B deletes it and claims the path; A then executes this `rmSync`, deleting B's live lock, and claims the now-free path. Both `acquireHostLock` calls return live handles. A deterministic two-child probe produced `{"bothAcquiredWhileAlive":true,"aPid":55412,"bPid":55413,"recordedOwner":55412}`. That breaks the single-server invariant this lock exists to enforce.

> **Fix RG1-FIX**
>
> Make stale-owner replacement atomic with respect to the exact file that was inspected. Prefer a maintained pure-JS lock implementation with atomic acquisition, stale recovery, heartbeat, and compromise detection. Add a child-process test that pauses one reclaimer after its stale read, lets another process reclaim, resumes the first, and asserts that exactly one handle exists.

lukastanisic99 (commented):
**Human note**

sure

**AI-generated review**

🟠 **[TR1] The tests do not exercise the cross-process guarantee or the claimed Windows support.** The exclusivity case calls `acquireHostLock` twice [inside one process](https://github.com/peer3to/state-channels-plus/blob/7b6e2612251c6442a5197582aab6519efde2e8d0/test/scripts/e2eParallelWorkerLease.test.ts#L184-L200), so both attempts have the same PID. The stale test uses a child only to obtain a PID after that child has exited. No test has two live processes contend or reclaim concurrently, which is why RG1 passes. Removing the Windows skips also does not add a Windows job; the repository workflow still runs only on [`ubuntu-latest`](https://github.com/peer3to/state-channels-plus/blob/7b6e2612251c6442a5197582aab6519efde2e8d0/.github/workflows/ci.yml#L12-L24).

> **Fix TR1-FIX**
>
> Drive the lock through real child processes. Cover live contention, concurrent stale recovery, owner crash, release/reacquire, and bypass behavior. Run that focused suite on Windows as well as Linux before claiming Windows support.

lukastanisic99 (approved):

PR #449, opened 2026-08-22
feat(contracts): carry the failed comparison in dispute and snapshot errors

sh3ll3x3c:
Phase 1 of [Trello 377](https://trello.com/c/7JIKsiuG) — populate Solidity errors with arguments so the on-chain state at a failure can be reconstructed. Related to #406 / #411.

## Summary

Fourteen custom errors on the dispute-upload, reduction and snapshot paths reverted with a name and nothing else. A rejected submission said *which rule* broke but never *which values* broke it, so reconstructing the on-chain state meant a follow-up chain read — if it was still possible at all.

Each now carries the operands of the check that rejected it, named as a pair so the comparison reads at a glance:

| Error | Arguments |
| --- | --- |
| `ErrorDisputerNotMsgSender` | `expectedDisputer, actualSender` |
| `ErrorDisputeAlreadyPosted` | `forkId, disputer` |
| `ErrorCantParticipateInDispute` | `channelId, participant` |
| `ErrorAuditingDataHashMismatch` | `expectedAuditingDataHash, providedAuditingDataHash` |
| `ErrorDisputePostedAuditingDataMismatch` | `expectedPostedAuditingData, actualPostedAuditingData` |
| `ErrorDisputeThrottled` | `disputer, throttleExpiry, currentTimestamp` |
| `ErrorInvalidLatestState` | `expectedStateMachineStateHash, actualStateMachineStateHash` |
| `RaceConditionReductionExpectationDoesntMatch` | `expectedReducedForkId, actualReducedForkId` |
| `RaceConditionSnapshotForkMismatch` | `currentForkId, submittedForkId` |
| `RaceConditionPendingInboundNotConsumed` | `submittedInboundMessageBlockHash, onChainInboundMessageBlockHash` |
| `RaceConditionDisputeTimeoutCalldataPosted` | `forkId, blockHeight, participant, blockCalldataCommitment` |
| `RaceConditionDisputeTimeoutPreviousBlockProducerPostedCalldataMismatch` | `previousBlockProducer, blockHeight, expectedPostedCalldata, actualPostedCalldata` |
| `RaceConditionDisputeTimeoutNotMinTimestamp` | `minTimestamp, currentTimestamp` |
| `RaceConditionDisputeTimeoutWindowCreatedTooEarly` | `windowCreationTimestamp, minTimestamp` |

18 construction sites across 5 facets.

**No client decoder is added.** `LoggerUtils.getCustomEvmErrorMetadata` (#406) already resolves argument names from the error's own ABI, so all fourteen log their operands for free. Client race handling keys on error **names**, and no 4-byte selector is hardcoded anywhere in `src/` or `test/`, so the signature changes are invisible to it. `GeneratedArtifacts.ts` is gitignored and regenerates.

`ErrorInvalidLatestState`'s site hoists `keccak256(encodedStateMachineState)` into a local shared by the condition and the error, so the hash is not computed twice.

## Scope

Errors whose failure carries no operand — empty-array guards like `ErrorNoDisputesProvided` — stay argument-less on purpose. The eleven declared-but-never-constructed error names are untouched.

`ErrorInvalidStateSnapshot` is deliberately excluded: its four call sites compare four different things (a snapshot-data hash, a block's snapshot hash, a genesis predicate, a timestamp), so one signature cannot serve them honestly. It needs a split decision, not arguments.

## Contract size — please read before extending this

`DisputeVerificationFacet` grows 46 bytes to **24541**, leaving **35 bytes** under the 24576 Spurious Dragon limit. Measured against `dispute` by comparing `deployedBytecode` length:

| Contract | base | now | delta |
| --- | ---: | ---: | ---: |
| DisputeManagerFacet | 15086 | 15316 | +230 |
| DisputeVerificationFacet | 24495 | **24541** | +46 |
| StateSnapshotFacet | 18143 | 18186 | +43 |
| JoinChannelFacet | 12908 | 12928 | +20 |

Nothing crossed the limit that was not already over it (`DisputeFraudProofFacet`, `StateChannelManagerProxy`, `LocalDiamond` were all over on `dispute` and are byte-identical here). But extending this to the remaining ~40 errors will push `DisputeVerificationFacet` over — that headroom needs addressing first.

## Solidity test expectations

`vm.expectRevert(bytes4)` matches the revert data **exactly** (forge-std reserves `expectPartialRevert` for prefix matching), so a bare selector no longer matches a revert that now carries arguments. Two sites are adjusted:

- `JoinChannelFacet.t.sol` — full data: `RaceConditionSnapshotForkMismatch(snapshot.forkId, keccak256("wrong-fork"))`
- `DisputeVerificationFacet.t.sol` — `vm.expectPartialRevert(...)`, pinning the error identity only

The timeout case does not assert its argument values. Passing the two `uint256` locals through `abi.encodeWithSelector` there encodes something the cheatcode reports as `(2, 2)` while the contract reverts with `(1, 2)` — literals in the same position pass, so it looks like a `viaIR` codegen quirk in that (large) test contract rather than a contract bug. Rather than ship a workaround for something not understood, that case keeps its original intent — error identity — and the argument values are asserted end to end in the e2e tests instead. Worth a look on its own if anyone wants to chase it.

## Test Plan

Four e2e upload-gate tests now assert the decoded arguments, not just the error name:

- `disputer.test.ts` — claimed disputer is `ZeroAddress`, actual sender is the submitting peer
- `channelId.test.ts` — the tampered channel id and the ineligible participant
- `disputeAuditingDataHash.test.ts` — the claimed hash differs from the hash the posted data produces; and both `postedAuditingData` entry-point mismatches assert the exact `(expected, actual)` pair
- `disputerThrottle.test.ts` (2 cases) — the throttled disputer, and the recorded expiry still ahead of the block timestamp

Their duplicated local `expectDecodedError` helper (3 copies) becomes one shared helper in `test/test_utils/customErrorAssertions.ts` that returns the decoded error.

Verified:

- `yarn test:parallel:distributed --e2e-only --grep "uploadRevert"` — **9 passing** on the pool (10 servers, 109s wall). Covers all four changed e2e files.
- `forge test` on both affected suites — `DisputeVerificationFacet.t.sol` 38 passing, `JoinChannelFacet.t.sol` 8 passing.
- Both forge suites also run **on the pool, 2/2 tasks passing**, via a throwaway branch merging this PR onto the infra stack tip (which carries #447's forge task support). The branch was discarded; this PR keeps its `dispute` base.
- `yarn hardhat compile --force` — clean
- `forge build --force` — clean (only pre-existing lint notes)
- `yarn tsc --noEmit` on `tsconfig.json` **and** `tsconfig.browser.json` — clean
- `yarn spec:impact` and `yarn spec:impact --staged` — no unmapped files, exit 0

CI on this PR fails in `Run tests` on `test/unit/BlockIngestService.test.ts:275` ("a confirmation carrying an on-chain timestamp → the stored block records it"). That failure is **pre-existing on `dispute`** — #422, also based on `dispute`, fails the same test. Because that step fails, the `Run e2e tests` step is skipped, so CI never exercised the tests in this PR; the pool run above is what covers them.

## Specification

- `Errors.sol.md` records the design decision: arguments carry the comparison, adding them changes the selector but not the name, and argument names read as `expected`/`actual` (or `current`/`submitted`) pairs.
- The four verification test reports have their overviews updated for the new assertions, and their declaration line references corrected.
- `yarn spec:refresh` regenerated the inverse reports; the only deltas are those line references.
- `test/test_utils/customErrorAssertions.ts` carries a `@spec-test-coverage-ignore` — it declares no runnable case.

## Note

`DisputeVerificationFacet.sol:252` has a commented-out `require(canParticipateInDisputes(...), ErrorCantParticipateInDispute())`. I left it as-is rather than silently updating dead code, so its signature is now stale — worth deleting or restoring in a separate pass.


sh3ll3x3c:
Verified the Solidity tests on Foundry.

Locally, both affected suites pass:

- `DisputeVerificationFacet.t.sol` — 38 passing
- `JoinChannelFacet.t.sol` — 8 passing

Also ran them through the distributed pool: **2/2 forge tasks passing** (one task per test contract). That needed #447's forge task support, which is not on `dispute`, so it was done on a throwaway branch merging this PR onto the infra stack tip. The branch was discarded — this PR keeps its `dispute` base.

Worth noting the runner reported `sum 2.36s` of actual forge execution against `wall 632s`; nearly all of that was syncing and building a fresh source tree on the worker.

Both `vm.expectRevert` sites were adjusted for this change, since `expectRevert(bytes4)` matches the revert data exactly and these errors now carry arguments — the join-channel case asserts the full data including both fork ids, and the timeout case pins the error identity with `expectPartialRevert`.

PR #450, opened 2026-08-22
PR 447 and 448 fixes + other parallel/distributed fixes

lukastanisic99:


daiagi:
## Human

ran the full gate from #453 on this branch. a few things i noticed while reading the runner code.

- the host lock moved to a heartbeat. that closes the pid-reuse hole, but as far as i can tell a live
server that stalls longer than the stale window looks the same to another server as a dead one, so
its lock can be taken while it is still running.

- on preparation: the workspace records the source fingerprint before it builds and marks itself
prepared after. if something interrupts in between, the retry can conclude there is nothing to
do. the worker will believe itself current while using stale source.

- `emptyForgeTierMessage` and the `includeMocha && files.length === 0` check in
`test-e2e-parallel.js` are both gone, removed in the same change that rewrites the forge parser.
those were the two things that failed the run when a tier discovered nothing. only the combined
`tasks.length === 0` check is left, and it passes as long as the other tier found something -> the
forge tier can drop to zero and the run still goes green on the mocha tests.

## AI

## 🏁 Race conditions

- 🔴 **[RR1] The host lock can be taken from a live server, which then dies uncaught.**
  [`lockSync`](https://github.com/peer3to/state-channels-plus/blob/e18dacc7/scripts/e2e-parallel/distributed/hostLock.js#L39-L44)
  passes no `onCompromised`, so proper-lockfile's default `(err) => { throw err; }` fires from an
  unref'd timer; there is no `uncaughtException` handler in `scripts/`. Reproduced with two real
  processes: A holds the lock and stalls 14s, B acquires at t+11s, both run ~3s, then A exits 1 and
  `shutdown()` never runs -> `pool.close()` and every `releaseLease`/`environment.stop()` are skipped.
  Second repro: A calls `release()` after unstalling and `rmdir`s B's live lock, killing B with
  `ECOMPROMISED`; base's `readLockOwner(lockPath) === process.pid` guard prevented exactly that, and
  [`release()`](https://github.com/peer3to/state-channels-plus/blob/e18dacc7/scripts/e2e-parallel/distributed/hostLock.js#L50-L56)
  no longer has it. `staleMs` is always the 10s default — `serverArgParser.js` exposes no key for it.
  [`workspaceLock.js:9`](https://github.com/peer3to/state-channels-plus/blob/e18dacc7/scripts/e2e-parallel/distributed/workspaceLock.js#L9-L12)
  has the same exposure. A solo stall is harmless; it needs a concurrent acquirer. The comment at
  `hostLock.js:27-29` says stale recovery renames the directory first — proper-lockfile 4.1.2 never
  renames anywhere.

## 🧱 Fundamental

- 🔴 **[FR1] An interrupted preparation marks the workspace prepared with a stale build.**
  [`selectPrepareScript`](https://github.com/peer3to/state-channels-plus/blob/e18dacc7/scripts/e2e-parallel/distributed/workspacePreparation.js#L46-L70)
  lost the `!cache.prepared` clause and can return `null`; `:151-153` also lost the
  `|| repository.prepareScript` fallback.
  [`isolatedGuest.js:142`](https://github.com/peer3to/state-channels-plus/blob/e18dacc7/scripts/e2e-parallel/distributed/isolatedGuest.js#L142)
  commits the manifest before building and `markPrepared` at `:184` is unconditional. Driven end to
  end: run2 commits S2 and the prepare throws; run3 sees `changed=[]` -> `null` -> no build; run4
  `prepared=true`. The workspace then runs the suite against the previous `dist/` indefinitely.
  `inspectWorkspace` only re-hashes `manifest.files`, and `dist/`, `out/`, `artifacts/` are gitignored
  so they are never re-verified. `RECOVERABLE_PREPARATION_FAILURE` does not set `environmentFailed`, so
  the `invalidate` at `server.js:732` never fires and the workspace is kept.

- 🟠 **[TO1] The discovery-parity test now builds Foundry from inside a scheduled task.**
  [`:462`](https://github.com/peer3to/state-channels-plus/blob/e18dacc7/test/scripts/e2eParallelForgeTasks.test.ts#L461-L468)
  replaced `if (!forgeArtifactsAreWarm()) this.skip()` with an unconditional `warmForgeArtifacts()`.
  Measured: 1.5s warm, **56.6s** with `out/` and `cache_forge/` removed, against the 60s global timeout
  -> 3.4s of headroom on an idle machine. Base also skipped when the binary was missing; head throws
  `Foundry forge is required for discovery parity`, so any machine without Foundry now hard-fails a
  test that used to skip. (The `out/` race is not reachable — every path with forge tasks pre-warms
  first, making the build a 0.5s no-op.)

- 🟠 **[TO2] Both per-tier empty-discovery guards were deleted, not just the forge one.** Only the
  combined `tasks.length === 0` check remains in `scripts/test-e2e-parallel.js`:
  $ node scripts/test-e2e-parallel.js --dry-run --forge-test-pattern "**/*.nosuchext"
    forge tasks : 0
    EXIT=0
The same holds for a Mocha glob that matches nothing, as long as the other tier found something. The
deleted comment stated the reason: "an empty forge tier is a defect, not an empty selection".

- 🟠 **[TO3] Three tests fail at this head.** `supports a consumer-defined test filename pattern`
passes on `dispute` and fails here -> regression. `accepts a repository with only default Mocha
tests` and `accepts a repository with only default Forge tests` are new and have never passed: they
run the parallel entry with `cwd` set to a bare temp repo with no `node_modules`, so
`shared/constants.js` cannot resolve hardhat. All three are deterministic over 3 runs each.

## ❓ Open questions

- `waitFrame`
([`e2eParallelProtocol.test.ts:724`](https://github.com/peer3to/state-channels-plus/blob/e18dacc7/test/scripts/e2eParallelProtocol.test.ts#L718-L730))
only waits for `PREPARED` and drops the `ERROR` frame, so
`"Distributed runner repository is missing"` surfaced as a bogus 10s timeout.

**Bottom line:** both distributed-infra failures hit from #453 are fixed here. Two 🔴 remain — the host
lock can be taken from a live server and then kills it uncaught (reproduced with two processes), and an
interrupted preparation permanently marks a workspace prepared against a stale build (reproduced end to
end). Neither is a test failure.

lukastanisic99:
@daiagi take a look again - hopefully it's fixed

scripts/e2e-parallel/distributed/hostLock.js
```diff
@@ -6,57 +7,98 @@ const HOST_LOCK_DIR = path.join(
...
+        if (ownerIsLive(observed, options)) {
+            throw new Error(contentionMessage);
+        }
+        recoverObservedOwner(lockPath, observed, options);
+    }
+}
+
+function claimLock(lockPath, owner) {
+    const stagingPath = `${lockPath}${CLAIM_MARKER}${process.pid}-${owner.token}`;
+    fs.writeFileSync(stagingPath, `${JSON.stringify(owner)}\n`, {
```
daiagi, line 96: `writeFileSync` sits outside the `try/finally` that removes the staging file -> a crash or ENOSPC here leaves a truncated `.claim-` behind, and `sweepRecoveryArtifacts` L282 then throws `Malformed lock owner record` on every later start.

for the work-root lock that file lives in the persistent cache dir, so the worker never recovers on its own.

daiagi (approved):
lgtm

PR #451, opened 2026-08-23
fix same-writer block race, inbound-run gaps, and dispute anchor audit

lukastanisic99:
- drop a local submission that lost the same-peer authoring race
- report incomplete inbound message-block runs instead of walking past the pinned snapshot; recover the gap from chain events
- run the inbound-anchor fraud check on both dispute paths and abstain when the pinned snapshot is unavailable locally
- keep finalized genesis deposits in the LocalDiamond balance mirror
- parallelize requestJoinSignature chain reads
- update spec layers, harden flaky join/dispute tests


src/stateManager/block/BlockProductionService.ts
```diff
@@ -34,14 +34,28 @@ export default class BlockProductionService {
...
+                    "Local submission lost a same-peer authoring race",
+                    {
+                        candidateHeight: Number(tx.header.transactionCnt),
+                        committedBlock: LoggerUtils.getBlockMetadata(
+                            staleLocalBlock,
+                            sm.storage
+                        )
+                    }
+                );
+                return undefined;
```
daiagi, line 57: `playTransaction` returns `undefined` here, but `LocalP2pSigner.sendTransaction` L116-122 assigns it to `_blockConfirmation`, never reads it, and returns the placeholder response anyway.

before this the loser reached `isMyTurn` and threw `NOT MY TURN`, so the caller's promise rejected. now it resolves and the tx is just gone. in the new unit test `add(2)` never executes and the app has no way to find out.

the height is read outside the mutex at L100-102, so this is the ordinary concurrent-submit path, not an edge case.

lukastanisic99: This was added as race tiebreaker for consumer applications that race on authoring a transition (e.g. poker timer expires, the client automated logic will produce a "fold", but there's a race and the user clicked "bet"), like this one of the 2 succeeds and the other is just dropped - returning undefined means that's the tx that was dropped

daiagi (approved):

PR #453, opened 2026-08-25
ci: run the full test suite on the distributed pool

daiagi:


lukastanisic99:
Can't get the CI to run on github - will check later, since it seems github is having problems at the moment - https://www.githubstatus.com/incidents/y1t7p9fzrlj2

lukastanisic99:
image
I got stuck after 250 tests - will investigate in the morning - probably test 251 failed and it got stuck fetching the logs.. 

What I don't like: needed to install foundry and compile contracts on the CI to discover the .loop{.it("....")} tests - not sure are they discovered or not nor how we got here since I remember adding to AGENTS.md that we want explicit .it() and not loop -> .it(...)

Probably will also have to add a flag to prevent log upload in the CI - not sure

lukastanisic99:
@daiagi take a look - it passed, but some tests are still flaky from time to time

lukastanisic99:
I'm merging this now, since I need the updated orchestrator to run tests on other branches - any changes/suggestions can be a followup

PR #456, opened 2026-08-25
V2 spec for watchtowers

lukastanisic99:
Really satisfied with the detailed specification including the test invariants and permutations. 
I have 1 more thing to 'experiment with', but will do it in a followup.


daiagi:
1. what is `assignmentEpoch`? it's a signed field that every verifier must reject on, and the spec never says what it is.

it's bound into the AFK block's signed bytes and carried on `WatchtowerDisputeApproval` (`watchtowers.md:219`, `data-types.md:217`), and `watchtowers.md:236` says a verifier MUST reject an artifact whose assignment epoch doesn't match. but there's no scope, no increment rule and no initial value anywhere, and "stale-epoch" implies an ordering nothing establishes. can't write the verifier.

it probably exists to fix the second half of this: a participant that leaves, swaps tower and rejoins the *same* channel ends up with two towers for one channel. `watchtowers.md:133-135` says a replacement applies only to channels joined afterward, `watchtowers.md:135-137` says the tower stays for the channel's entire lifetime. both sentences are in REQ-WT-1. binding the tower key in the on-chain join record would make the join instance the epoch and settle both.

2. how many times can a watchtower evict its own participant? OQ-52 leaves it open and it needs closing before implementation.

a tower partitioned from its peer but still connected to everyone else produces a *valid* AFK block removing a peer that did nothing wrong (`watchtowers.md:498-503`). if the answer is "once per fork" that costs the peer one bad slot. if repeats are allowed the tower retries at every slot that peer authors, and the retries are free - class 1 is per-slot (`watchtowers.md:385-387`) and only bites if the tower also acknowledged the competing normal block at that same slot, which a partitioned tower never sees.

not a tuning parameter either. the OQ pointer sits inside a rejection rule (`watchtowers.md:159-161`, "rejected by every validation context"), so two implementations that guess differently disagree on whether that peer is still a member.

3. if your tower is having a bad day, you pay for it. you file an ordinary timeout, your own tower's availability receipt kills that timeout, and you eat the failed-dispute penalty. the spec already says it: a tower "can spend the participant's slot or draw the disputer penalty onto it - a service failure inside the selection trust boundary with no protocol compensation" (`watchtowers.md:535-537`).

no malice needed, just an outage. cheap fix: if a timeout is defeated *only* by the disputer's own tower's credit, with no direct acknowledgement from the disputer, it shouldn't carry the disputer penalty. the timeout still dies, the peer just stops paying for its own tower's downtime.

4. REQ-DIS-3 (`disputes.md:553`) defers the disputer-penalty destination to "open question §4.2". §4.2 doesn't exist - headings jump from 4.1 (`disputes.md:126`) to 4.3 (`disputes.md:159`).

5. REQ-TRUST-3 says safety needs at least one threshold-required peer whose *full authority path* is honest, and when a peer delegates, its tower is part of that path (`trust-model.md:117-123`). twelve lines down: "if no threshold-required peer has an honest full authority path, the trust assumption provides no safety for that channel" (`trust-model.md:131-133`).

with a shared default tower that's one operator on every authority path, so the whole assumption reduces to that one operator. fine as a trust assumption, but it should be written down instead of left as two sentences that never meet. related: the §9 threat table (`trust-model.md:284-292`) has no watchtower row at all.

smaller: nobody is paid to submit a contradiction proof, and proving one never restores the removed participant or reopens settlement (`watchtowers.md:424-428`) -> the victim funds a public good. "availability set" used 3x, defined nowhere. `trust-model.md:71` points Future Work at "§9", which is the threat model.

lukastanisic99:
@daiagi thanks for the review - I've pushed clarifications

daiagi (approved):
lgtm

docs/spec/specification/protocol-model/data-types.md
```diff
@@ -234,9 +260,9 @@ Each row is a planned black-box test obligation, not an additional specification
...
...ned AFK block bytes bind the exact represented participant, every wrong channel, fork, height, or target presentation fails, and a presentation against a membership interval whose accepted join established a different tower fails                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          |
```
momoshell: `P4 — correct signature domain` and `P9 — wrong signature domain` can't be written: `OQ-29-EFY4NF` is open and says there is no domain — plain EIP-191, no chainId, verifying contract, version or type tag.

It matters more now than it did for blocks alone, since one tower signature can satisfy a whole availability threshold: recreate the same channel, fork and tower binding on another deployment and a receipt or approval replays there.

Either give the new artifacts their own domain (chainId + manager address + version + type tag), or mark P4/P9 as blocked on OQ-29 so nobody tries to implement the oracle.

docs/spec/audit/specification.md
```diff
@@ -5,3 +5,116 @@
...
+tower-authored removal design are all decided
+([`OQ-47-QMYM54`](../specification/open-questions.md#oq-47-qmym54),
+[`OQ-43-HWRTNF`](../specification/open-questions.md#oq-43-hwrtnf),
+[`OQ-46-ZXR2V3`](../specification/open-questions.md#oq-46-zxr2v3), and
+[`OQ-45-23GGV6`](../specification/open-questions.md#oq-45-23ggv6), all resolved);
+[`OQ-48-CS3JNE`](../specification/open-questions.md#oq-48-cs3jne) stays separate future-work
+context, and the bounded open details are
+[`OQ-49-2Z3FAS`](../specification/open-questions.md#oq-49-2z3fas),
+[`OQ-50-YSDG8S`](../specification/open-questions.md#oq-50-ysdg8s), and
+[`OQ-51-BCKA50`](../specification/open-questions.md#oq-51-bcka50); none gates the version-one
```
momoshell: This says OQ-49/50/51 "none gates the version-one watchtower requirements." Two look gated *by* the new powers.

**OQ-50** — a tower approval finalises with no kill-period challenge (`REQ-DIS-14-032T4M`). If it finalises against checkpoint C0 and a snapshot advances to C1, applying the C0 output rolls back C1; rejecting it strands a finalised dispute. Before approvals there was no signed final output to strand.

**OQ-51** — the equal-height tie decides *which participant is removed*, which is now a tower power.

Either add the argument for why they stay non-gating, or move them to gating for this PR.

docs/spec/specification/disputes/disputes.md
```diff
@@ -143,6 +156,34 @@ feeds the next dispute.
...
+  slashes the submitter when the submitter is dispute-eligible, and it never touches the accused
+  dispute ([fraud-proofs.md](./fraud-proofs.md),
+  [`OQ-2-7WTV16`](../open-questions.md#oq-2-7wtv16), resolved).
+- The base protocol defines **no separate dispute bond, no challenger reward, and no universal
+  recipient for slashed value**. The kill's slash-set entry is consumed later, at reduction
+  (§4.3, §5), where the application state machine's `slashParticipant` hook owns the slashed
+  participant's balance disposition
+  ([state-machines.md §6.3](../protocol-model/state-machines.md#63-_slashparticipantaddress--punitive-removal)).
+
+The remaining undecided kill-window edges — the window-reopen rule after all commitments are
```
momoshell, line 195: "kill-versus-counter-dispute ordering" is undecided here under `OQ-1-NTJBA1`, while `REQ-WT-9-GKFQXZ` newly lets a tower submit its participant's counter-dispute.

With no admission or result rule, transaction ordering decides commitment membership and kill-period timing — two implementations can disagree about whether the dispute is live.

Either decide the ordering in this PR, or state that delegated counter-dispute submission is gated on OQ-1.

docs/spec/specification/protocol-model/identity.md
```diff
@@ -138,13 +195,13 @@ matters, e.g. counting threshold signatures).
...
...y's signatures act as the participant; any other key's do not, regardless of claimed identity fields. | <a id="inv-id-1-b4fxj4.t1.p1"></a>`INV-ID-1-B4FXJ4.T1.P1` — signed block accepted from the key holder; <a id="inv-id-1-b4fxj4.t1.p2"></a>`INV-ID-1-B4FXJ4.T1.P2` — claimed-identity fields cannot override recovery; <a id="inv-id-1-b4fxj4.t1.p3"></a>`INV-ID-1-B4FXJ4.T1.P3` — a "stolen key" signature is indistinguishable and accepted (documenting the compromise model); <a id="inv-id-1-b4fxj4.t1.p4"></a>`INV-ID-1-B4FXJ4.T1.P4` — signed transaction accepted from the key holder; <a id="inv-id-1-b4fxj4.t1.p5"></a>`INV-ID-1-B4FXJ4.T1.P5` — signed open accepted from the key holder; <a id="inv-id-1-b4fxj4.t1.p6"></a>`INV-ID-1-B4FXJ4.T1.P6` — signed join accepted from the key holder; <a id="inv-id-1-b4fxj4.t1.p7"></a>`INV-ID-1-B4FXJ4.T1.P7` — signed dispute accepted from the key holder; <a id="inv-id-1-b4fxj4.t1.p8"></a>`INV-ID-1-B4FXJ4.T1.P8` — central reuse policy: a signature produced by any process from a manually reused participant key acts as that participant with full authority and liability, regardless of claimed peer or tower role; reuse or address overlap alone triggers no slash. |
```
momoshell: The overlapping-authority rules above are specific: one signature supplies that identity's own finality vote **and** per-peer tower availability credit, with the two punishment predicates evaluated independently. P1–P8 don't reach any of it — they test registered-key-versus-other-key, and P8 only says overlap alone triggers no slash.

An implementation written `if participant … else if tower` passes every permutation, because the fixtures use distinct keys.

Three permutations would close it: combined credit from one confirmation, cross-count deduplication, and each punishment predicate firing independently on one key.

docs/spec/specification/runtime/watchtowers.md
```diff
@@ -0,0 +1,617 @@
...
+## Registration, selection, and lifecycle
+
+This section is explanatory; the binding rules are
+[`REQ-WT-1-TXW328`](watchtowers.md#req-wt-1-txw328),
+[`REQ-WT-5-T5ZFTZ`](watchtowers.md#req-wt-5-t5zftz),
+[`REQ-WT-2-HNZA3Y`](watchtowers.md#req-wt-2-hnza3y),
+[`REQ-WT-10-GNG79P`](watchtowers.md#req-wt-10-gng79p), and
+[`REQ-WT-8-W3YP4R`](watchtowers.md#req-wt-8-w3yp4r) below.
+
+A tower identity exists by paying a fixed, minimal, one-time registration bond that is permanently
```
momoshell, line 86: Three things worth pinning down for the poker case.

**Can a spectator see hole cards?** The tower spectates and keeps a rolling evidence set per participant, so a shared default tower holds several players' views at one table. §7 only says "the same view as a spectator." If cards are mental-poker encrypted, saying so here settles it — that one fact is the difference between a shared tower being fine and being a collusion engine.

**"Non-punitive" needs a game-layer caveat.** It's the stated mitigation for a tower folding a compliant player, but `REQ-SM-8-8CHSQ8` leaves disposition to the application. Mid-hand, the pot is gone either way. Worth saying the claim covers held balance, not game outcome.

**Say why delegation rather than checkpoint complaints.** [ROYALE](https://eprint.iacr.org/2018/157) solves the same offline-player problem on-chain with complaints plus penalties; [Pisa](https://eprint.iacr.org/2018/582.pdf) makes the delegate liable for not acting. A paragraph on why neither was chosen would close the design section.

docs/spec/specification/protocol-model/data-types.md
lukastanisic99: For this change, we will not add deployment-domain checks. We will rely on the existing channel ID, fork ID and height bindings, under the chosen design assumption that channel and fork identifiers are unique across the relevant deployments. We treat accidental hash collisions as negligible, but hashing alone does not establish that uniqueness: identical or reused inputs can reproduce an identifier without a collision. P4/P9 should be clarified against the existing signing forms and bindings in REQ-ID-4, rather than require new chain, manager or version fields. This records the design assumption; it is not an independently verified cross-deployment replay guarantee and does not resolve OQ-29.

docs/spec/audit/specification.md
lukastanisic99: Luka’s decision combines three changes. First, during a committed dispute’s open kill period, anyone may kill it if its named disputer is already in that same channel’s on-chain slash set, regardless of which adjudicated fraud caused the slash. This is a new kill ground, not merely exclusion of already-slashed participants from new uploads. Double-signing is one way to establish the slash, but this kill requires only the existing on-chain slash membership check, not a new double-sign proof payload. Remove the target commitment without duplicating the existing slash or penalty. Preserve historical signature validity and the ordinary commitment and live-window checks. This neither extends kill periods nor reopens finalized disputes; malicious selected-tower approval remains an accepted delegated-trust risk.

Second, a tower’s dispute-finality approval must sign both the exact dispute and the exact on-chain snapshot reference it relied on. Count that tower credit only when the reference equals the current snapshot at execution. Direct participant signatures retain their ordinary rules without this extra tower-specific condition. If Bob settles first, a partitioned tower’s old-snapshot approval contributes no immediate-finality credit. If the remaining credits do not reach threshold, the dispute remains an ordinary challengeable commitment provided ordinary admission checks pass. Its signed bytes and output hash are not rewritten, and stale approval does not make invalid dispute inputs valid. Anyone, including Bob or his tower, can use the new kill ground during the open kill period once that named disputer is recorded in the same channel’s on-chain slash set.

Third, freeze snapshot updates whose destination is a disputed fork from its first on-chain commitment, including after finalization. Adoption may still leave that fork for a resolved successor whose destination is not itself disputed, following verified reduction links past disputed intermediate forks. If the dispute lands first, Bob’s later same-fork settlement fails and settlement proceeds through the resolved successor. AFK removal remains subject to the existing reduction rules: channel slashes suppress additional timeout removal; an AFK removal already present in the selected valid history is not undone by that suppression. Every accepted result and adoption must preserve the exact settled checkpoint and message history, including the canonical checkpoint bridge. Delayed settlement and withdrawals and the pre-settlement partition race are accepted; no short-duration or single-lost-transaction bound is claimed. OQ-51’s equal-height different-target question remains separate and open.

docs/spec/specification/protocol-model/identity.md
lukastanisic99: Agreed. The overlap rule is normative and no permutation stages one key with both authorities. I will add the three cases under the invariant that owns the rule.

docs/spec/specification/runtime/watchtowers.md
lukastanisic99: I’ll clarify the poker application’s privacy model: before showdown, only the player holds their hole-card decryption key. Neither their selected tower, a shared tower, nor an independent spectator has that key or sees plaintext hole cards. Public state-machine/on-chain data contains encrypted cards; player-side off-chain/ZK logic handles private information. At showdown, the application requires disclosure of decryption material in state so observers can see the cards and the state machine can determine the winner. This is my clarification of the poker application design, not a universal SDK secrecy guarantee.

By “non-punitive,” I mean removal without protocol slashing. The application can still fold the removed player and apply ordinary game losses. The SDK is general; it does not promise that removal preserves a poker outcome or prevents those losses.

I deliberately chose optional participant-selected delegation with broader authority to maximize realtime off-chain progress and settlement and minimize fallback. That requires greater trust in the selected delegate. Independent observation and permissionless proof submission do not confer that delegated authority. Each participant chooses their delegate; one tower can serve many participants. No mandatory participant service payment is intended, but providers may charge and this is not an economic sustainability guarantee. Anonymous sampled delegation remains future work, not an implemented V1 feature.

I’ll keep the comparison brief. [ROYALE](https://eprint.iacr.org/2018/157.pdf) requests missing contributions on chain and can continue after a valid response; if a required share remains missing, its penalty/compensation path halts rather than guarantees decryption. [Pisa](https://eprint.iacr.org/2018/582.pdf) uses a paid custodian, signed receipts and fresher-state dispute responses; basic failure recourse burns its deposit, while compensation is a separate extension. poker ZK design is intended to guarantee continuation under its assumptions even when penalties occur; Our broader optional authority is a deliberate liveness/trust tradeoff, not blanket superiority or an adoption of all Pisa guarantees.

docs/spec/specification/disputes/disputes.md
lukastanisic99: A delegated counter-dispute is the participant's own submission (REQ-WT-9, REQ-DIS-15: same disputer, slot, stake and outcome, tower as transport only), so it inherits exactly what OQ-1 decides for participant counter-disputes, and both places state that OQ-1 applies unchanged. Gating only the delegated form would describe a dependency that does not exist; OQ-1 stays the single owner of the ordering question for both.

PR #457, opened 2026-08-26
refactor(transport): separate identity verification from channel admission

momoshell:
## Summary

Separates two things the handshake used to do at once: proving who a peer is, and admitting
that peer to the channel. Completing a handshake now only registers the peer's identity in
`ProfileManager`; joining the broadcast set is a separate decision made by `P2PManager`, based
on whether the peer is actually in the channel's on-chain participant union. It also gives
`ProfileManager` sole ownership of Holepunch ban policy, so a normal transport close stops
permanently banning the peer it just closed.

This is the transport and handshake half of #422, split out on Luka's review there. It carries
no discovery code — no lobby, no ads, no acquisition. Those follow separately, and the reason
they can follow separately is precisely this change: once a verified peer is not automatically a
channel peer, a service can meet strangers on the shared swarm without leaking channel traffic
to them.

### Why the handshake stopped deciding

`InitHandshakeService` used to call `addConnection` itself the moment a handshake finished. That
conflated two sets that mean different things. `ProfileManager` answers "who is this, and how do
I reach them"; `openConnections` answers "who is in this channel's conversation" and drives
broadcasts, `getConnectedPeers` and connection-wide cleanup. Registering an identity should not
grant channel traffic.

So the service now emits `handshakeCompleted` and stops there. `P2PManager` subscribes and
decides. A peer that resolves as a channel participant is promoted; a peer that does not is
deferred rather than promoted speculatively, because a promotion decided once at handshake time
is exactly how a stranger met before we join a channel would keep receiving our broadcasts
forever afterwards. An accepted spectator is promoted separately, at the point we actually
accept that relationship, which can be long after the handshake.

The hook carries only an address. It crosses the runtime port, which structured-clones
everything, so it could never carry a live transport — the transport is resolved host-side
through `ProfileManager`.

### Membership is one question, asked the same way twice

The immediate and deferred promotion paths had drifted onto different definitions of membership.
The immediate path asked the on-chain participant union. The deferred path asked whether the
peer could participate in disputes — which is a narrower, later-forming property derived from
the latest inbound message block hash, and is false at genesis even for a founding participant.
The immediate path already carried a comment explaining why that narrower question stalled a
channel at height zero; the deferred path was still asking it.

Both now go through one `isChannelConversationMember`, so they cannot drift again.

### A deferred peer needs someone to tell it the news

Deferring correctly creates a second problem: something has to make us look again. Our own
status changing covers "we advanced". It does not cover the case that matters most — membership
also grows when *someone else* joins, which moves the joiner's status and not ours. A peer that
defers a stranger and then sits idle, whether a spectator or a participant watching a newcomer
arrive, has no trigger left and keeps that peer deferred permanently.

Only the joiner knows when its join actually lands, so the joiner is what closes it:

```ts
await tx.wait();

// The join is now on chain, so peers that deferred a promotion
// decision about us can resolve it.
sm.p2pManager.announceChannelMembership();   // ← fire-and-forget, to handshaked peers
```

The announcement is a hint, never authority. The receiver re-reads the on-chain participant
union exactly as it would have anyway, so a peer that lies here costs one chain read and gains
nothing.

It is sent after the transaction is mined rather than on a status change, and that detail is
load-bearing. `PENDING_PARTICIPANT` is set when the join is *submitted*, so announcing then
races the chain: the receiver looks, does not yet find the joiner in the union, and has no
reason to look again. A joiner also need never reach `PARTICIPATING`, so there is no later
status to fall back on.

### Ban policy belongs to the identity, not the transport

`HolepunchTransport._close` used to call `peerInfo.ban(true)` unconditionally, so every close
permanently banned that peer at the DHT level — including an expected upgrade from Holepunch to
WebRTC. A close now only tears down the socket.

`ProfileManager` owns the policy instead, because there is one profile per identity and it
survives the transport being replaced. Three cases: an explicit blacklist bans and never
unbans; a successful WebRTC upgrade bans Holepunch temporarily so the swarm does not reconnect
the relay underneath a healthy WebRTC transport; and a closing WebRTC transport lifts that ban
so Holepunch can serve as fallback again, unless the peer is blacklisted, in which case the
blacklist wins.

The profile keeps the Holepunch peer-info handle so it can still apply policy after the active
transport has moved on to WebRTC.

### One relayer failover implementation

`HolepunchRelay` carried its own relayer selection, exclusion and backoff inline. That logic
moves to `RelayerPool` and `HolepunchRelay` uses it, so a second caller does not grow a second
copy that drifts. The extracted version is hardened in one place: a single pending-retry flag
deduplicates a socket that fires both `error` and `close` for one failure, which previously
double-counted toward the exhaustion backoff.

Also adds a per-topic `Holepunch.leave`, so a caller that joined one topic can leave just that
one instead of tearing down the swarm.

## Tests

`test/rpc/initHandshake/InitHandshakeLifecycle.test.ts` covers both halves of the split: a
handshake that registers a profile without promoting when nothing subscribes; the normal channel
path still promoting and syncing; a non-participant, non-spectator staying out of
`openConnections`; an accepted spectator being promoted even though it is not a dispute
participant; and a stranger not being promoted retroactively when we ourselves become a
participant.

The deferred-promotion fix has a test that fails without it — two spectators that never accepted
each other, one joins for real, and the other must promote it while its own status never moves.
That test failed against every event-driven trigger tried before the announcement
(`onStatusChanged` alone, adding `onSetState`, adding the chain-driven
`onStateSnapshotUpdated`), which is what established that no existing signal reaches an idle
peer.

`HolepunchBanPolicy`, `HolepunchLeaveTopic` and `RelayerPool` cover the transport changes.

Running the affected suites on this branch: **72 passing, 1 failing**. The failure is
`JoinChannelSignatureRequest`, which is a pre-existing load flake on `dispute` — running the
same selection at the same concurrency against the merge base and against this branch gives
155 passing / 6 failing on both sides, with overlapping failure sets.

Unrelated but worth knowing: `yarn tsc -p tsconfig.json` is currently red on `dispute` itself.
`test/scripts/e2eParallelProtocol.test.ts` gained `import * as tar from "tar"` without
`@types/tar` being installed.

Related: #422


lukastanisic99:
just rebased on top of the `dispute` branch

PR #461, opened 2026-08-27
cross-thread crash log collection

daiagi:
a throw in any thread now collects and uploads the logs of every thread, and
`logger.uploadLogs()` returns what actually happened so a report-a-bug button can
tell the truth.

each thread keeps its own logs, so a collection has to ask all of them. it goes out
over the connections the threads already have, skips the one it came from, and each
thread answers once - so it always finishes. a thread sends only what is new since
its last upload, compressed, and the pieces are stitched back together on read.

`yarn test:browser:worker` (Playwright smoke, `test/browser/crash-log-smoke.js`) is a separate manual gate, not part of `test:parallel:distributed`.


lukastanisic99:
@daiagi 1 conflict after merging your CI branch - take a look while I review other stuff

daiagi:
@lukastanisic99  Conflict resolved

lukastanisic99 (commented):
**Human note**

Valid test case;

**AI-generated review**

🟠 **[TO1] The crash tests do not prove the full topology or fatal lifecycle.** The SDK crash checks main+SDK although a VM realm is configured; the VM crash checks VM+SDK but not main. The “pending calls do not hang” case has no request pending when the crash occurs—it polls new calls only after the trigger returns. Worker upload completion is inferred from a request-count threshold, which is ordering-sensitive. Caller-owned logger preservation is also untested.

> **Fix TO1-FIX**
>
> Add a real main → SDK → VM fatal workflow with unique markers from all realms, an actually in-flight request at the crash boundary, deterministic upload acknowledgements, worker-exit assertions, and a caller-owned logger that remains usable afterward.



---

**Human note**

yes;

**AI-generated review**

🟡 **[TY2] Two promised boundaries have no executable evidence.** `REQ-LOG-8` requires every supported host, including browser realms, but all new runtime cases use Node. The crash-log server's missing/wrong bearer-token `401` paths and recovery with a valid token are also untested.

> **Fix TY2-FIX**
>
> Add browser inline+worker context/flush/cleanup coverage, plus missing-token, wrong-token, correct-token, and subsequent-valid-upload route cases. Until browser execution exists, mark that requirement partial rather than covered.



---

**Human note**

[THEORETICAL] - (the following is GPT writing my words): "When merge can isolate a malformed serialized log entry, keep its sequence position with a synthetic crash-log-server entry that clearly says decoding failed. Include the original serialized value only when it can be safely bounded and escaped. If a whole chunk cannot be decompressed, emit a visible chunk-level omission marker for the affected sequence range. Nothing should disappear silently."

**AI-generated review**

🟠 **[TY3] Partial server reads are silent at both boundaries.** POST validation accepts any decodable array with the expected length, but merge later drops individual malformed entries without adding them to the skipped count. When a whole chunk is skipped, the server sets `x-skipped-chunks`, yet `fetchLogEntries` ignores that header and writes an artifact that looks complete.

> **Fix TY3-FIX**
>
> Reject malformed entries during POST or count them as explicit omissions. Make the fetch tool fail or write a clear truncation marker whenever the server reports skipped data. Cover malformed-entry and skipped-chunk downloads through the route and fetch surfaces.



---

**Human note**

Nit

**AI-generated review**

🟡 **[AY1] Three new tests bypass repository test rules.** `WorkerContractExecutor.test.ts` and `crashLogServer.test.ts` keep reusable staging/helpers inside declarations instead of typed fixture modules. `logEncoder.test.ts:169-178` uses the forbidden `as unknown as LogEntry` structural cast rather than reaching the real public logger boundary. Static literal declarations and timeout rules are otherwise followed.

> **Fix AY1-FIX**
>
> Move reusable worker/server staging into fixture files and replace the encoder cast with a black-box logger case that supplies the hostile value through its public input.



---

**Human note**

yes;

**AI-generated review**

🟠 **[DR1] The new requirements are not connected to their actual evidence or implementation owners.** The PR adds normative `INV-LOG-*`/`REQ-LOG-*` families, but their requirement permutations are unassigned. Direct evidence files—including `E2E-CrashLogUpload`, `LogUploaderDelta`, `LoggerThreadContext`, and `logStore`—use `@spec-test-coverage-ignore` and have no maintained verification reports. Only three source reports were updated among the changed logging/runtime owners; `Logger`, `LogUploader`, `logStore`, `logEncoder`, `crash-log-server`, and `logChunks` have no accurate conformance/obligation chain. The audits and generated inverse reports were not refreshed.

> **Fix DR1-FIX**
>
> Create/update the source subjects and inventories, assign every requirement permutation to exact declarations, remove coverage-ignore from direct evidence, add verification reports, update the maintained audits, then run and inspect `yarn spec:refresh` and both impact audits. Do not edit the engineer approval register.



---

**Human note**

Yes;

**AI-generated review**

🟡 **[DY2] Existing verification reports no longer describe their suites.** `E2E-RuntimeTransportModes` omits the new cases and its anchors shifted; `WorkerContractExecutor` still describes a deleted dynamic declaration and misses fatal/log cases; `logEncoder` misses new declarations/anchors; and `LogUploader` still claims an empty-batch behavior where the test now uploads a routine entry.

> **Fix DY2-FIX**
>
> Reconcile every changed declaration with its exact verification-report name, anchor, setup, oracle, and Covers row before refreshing the inverse reports.



---

**Human note**

My design comment written by GPT; Take a look;

**AI-generated review**

🟠 **[GO1] Prefer merge-on-write storage.** The server already decompresses every incoming delta to validate it. Under the existing per-channel serialization and locking, it should also read the persisted compressed stream for the same logical thread/store, decompress it, combine the existing and new entries, deduplicate and order them with the same identity and ordering rules, recompress the result, and atomically replace the persisted stream. This merge-on-write approach, or incremental compaction, keeps one compressed stream instead of an unbounded number of small chunks, improves compression from the larger input, and avoids merging the same historical chunks on every download. It needs atomic, crash-safe writes, bounded memory use, no lost acknowledged entries under concurrent uploads, and benchmarks showing that the storage and read savings justify the write amplification.

> **Fix GO1-FIX**
>
> Implement incremental compaction under the current lock with crash-safe atomic replacement and bounded memory. Add concurrent-upload and restart tests that prove no acknowledged entry is lost and ordering stays stable, then benchmark whether the storage and read savings justify the extra write work.



---

**Human note**

My design comment written by GPT; Take a look;

**AI-generated review**

🟠 **[GO2] Expose raw per-thread compressed streams as an additional API.** Keep the current backward-compatible merged endpoint as the default. Add another endpoint that returns compressed per-thread streams or chunks and their metadata without server-side cross-thread merging. Clients can download and decompress them, then optionally apply the shared deduplication and ordering algorithm. This supports advanced tooling and avoids mandatory server recompression while preserving the simple existing API. If merge-on-write keeps one consolidated stream per thread/store, this raw API should expose those consolidated streams, not the original upload chunks. Retaining the original chunks would conflict with the single-stream storage goal.

> **Fix GO2-FIX**
>
> Add an opt-in raw endpoint that returns each per-thread compressed stream and its metadata without changing the existing merged endpoint. Test that the default response stays compatible, raw metadata round-trips, and clients can apply the shared deduplication and ordering rules. When storage is consolidated, return the consolidated streams only.



---

**AI-generated review**

🟠 **[GO3] Make worker protocols composable through the shared typed RPC/service abstraction.** The log-control payload is shared today, but `RuntimeLogControlMessage` and `WorkerLogControlMessage` are manually added to their protocol unions, and each client/host pair dispatches its own special `logControl` case. Every new cross-worker service will repeat this wiring. Refactor the SDK's general worker communication model to follow the existing custom-RPC and harness-control pattern: treat each worker `MessagePort` as one trusted loopback transport, or telephone line, give the worker protocol one explicit top-level RPC/service envelope, and compose a typed service manifest at each worker entry point. The manifest should declare which services the worker hosts. Including a logging service should automatically provide the typed logger-control client and host plus its cascade behavior, without editing unions and switches in every worker. Reuse the `ARpcService`, `ARpcMethods`, `RpcHandler`, and `LoopbackTransport` ideas, or extract a transport-neutral equivalent, rather than creating another dispatcher. The result should share correlation, error, and dispatch machinery, keep service composition type-safe, and let the same service run over local worker ports or an explicitly supported remote transport. Keep that trust boundary explicit: a local worker port is trusted loopback, while remote exposure still requires authentication, authorization, and the normal remote-transport guards.

The shared abstraction should also own a typed local topology registry for each worker realm. It should expose an optional parent endpoint and a collection of child endpoints, each backed by its port. When more than one child can exist, each child needs a stable role or instance identity. Derive every endpoint as a typed RPC/service client from the remote worker's service manifest. Callers can then address the parent, one child, selected children, or all children without constructing raw union messages. Register and remove endpoints with the worker lifecycle so a dead port cannot remain addressable. Local parent/child adjacency is enough; a worker does not need to know the full global graph.

Build tree-wide operations above this neighbor registry. Log flushing should forward to every neighbor except the incoming edge, aggregate typed results, and keep per-origin or per-epoch state when concurrent traversals need it. The transport registry should stay service-neutral and must not hard-code logging behavior.

> **Fix GO3-FIX**
>
> Define the shared service envelope, router, and typed service registry. Add typed parent and child endpoint registration from remote manifests, stable child identities, lifecycle cleanup, and explicit parent, one-child, selected-children, and all-children addressing. Adapt the SDK-runtime and contract-executor protocols, register log control through manifests at the main, SDK, and VM entry points, and remove the duplicated `RuntimeLogControlMessage` / `WorkerLogControlMessage` unions and `logControl` switches. Implement log traversal above the registry. Add compile-time capability checks, local-loopback and guarded-remote tests, and traversal tests for one parent, multiple children, incoming-edge exclusion, child removal, result aggregation, and service capability mismatches.

src/utils/logging/LogFlushBus.ts
```diff
@@ -0,0 +1,371 @@
...
+        if (!fromPort) {
+            // local trigger -> reach every port
+            this.queuedExcluded = new Set();
+            return;
+        }
+        if (!this.queuedRound || !this.queuedExcluded) {
+            this.queuedExcluded = new Set([fromPort]);
+            return;
+        }
+        this.queuedExcluded.add(fromPort);
```
lukastanisic99, line 295: **Human note**

yes

**AI-generated review**

🔴 **[FR1] Folding simultaneous origins removes every requesting subtree from the shared round.** In a child-A → hub → child-B topology, A and B can both start collection while the hub is busy. The hub unions both source ports into `queuedExcluded`, so its next round forwards to neither child. Each caller receives only its own realm plus the hub, never the sibling. That violates `INV-LOG-1` and `REQ-LOG-2`, which require one collection request to reach every linked realm. The new test at `LogFlushBus.test.ts:260` checks forwarding counts, not the returned/stored realm markers, so it codifies the omission.

> **Fix FR1-FIX**
>
> Preserve independent origin rounds or merge completed subtree results without excluding a subtree another requester needs. Add a three-realm star test where both leaves originate during an active hub round; each result and stored upload must contain all three unique markers.

src/evm/contractExecutor/node/ContractExecutorWorkerEntry.ts
```diff
@@ -11,6 +12,34 @@ if (!parentPort) {
...
+    // empty round and then kill the thread mid-upload
+    setImmediate(() => {
+        // ask every realm to upload, but wait only for this one's POST. the
+        // realms across the port are still running and finish on their own,
+        // so their acks would only delay the exit
+        void realmLogFlushBus.flushAll(reason).catch(() => undefined);
+        void realmLogFlushBus
+            .flushOwnRealm()
+            .catch(() => undefined)
+            .finally(() => process.exit(1));
```
lukastanisic99: **Human note**

Make sure cleanup is safe - uv_loop_close was annoying

**AI-generated review**

🔴 **[FR2] The fatal path force-exits a worker with live handles.** `src/evm/node/workerShutdown.ts` records that worker-side `process.exit()` can abort the whole process with `uv_loop_close() while having open handles` under parallel load and requires natural drain. This worker owns a parent port, uploader, monitor, and EVM state when the callback runs. The new fatal test proves later calls reject, but it does not prove the parent process survives with those handles active.

> **Fix FR2-FIX**
>
> Report the fatal error, flush, dispose every worker-owned resource, set `process.exitCode = 1`, close the worker port, and let the loop drain. Prove it in a child-process integration test with live logging and EVM handles; the worker must exit non-zero without aborting the parent.

src/evm/contractExecutor/WorkerContractExecutor.ts
```diff
@@ -72,6 +76,7 @@ export default class WorkerContractExecutor extends AContractExecutor {
             ),
             config
         });
+        executor.attachLogPort();
```
lukastanisic99, line 109: **Human note**

yes, but it can be addressed with my `GO3` proposal bellow

**AI-generated review**

🟠 **[FO3] Initialization failure leaks an unreachable worker and loses linked-realm crash logs.** `create()` attaches the parent log port only after `init` succeeds and has no failure cleanup. The worker host creates its logger and log port before `createEvm`; if that initialization rejects, the request returns an error but the static factory rejects without returning an executor that the caller can dispose. The worker, crash hooks, logger, and parent port remain live, while any fatal `logControl` messages sent during initialization are dropped because `logPortHandle` is still absent.

> **Fix FO3-FIX**
>
> Attach the log port before initialization and wrap the factory sequence in cleanup that closes the worker and both port handles on every failure. Inject an EVM initialization failure and assert no worker survives and all linked-realm logs are collected.

src/utils/logging/logEncoder.ts
```diff
@@ -98,8 +115,28 @@ function sanitizeForEncoding(value: unknown, seen: WeakSet<object>): unknown {
...
+    if (typeof value === "string") return value;
+    if (value instanceof Error) return value.message;
+    if (value && typeof value === "object" && "message" in value) {
+        const inner = (value as { message?: unknown }).message;
+        if (typeof inner === "string") return inner;
+    }
+    try {
+        return JSON.stringify(value) ?? String(value);
+    } catch {
+        return String(value);
```
lukastanisic99: **Human note**

Ok - true but, but if it's not simple to do then it can sit in trello untill it's an actual problem

**AI-generated review**

🟠 **[SO2] A hostile value can make the logger throw while handling the original failure.** `Error.message`, the `in` check/property read, and the fallback `String(value)` can all invoke user code. A throwing getter, Proxy trap, or `Symbol.toPrimitive` escapes `toMessageString`, so `logger.warn(value)` can disturb the session in direct conflict with `REQ-LOG-3`.

> **Fix SO2-FIX**
>
> Guard every conversion and use a final literal fallback that cannot invoke the value. Exercise hostile getters, proxies, `toJSON`, and `Symbol.toPrimitive` through the public logger, not by casting a malformed `LogEntry` directly into the encoder.

src/utils/logging/LogUploader.ts
```diff
@@ -94,37 +112,74 @@ export abstract class LogUploader {
...
+            // checked before the jitter sleep -> a fan-out onto an idle realm
+            // costs neither HTTP nor wall time
+            if (
+                this.logStore.getLogsSince(this.lastUploadedSeq).entries
+                    .length === 0
+            ) {
+                return { ok: true, entries: 0 };
             }
 
+            await sleep(this.getJitterMs());
```
lukastanisic99, line 187: **Human note**

Nice race condition caught by the agent, but mostly theoretical, but yes it's true and I don't think it's a hard fix to include;

**AI-generated review**

🟠 **[RR1] Context and entries come from opposite sides of the jitter sleep.** The uploader captures channel, peer, thread, and upload key before sleeping, then reads the delta afterward. `Logger` shares and mutates the context object. If a session switches channel or peer and logs during the sleep, those new entries are posted under the old route, and the old-key watermark advances past them. The misplaced copy remains in the wrong crash-log bucket.

> **Fix RR1-FIX**
>
> Snapshot routing context and the matching delta atomically after jitter, with no await between them; use context epochs or split batches when entries span a context transition. Hold the jitter, switch channel and peer, log, release it, and prove no old-key request contains the new-context entry.

src/evm/p2pRuntime/P2pRuntimeHost.ts
```diff
@@ -236,35 +243,59 @@ export async function startP2pRuntimeHost<
...
-            { peerId: payload.peerId, peerAddress: signerAddress },
+            {
+                peerId: payload.peerId,
+                peerAddress: signerAddress,
+                threadName: "sdk"
+            },
             { component: "P2pRuntimeHost" },
-            { attachErrorListener: false }
+            // inline -> the main realm's logger already has the crash hooks
+            { attachErrorListener: Boolean(threadLabel) }
```
lukastanisic99, line 204: **Human note**

Yes!

**AI-generated review**

🟠 **[RY2] An SDK-worker uncaught error is reported but the corrupted worker keeps serving requests.** Installing `uncaughtException` and `unhandledRejection` listeners suppresses Node's fatal default. This callback only posts `hostError`; it does not stop the runtime, reject the client, or close the worker. The fixture receives the error and then disposes the worker normally, so it never proves fatality.

> **Fix RY2-FIX**
>
> Give the SDK worker one fatal owner: report and flush once, reject the client, dispose the runtime, close its ports, and drain. Assert that calls made after the injected fatal reject and no worker remains alive.

src/evm/P2pInstance.ts
```diff
@@ -91,6 +97,10 @@ export default class P2pInstance<
         } finally {
             this.webRTCBridgeHandle?.dispose();
             this.webRTCBridgeHandle = undefined;
+            // leaves the realm bus with the session: otherwise every closed
+            // session keeps its store and its process crash hooks, and every
+            // later round re-uploads it
+            if (this.ownsLogger) this.logger.dispose();
```
lukastanisic99: **Human note**

yes;

**AI-generated review**

🟡 **[RY4] One listener-removal rejection can dispose the logger while client teardown is still running.** `Promise.all` rejects immediately, then `finally` removes the logger from the realm bus even though `client.dispose()` continues in the background and may still need to report or flush teardown failures.

> **Fix RY4-FIX**
>
> Await all three cleanup operations independently or with `allSettled`, dispose the owned logger only after every operation settles, then rethrow the collected failure. Cover a rejecting listener removal with a delayed client disposal.

src/utils/logging/logStore.ts
```diff
@@ -17,7 +35,11 @@ export class LogStore {
         const serializedLog = encodeLogEntry(logEntry);
 
         const entrySize = serializedLog.length * 2; // 2* since the serialized string is UTF-16
```
lukastanisic99, line 38: **Human note**

True;

**AI-generated review**

🟡 **[PO1] The client and server enforce the same numeric limit in different byte units.** `LogStore` budgets UTF-16 code units, while the compressed wire payload inflates to UTF-8 and the server caps those bytes. BMP non-ASCII text can use two bytes in the store estimate but three on wire, plus aggregate JSON overhead. A valid store below 10 MiB can therefore produce a chunk the server rejects.

> **Fix PO1-FIX**
>
> Use the same UTF-8 measurement at storage, upload, and server boundaries, or size the receiver limit to cover the maximum encoded store. Add a near-cap multibyte route test.

src/utils/logging/logStore.ts
```diff
@@ -1,10 +1,28 @@
...
+    fromSeq: number;
+    toSeq: number;
+};
+
 // Shared log storage helper (instance-based)
 export class LogStore {
-    private logs: Array<{ entry: LogEntry; sizeInBytes: number }> = [];
+    /** distinguishes one process's stream from the next: seq restarts at 0 per
+     *  store, so without it a second run overwrites the first's chunks */
+    public readonly storeId: string = Array.from({ length: 2 }, () =>
```
lukastanisic99: **Human note**

Theoretical;

**AI-generated review**

🟡 **[PO2] The sole run discriminator has only about 32 random bits.** A collision makes independent processes share the same chunk namespace, so sequence deduplication can merge or overwrite unrelated crash logs. This conflicts with `REQ-LOG-6`'s no-mixing guarantee.

> **Fix PO2-FIX**
>
> Generate a 128-bit Web Crypto identifier or UUID and validate its exact format at the server boundary.

daiagi:
**[TY3] Partial server reads are silent at both boundaries**

**Human:** fetch half fixed. entry half I'd leave - nothing can trigger it except a hand-crafted request against yourself.

**Agent:** fetch half (a362da94): `fetchLogEntries` reads `x-skipped-chunks`, `persistLogEntries` writes an `Incomplete log` line at the top of the file, summary carries the total, `fetch-log-file` reads the same header. covered by `fetch-logs > marks a persisted log the server read short` plus a control case asserting the marker's absence.

entry half: POST validation checks chunk decodability and entry count, not per entry - the `isDecodableEntry` comment claiming otherwise is corrected. but nothing in the pipeline can produce an entry that fails the check. `Logger.log` L204-213 fills every field it looks at unconditionally (`time` through `String()` behind a catch, `meta` from a rest arg, `stack` from `new Error().stack`), `message` coerced by `toMessageString`. gzip's CRC + `JSON.parse` mean transit damage kills the whole chunk, and whole chunks are already counted (`logChunks.js` L166). so the drop at L173 needs a hand-crafted POST with a valid token, costing the poster one entry in their own directory - which garbage in the message does anyway. `decodeLogs` takes the same line deliberately (`logEncoder.ts` L14-21: one unreadable entry drops itself, not the file), so merge is consistent with it. chunk-level omission marker already exists as `x-skipped-chunks`, now honoured end to end; happy to fold an entry count into it if you want it visible.

src/utils/logging/LogFlushBus.ts
daiagi: **Luka:** confirmed, fixed. no logs were actually lost, only the count each asker got back was wrong. not patched here - the stacked PR removes the shared-round mechanism entirely, so this class of bug can't recur. both PRs go up together.

**Agent:** `foldExclusion` only ever excluded a port a request *arrived* on, and a request only arrives from a round that already covers that subtree, so the union of rounds always covered the tree - nothing was dropped. what broke was the count: B saw `ok: 1` for a 3-realm tree. that's `REQ-LOG-2` and the `Log flush round reached` record from `REQ-LOG-9`, not `INV-LOG-1`. the fold exists only because rounds are shared between askers; the stacked PR (GO3) makes every request its own round over the RPC core, so there is nothing to fold and nothing per-source to track. old test at L260 asserted the bug as intent; stacked PR replaces it with `two children asking during a hub round are each told the whole tree` (every asker gets `ok: 3`).

src/evm/contractExecutor/node/ContractExecutorWorkerEntry.ts
daiagi: **Luka:** not reproducible on this node version - probed it directly. kept the exit, fixed the comment that was wrong about the cause.

**Agent:** `process.exit()` inside a worker goes through `Worker::Exit -> CleanupHandles`, which closes the thread's handles and ends the thread only. probed on node 22.22: a worker holding a timer, a listening server, an open socket and an in-flight POST, x12 concurrent, real main -> sdk -> vm nesting x6 - parent exit 0 every time. the 17 `uv_loop_close` aborts in retained runs all share one handle dump (`udp` + `poll (active)`, hyperswarm/udx native handles), and run-372's trigger is the *main* thread throwing on the event-loop watchdog and exiting with the sdk worker alive - the comment's third clause, not the second. vm worker owns no native handles. corrected the comment in 03a64891; `ends the worker when init fails instead of leaking it` proves the real dangerous case (a leaked worker at process exit) via a child process that has to be able to end.

src/evm/contractExecutor/WorkerContractExecutor.ts
daiagi: **Luka:** confirmed, fixed in the stacked PR by removing the ordering dependency rather than patching it here.

**Agent:** port was attached after `init` resolved while the worker attached its end at the top of `init`, so for the whole `createEvm` window a vm crash posted `logControl` into `this.logPortHandle?.receive` with the handle undefined - dropped, and the vm's own upload carried no identity. the ordering existed only because identity traveled as a separate message. stacked PR makes the vm's log identity a parameter of `init`, so link and init can post in either order, and a rejected init disposes the executor before the error propagates. two tests there: `a crash while the evm is still being built reaches the realms above`, `ends the worker when init fails instead of leaking it`.

src/utils/logging/logEncoder.ts
daiagi: **Luka:** fixed, was cheap enough to just do now - 933b95eb.

**Agent:** `toMessageString` runs the value through the same accessor-refusing sanitizer as meta and takes its `message` field or the sanitized JSON; a getter, `toJSON` or `Symbol.toPrimitive` on a hostile message is never invoked, a primitive is stringified directly. two cases: object whose getter/toJSON/toPrimitive all throw goes through `logger.warn` without throwing, decodes as `[accessor]`; Error whose message getter throws decodes as `[unreadable]`. `UNIT-TEST-LOG-ENCODER-1.P14/P15`.

src/utils/logging/LogUploader.ts
daiagi: **Luka:** fixed, though the actual harm was smaller - nothing was lost, just duplicated into the old bucket. 4b3e6b74.

**Agent:** identity and delta now read together after the jitter sleep; before the sleep the cursor is only read (to skip an idle realm without paying it), never moved. `files an upload under the identity current after the jitter` sets the peer while an upload is already sleeping: one POST under the new peer, second upload sends nothing. fails against the old order with two POSTs. `UNIT-TEST-LOG-UPLOADER-1.P13`; Missing-behavior entry in `LogUploader.ts.md` removed.

src/evm/p2pRuntime/P2pRuntimeHost.ts
daiagi: **Luka:** open design question, not a patch here - need your call on the fatal-policy shape before building.

**Agent:** sdk worker surviving a fatal is deliberate mode parity - `startP2pRuntimeWorker.ts:24` and `P2pRuntimeClient.ts:329` both say errors surface as if the host ran inline, and inline mode's crash hooks survive too. ending the worker on any unhandled rejection while inline keeps going would break `INV-RUNTIME-1`. right shape is a fatal policy at the `P2pInstance` level that behaves identically in both modes. vm worker differs - stateless compute service, one owner - so ending it there is correct, and that's what it does.

src/evm/P2pInstance.ts
daiagi: **Luka:** fixed. one honest gap left, no test path exists for it yet.

**Agent:** `dispose` runs all three teardowns under `allSettled`, drops the bridge and the owned logger only after every one settled, then rethrows the first rejection. `REQ-LOG-1` in words, same rule the host already applies one level down at `P2pRuntimeHost.ts:264`. gap: the rejecting path has no executable evidence - none of the three members can currently reject through the real collaborators (ethers' `off` never rejects, the client swallows a host that's already gone), a test would need a mock. recorded in the conformance row. success path covered end to end by `a closed session leaves no root on the flush bus`.

src/utils/logging/logStore.ts
daiagi: **Luka:** declining - the finding's mechanism is wrong, the actual thing you want is already tested.

**Agent:** server cap is `MAX_CHUNK_BASE64_BYTES` on the base64 of the *deflated* payload (`crash-log-server.js:22,271`), not utf-8 of raw json, and `compressToBase64` really deflates. a 10 MiB utf-16 store is at most 5M chars; even incompressible that base64s to ~6.7 MiB, under the 8 MiB cap. compression sits between the two units the finding compares. what `REQ-LOG-7` actually asks - bounding the *inflated* size at the receiver - is `inflateBounded` + `maxInflatedBytes` in `logChunks.js`, and it's tested.

src/utils/logging/logStore.ts
daiagi: **Luka:** fixed, cheap enough despite being theoretical - c1a745e8.

**Agent:** 64 bits from `crypto.getRandomValues` rather than `randomUUID` - `getRandomValues` exists in workers and on a plain-http origin where `randomUUID` does not, and `logStore.ts` is browser-compiled. `draws a 64-bit store id that no two stores share` checks the format and 200 distinct draws. `UNIT-TEST-LOG-STORE-1.P5`; Missing-behavior entry in `logStore.ts.md` removed.

daiagi:
**[TO1] The crash tests do not prove the full topology or fatal lifecycle**

**Human:** confirmed, fixed. topology proof is in the stacked PR.

**Agent:** `rejects a call still in flight when the worker crashes` closes the pending-call gap - crashing precompile now keeps answering for 30s, so the call that scheduled the crash is still unanswered when the thread ends and must reject rather than hang. `keeps the caller's logger working after the worker crashed` closes the caller-owned-logger gap: after crash+dispose the host logger logs, a realm-wide flush reports nothing timed out, entry reaches the receiver. three-realm topology proof (`a crash while the evm is still being built reaches the realms above`) is in the stacked PR with FO3. `UNIT-TEST-WORKER-CONTRACT-EXECUTOR-1.P12/P13`.

daiagi:
**[TY2] Two promised boundaries have no executable evidence** 

**Human:** both boundaries covered now.

**Agent:** token routes (810be5c1): no token -> 401 not stored, wrong token -> 401, correct token -> 200 readable, refused-then-accepted send stores only the accepted store. `REQ-LOG-7.T1.P4/P5/P6`. browser host (5ddf866b): Playwright runner (`test:browser:worker`) starts real `crash-log-server.js` on a fresh dir, drives a browser main realm with a vm worker beneath it; worker crashes on a timer and collects on its own, `logger.uploadLogs` runs a collection over the worker port. asserted: round reports both realms ok with nothing timed out, server holds a `vm` chunk and a `main` chunk carrying the marker under the main realm's identity. `REQ-LOG-8.T1.P5`. caught two real bugs: runner had no `@platform/evmJumpdestCache` alias (existing worker smoke could never have run), `BrowserLogUploader` only listened on `window` (vm worker reachable only via a shim) - now listens on whichever scope is running.

daiagi:
**[AY1] Three new tests bypass repository test rules** 
**Human:**  4c4d7952 here, worker staging in the stacked PR.

**Agent:** `as unknown as LogEntry` cast removed - non-string message case now goes in through a real logger's `warn` and reads the stored entry back. worker staging moved into `LogFlushBus.fixture.ts` (`hostRealmOn`) and `workerAnswerPrecompile.ts` (`crashingWorkerPrecompile`, `unloadableWorkerPrecompile`); server staging into `crashLogServer.fixture.ts`, shared by the server suite and the new fetch suite.

daiagi:
**[DR1] The new requirements are not connected to their actual evidence or implementation owners**

**Human:** fixed - d9f58b2d.

**Agent:** `yarn spec:impact --base origin/dispute` exits 0, no unmapped files. `Logger`, `LogUploader`, `logStore`, `logEncoder` now carry real responsibility/decisions/linked-requirements/conformance rows/obligations (`UNIT-TEST-LOGGER-1`, `UNIT-TEST-LOG-UPLOADER-1`, `UNIT-TEST-LOG-STORE-1`, `UNIT-TEST-LOG-ENCODER-1`); `WorkerContractExecutor` and `P2pInstance` carry the log requirements they implement. six direct-evidence suites lost their false `coverage-ignore` and have verification reports. `crashLogServer`, `logChunks`, `fetchLogs` map to `REQ-LOG-5/6/7` permutations. `spec:refresh` run, byte-stable, approvals register untouched.

daiagi:
**[DY2] Existing verification reports no longer describe their suites** 

**Human:** fixed

**Agent:** dynamic-declaration row was base's - b7e78b61 unrolled the loop and left the report behind. every anchor in `E2E-RuntimeTransportModes`, `WorkerContractExecutor`, `logEncoder`, `LogUploader`, `LogFlushBus` now matches its declaration; dynamic row replaced by two static ones; four transport-mode log cases and five worker crash cases have rows; `LogUploader`'s overview no longer claims an empty batch.

daiagi:
**[GO1] Prefer merge-on-write storage**

**Human:** defer to follow up

**Agent:** surface is self-contained and all new in this PR - write path `crash-log-server.js` L462-476, read merge L582-619, `logChunks.js`, four fetch clients. scope for the follow-up: per-thread streams only, one framed length-prefixed record file per `(channel, peer, thread, store)`, each record carrying `fromSeq`/`toSeq` so eviction gaps survive; cross-thread interleave stays merge-on-read, a channel-wide file would serialize every peer's upload through one lock. two things current layout gets free that the rewrite must add back: retried delta is idempotent today only because it overwrites its own chunk by name (L475-476) - append needs a per-stream `lastToSeq` and a skip for `toSeq <= lastToSeq`; and there's no per-stream write lock today, only the global `uploadInFlight` counter.

daiagi:
**[GO2] Expose raw per-thread compressed streams as an additional API**

**Human:** same follow-up as GO1, sequenced after it.

**Agent:** on per-thread streams the raw endpoint is just the stream file itself, so building it before the layout lands means building it twice.

daiagi:
**[GO3] Make worker protocols composable through the shared typed RPC/service abstraction**

**Human:** done, as the PR stacked on this one.

**Agent:** trusted `MessagePortTransport` under a `PortRpcRouter` on every worker port, each side's operations as ordinary services on a typed root, log flushing as one `logControl.flush` call above a neighbour registry - no unions, no per-worker `logControl` switch. FR1 and FO3 fixed there by removal rather than patched here: no shared rounds so nothing per-source to track; vm's log identity is a parameter of `init` so no port ordering left to get wrong.

lukastanisic99:
@daiagi we merged our discovery feature into `dispute`, it ended up taking a lot more effort and fixing deep nested bugs across the stack, but good thing is over 2k tests now are running stable constantly - see is it easier for you to rebase your stack on all of that or just redo it - resolving conflicts will probably be a mess;

lukastanisic99:
@daiagi not sure did it resolve the conflicts correctly since now a lot of tests are failing - take a look and let me know when it's ready for review

daiagi:
I know. Handling it

lukastanisic99:
@daiagi did you double check how you resolved the conflicts, since those parallel tests that are failing should pass more or less deterministically

docs/spec/audit/open-findings.md
```diff
@@ -38,9 +38,10 @@ Each should become a normal fix + test once confirmed.
...
...                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               | Decided (engineer review): the spec is the intent — make `FraudProofStorage` first-write-wins and keep the participant index exactly consistent with the stored set. Closes when that lands.                                                                                                                                                                                                                  |
```
lukastanisic99, line 41: **Human note**

ye I'm suss about how the agent resolved conflicts

**AI-generated review**

🔴 **[FR1] 95% — The rebase un-resolves `FIND-STORAGE-4`, which `dispute` resolved in #471, while the source it describes is identical to `dispute`.** On `dispute` this row reads "_Resolved ([#471](…))._ `FraudProofStorage.storeFraudProof` now returns early on an existing content hash"; at head it is the pre-#471 text claiming `storeFraudProof` "overwrites unconditionally". `git diff origin/dispute...HEAD -- src/storage/FraudProofStorage.ts` is empty, so the register now asserts a defect the code does not have. This is exactly the conflict-resolution miss your Sep 9 comment asked about, and it is the only row in the file that changed. A second row (`FIND-RUNTIME-1`, L44) is a legitimate new finding and is fine.

> **Fix FR1-FIX**
>
> Restore line 41 from `origin/dispute` verbatim (`git checkout origin/dispute -- docs/spec/audit/open-findings.md` then re-add the `FIND-RUNTIME-1` row). Acceptance: `git diff origin/dispute...HEAD -- docs/spec/audit/open-findings.md` shows only the added `FIND-RUNTIME-1` row.

src/utils/logging/LogFlushBus.ts
```diff
@@ -0,0 +1,371 @@
...
+                this.activeLocalFlush = undefined;
+            }
+        });
+        this.activeRound = active;
+        return active;
+    }
+
+    // a port is skipped only when every folded trigger came from it, so widen
+    // by union - a reset would forward the round back to the realms that asked
+    private foldExclusion(fromPort: LogControlPort | undefined): void {
```
lukastanisic99, line 285: **AI-generated review**

🟠 **[FO2] 85% — Carried from `1-review.md` FR1, downgraded: coverage holds, the reported counts do not.** With children A and B both originating during a hub round, the hub's queued round excludes both ports, so A's ack carries hub+A and B's carries hub+B while every store was in fact uploaded (A's and B's own rounds covered their subtrees). `REQ-LOG-2` (each asker told about the whole session) is therefore violated in the number, not in the data. The author defers the fix to #466, which removes shared rounds. If #461 is meant to land alone, the count is wrong in production until #466 merges. The test at [LogFlushBus.test.ts L259](https://github.com/peer3to/state-channels-plus/blob/e3cb5a32828e2da56e6d2d1c99bbe376874b3ee7/test/utils/logging/LogFlushBus.test.ts#L259) still asserts forwarding counts rather than the returned totals.

> **Fix FO2-FIX**
>
> Cheapest here: when folding, remember the result each excluded asker would have needed and sum the completed active round into the folded ack (`activeRound` result + queued result), or simply do not exclude on fold and accept one echo. Either way add the three-realm test where both leaves originate mid-round and assert each receives `ok: 3`. If deferred, apply XO1-FIX.

src/evm/contractExecutor/WorkerContractExecutor.ts
```diff
@@ -102,6 +106,7 @@ export default class WorkerContractExecutor extends AContractExecutor {
             config,
             clockAdjustmentSeconds
         });
+        executor.attachLogPort();
```
lukastanisic99, line 109: **AI-generated review**

🟠 **[FO3] 90% — Carried from `1-review.md`: the parent log port is attached only after `init` resolves, so a VM crash during `createEvm` posts `logControl` into an undefined handle and a rejected `init` leaks the worker.** `create()` awaits `workerReady` and the `init` request, then attaches; the worker host attaches its end at the top of `init` ([ContractExecutorWorkerHostCore L89-96](https://github.com/peer3to/state-channels-plus/blob/e3cb5a32828e2da56e6d2d1c99bbe376874b3ee7/src/evm/contractExecutor/worker/ContractExecutorWorkerHostCore.ts#L89-L96)). Nothing in `create` disposes the executor when `init` rejects. The author fixes both in #466 by making the VM's log identity a parameter of `init`; nothing changed here.

> **Fix FO3-FIX**
>
> Minimal here: attach the port before sending `init` and wrap `create` in `try { … } catch (e) { await executor.dispose(); throw e; }`. Add "ends the worker when init fails instead of leaking it" (the author says it exists in #466) to this branch. If deferred, say so in the PR body and mark `REQ-LOG-1` `Partial` in `WorkerContractExecutor.ts.md`.

src/utils/logging/LogFlushBus.ts
```diff
@@ -0,0 +1,371 @@
...
+        this.portOwners.set(port, root);
+        this.postContextOn(port, root.getSharedContext());
+        return {
+            receive: (message) => this.receive(message, port),
+            remove: () => {
+                this.portOwners.delete(port);
+                // a removed port can never ack -> settle its waiters now, not
+                // at the timeout
+                for (const waiting of this.pendingAcks.values()) {
+                    waiting.get(port)?.(emptyFlushResult());
```
lukastanisic99, line 81: **Human note**

Nit, but all these nits that something doesn't cleanup/abort/fails safely were causes of many flakes - the logger is not mission critical software, but still the same pattern

**AI-generated review**

🟡 **[RY1] 85% — Carried from `1-review.md` RY3 (never posted): a port removed mid-round settles as "nothing", not as a missing realm.** A worker that crashes and drops its port while a round is waiting resolves that hop with all-zero counts, so the `Log flush round reached` record cannot distinguish a complete collection from one that lost the crashing realm. `REQ-LOG-2` says the caller must be told how many "never answered".

> **Fix RY1-FIX**
>
> Settle a removed port with `{ ok: 0, failed: 0, timedOut: 1, entries: 0 }` (or a new `unreachable` counter if you want to keep `timedOut` literal), and add a remove-during-active-round test asserting the round result names it.

lukastanisic99 (commented):
**Human note**

Account for this if we're not doing it here, so we don't forget;

**AI-generated review**

🟠 **[XO1] 90% — `REQ-LOG-2` promises each concurrent asker the whole session; the bus still gives a folded asker its own subtree's count.** [log-collection.md L44-48](https://github.com/peer3to/state-channels-plus/blob/e3cb5a32828e2da56e6d2d1c99bbe376874b3ee7/docs/spec/specification/runtime/log-collection.md#L44-L48): "When two threads ask at the same time, each MUST be told about the whole session". [`foldExclusion` L285-296](https://github.com/peer3to/state-channels-plus/blob/e3cb5a32828e2da56e6d2d1c99bbe376874b3ee7/src/utils/logging/LogFlushBus.ts#L285-L296) unions the source ports of folded requests, so the queued round skips those ports and the acks those askers receive omit each other's subtrees. The author confirms the count is wrong and defers the fix to #466. The mirror `LogFlushBus.ts.md` should say `Contradicts` for `REQ-LOG-2` until #466 lands; the spec text is the intended contract and stays.

> **Fix XO1-FIX**
>
> Either fix here (see FO2-FIX) or mark `REQ-LOG-2` as `Contradicts` in `LogFlushBus.ts.md` with the #466 reference, so the register does not read `Covered` for a known gap.



---

**Human note**

It doesn't have to be backwards compatible, but it has to be atomic - marked as 'human assesment needed'

**AI-generated review**

🟠 **[FO4] 85% — Carried from `1-review.md` (never posted): client, server, and on-disk formats must be rolled out together.** **Human assessment needed** Decision: accept a coupled rollout (server first; pre-existing flat files become unreadable; old clients get 400 until upgraded), or require the new server to accept old-shape uploads and read old flat files? Evidence: [crash-log-server.js L252-260](https://github.com/peer3to/state-channels-plus/blob/e3cb5a32828e2da56e6d2d1c99bbe376874b3ee7/scripts/logging/crash-log-server.js#L252-L260) returns 400 when `threadName`, `storeId`, `fromSeq`, or `toSeq` is missing; the reader walks only `<thread>/<store>/<from>-<to>` chunk directories; `decodeLogEntry` requires `wallTimeMs`. The server is an internal script the team deploys, so a coupled rollout may be fine, but nothing in the PR says so. Recommended: accept the coupling and write the deploy order (server, then clients) into `scripts/logging/README` or the spec's assumptions, and add a one-line 400 body that names the missing field so an old client's failure is diagnosable.

> **Fix FO4-FIX**
>
> Conditional on the decision. Coupled: document the order and the "old data unreadable" consequence; no code change. Compatible: accept uploads without the four fields as a full-store `main` upload under a synthetic store id, and read `<peer>/*.log` flat files as a `main` stream with `wallTimeMs` defaulted from `time`; cover old-client → new-server and old-disk → new-reader in `crashLogServer.test.ts`.



---

**Human note**

Take a look how things merged and is it a 'flake' that I somehow never observed or what

**AI-generated review**

🟡 **[TO1] 85% — GitHub's `test` check is red on this head while the canonical gate passes locally on the same SHA.** The CI job `Run full test suite on the distributed pool` failed for `e3cb5a32` (run 34313025244); locally, compile, both typechecks, the 149-test focused selection, and the full distributed gate (2152 passing, 0 failing, `logs/run-196/`) pass. The two starvation retries in my run are in `E2E-InitHandshake` and `E2E-LobbyMatching`, files this PR does not touch. Without the CI job's failing test names (the log needs a token) I cannot say whether CI hit pool starvation or something environment-specific; nothing in the diff reproduces a failure here.

> **Fix TO1-FIX**
>
> Author to paste the failing CI test names into the PR and re-run the job. If they carry the starvation signature (watchdog / event-loop delay / `exit null`), treat as pool load; if they are `WorkerContractExecutor` or `E2E-RuntimeTransportModes` cases, that points back at the rebase resolution.



---

**AI-generated review**

🟡 **[TY2] 85% — Unit and E2E gaps that remain after the new suites.** (a) `P2pInstance.dispose` rejecting path: no test (author admits none of the three teardowns can reject through real collaborators; record it as an accepted gap in the mirror rather than an unstated one). (b) Port removed during an active round (RY1). (c) Three-realm star with two leaves originating mid-round (FO2). (d) `init` rejection disposing the worker (FO3). (e) `INV-LOG-2`, `REQ-LOG-1`, `REQ-LOG-10` have 0 mapped tests after refresh; `REQ-LOG-10` (ending thread waits only for its own upload) is implemented in `ContractExecutorWorkerEntry` and has a `LogFlushBus.test.ts` case per the author's Aug 31 note, so this is a mapping loss from DO1 rather than a missing test. E2E: `E2E-CrashLogUpload` covers the two-thread topology in worker and inline modes; the three-thread topology E2E is in #466.

> **Fix TY2-FIX**
>
> Add (b), (c), (d) as static `it` cases in `LogFlushBus.test.ts` / `WorkerContractExecutor.test.ts`; record (a) in `P2pInstance.ts.md`; (e) resolves with DO1-FIX.



---

**AI-generated review**

🟡 **[TY3] 90% — The new suites follow the harness rules; one assertion was removed with the disputed finding.** 27 `LogFlushBus` cases, 23 `WorkerContractExecutor` cases, 18 server cases, 9 chunk cases, 8 delta cases, 6 E2E cases, all static `it`, no `this.timeout`, no `console.*`, no casts, fixtures under `test/fixtures/logging/`. The Playwright smoke (`test/browser/crash-log-smoke.js`) is outside Mocha discovery and runs only via `test:browser:worker`, so it is not part of the canonical gate; the PR body should say so.

> **Fix TY3-FIX**
>
> One sentence in the PR body naming `yarn test:browser:worker` as a separate, manual gate.



---

**Human note**

Nit

**AI-generated review**

🟠 **[DO1] 95% — Every anchor in the logging verification reports is one line above its declaration, so 65 declarations are unreferenced again after the rebase.** Example: [`LogFlushBus.test.ts.md`](https://github.com/peer3to/state-channels-plus/blob/e3cb5a32828e2da56e6d2d1c99bbe376874b3ee7/docs/spec/verification/tests/test/utils/logging/LogFlushBus.test.ts.md) anchors "uploads a connected realm's logger" at `#L94 (line 94)`; at head the declaration is [L93](https://github.com/peer3to/state-channels-plus/blob/e3cb5a32828e2da56e6d2d1c99bbe376874b3ee7/test/utils/logging/LogFlushBus.test.ts#L93). The refreshed `generated/verification-coverage.md` lists 25 `LogFlushBus`, 20 `WorkerContractExecutor`, 12 `crashLogServer`, 6 `logChunks`, 1 `fetchLogs`, and 1 `E2E-CrashLogUpload` declarations under "Tests not referenced". This is the regression of DR1/DY2 that commit `c5fe858d` fixed once and the final rebase commit undid (a one-line shift in each test file). `docs/spec/AGENTS.md`: repair a shifted anchor when the declaration has one unique match; every one of these does. `spec:refresh` is deterministic and CI does not fail on unreferenced declarations, which is why nothing caught it.

> **Fix DO1-FIX**
>
> Re-anchor the six reports to the head declarations (a one-line shift per file), run `yarn spec:refresh`, and confirm the "Tests not referenced" section lists none of these files. Acceptance: `spec:impact` mapped-test counts for `REQ-LOG-*` rise and `INV-LOG-2`/`REQ-LOG-1`/`REQ-LOG-10` are no longer 0.



---

**Human note**

Who cares if it's the same - the only question is, are we again having formater problems?

**AI-generated review**

🟡 **[DY2] 90% — Two Solidity mirrors changed only in table formatting.** `UtilityFacetInterface.sol.md` and `FraudProofFacet.t.sol.md` differ from `dispute` by column widths alone, the same formatter churn as MY1. They will show as pending review for no content change.

> **Fix DY2-FIX**
>
> `git checkout origin/dispute -- docs/spec/implementation/source/contracts/V1/StateChannelDiamondProxy/UtilityFacetInterface.sol.md docs/spec/verification/tests/test/V1/StateChannelDiamondProxy/FraudProofFacet.t.sol.md`.



---

**Human note**

Hmm this is maybe the cause of always having the diff in the format?

**AI-generated review**

🟡 **[MY1] 95% — Three contracts and a Forge test are reformatted with no semantic change.** `DisputeFraudProofFacet.sol`, `JoinChannelFacet.sol`, `LocalDiamond.sol`, and `JoinChannelFacet.t.sol` differ only in line wrapping (a different `forge fmt` version). A logging PR has no reason to touch them, and they will conflict with any concurrent contract PR.

> **Fix MY1-FIX**
>
> `git checkout origin/dispute -- contracts test/V1` and, if the local formatter disagrees with `dispute`, pin the `forge fmt` version instead of committing its output.



---

**Human note**

Human decision needed;

**AI-generated review**

🟠 **[OO1] 90% — Fatal policy for the SDK worker.** **Human assessment needed** Decision: after an unhandled error inside the SDK worker, end the worker (reject the client, dispose the runtime, drain) as the VM worker now does, or keep serving so threaded and inline modes behave identically? Evidence: [`startP2pRuntimeWorker.ts L27-31`](https://github.com/peer3to/state-channels-plus/blob/e3cb5a32828e2da56e6d2d1c99bbe376874b3ee7/src/evm/p2pRuntime/worker/startP2pRuntimeWorker.ts#L27-L31) posts `hostError` and continues; the author's Aug 31 reply argues parity with inline mode (`INV-RUNTIME-1`) and proposes a `P2pInstance`-level policy that behaves the same in both modes. The VM worker is treated differently because it is a stateless compute service. Recommended: accept parity for this PR and file the `P2pInstance`-level fatal policy as an implementation open question with the two options, so the decision is recorded rather than implied.

> **Fix OO1-FIX**
>
> Decide. Parity: add `OQ-IMPL-*` for the fatal policy and close this thread. Terminate: give the SDK worker one fatal owner (report, flush, reject client, dispose, drain) with a test that calls after the injected fatal reject.

contracts/V1/StateChannelDiamondProxy/LocalDiamond.sol
```diff
@@ -318,7 +318,8 @@ contract LocalDiamond is StateChannelManagerProxy {
         );
 
         // Perform the low-level call with a gas limit
-        (bool success, bytes memory returnData) = disputeVerificationFacetAddress.delegatecall{gas: _getGasLimit()}(data);
+        (bool success, bytes memory returnData) =
+            disputeVerificationFacetAddress.delegatecall{gas: _getGasLimit()}(data);
```
lukastanisic99, line 322: ye all the solidity changes are just formating

src/utils/logging/logControl.ts
```diff
@@ -0,0 +1,82 @@
...
+export type LogControlMessage =
+    | LogFlushRequest
+    | LogFlushAck
+    | LogContextUpdate;
+
+/** this realm's end of a thread boundary */
+export interface LogControlPort {
+    post(message: LogControlMessage): void;
+    remoteRealm: LogRemoteRealm;
+}
```
lukastanisic99, line 54: I hope in the worker RPC pr that follows we're not adding more ports and safely reuse the single port that's implicitly available for all worker<>worker communication

src/utils/logging/logControl.ts
```diff
@@ -0,0 +1,82 @@
...
+
+/** the sender's channel or identity changed */
+export interface LogContextUpdate extends LogControl<"contextUpdate"> {
+    context: SharedLoggerContext;
+}
+
+export type LogControlMessage =
+    | LogFlushRequest
+    | LogFlushAck
+    | LogContextUpdate;
```
lukastanisic99, line 48: Probably this entire file is not needed with the RPC worker - The services and implicit availability of typed request/response make all of this so elegant in the other design - one of the better design decision that has been taken in the SDK

src/evm/p2pRuntime/P2pRuntimeHost.ts
```diff
@@ -155,35 +162,59 @@ export async function startP2pRuntimeHost<
...
 
+        // threadLabel is set only by startP2pRuntimeWorker -> host is threaded
+        if (threadLabel) {
+            hostLogHandle = logger.addLogPort({
+                post: (message) => port.post({ type: "logControl", message }),
+                remoteRealm: "parent"
+            });
+            removeLogWiring = () => hostLogHandle?.remove();
+        } else if (ctx.contextFollower) {
+            removeLogWiring = logger.followContextTo(ctx.contextFollower);
```
lukastanisic99, line 215: These runtime/host/client files are probably the most bloated parts of our SDK code - can't wait to look at the next PR and see all of this crt+a,del

docs/spec/audit/open-findings.md
daiagi: **Human:** confirmed, fixed on #466. **Agent:** `FraudProofStorage.ts` had no diff against `dispute`; only the register row was rolled back by the rebase. restored from `origin/dispute`, `FIND-RUNTIME-1` re-added, diff against dispute is that one row.

src/utils/logging/LogFlushBus.ts
daiagi: **Human:** not a #466 problem - there is no fold there. **Agent:** #466 gives every request its own fan-out, so a folded asker doesn't exist: [LogFlushBus.ts L225-L244](https://github.com/peer3to/state-channels-plus/blob/74e3814e51f70085637c51c5ffd7859bf3a0fa4f/src/utils/logging/LogFlushBus.ts#L225-L244), test [`two children asking during a hub round are each told the whole tree`](https://github.com/peer3to/state-channels-plus/blob/74e3814e51f70085637c51c5ffd7859bf3a0fa4f/test/utils/logging/LogFlushBus.test.ts#L265). `REQ-LOG-2` stays `Covered` on #466. #461 alone would need the `Contradicts` row; we're merging the stack.

src/evm/contractExecutor/WorkerContractExecutor.ts
daiagi: **Human:** fixed on #466. **Agent:** identity rides in `init`, so link and init order no longer matters, and a rejected init disposes the executor before the error propagates: [WorkerContractExecutor.ts L90-L94](https://github.com/peer3to/state-channels-plus/blob/74e3814e51f70085637c51c5ffd7859bf3a0fa4f/src/evm/contractExecutor/WorkerContractExecutor.ts#L90-L94). tests [`a crash while the evm is still being built reaches the realms above`](https://github.com/peer3to/state-channels-plus/blob/74e3814e51f70085637c51c5ffd7859bf3a0fa4f/test/evm/WorkerContractExecutor.test.ts#L427) and [`ends the worker when init fails instead of leaking it`](https://github.com/peer3to/state-channels-plus/blob/74e3814e51f70085637c51c5ffd7859bf3a0fa4f/test/evm/WorkerContractExecutor.test.ts#L482).

src/utils/logging/LogFlushBus.ts
daiagi: **Human:** agreed on the pattern; #466 already settles it structurally, test added there. **Agent:** on #466 a closed link rejects the pending flush at once ([PortRpcRouter.ts L128-L147](https://github.com/peer3to/state-channels-plus/blob/74e3814e51f70085637c51c5ffd7859bf3a0fa4f/src/rpc/PortRpcRouter.ts#L128-L147)) and the bus counts that realm as never answered ([LogFlushBus.ts L235-L244](https://github.com/peer3to/state-channels-plus/blob/74e3814e51f70085637c51c5ffd7859bf3a0fa4f/src/utils/logging/LogFlushBus.ts#L235-L244)) - no all-zero settle left. new `a link that closes during the round counts as never answered`: ok 1, timedOut 1, resolves before the ack timeout.

contracts/V1/StateChannelDiamondProxy/LocalDiamond.sol
daiagi: **Human:** yes, but not two forge versions - merge commits skip the pre-commit hook. fix is #480, the two mirrors are reverted on #466. **Agent:** forge 1.1.0, 1.2.3 and 1.3.0 all format these four files to the same bytes, and those bytes are what #461/#466 carry. `dispute` got the unformatted copies through merge commits, which never run lint-staged, so nothing ever formatted them there. #480 pins v1.2.3 (README already said so; 1.8.1 is the real break, 18 files) and formats the four files - byte-identical to this branch, so this diff closes when #480 lands. the two `.sol.md` mirrors were column-width churn and are restored from dispute here.

src/utils/logging/logControl.ts
daiagi: **Agent:** on #466 both `worker/protocol.ts` files are gone (386 lines), `P2pRuntimeHost.ts` 695 -> 502, `P2pRuntimeClient.ts` 326 -> 293, `WorkerContractExecutor.ts` 386 -> 261, `LogFlushBus.ts` 371 -> 252. no new port: the log port is the existing worker link ([logControlPort.ts L8-L11](https://github.com/peer3to/state-channels-plus/blob/74e3814e51f70085637c51c5ffd7859bf3a0fa4f/src/utils/logging/rpc/logControl/logControlPort.ts#L8-L11)), `new MessageChannel` count unchanged. `logControl.ts` is 51 lines of types now (`LogFlushResult`, `LogControlPort`, `sumFlushResults`).

src/utils/logging/logControl.ts
daiagi: **Agent:** none added - the log port is the existing worker link ([logControlPort.ts L8-L11](https://github.com/peer3to/state-channels-plus/blob/74e3814e51f70085637c51c5ffd7859bf3a0fa4f/src/utils/logging/rpc/logControl/logControlPort.ts#L8-L11)); `new MessageChannel` count is unchanged on #466.

src/evm/p2pRuntime/P2pRuntimeHost.ts
daiagi: **Agent:** on #466: `P2pRuntimeHost.ts` 695 -> 502, `P2pRuntimeClient.ts` 326 -> 293, `WorkerContractExecutor.ts` 386 -> 261, both `worker/protocol.ts` files gone (386 lines). the services under `src/evm/p2pRuntime/rpc/` are what replaced the unions and switches.

daiagi:
**[XO1] `REQ-LOG-2` folded-asker count**

**Human:** not a #466 problem - there is no fold there. **Agent:** #466 gives every request its own fan-out, so a folded asker doesn't exist: [LogFlushBus.ts L225-L244](https://github.com/peer3to/state-channels-plus/blob/74e3814e51f70085637c51c5ffd7859bf3a0fa4f/src/utils/logging/LogFlushBus.ts#L225-L244), test [`two children asking during a hub round are each told the whole tree`](https://github.com/peer3to/state-channels-plus/blob/74e3814e51f70085637c51c5ffd7859bf3a0fa4f/test/utils/logging/LogFlushBus.test.ts#L265). `REQ-LOG-2` stays `Covered` on #466. #461 alone would need the `Contradicts` row; we're merging the stack.

daiagi:
**[FO4] client / server / on-disk rollout**

**Human:** coupled + atomic, as you said. **Agent:** assumption added to `log-collection.md`: receiver and senders ship as one unit, old layout unreadable, other-version sender gets a 400 naming the first missing field (`Missing storeId` etc, was `Incorrect request data`). two validate tests assert the field name.

daiagi:
**[TO1] CI red on `e3cb5a32`**

**Human:** not the rebase - runner self-test flake. **Agent:** the two reds are the distributed runner's own tests: `e2eParallelDistributedE2E > reports every real worker failure cause after quarantine and protocol loss` (`AssertionError: Target cannot be null or undefined` at L636 - the run *succeeded*, the 25 ms discovery refresh re-dialed before the malformed frame landed) and `e2eParallelDistributedLifecycle > keeps concurrent attempt artifact transfers isolated by request` (`Timed out waiting for worker-a LEASE_GRANTED`). both on server-1, 12 s apart. this PR touches nothing under `scripts/` but `scripts/logging/`, nothing in `test/fixtures/distributed`. #466's run on the superset tree passed 2 min later; both tests pass locally on this tip. new tests with the discovery merge, so 'never observed' is expected.

daiagi:
**[TY2] remaining unit / E2E gaps**

**Human:** (a) already in the mirror, (b) test added on #466, (c)(d) exist on #466, (e) closes with DO1. **Agent:** (a) [P2pInstance.ts.md, REQ-LOG-1 row](https://github.com/peer3to/state-channels-plus/blob/74e3814e51f70085637c51c5ffd7859bf3a0fa4f/docs/spec/implementation/source/src/evm/P2pInstance.ts.md#L74) records the rejecting path as an accepted gap. (c) L265, (d) L482 - links in the FO2/FO3 replies.

daiagi:
**[TY3] browser smoke is a separate gate**

**Human:** added to the PR body.

daiagi:
**[DO1] unreferenced logging declarations**

**Human:** fixed on #466. **Agent:** `spec:refresh` is byte-stable, so the list was real: LogFlushBus was re-anchored on #466 already; WorkerContractExecutor had lost dispute's rows in the rebase and pointed P8/P12 at deleted tests; the crashLogServer/logChunks/fetchLogs/E2E rows anchored correctly but had no test id. all 24 executor declarations now map (P1-P23), the script rows map to new `REQ-LOG-5/6/7.T1` permutations. 'Tests not referenced' lists none of the six files.

daiagi:
**[MY1 / DY2] solidity reformat + mirror churn**

**Human:** yes, but not two forge versions - merge commits skip the pre-commit hook. fix is #480, the two mirrors are reverted on #466. **Agent:** forge 1.1.0, 1.2.3 and 1.3.0 all format these four files to the same bytes, and those bytes are what #461/#466 carry. `dispute` got the unformatted copies through merge commits, which never run lint-staged, so nothing ever formatted them there. #480 pins v1.2.3 (README already said so; 1.8.1 is the real break, 18 files) and formats the four files - byte-identical to this branch, so this diff closes when #480 lands. the two `.sol.md` mirrors were column-width churn and are restored from dispute here.

daiagi:
**[OO1] sdk-worker fatal policy**

**Human:** terminate. next PR in the stack, on top of #466. **Agent:** one fatal owner in the sdk worker: report + flush once, reject the client's pending and later calls, dispose the runtime, close the port, drain the thread; proof = calls after the injected fatal reject and no worker survives. anything that reaches the worker's unhandled hook ends the peer - a rethrow that gets there is a bug, and ending the peer is what makes it visible. the vm worker dying ends the sdk host the same way (today the host hangs). inline isn't a counter-argument: with the default config an unhandled rejection already exits the process there. on #466 a closed port already rejects every pending request at once ([PortRpcRouter.ts L128-L147](https://github.com/peer3to/state-channels-plus/blob/74e3814e51f70085637c51c5ffd7859bf3a0fa4f/src/rpc/PortRpcRouter.ts#L128-L147)), so the client side is mostly there.

daiagi:
@lukastanisic99  to clarify, all the fixes are on #466 and the test suite is green.
I didn't push anything here
I think the test suite still has some low number of flakes. see my PR #480. it has 1 failing test:
```
    - test:ReductionManager.test.ts:disposal after the completion exists settles the attempt as undefined and installs nothing
    ```
    
    which has nothing to do with that minimal PR

lukastanisic99:
> @lukastanisic99  to clarify, all the fixes are on #466 and the test suite is green.
> I didn't push anything here
> I think the test suite still has some low number of flakes. see my PR #480. it has 1 failing test:
> ```
>     - test:ReductionManager.test.ts:disposal after the completion exists settles the attempt as undefined and installs nothing
>     ```
>     
>     which has nothing to do with that minimal PR

Ye I'll take a look later - I'm playing around with some ideas for #466 so we'll compare later - Momcilo also found some other flakes and is fixing them, so whenever new flakes arise we fix them on the spot so it's always stable

PR #462, opened 2026-08-27
test: drop loop-generated it() and the junk fuzz suite

daiagi:
follow-up from #453 review comment.

- deleted E2E-Fuzz-Dispute-MVP.test.ts - env-var-sized loop of it(), no assigned test IDs, junk
- DisputeManager.test.ts: unrolled the two for-loops that generated it() names into explicit static cases (killDispute race-condition custom errors, unrecognized apply failure at send/wait)
- ci.yml: cache foundry toolchain + compiled contracts/typechain output so discovery doesn't force a full recompile every run


lukastanisic99:
Didn't yet look, just read the PR description - you can probably get rid of foundry and compiling in the CI now that all .it() are top level discoverable - the agent added foundry and compilation to the CI since it needed the artifacts for some reason to discover tests - all the real compilation and foundry invocation happens on the workers

daiagi:
you were right, discovery is static parsing only, never imports test files or invokes forge. dropped foundry+compile from ci.yml. also unrolled 2 more dynamic-title loops (E2E-Spectate, WorkerContractExecutor) that were forcing it.

PR #463, opened 2026-08-27
refactor(contracts): route proxy calls through selector fallback dispatch instead of typed forwarders

sh3ll3x3c:
Stacked on #459 (base: `refactor/facet-size`). Completes the deployable side of [Trello 349](https://trello.com/c/hNitQoP8): after #459 put every facet under the EIP-170 limit, `StateChannelManagerProxy` was the last deployable contract over it (29,168 bytes).

## Summary

The proxy's 28 typed forwarder bodies (`abi.encodeCall` + `_delegatecall` + `abi.decode`, ~200–400 bytes each) are replaced by selector routing in the already-existing fallback: `_facetForSelector(msg.sig)` maps 56 selectors to their facet and the fallback delegatecalls it with raw `msg.data`. Each selector is written as `Facet.fn.selector`, so the compiler derives it — a signature change updates the routing with it; nothing is hand-maintained. Unknown selectors still fall through to the consumer facet, and revert data still bubbles verbatim (the dispute client keeps decoding custom errors from these reverts). `facetAddressForSelector(bytes4)` exposes the table for tests.

The state-reading view wrappers from #459 moved verbatim to `UtilityFacet` (which now extends `StateChannelCommon`; `UtilityFacetInterface` breaks the definition cycle and keeps facet and base in step via `override`). With no proxy-declared function calling Common's internals, Common's bodies dead-code-eliminate out of the proxy.

Sizes (forge, deployed runtime):

| Contract | Before | After | Margin |
| --- | --- | --- | --- |
| StateChannelManagerProxy | **29,168 (over)** | 13,805 | +10,771 |
| UtilityFacet | 7,581 | 14,894 | +9,682 |
| JoinChannelFacet | 7,340 | 7,262 | +17,314 |
| LocalDiamond (test-only) | 39,930 | 29,211 | exempt |

No other contract changed size; the tightest deployable contract is now `DisputeFraudProofFacet` at 22,176. The only over-limit contracts are test-only (`LocalDiamond`, `DisputeExpiryGuardHarness`).

## Design notes

- **Interface inheritance**: the proxy no longer inherits `StateChannelManagerInterface`. Solidity cannot leave an inherited declaration unimplemented, so the interface became a caller-side typing artifact nothing implements — the union of proxy-implemented and routed functions, carrying `StateChannelManagerEvents` so `.filters` keeps working. Its ABI is verified to be an exact superset of the old proxy ABI (identical names, params, and state mutability; only `facetAddressForSelector` added), so no TS or typed self-call changes shape. TS and the facets' `StateChannelManagerInterface(address(this))` self-calls bind it; `connectLocalDiamond` merges LocalDiamond's own ABI with the routed one.
- **Deliberate exclusions**: 20 facet functions are intentionally not routed, each with a documented reason in the test fixture — UtilityFacet's 13 stateless helpers (Common calls them on the facet address directly; `tryDecodeBlock` would misbehave through the proxy because its `this.decodeBlock` self-call would land there), DisputeVerificationFacet's 6 internal pipeline steps including `killDispute`, and `FraudProofFacet.runFraudProof`. Routing them would widen the attack surface for no caller.
- **Payable audit**: no `payable` function exists anywhere in `contracts/`; the fallback stays non-payable and value-handling semantics are unchanged.
- **LocalDiamond**: its debug `isBlockAuthentic` is now LocalDiamond's own public function (declared functions dispatch before the fallback), so local deployments keep the debug behavior TS sees today.
- Moved wrapper bodies are byte-identical to the originals; storage layouts verified identical slot-for-slot across proxy, facets, and LocalDiamond.

## Tests

- **`SelectorRouting.test.ts`** (14 cases) is the drift-catcher: it builds every facet's selector set from the real compiled ABI and asserts each selector resolves to that facet via `facetAddressForSelector`, that exclusions and unknown selectors fall through to the consumer facet, that the proxy's own selectors are absent from the table, and that no selector is defined by two facets. A facet function added without a routing entry (or documented exclusion) fails it.
- **`LocalDiamondBinding.test.ts`** (5 cases) covers the merged local ABI, including a real crash found during this work (`format("sighash")` throwing on the constructor fragment).

## Test plan

- Full distributed gate on the pool (all mocha unit + e2e + forge tiers): **1481 passing, 0 failing** — every routed selector exercised end-to-end through the new dispatch.
- `forge build --sizes`: table above; no facet regressed.
- `yarn tsc --noEmit` on `tsconfig.json` **and** `tsconfig.browser.json` — clean.
- `yarn spec:refresh` green and idempotent; new/updated implementation reports for the moved surface.

## Follow-ups (not in this PR)

- **No build gate enforces EIP-170** — hardhat runs with `allowUnlimitedContractSize` and foundry has no size check, so nothing catches the next size regression. A `forge build --sizes` CI step would close that.
- `StateChannelManagerProxyOpen.test.sol` is invisible to the spec test-discovery (`.test.sol` vs `.t.sol`) — recorded as `OQ-VER-DISCOVERY-1-9JWTRH` rather than papered over.
- `foundry.lock` (emitted by newer Foundry) is neither committed nor gitignored.
- `UNIT-TEST-UTILITY-FACET-2` permutations for the moved views under delegatecall are recorded as an honest coverage gap; nothing reads those views through the manager today.


contracts/V1/StateChannelDiamondProxy/StateChannelManagerProxy.sol
```diff
@@ -621,143 +275,83 @@ contract StateChannelManagerProxy is StateChannelManagerInterface, StateChannelC
...
-            }
-        }
-        return abi.decode(returnData, (bool));
+    /// @dev Selector routing table. Selectors are derived by the compiler from the
+    ///     facet function types, so a signature change updates the routing with it -
+    ///     nothing here is hand-hashed. The state-changing facets come first so
+    ///     on-chain calls pay the fewest comparisons; the utility views are read
+    ///     off-chain. Unknown selectors keep the historical behaviour of falling
+    ///     through to the consumer facet.
+    function _facetForSelector(bytes4 sig) internal view returns (address) {
```
lukastanisic99, line 286: This can be O(1) for all calls - you don't have to compare everything. Take a look at `PokerDiamond.sol` in poker contracts. There we did the true Diamond implementation following EIP-2535; All the selectors are stored in a `storage map` (selector->facet); The cost is on the deployer to populate the storage map and later reads are O(1). Also take a look how we did storage in poker contracts for our Diamond. It's interesting since it allows us to have many storage layouts so contract upgrades can fine tune which storage slots overlap - you don't have to do it here in this PR, it's not critical at the moment it's just a nice to have upgradability option for later

contracts/V1/StateChannelManagerInterface.sol
```diff
@@ -81,27 +143,75 @@ abstract contract StateChannelManagerInterface {
...
+    // ********** routed to DisputeVerificationFacet **********
+
     function challengeDisputeReduction(
         Dispute[] memory disputes,
         StateSnapshot memory latestStateSnapshot,
         bytes memory encodedStateMachineState,
         MessageBlock[] memory inboundMessageBlocks
     ) public virtual;
 
+    function reduce(Dispute[] memory disputes) public virtual returns (ReduceOutput memory);
```
lukastanisic99: **Human note**

Just bringing this to your attention, but probably it's accounted for since tests are passing;

**AI-generated review**

🟠 **[FO1] Ten interface declarations do not match the routed facet mutability.** Eight view/pure facet functions are nonpayable or `view` in the new interface, so ethers can choose a transaction where the old binding performed an `eth_call`; two more pure functions are declared `view`. The mismatches include `reduce`, `hasInvalidTimestamp`, the state-proof validation helpers, `isGenesisSnapshotWithoutTimeCheck`, and `isSnapshotNewer`. This breaks the PR's stated ABI-compatibility contract even though selectors stay identical.

> **Fix FO1-FIX**
>
> Make every interface declaration match its routed facet exactly. Add an exhaustive ABI reconciliation test that compares signature and `stateMutability`, not only selectors, for every proxy-owned and routed function.

lukastanisic99 (commented):
**Human note**

NIT, but true

**AI-generated review**

🟡 **[AY1] `LocalDiamondBinding.test.ts` keeps reusable test logic inside the declaration file.** `fragmentKeys` is a general ABI helper used by the cases below it. `test/AGENTS.md` requires reusable helpers and builders to live in fixtures so test files contain direct declarations and their assertions.

> **Fix AY1-FIX**
>
> Move `fragmentKeys` to the ABI or selector-routing fixture and import it into the test.



---

**Human note**

NIT

**AI-generated review**

🟡 **[DY2] The maintained audit summaries were not refreshed after the routing and ABI change.** For example, `docs/spec/audit/verification.md` still reports 924/4464 planned permutations and 535/1001 mapped declarations, while the regenerated coverage report now says 791/4562 and 490/1194. The PR changes a contract trust boundary and several generated inverse reports, but none of the maintained implementation, verification, or security audits changed.

> **Fix DY2-FIX**
>
> Refresh the affected maintained audits against the final generated reports and routing evidence. Do not edit the engineer-owned approval register.



---

**Human note**

Yes, it would be good to have a test that catches any production contract over the limit;

**AI-generated review**

🟡 **[OQ1] Should contract-size limits become an executable deployment gate in this PR?** The change exists because the old proxy exceeded EIP-170, but the repository still relies on compile output and a manual size table. A later facet or init-code growth can recreate the same deployment failure without a focused assertion.

> **Recommended decision**
>
> Add a small build-time test that checks every deployable runtime against EIP-170 and relevant creation bytecode against EIP-3860. If that is deliberately deferred, track it as a concrete follow-up rather than only a PR-body note.

scripts/V1/deploy.ts
```diff
@@ -236,12 +239,18 @@ export async function deploy(
             ]
         }
     );
+    // The proxy artifact only carries the selectors the proxy implements itself;
+    // the rest are routed to facets, so bind the diamond's full surface here.
+    return {
+        address,
+        contract: StateChannelManagerInterface__factory.connect(address, signer)
```
lukastanisic99, line 246: **Human note**

Also bringing this to your attention, even though you said error handling will follow in a new PR;

**AI-generated review**

🟠 **[FO2] The replacement public binding has lost the proxy's custom-error ABI.** `StateChannelManagerInterface` declares functions and events but no errors, while the old proxy artifact exposed twelve custom errors. Returning this factory means callers can no longer use `contract.interface.parseError`, `getError`, or `decodeErrorResult` for failures that were previously part of the binding. Internal code that uses a separate global error list does not restore that public compatibility.

> **Fix FO2-FIX**
>
> Bind deployments with a merged diamond ABI that includes interface functions/events and all proxy/routed facet errors, or generate one dedicated full ABI artifact. Test that every old proxy function, event, and error remains present, together with newly routed facet errors.

test/fixtures/ProxySelectorRoutingFixture.ts
```diff
@@ -0,0 +1,183 @@
...
+        if (reason === undefined) continue;
+        expect(
+            await diamond.facetAddressForSelector(fragment.selector),
+            `${spec.facetName}.${fragment.format("sighash")} (${reason})`
+        ).to.equal(consumerFacetAddress);
+    }
+}
+
+/** Selectors defined by more than one facet ABI, as `selector -> facet names`. */
+export function facetSelectorCollisions(): Map<string, string[]> {
```
lukastanisic99, line 241: **Human note**

The current tests do not catch a selector collision between the proxy and a routed facet, which can leave a facet operation unreachable while the suite still passes.

**AI-generated review**

🟠 **[TO1] The advertised ABI/selector drift catcher leaves the main compatibility boundaries unchecked.** The current suite checks facet-versus-facet selector collisions and individual routing, but it does not prove that the interface callable set equals proxy-owned plus routed functions, compare mutability, or reject a routed selector that collides with a proxy-owned function. In that last case the proxy body wins over fallback dispatch while `facetAddressForSelector` still reports a facet, so the existing tests can all pass with an unreachable routed operation.

> **Fix TO1-FIX**
>
> Add one exhaustive test over the compiled ABIs: exact callable-set equality in both directions, exact mutability equality, and zero collisions between proxy-owned and routed facet selectors. Include the consumer ABI if it is also required to stay collision-free.

docs/spec/verification/open-questions.md
```diff
@@ -29,3 +30,33 @@ serialized output state and therefore the successor `forkId`; the empty-timeout
...
+
+1. **Rename the file to `StateChannelManagerProxyOpen.t.sol`** (a repository change outside this
+   tree). It restores the directory's convention, makes the file discoverable with no tooling
+   change, and then requires a verification report with a row per declaration. Preferred by the
+   surrounding convention, but it is a code change an agent must not make unilaterally.
+2. **Widen `TEST_FILE_RE` to accept `*.test.sol`.** No repository change, and it is strictly
+   stricter (it adds an obligation rather than hiding one), but it blesses two spellings for the
+   same thing and leaves the naming inconsistency in place.
+
+Blocking effect: until it is resolved, `yarn spec:impact` reports this file as an unmapped changed
```
lukastanisic99: **Human note**

Yes, let the agent update the docs to account for change;

**AI-generated review**

🟠 **[DO1] The required spec-impact audit is knowingly left blocked.** `yarn spec:impact --base origin/refactor/facet-size` exits 1 on three changed paths: the moved `EventSyncService.ts`, the moved `ValidationService.ts`, and `StateChannelManagerProxyOpen.test.sol`. The two source reports remain at their old directory paths. The Solidity test does execute under Forge, but specification discovery accepts `.t.sol`, not `.test.sol`, so its evidence cannot be inventoried or assigned. Recording the blocker as an open question does not satisfy the repository's traceability requirement.

> **Fix DO1-FIX**
>
> Move and refresh the two source reports and their directory inventories. Rename the Foundry test to the repository's `.t.sol` convention and add its verification report/mappings, or update discovery to support `.test.sol`. The base impact command must finish with no unmapped changed files.

contracts/V1/StateChannelDiamondProxy/StateSnapshotFacet.sol
```diff
@@ -119,7 +119,7 @@ contract StateSnapshotFacet is StateChannelCommon {
         StateSnapshot[] memory milestoneSnapshots,
         StateSnapshot memory thresholdStateSnapshot
     ) internal returns (bool) {
-        bool isValid = StateChannelManagerProxy(address(this))
+        bool isValid = StateChannelManagerInterface(address(this))
```
lukastanisic99, line 122: Just a few quick thoughts:
Maybe you'll need to use facets in some place directly and do the delegatecall to them directly instead of going through the root proxy (StateChannelManager); 

The root proxy exposes the public interface (selectors), but in some cases you might end up using other facet functions that are not intended to be called publicly. (e.g. imagine one facet exposes a `withdraw` function which other facets can use, if you added that selector to the root proxy, anyone could call `withdraw` directly which you probably don't want, but you want to reuse the code and have protected calls still be able to use it under the same delegatecall); 

Why I'm saying this is that currently some functions have an `onlySelf` modifier, that only allows the StateChannelManager to call them directly, even though they are part of the public interface, this works for the stateChannelManager, but if you want similar behavior for `facet-to-facet` calls you'd probably have to do the above

Image

contracts/V1/StateChannelManagerInterface.sol
sh3ll3x3c: Fixed in 4c07c9e4. All ten declarations now match the routed facet exactly: eight went nonpayable -> view/pure (`reduce`, `hasInvalidTimestamp`, `hasStateProofHeaderMismatch`, `isDisputeInboundHashValid`, `isCorrectLatestState`, `areSignedBlocksLinkedAndVerified`, `isInvalidBlockStructureInStateProof`, `findFirstInvalidBlockStructureInStateProof`) and two view -> pure (`isGenesisSnapshotWithoutTimeCheck`, `isSnapshotNewer`). Four param/return-name drifts were also synced. Every TS call site of the newly-view functions already used `.staticCall`, so runtime behavior is unchanged. The reconciliation test now compares full signature and stateMutability for every declaration, not just selectors.

test/fixtures/ProxySelectorRoutingFixture.ts
sh3ll3x3c: Fixed in 4c07c9e4. Five new static tests in `SelectorRouting.test.ts` over the compiled ABIs: (1) no routed facet selector is shadowed by a proxy-owned function (the unreachable-routed-operation case); (2) the interface declares every proxy-owned + routed function; (3) the interface declares nothing that is neither implemented nor routed (minus the documented fixture exclusions); (4) exact stateMutability equality per declaration; (5) full-signature equality per declaration. Before the fix these reproduced exactly the ten mutability mismatches from FO1; zero after.

docs/spec/verification/open-questions.md
sh3ll3x3c: Fixed in da0907f8 + 35c86a7a. The two source reports moved to their relocated `stateManager/eventSync/` and `stateManager/ingest/` paths with inventories and inbound links updated. `StateChannelManagerProxyOpen.test.sol` is renamed to `.t.sol` (forge still discovers and passes it), with a new verification report and permutation mapping. The open question is resolved in favor of the rename. `yarn spec:impact`, `--staged`, and `--base origin/refactor/facet-size` all exit 0 with no unmapped changed files; inverse reports regenerated via `yarn spec:refresh`. The touched maintained docs are pending your reapproval in `generated/pending-review.md`.

scripts/V1/deploy.ts
sh3ll3x3c: Acknowledged — deferring this to the error-handling follow-up PR as discussed. The merged-ABI binding (interface functions/events + proxy/routed facet errors) will land there together with a test that every old proxy error stays parseable.

contracts/V1/StateChannelDiamondProxy/StateChannelManagerProxy.sol
sh3ll3x3c: Keeping the `_facetForSelector` chain in this PR per your note. Will look at `PokerDiamond.sol` (EIP-2535 selector->facet storage map populated at deploy) and its multi-layout storage approach for a follow-up — the fixture/tests added here already enforce the selector surface, so swapping the dispatch to an O(1) map later is contained.

contracts/V1/StateChannelDiamondProxy/StateSnapshotFacet.sol
sh3ll3x3c: Good point, noted. That matches the current shape: a facet function meant only for facet-to-facet reuse gets no routing entry (with a reasoned exclusion in `ProxySelectorRoutingFixture.ts`), and other facets reach it by delegatecall to the facet address directly — so it never becomes publicly callable, unlike routing it and guarding with `onlySelf`. When we need protected facet-to-facet calls beyond what `onlySelf` covers, direct facet delegatecall is the pattern we will use.

PR #464, opened 2026-08-27
V2 of #457

lukastanisic99:


src/P2PManager.ts
```diff
@@ -138,243 +106,52 @@ class P2PManager<TCustomRpc extends MainRpcService = MainRpcService>
...
-     * trigger. The announcement is only a prompt to look again; what
-     * decides is the on-chain union read below.
-     */
-    public async reevaluatePendingChannelMembership(): Promise<void> {
-        const stateManager = this.stateManager;
-        if (stateManager.isDisposed) return;
-        if (this.pendingChannelMembershipTransports.size === 0) return;
+        const status = stateManager.status;
+        const isChannelOpened = status === Status.OPENED;
+        this.addConnection(transport);
```
momoshell, line 189: One thing I think may have been lost in the rework.

`addConnection(transport)` here now runs for every completed handshake, and `isChannelConversationMember` is gone — so there's no longer a point where we decide whether a peer belongs in the conversation.

To be fair, this isn't a regression against `dispute`; promotion is unconditional there too (`InitHandshakeService.ts:427`). But it was the substance of #457 — the commits `promote only peers in the channel conversation` and `gate promotion on channel participation, not dispute eligibility` added exactly this gate, and V2 removes it. If V2 is meant to supersede #457, that improvement goes with it.

Worth flagging that this set is load-bearing beyond broadcasts: `IsForkDisputedService.ts:56` snapshots `getConnectedPeers()` and waits for each to acknowledge a disputed fork within `2 × agreementTime`, blacklisting non-responders — so promoted strangers add round-trips on the dispute path.

**Update after reviewing #468/#469:** #468 adds `Status.DISCOVERING` and a branch that keeps lobby transports out of the connection set — that addresses the lobby case directly, and it is the design the #422 review asked for. What remains is narrower: promotion is still unconditional for every *other* status, and `isChannelConversationMember` is absent from the whole stack. So a peer handshaking on the channel topic while we are `OPENED` or `PARTICIPATING` is still promoted with no check.

Reproduction if useful: the deleted test `registers a non-participant, non-spectator peer without promoting it into openConnections` passes on `feat/handshake-promotion-foundation` and fails here, unchanged. Happy to be wrong if there's a downstream gate I've missed.

test/rpc/initHandshake/InitHandshakeLifecycle.test.ts
```diff
@@ -1,420 +1,64 @@
...
-
-        expect(
-            result.profileRegistered,
-            "the peer's identity should still be known to ProfileManager"
-        ).to.equal(true);
-        expect(
-            result.isOpenConnection,
-            "a non-participant that never established a spectate relationship must not receive channel traffic"
-        ).to.equal(false);
+        expect(result.connected).to.equal(true);
```
momoshell, line 20: Related to the above, and worth catching early: these five tests pin the current behaviour as expected.

`result.connected` comes from `P2PManagerFixture.ts:95` → `getConnectedPeerAddresses()` → `openConnections`, and lines 21/30/39/50/59 all assert `true`, including from `NOT_OPENED`. So if a membership check comes back, these five go red and will look like a regression rather than a fix.

The status-routing coverage itself is genuinely useful and worth keeping — it's the admission question that's no longer covered. The five tests that did cover it went with this commit:

- `registers the profile without promoting the transport when nothing subscribes to handshakeCompleted`
- `registers a non-participant, non-spectator peer without promoting it into openConnections`
- `promotes an accepted spectator into openConnections even though it is not a dispute participant`
- `does not retroactively promote a stranger once we ourselves become a participant`
- `promotes a deferred peer once that peer joins, even though our own status never changes`

Would it make sense to keep the routing tests and bring one or two of these back alongside them?

**Update after reviewing #468/#469:** this file is byte-identical across all three PRs, so the five `connected === true` assertions still stand. Worth noting that #468's new `DISCOVERING` no-promotion branch has no test of its own either — nothing anywhere asserts that a `DISCOVERING` peer stays out of `openConnections`, and `E2E-LobbyMatching.test.ts` never inspects the connection set. So the fix to the comment above is currently unprotected by the same gap.

src/ProfileManager.ts
```diff
@@ -47,6 +41,41 @@ class ProfileManager {
...
+    public authenticateTransport(
+        transport: ATransport,
+        evmAddress: Address
+    ): PeerProfile | undefined {
+        const normalizedAddress = getChecksumAddress(evmAddress);
+        const existingProfile =
+            this.mapEvmAddressToProfile.get(normalizedAddress);
+        if (existingProfile) {
+            if (existingProfile.isBlackListed) {
+                this.blacklistPeer(transport);
```
momoshell, line 64: Two small things here, both non-blocking.

`blacklistPeer(transport)` resolves the temporary per-transport profile created by `registerTransport`, not the identity profile matched on line 50. That profile is already blacklisted so the outcome is right, but it's the DHT ban doing the work rather than this call as written.

And `close(true)` on line 54 sets `isExpected`, which suppresses `onDisconnection` (`ATransport.ts:57-62`). Since this path is a rejection rather than an upgrade — would plain `close()` be more accurate, so the app learns the peer was dropped?

PR #465, opened 2026-08-28
refactor(contracts): make proxy routing upgrade-ready

lukastanisic99:
## TLDR;
- replace linear dispatch with a constructor-built selector map
- add canonical full-error manager bindings
- enforce artifact size limits and update traceability

## Summary

- Replace the proxy’s 56-branch selector dispatch with a constructor-built selector map.
- Add one canonical manager binding with all reachable custom errors.
- Enforce EIP-170 runtime and EIP-3860 initcode limits for compiled artifacts and artifact-backed deployments.
- Update tests, deployment fixtures, browser paths, and specification traceability.

### Proxy routing

- Register all routed facet selectors during proxy construction using compiler-derived `Facet.function.selector` values.
- Resolve routed calls through one storage lookup.
- Reject duplicate selector registration with `ErrorDuplicateSelectorRegistration(bytes4)`.
- Preserve the existing consumer-facet fallback for unknown selectors.
- Keep the route table constructor-only. Governance-controlled route updates remain future work.

The routing benchmark was deliberately removed. We are choosing the selector map for constant lookup cost and future upgradeability regardless of gas break-even.

### Manager ABI

- Add `connectStateChannelManager` as the canonical binding for deployed manager addresses.
- Keep its function and event surface equal to `StateChannelManagerInterface`.
- Merge the generated reachable-error ABI into the binding.
- Include errors from `StateProofFacet` and `UtilityFacet`, including the ECDSA errors.
- Preserve the full ABI when it crosses the runtime worker port.
- Reuse the same manager ABI in `LocalDiamond`.

### Shared ownership

- Add one production-owned routed-facet inventory used by deployment and selector-routing fixtures.
- Add shared ABI merge helpers.
- Remove direct `StateChannelManagerInterface__factory.connect` usage for deployed manager addresses.
- Avoid exporting routed facet artifacts through the shared browser utility barrel.

### Contract-size checks

- Add shared EIP-170 and EIP-3860 size policy helpers.
- Scan every compiled production artifact.
- Reject oversized runtime or full deployment initcode in `deployArtifact` before submitting a transaction.
- Keep `LocalDiamond` as an explicit local-only exemption.
- Return structured size errors containing the contract name, measured size, limit, and excess bytes.

## Testing

- `yarn compile` — pass
- Foundry proxy registration tests — 2/2 pass
- Focused routing, ABI, error, runtime, size, and deployment tests — 64/64 pass
- Review follow-up unit tests — 20/20 pass
- Node and browser TypeScript checks — pass
- Browser build and WebRTC suite — pass
- `yarn spec:ids:check` — 5,615 definitions, 0 issues
- `yarn spec:impact --base origin/refactor/facet-size` — pass
- Two byte-identical `yarn spec:refresh` runs — pass
- `git diff --check` — pass

A pre-review distributed run passed 1,507/1,507. Later distributed runs recorded three unrelated timing failures in unchanged E2E cases:

- `run-482`: malicious-snapshot wait and dispute cleanup timeout
- `run-484`: last-milestone fork-change barrier timeout

All three passed on focused rerun (`run-483` and `run-485`). No source or fixture changes were made for them.


test/browser/run-p2p-webrtc-e2e.mjs
```diff
@@ -63,6 +63,10 @@ const platformAliases = {
         projectRoot,
         "src/utils/browser/DeployUtils.ts"
     ),
+    "@platform/evmJumpdestCache": path.join(
```
sh3ll3x3c, line 67: **AI-generated review (Claude, from 1-review.md)**

🔴 **[AR1] The spec-impact audit fails against this PR's real base.** `yarn spec:impact --base 35c86a7a` (the actual merge-base with `refactor/proxy-fallback-dispatch`) exits 1: `Unmapped changed files — review blocked: test/browser/run-p2p-webrtc-e2e.mjs`. This `@platform/evmJumpdestCache` alias is an unrelated drive-by fix (pre-existing gap from 08c97885) with no mapped ID or exclusion. The PR body's passing check used `--base origin/refactor/facet-size`, which spans #463's already-audited changes and masked this file.

> **Fix AR1-FIX**
>
> Map the file (or record a justified exclusion) — or split the alias fix into its own commit/PR with its own spec mapping. Acceptance: `yarn spec:impact --base <merge-base with refactor/proxy-fallback-dispatch>` exits 0 with no unmapped section.

docs/spec/verification/tests/test/V1/DiamondProxy/StateChannelManager/SelectorRouting.test.ts.md
```diff
@@ -58,6 +61,7 @@ report but are kept here.
...
 | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| [`StateChannelManagerProxy selector routing > uses the canonical routed-facet inventory for every deployed facet`](../../../../../../../../test/V1/DiamondProxy/StateChannelManager/SelectorRouting.test.ts#L37) (line 37)                   | [`UNIT-TEST-MANAGER-PROXY-2-KJRMB8.P30`](../../../../../../implementation/source/contracts/V1/StateChannelDiamondProxy/StateChannelManagerProxy.sol.md#unit-test-manager-proxy-2-kjrmb8.p30) |
```
sh3ll3x3c, line 64: **AI-generated review (Claude, from 1-review.md)**

🔴 **[AR2] Inserting tests shifted line anchors and silently killed 25 previously-resolved evidence links.** The generators bind a Covers row by exact line. This report still anchors the pre-existing rows at L36…L198, but the new inventory test at the top of the describe pushed the real `it`s to L43…L205 (e.g. "routes every dispute manager selector" is now L43, the row says L36). The five parity rows from the base branch (KJRMB8.P24–P28) are among the dead links. Same rot in `LocalDiamondBinding.test.ts.md` (5 rows), `UniversalDeployment.test.ts.md` (L70 → actual 110), and `RuntimeChainContext.test.ts.md` (2 rows). The loss is masked because the new tests raised the headline count (base `generated/traceability.md` resolved 19/19 rows here; HEAD resolves only the new L37 row).

> **Fix AR2-FIX**
>
> Re-line every declaration row in the four reports to the actual `it(...)` lines, then `yarn spec:refresh` and confirm `generated/traceability.md` resolves all rows again.

docs/spec/verification/tests/test/unit/ContractSize.test.ts.md
```diff
@@ -0,0 +1,30 @@
...
+| ---------------------------------------------------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| [`contract size > keeps every compiled production artifact within both deployment limits`](../../../../../../test/unit/ContractSize.test.ts#L31) (line 31) | [`UNIT-TEST-CONTRACT-SIZE-1-MX797V.P1`](../../../../implementation/source/src/utils/contractSize.ts.md#unit-test-contract-size-1-mx797v.p1), [`REQ-CONTRACT-SIZE-1-881Q6E.T1.P1`](../../../../specification/enforcement/contracts.md#req-contract-size-1-881q6e.t1.p1), [`REQ-CON-2-CBVFV9.T1.P1`](../../../../implementation/views/architecture/contracts/architecture.md#req-con-2-cbvfv9.t1.p1) |
```
sh3ll3x3c, line 22: **AI-generated review (Claude, from 1-review.md)**

🔴 **[AR3] The committed `generated/*` reports are stale — a refresh at this commit rewrites them and drops evidence.** `generate-verification-coverage.js --check`, `generate-traceability-views.js --check`, and `generate-audit-summary.js --check` all fail at HEAD ("stale generated report"). Regenerating in a scratch copy drops all 7 `ContractSize.test.ts` rows (evidenced 814 → 795; `REQ-CONTRACT-SIZE-1-881Q6E` 10/10 → 3/10) because the line anchors in this report are off by 2 from the actual test lines (68/77/92/112/129/143/162 vs actual 70/79/94/114/131/145/164). The PR body's "two byte-identical `yarn spec:refresh` runs" does not hold at ad2e0ee4.

> **Fix AR3-FIX**
>
> Correct this report's line anchors (together with AR2's re-lining), run `yarn spec:refresh`, commit the regenerated reports, and re-run the seven `--check` commands to prove they pass at the committed SHA.

contracts/V1/StateChannelDiamondProxy/StateChannelManagerProxy.sol
```diff
@@ -277,81 +334,16 @@ contract StateChannelManagerProxy is StateChannelCommon {
...
     // ********** private/internal functions **********
 
-    /// @dev Selector routing table. Selectors are derived by the compiler from the
-    ///     facet function types, so a signature change updates the routing with it -
-    ///     nothing here is hand-hashed. The state-changing facets come first so
-    ///     on-chain calls pay the fewest comparisons; the utility views are read
-    ///     off-chain. Unknown selectors keep the historical behaviour of falling
-    ///     through to the consumer facet.
+    /// @dev Registers a compiler-derived selector once during construction.
+    function _registerRoute(bytes4 selector, address facet) internal {
```
sh3ll3x3c, line 338: **AI-generated review (Claude, from 1-review.md)**

🟠 **[FO1] A registered route to an address with no code dispatches as a silent empty success.** `_registerRoute` accepts `address(0)` or any codeless address, and `_delegatecall` (GeneralUtils.sol:6) has no code-existence check — delegatecall to a codeless target *succeeds* with empty returndata, so a misdeployed facet turns e.g. `uploadDispute` into a no-op that looks like success on-chain. The hole predates this PR (the facet address vars were equally unvalidated), but the constructor registration is the natural choke point to close it — and the new Foundry test `test_facetAddressForSelector_configuredZeroFacet_returnsZero` currently *blesses* the zero-address route as configured.

> **Fix FO1-FIX**
>
> In `_registerRoute`, revert (e.g. `ErrorRouteTargetHasNoCode(bytes4,address)`) when `facet.code.length == 0`. Replace the zero-facet Foundry test with one asserting construction reverts for a codeless route, plus one asserting a dispatched call through a real map entry executes on the facet.

src/evm/p2pRuntime/P2pRuntimeClient.ts
```diff
@@ -135,11 +133,10 @@ class P2pRuntimeClient<T = ethers.Contract> {
             options.provider,
             options.signerAddress.toString()
         );
-        this.stateChannelManagerContract =
-            StateChannelManagerInterface__factory.connect(
-                options.scm.address,
-                this.chainSigner
-            );
+        this.stateChannelManagerContract = connectStateChannelManager(
```
sh3ll3x3c, line 136: **AI-generated review (Claude, from 1-review.md)**

🟠 **[FO2] "Full ABI preserved across the runtime port" is only true for the client half.** The client now binds via `connectStateChannelManager`, but `P2pRuntimeHost` (untouched by this PR, L274) still rebuilds the SCM from the app-shipped `payload.scm.abiJson` — whatever the consumer happened to bind. A consumer not using the new binding gets a host-side contract without the error union. Meanwhile the client now *ignores* `scm.abiJson`, yet `SerializedContract.abiJson` (types.ts:40) is still required and documented as the binding source — the two halves of one port now disagree about what the field means. The `RuntimeChainContext` tests feed the merged ABI themselves, so they can't catch this.

> **Fix FO2-FIX**
>
> Have the host build the SCM with `connectStateChannelManager(payload.scm.address, chainSigner)` too (dropping the `as unknown as` cast there), then either delete `scm.abiJson` or re-document it as decoding metadata only. Acceptance: a `RuntimeChainContext` test that ships a minimal interface-only `abiJson` and still decodes a facet error on the host side.

test/V1/UniversalDeployment.test.ts
```diff
@@ -192,6 +232,65 @@ describe("Universal Deployment", () => {
...
+                    encodedOpenChannel: firstSignature.encoded,
+                    signatures: [invalidSignature, validSignature]
+                });
+            } catch (error) {
+                facetFailure = error;
+            }
+            const facetError = diamondContract.interface.parseError(
+                facetFailure.data
+            );
+            expect(facetError?.name).to.equal("ECDSAInvalidSignatureLength");
```
sh3ll3x3c, line 293: **AI-generated review (Claude, from 1-review.md)**

🟠 **[TO1] No test decodes a revert that actually originates inside a routed, delegatecalled facet.** `StateChannelManagerBinding.test.ts` proves the merged ABI *parses* every error fragment, and this `ECDSAInvalidSignatureLength` decode comes from a *direct* call to an unrouted `UtilityFacet` function. The chain this PR exists for — fallback → selector map → delegatecall → revert bubbling → `parseError` on the merged binding — is never exercised end to end.

> **Fix TO1-FIX**
>
> Add one unit test that calls a routed function through `connectStateChannelManager` at the proxy address with inputs that make the *facet* revert with a facet-declared error, and assert the typed error name and arguments decode. One test covers the routed-revert integration contract.

src/utils/contractSize.ts
```diff
@@ -0,0 +1,106 @@
...
+
+type ArtifactCandidate = {
+    contractName?: unknown;
+    bytecode?: unknown;
+    deployedBytecode?: unknown;
+};
+
+export type ContractSizeKind = "runtime" | "initcode";
+
+export class InvalidContractArtifactError extends Error {
```
sh3ll3x3c, line 18: **AI-generated review (Claude, from 1-review.md)**

🟠 **[MO1] The size gate is now unconditional public API with uncatchable error types.** Every consumer `deployArtifact` call hard-fails on an oversized artifact — including deliberate dev deployments on `allowUnlimitedContractSize` chains — and `ContractSizeLimitError` / `InvalidContractArtifactError` are exported from this file but reach neither `src/utils/index.ts` nor `src/index.ts`, so package consumers cannot `instanceof`-catch the structured error the PR body advertises.

> **Fix MO1-FIX**
>
> Re-export both error classes (and the two limit constants) through the utils barrel and `src/index.ts`; add a binding test importing them from the package root.

scripts/V1/deploy.ts
```diff
@@ -70,12 +54,22 @@ export async function deployArtifact<T>(
         txOverrides?: ethers.TransactionRequest;
     }
 ): Promise<{ address: string; contract: T }> {
+    assertArtifactRuntimeSize(artifact);
```
sh3ll3x3c, line 57: **AI-generated review (Claude, from 1-review.md)**

🟠 **[OO1] Open question: should `deployArtifact` offer a size-gate opt-out for dev chains, or is hard-fail everywhere the intended policy?** Production chains reject oversized code anyway, so the gate mainly bites hardhat `allowUnlimitedContractSize` workflows (oversized consumer prototypes, instrumentation builds), which now have no escape hatch except bypassing `deployArtifact` entirely. Recommended: keep the gate default-on, add an explicit `options.skipSizeChecks` that logs loudly — with MO1's exported types the failure is then catchable either way. If unconditional is intended, document the constraint so consumers aren't surprised.

sh3ll3x3c (approved):

PR #466, opened 2026-08-31
compose the worker protocols as services over the rpc port router

daiagi:
the sdk host, the vm worker and log collection no longer speak hand-rolled request unions over the worker ports. each link is a `MessagePortTransport` (trusted) under a `PortRpcRouter`, and each side's operations are ordinary `*Service` + `*RpcMethods` on a root, typed end to end -> a new host operation is a method, not a message type + switch + client stub.

- `ARpcRouter` extracts the request/response core (ids, pending map, timeouts, reply-before-request, dispatch) out of `P2PManager`, which now supplies only peer policy hooks. peer behavior unchanged (55 rpc-layer tests green).
- `PortRpcRouter` is the same core for worker links: no guards, structured-clone frames, unknown method answered not disconnected, `holdInbound` until the root can answer.
- sdk host root: `lifecycle` / `p2pSigner` / `chainSigner` / `deploySigner` / `hostRpc` / `logControl`; client root: `runtimeEvents` / `logControl`. readiness is the `deployComplete` reply, no `ready` frame. the WebRTC bridge port rides the bootstrap message.
- vm worker root: `contractExecutor` / `logControl`; readiness is the `init` reply.
- log collection is `logControl.flush(reason)` over `WorkerLinks`; the reply is the ack, `LogFlushBus` keeps no ids.
- rich `SerializedError` only over trusted transports; peers still get the message only.
- the two #461 workarounds GO3 was meant to delete are gone: the bus has no round scheduling (every request is its own fan-out, the uploader's depth-one queue is the only coalescing -> FR1's per-source queue has nothing to fix), and the vm's log identity rides in `init` (link and init in any order -> FO3's port ordering is gone).

- the browser vm worker boots over the router too (node-globals + Buffer shim at the entry, no `setImmediate` in shared files); caught by the new browser crash-log smoke in #461.

deleted: `p2pRuntime/worker/protocol.ts`, `contractExecutor/worker/protocol.ts`, `ContractExecutorWorkerHostCore.ts`.

tests: `PortRpcRouter` (11), `MessagePortTransport` (4), `P2pRuntimeClient` (4, fake host speaks the real envelope), `LogFlushBus` fixture over routers; e2e `p2pSetup runtime modes` + `crash log upload` 21/21; full e2e gate run.

spec: source reports for the 28 new files, views rewritten (runtime-and-concurrency §1–4, §7 done), `INTEGRATION-TEST-RPC-7`, verification reports; `spec:refresh` green. `spec:impact` flags the 3 deleted files as unmapped -> the tool has no exclusion for deletions (same as #461).


daiagi:
@lukastanisic99  conflicts resolved and the stack is green, one unrelated failure on 461. but as an atomic stack, green

src/P2PManager.ts
```diff
@@ -284,136 +264,87 @@ class P2PManager<TCustomRpc extends MainRpcService = MainRpcService>
...
-                transport,
-                timeout
-            });
+    protected scheduleTimeout(
+        fn: () => void,
+        ms: number,
+        label: string
+    ): unknown {
+        return this.stateManager.timeoutManager.scheduleTask(fn, ms, label);
+    }
```
lukastanisic99, line 275: Why is this wrapper function added? Just call timeoutManager directly where needed

src/P2PManager.ts
```diff
@@ -284,136 +264,87 @@ class P2PManager<TCustomRpc extends MainRpcService = MainRpcService>
...
-            try {
-                transport.send({ ...rpc, requestId });
-            } catch (e) {
-                if (this.pendingRpcRequests.delete(requestId)) {
-                    this.stateManager.timeoutManager.cancelTask(timeout);
-                    reject(e instanceof Error ? e : new Error(String(e)));
-                }
-            }
-        });
+    protected cancelTimeout(handle: unknown): void {
```
lukastanisic99, line 277: -||-

src/P2PManager.ts
```diff
@@ -284,136 +264,87 @@ class P2PManager<TCustomRpc extends MainRpcService = MainRpcService>
...
-        }
-        this.pendingRpcRequests.delete(response.requestId);
-        this.stateManager.timeoutManager.cancelTask(pending.timeout);
-        if (response.ok) {
-            pending.resolve(response.result);
-        } else {
-            pending.reject(
-                new Error(response.error ?? "RPC request failed on the peer")
+    // `agreementTime` is in seconds; the RPC timeout is in milliseconds.
+    protected defaultRequestTimeoutMs(): number {
```
lukastanisic99, line 282: this is not used here and I see it also being declared in 3 other files;

src/P2PManager.ts
```diff
@@ -284,136 +264,87 @@ class P2PManager<TCustomRpc extends MainRpcService = MainRpcService>
...
+        return this.stateManager.timeConfig.agreementTime * 1000;
+    }
+
+    // Only the peer we sent the request to may settle it. Compare by peer
+    // identity (not transport object) so a transport upgrade for the same
+    // peer (e.g. HOLEPUNCH -> WEBRTC) still settles the pending request.
+    protected isResponseFromRequestee(
```
lukastanisic99, line 289: Again not used and ARpcRouter defines its own - not sure why the LLM did all of this - for some reason it loves creating bloat and just 'adding stuff' - The LLM needs to read this https://dokumen.pub/a-philosophy-of-software-design-2nd-edition-2nbsped-173210221x-9781732102217.html

src/P2PManager.ts
```diff
@@ -284,136 +264,87 @@ class P2PManager<TCustomRpc extends MainRpcService = MainRpcService>
...
         }
+        this.disconnectConnection(transport);
     }
 
-    private rejectPendingRpcRequestsForTransport(
+    /** a peer whose frame or handler failed is disconnected */
+    /** a frame the router refused is a protocol violation: the peer is dropped
+     *  and banned, as it was before the router. one of our own handlers
+     *  failing is not the peer's doing, so that only drops the line. */
+    public onServiceFailure(
```
lukastanisic99, line 323: If we now have a `router` and we're passing everything to the router, there is no reason to have this here - again this one is not used, since the router defines its own uses its own; ARpcService uses the router implementation of this; RortRpcRouter also defines this

lukastanisic99:
I'm thinking about this one - I'll comment later

src/P2PManager.ts
daiagi: **Human:** not a wrapper - `ARpcRouter` requires every router to supply these. **Agent:** `protected abstract` on [ARpcRouter.ts L100-L105](https://github.com/peer3to/state-channels-plus/blob/74e3814e51f70085637c51c5ffd7859bf3a0fa4f/src/rpc/ARpcRouter.ts#L100-L105), called from the shared request loop at [L169](https://github.com/peer3to/state-channels-plus/blob/74e3814e51f70085637c51c5ffd7859bf3a0fa4f/src/rpc/ARpcRouter.ts#L169) and [L196-L232](https://github.com/peer3to/state-channels-plus/blob/74e3814e51f70085637c51c5ffd7859bf3a0fa4f/src/rpc/ARpcRouter.ts#L196). the router is shared with `PortRpcRouter` (plain `setTimeout`, [L160-L166](https://github.com/peer3to/state-channels-plus/blob/74e3814e51f70085637c51c5ffd7859bf3a0fa4f/src/rpc/PortRpcRouter.ts#L160-L166)) and can't see `stateManager.timeoutManager`, so the peer side supplies them. the `Parameters<...>` cast in `cancelTimeout` was bloat -> `handle as NodeJS.Timeout`.

src/P2PManager.ts
daiagi: **Agent:** same as `scheduleTimeout` above: abstract on `ARpcRouter`, cancelled from its request loop at L196/L211/L232. the `Parameters<...>` cast is gone -> `handle as NodeJS.Timeout` (74fbfb6b).

src/P2PManager.ts
daiagi: **Agent:** abstract at [ARpcRouter.ts L107](https://github.com/peer3to/state-channels-plus/blob/74e3814e51f70085637c51c5ffd7859bf3a0fa4f/src/rpc/ARpcRouter.ts#L107), read at [L162](https://github.com/peer3to/state-channels-plus/blob/74e3814e51f70085637c51c5ffd7859bf3a0fa4f/src/rpc/ARpcRouter.ts#L162) for every request without an explicit timeout. the three files are the abstract + the two implementations (port: `options.defaultTimeoutMs ?? null`, peer: `agreementTime * 1000`).

src/P2PManager.ts
daiagi: **Agent:** the base default is `expected === actual` ([ARpcRouter.ts L129-L135](https://github.com/peer3to/state-channels-plus/blob/74e3814e51f70085637c51c5ffd7859bf3a0fa4f/src/rpc/ARpcRouter.ts#L129-L135)), called at [L206](https://github.com/peer3to/state-channels-plus/blob/74e3814e51f70085637c51c5ffd7859bf3a0fa4f/src/rpc/ARpcRouter.ts#L206). the peer override compares by peer identity so a HOLEPUNCH -> WEBRTC upgrade still settles the request - that's the pre-router behaviour, and the comment above it says so.

src/P2PManager.ts
daiagi: **Agent:** abstract at [ARpcRouter.ts L95](https://github.com/peer3to/state-channels-plus/blob/74e3814e51f70085637c51c5ffd7859bf3a0fa4f/src/rpc/ARpcRouter.ts#L95), called at [L264-L348](https://github.com/peer3to/state-channels-plus/blob/74e3814e51f70085637c51c5ffd7859bf3a0fa4f/src/rpc/ARpcRouter.ts#L264-L348); `ARpcService` calls `this.router.onServiceFailure`, which *is* this method when the router is a `P2PManager`. port router logs, peer router bans on a refused frame / drops on our own handler failing. the stale second doc comment above it is removed.

daiagi:
superseded by #481, same goal on the one-router model. comparison in the new PR body; the review threads here stay as the record.

PR #468, opened 2026-09-01
Discovery v1

lukastanisic99:
## Summary

Adds the first complete lobby-discovery flow and an authoritative on-chain registry of open channels.

Applications can join a caller-supplied lobby topic, match one authenticated peer, negotiate and open a channel, then continue using the selected transport after leaving the lobby topic.

### Open-channel registry

- Tracks open channel IDs in the state-channel manager.
- Adds paginated `getOpenChannelCount()` and `getOpenChannelIds(offset, limit)` reads.
- Appends channels only after a successful open.
- Removes fully closed channels with swap-and-pop index repair.
- Supports closing and later reopening the same channel ID.
- Keeps event indexers as the normal discovery source, with the registry as an authoritative fallback.

### Lobby matching

- Adds `joinLobby(topic, options)` and `leaveLobby(topic)` to local, client, worker, and host signer paths.
- Uses an application-provided 32-byte rendezvous topic with no default global lobby.
- Matches only authenticated peers on the exact active topic.
- Assigns advertiser and selector roles with monotonic epochs.
- Uses exclusive pick reservations and mutually acknowledged commit challenges.
- Supports optional finite matching timeouts; omitted or `null` matching stays active until success or explicit leave.
- Allows cancellation only before match commitment.
- Retries unsigned negotiation failures through a fresh matching session.

### Negotiation and transport ownership

- Derives a unique channel ID from the authenticated peer pair and committed lobby transcript.
- Keeps matching separate from channel selection and negotiation.
- Promotes only the selected peer into the channel workflow.
- Closes non-selected and abandoned lobby transports.
- Preserves a selected peer across a healthy transport replacement.
- Leaves the lobby topic after opening without closing the transport now owned by the channel.
- Adds deferred admission for authenticated negotiation frames that arrive before local initialization.

### Lifecycle and SDK behavior

- Adds `Status.DISCOVERING` for lobby participation without a selected channel.
- Keeps ordinary channel connection separate from lobby discovery.
- Routes discovery topics through one validated `P2PManager` join/leave boundary.
- Returns `{ channelId, peerAddress }` only after the channel opening is observed.
- Exposes matching timeout and opening amount through serializable worker options.

## Public API

```ts
joinLobby(
    topic,
    {
        amount?: number;
        matchTimeoutMs?: number | null;
    }
): Promise<{ channelId: string; peerAddress: string } | undefined>;

leaveLobby(topic): Promise<boolean>;

### CODE REVIEW

File: contracts/V1/StateChannelDiamondProxy/StateChannelManagerStorage.sol [hunk 1]
```diff
@@ -41,6 +41,10 @@ contract StateChannelManagerStorage {
     /// @dev stateSnapshot Data
     mapping(bytes32 channelId => StateSnapshot) stateSnapshots;
 
+    /// @dev Enumerable set of channels whose snapshot currently exists on-chain.
```
momoshell, line 44: `openChannelIds` and `openChannelIndexPlusOne` are inserted before the existing block-storage fields, which shifts everything after them. `forge inspect StateChannelManagerStorage storage-layout` on this branch vs #464:

```
                           #464   #468
blockCalldataCommitments     18  →  20
disputeData                  19  →  21
disputerThrottle             20  →  22
selectorRoutes               21  →  23
```

`selectorRoutes` is the diamond's facet routing table, so an in-place proxy upgrade would read routing — and dispute data — from the wrong slots.

I couldn't find any deployment artifacts in the repo, so I don't think this breaks anything live today; it's more that it's free to fix now and impossible later. Would appending both variables after `selectorRoutes` work for you?

src/types/flags.ts
```diff
@@ -9,9 +9,16 @@ export enum BlockValidationResult {
 }
 
 export enum Status {
+    /** Active caller-owned lobby topic with no selected channel ID. */
+    DISCOVERING,
```
momoshell, line 13: Adding `DISCOVERING` is exactly what the review asked for — but putting it first renumbers all five existing values.

`Status` is exported publicly (`src/index.ts:72`) and crosses the runtime port via `getChannelStatus` (`P2pRuntimeHost.ts:587` → `ClientP2pSigner.ts:231`), so anything holding a numeric value from a previous version now reads `OPENED` as the old `SYNCED`. I couldn't find internal persistence, so the exposure looks like the public API rather than your own state.

Would placing `DISCOVERING` after `PARTICIPATING`, or pinning explicit numeric values, give you the same thing without the renumber?

src/rpc/MainRpcService.ts
```diff
@@ -24,6 +26,8 @@ class MainRpcService {
     spectateService: SpectateService;
     isForkDisputedService: IsForkDisputedService;
     joinChannelService: JoinChannelService;
+    lobbyMatchingService: LobbyMatchingService;
```
momoshell, line 29: This is really a question rather than a defect, and it may already have been settled off-thread.

The #422 review asked to reuse open-channel negotiation and specifically not to add a separate lobby service and RPC surface, and to keep the interim step dumb and swappable. This branch adds a routable `lobbyMatchingService` with `advertise`/`pick`/`commit`, plus a persistent on-chain open-channel registry in `UtilityFacet.sol:336` that the TypeScript flow doesn't read. There's also no `getLogs` discovery path — `joinLobby` matches a peer and opens a new channel rather than discovering existing ones (`LocalP2pSigner.ts:172`).

Is this a deliberate change of direction that's been agreed, or should the interim step still be the smaller log-based one?

src/P2PManager.ts
```diff
@@ -125,6 +125,29 @@ class P2PManager<TCustomRpc extends MainRpcService = MainRpcService>
 
         const status = stateManager.status;
         const isChannelOpened = status === Status.OPENED;
+        if (status === Status.DISCOVERING) {
```
momoshell: Nice — this is the status routing the #422 review asked for, and it addresses the lobby half of the promotion question I raised on #464.

One gap: I can't find a test that pins it. Nothing asserts a `DISCOVERING` peer stays out of `openConnections`, and `E2E-LobbyMatching.test.ts` never touches the connection set, so deleting this branch would leave the suite green. Worth one test that fails without it?

contracts/V1/StateChannelDiamondProxy/StateChannelManagerStorage.sol
lukastanisic99: Take a look at our poker-contracts and the storage pattern we use there - we'll end up doing a similar one here where the first storage slot is computed as a deterministic hash - so we can mix multiple storage layouts and always target what we want exactly. That + the upgradable pattern where we dynamically map function selectors to facet addresses is all we need to support any upgradability in the future; It's more of a cleanup task so feel free to do it if you want;

src/types/flags.ts
lukastanisic99: My agent said the same, but it doesn't matter we don't have anything in production at the moment, so better to have a natural gradation and all of the 'backwards compatibility` we can ignore for now - before production we can add versioning and make it explicit what's compatible and what not

src/rpc/MainRpcService.ts
lukastanisic99: The lobby service is not really a 'lobby service' it's a matchmaking service that's general and can be reused to quickly match peers over a shared topic (or at least I hope it's quickly since I designed it and didn't use any preexisting algo for it, but the base idea is simple) - the negotiation service requires a match and negotiates the open with a specific peer; Before they were all one service, like this they are modular and can be used and combined with other services + they were too big to be in 1 service - they are simple to understand, but handle many failure/cancelation cases;

src/P2PManager.ts
lukastanisic99: #469 does this better, it doesn't use the status it checks is the matchmaking service in use currently and routes there regardless of status - this way we're not dependent on status, but actual intention of use;

src/rpc/MainRpcService.ts
momoshell: That answers it, thanks.

My comment bundled three things into one, which wasn't helpful. Two are still open and both are smaller:

- `UtilityFacet.sol:336` adds a persistent on-chain open-channel registry that nothing in the TypeScript flow reads. Is that for a consumer landing later, or currently unused state?
- There's no `getLogs` path — `joinLobby` matches a peer and opens a new channel rather than finding existing ones. Is log-based discovery still planned as the interim step, or has matchmaking replaced that idea?

src/P2PManager.ts
momoshell: I also checked whether "regardless of status" could catch a channel peer mid-match — it can't: `joinLobby` refuses when a channel is selected, and the targeted path only reaches the match from `NOT_OPENED`.

One thing does carry over to #469: I still can't find a test asserting a routed peer stays out of `openConnections`, so the branch could be deleted and the suite would stay green.

PR #469, opened 2026-09-01
feat(discovery): support targeted pre-open channel joins

lukastanisic99:
## Summary

Adds a public flow for connecting to one known channel ID before or after it opens.

The caller can independently request opening, synchronization, and membership. The flow reuses the existing lobby matcher, open-channel negotiation, spectator sync, join authorization, and membership transaction owners.

This PR is stacked on #468 and should be reviewed as the `discovery-v1...lobby-opening` diff.

## Public API

```ts
connectToChannel(
    channelId,
    {
        autoOpen?: boolean;
        shouldJoin?: boolean;
        balance?: BalanceStruct;
        timeoutMs?: number | null;
    }
): Promise<boolean>;

cancelConnectToChannel(channelId): Promise<boolean>;

### CODE REVIEW

File: src/evm/signer/LocalP2pSigner.ts [hunk 1]
```diff
@@ -150,56 +157,161 @@ class LocalP2pSigner<TCustomRpc extends MainRpcService = MainRpcService>
...
+        options: ConnectToChannelOptions | null = {}
+    ): Promise<boolean> {
+        options ??= {};
+        const normalizedChannelId = ethers.hexlify(channelId);
+        if (!ethers.isHexString(normalizedChannelId, 32))
+            throw new Error("Channel ID must be exactly 32 bytes");
+        const stateManager = this.p2pManager.stateManager;
+        let openedGenesis = false;
+
+        if (String(stateManager.channelId) !== normalizedChannelId) {
```
momoshell, line 173: Switching channels can skip sync entirely, I think.

`clearChannelId()` resets the id, dispute manager, event sync and listener, but not `status`. `refreshOpenedStatusFromChain()` then declines to move it — its `// Only move to OPENED if we haven't already synced/applied state` guard only fires when status is `NOT_OPENED`.

So going from channel A (`SYNCED`/`PARTICIPATING`) to channel B leaves A's status in place, this method's `NOT_OPENED` branch is skipped, and it returns `true` having never synced B — with A's state still loaded under B's id.

That guard reads correct for a single-channel runtime; targeted joins are what make it reachable. Should `clearChannelId()` reset status too?

src/evm/signer/LocalP2pSigner.ts
```diff
@@ -150,56 +157,161 @@ class LocalP2pSigner<TCustomRpc extends MainRpcService = MainRpcService>
...
+            } else {
+                await stateManager.refreshOpenedStatusFromChain();
+                if (stateManager.status === Status.NOT_OPENED) {
+                    const outcome =
+                        await this.p2pManager.localRpc.openChannelNegotiationService.initMatchedNegotiation(
+                            match,
+                            {
+                                mode: "targeted",
+                                channelId: normalizedChannelId,
+                                balance:
```
momoshell, line 204: `balance: options.balance ?? this.defaultBalance()` feeds the genesis opening regardless of `shouldJoin`.

So a balance passed with `shouldJoin: false` isn't dormant — it becomes the opening balance, and a zero one makes negotiation fail rather than falling back to the default. Should `balance` be ignored unless `shouldJoin` is set?

src/rpc/services/openChannelNegotiation/OpenChannelNegotiationService.ts
```diff
@@ -350,6 +387,9 @@ export default class OpenChannelNegotiationService extends ARpcService<
             decoded,
             this.p2pManager.stateManager.signer
         );
+        if (this.state.attempt !== attempt || attempt.observedOpenHandoff) {
```
momoshell, line 397: Minor, and possibly a deliberate trade.

Between the `observedOpenHandoff` check here and the submit, `onChannelOpened` can land while `classifyObservedOpening()` is still awaiting `getParticipants` — so the guard passes and we submit an opening for a channel that's already open.

The revert is caught and handled (`RaceConditionChannelAlreadyOpen` → `completeObservedTargetOpen` for targeted mode), so nothing breaks; it just costs a reverted transaction's gas. Worth awaiting classification before releasing signing, or is the revert an acceptable price for the simpler flow?

src/rpc/services/lobbyMatching/LobbyMatchingService.ts
```diff
@@ -554,6 +607,7 @@ export default class LobbyMatchingService extends ARpcService<LobbyMatchingRpcMe
                 advertiserChallenge: pick.advertiserChallenge
             });
         } catch (error) {
+            this.commitInFlight = false;
```
momoshell, line 601: This early return looks like it can strand a cancellation.

If cancellation is requested while `commitInFlight` and the committed peer then disconnects: `onProfileDisconnected` clears `inFlightSelection` and kicks `selectNextCandidate()`, the commit request rejects, and this catch takes the `this.inFlightSelection !== selection` branch — returning before the `pendingCancellation` → `cleanup(true)` path below.

Nothing else settles it: `onProfileDisconnected` doesn't, and `selectNextCandidate()` with an empty candidate list only schedules a role switch. `resolvePendingCancellation` is otherwise reached only from `finishMatch`. So the promise from `cancelMatching()` — and the targeted connect awaiting the match — stay pending indefinitely.

Would resolving the cancellation before this early return cover it?

test/evm/DiscoveryRuntimePort.test.ts
```diff
@@ -5,6 +5,578 @@ import { Status } from "@/types";
...
+        ).to.equal(Status.SYNCED);
+        const participants = await h.channelManager.getParticipants(channelId);
+        expect(
+            participants
+                .map(String)
+                .map((address: string) => address.toLowerCase())
+        ).not.to.include(h.peers[2].address.toLowerCase());
+    });
+
+    it("autoOpen without shouldJoin opens and syncs without membership", async function () {
```
momoshell, line 498: This test asserts success and participant count only — swapping the supplied balance for the default leaves it green, so it can't catch the `shouldJoin` balance issue in `LocalP2pSigner.ts:199`. Asserting the resulting genesis balances would make it discriminating.

src/evm/signer/LocalP2pSigner.ts
momoshell: Worth flagging an interaction, since it cuts against the fix: the reason the matchmaking branch can't currently catch an existing channel's peers is *this* bug — status never returns to `NOT_OPENED`, so `connectToChannel` never reaches the match on a channel switch.

If `clearChannelId()` starts resetting status, that branch becomes reachable while the previous channel's transports are still in `openConnections` — nothing in the switch path calls `disconnectAll()` (only `dispose()` does, `P2PManager.ts:113`). So the fix probably wants to drop or re-evaluate the old channel's connections at the same time.

lukastanisic99: Good catch. The current path can retain A's status and applied state after selecting B, skip B's initial sync, and return true. Resetting only the status is not enough because the old channel state is still loaded. I will reject every different target on a live attached runtime. A new leaveChannel flow will first follow the normal exit path — request exit, produce and finalize the exit block, submit the snapshot, and return to SYNCED — then dispose the runtime. 

`A new runtime is required for another channel.` - this is a quick solution for this PR, but your or @sh3ll3x3c can add cleanup logic that doesn't require us to dispose the full runtime and redeploy everything - I created a Trello card for this

src/evm/signer/LocalP2pSigner.ts
lukastanisic99: Somehow this comment convinced 2 of my GPT agents that it's correct, but Fable correctly pointed out that it's not.. 

This is already the behavior. autoOpen is an active genesis operation, so it may use the supplied balance; shouldJoin separately uses a supplied balance for later membership. An observer-only call reaches neither path. I will strengthen the balance oracle, but no production behavior change is needed here.

test/evm/DiscoveryRuntimePort.test.ts
lukastanisic99: Yes

src/rpc/services/lobbyMatching/LobbyMatchingService.ts
lukastanisic99: good catch

src/rpc/services/openChannelNegotiation/OpenChannelNegotiationService.ts
lukastanisic99: yes

lukastanisic99:
Overall all of this + MANY more things are addressed in #472

PR #470, opened 2026-09-02
gitignore generated spec reports

daiagi:
- docs/spec/generated/ is pure output of yarn spec:refresh; commits touching it caused merge conflicts
- gitignore it and untrack the 7 generated files
- add a spec CI job running the two diff-scoped checks: yarn spec:ids:check and yarn spec:impact --base <PR base>. both cheap
- note in docs/spec/AGENTS.md that generated/ is now local-only, rebuilt with yarn spec:refresh


lukastanisic99:
yes, just resolve conflicts

PR #472, opened 2026-09-05
PR #469 fixes, Leave channel and FIX ALL (hopefully) FLAKES | Many bug fixes

lukastanisic99:
It started with a few simple changes to address review comments on #469, then again a few tests were flaky and it sent me on a 4 day rabbit hole to FIX stuff across the stack from genuine src code bugs to tests bugs; This one was really exhausting;

I don't think any PR description is suitable - everything was touched - literally; Everything is documented, so your agents can see what they need; 

In the end, I reran tests 15+ times in a row without flakes. I've increased concurrency on our distributed servers for testing + added my devices (during testing) to the swarm, so there were a mix of high performance devices to low performance devices and it all held;

PR #475, opened 2026-09-07
address PR comments and other fixes

lukastanisic99:


sh3ll3x3c (commented):
**Bottom line:** rework before landing. Verification run during this review on the exact head: `yarn compile` OK; `yarn tsc --noEmit` for `tsconfig.json` and `tsconfig.browser.json` OK; `yarn lint:imports` OK; canonical gate `yarn test:parallel:distributed` (`./logs/run-123/`, 2046 tasks incl. forge) **2045 passing / 1 failing** (wall 716 s, 43.6× speedup); `yarn spec:impact --base origin/simplification-refactor-v1` **exits 1** with one file "review blocked"; `yarn spec:refresh` rewrites three committed generated reports (tree restored afterwards). Three 🔴s block, two of them the same defect: **FR1 = TR1** (the PR's own new probe test fails on the pool because two concurrent syncs sharing the local EVM reject an honest responder and blacklist it) and **AR1** (the traceability gate is red again).

Reviewed SHA `8cb0c062bbeef2fb305b754c23ff03f818dba906` (= PR head), 189 files. Inline comments carry the code-anchored 🔴/🟠 findings (FR1, FO2–FO4, SO1, TR1, TO2–TO6, CO1, CO2, OO1, OO2). Below are the ones with no single diff line.

#472 promises at this head: 39 addressed · 7 partially · 4 not · 8 declined (declines hold). Partial: TR1, AR1, AO2, AO5b, AY11, MO1 (PR body), TY13. Not: AR1's new drift (below), AO3, SY1 comment, AY8 `Block` annotation.

## 📏 Specification traceability

🔴 **[AR1] The traceability gate is red again on this head, on the same defect class #472 reported.** `yarn spec:impact --base origin/simplification-refactor-v1` exits 1 with `## Unmapped changed files — review blocked` → `test/evm/ContractExecutor.test.ts`; `yarn spec:refresh` rewrites `audit-summary.md`, `traceability.md`, `verification-coverage.md` (complete paths 41 → 38 of 280, mapped declarations 1624 → 1604). Cause: this PR moved test bodies without repointing their rows. [`ContractExecutor.test.ts.md` L19](https://github.com/peer3to/state-channels-plus/blob/8cb0c062bbeef2fb305b754c23ff03f818dba906/docs/spec/verification/tests/test/evm/ContractExecutor.test.ts.md#L19) says line 73; the `it(` is at 72 (all 20 rows off by one after a deleted import). `DisputeWindowAdmission.t.sol.md` points at L124 and L128; the functions start at L130 and L134 (13 of 16 rows off by six or seven). Both reports were correct on base. 33 declarations lose their evidence (`REQ-DIS-2.T1.P17/P19-23`, `REQ-DISPUTE-PIPE-9.T1.P7-11/P14/P15`, `REQ-TIME-5.T1.P1/P2`, `UNIT-TEST-CONTRACT-EXECUTOR-1.P1/2/3/6`, `UNIT-TEST-DISPUTE-WINDOW-ADMISSION-1.P1-5/P8/P9`, `UNIT-TEST-STATE-CHANNEL-COMMON-1.P9/P11-15`).

> **Fix AR1-FIX**
>
> Repoint both reports to the `it(` / `function test` lines, run `yarn spec:refresh`, commit the generated diff, and confirm `yarn spec:impact --base origin/simplification-refactor-v1` exits 0 with no "review blocked" section. Consider a pre-commit check that fails when a row's anchor line does not contain `it(`.


🟠 **[AO2] Three of the five reports from #472's AO2 are still stale and untouched**: `harness/JoinActions.test.ts.md` (7 rows), `unit/BlockProductionService.test.ts.md` (1), `onChainSlashes.test.ts.md` (1). `DisputeManager` and `latestStateSnapshotHash` were repointed.

> **Fix AO2-FIX**
>
> Repoint the nine rows in the same pass as AR1.


🟠 **[AO3] The one real new duplicate owner is unchanged.** `UNIT-TEST-REDUCTION-EXECUTOR-1-DGAD37.P9` is still credited by two rows, [`ReductionManager.test.ts.md` L64 and L67](https://github.com/peer3to/state-channels-plus/blob/8cb0c062bbeef2fb305b754c23ff03f818dba906/docs/spec/verification/tests/test/stateManager/ReductionManager.test.ts.md#L64-L67). (Our #472 count of 4 → 23 was wrong; the consistent count is 25 → 26, this is the one.)

> **Fix AO3-FIX**
>
> One owning row; the other shows `—`.


🟠 **[AO5b] One of eleven implementation-layer gaps closed.** Only `REQ-DISPUTE-PIPE-8` gained its Linked-requirements entry; `REQ-DISPUTE-PIPE-9`, `REQ-TIME-5`, `REQ-RUNTIME-6`, `REQ-TJOIN-2/4`, `INV-TJOIN-1` still read `none — gap` in `generated/traceability.md`, and `REQ-RUNTIME-5`, `INV-ENFDIS-1`, `INV-DISPUTE-PIPE-1`, `REQ-LIF-7` still have no evidence.

> **Fix AO5b-FIX**
>
> Add each ID to its owning file's Linked-requirements table; declare tests or an explicit gap for the four unevidenced IDs; refresh.


## 🧹 Scope

🟠 **[MO1] The PR body is empty, and the diff carries three peer-observable changes nobody asked for.** (1) [`getActiveValidationStrategy`](https://github.com/peer3to/state-channels-plus/blob/8cb0c062bbeef2fb305b754c23ff03f818dba906/src/stateManager/StateManager.ts#L483-L487) now hands `PENDING_PARTICIPANT` the live strategy, so a joining peer submits disputes, blacklists suppliers and applies the agreement window (base: `PARTICIPATING` only). (2) Sync: the requested height is now a minimum and a verified successor fork satisfies a pinned-fork request ([L755](https://github.com/peer3to/state-channels-plus/blob/8cb0c062bbeef2fb305b754c23ff03f818dba906/src/rpc/services/spectate/SpectateService.ts#L755), [L681-L686](https://github.com/peer3to/state-channels-plus/blob/8cb0c062bbeef2fb305b754c23ff03f818dba906/src/rpc/services/spectate/SpectateService.ts#L681-L686)); `REQ-SYNC-1` was rewritten from "exact target" to "minimum target" to match. (3) [`AStateMachine.removeParticipant`](https://github.com/peer3to/state-channels-plus/blob/8cb0c062bbeef2fb305b754c23ff03f818dba906/contracts/V1/AStateMachine.sol#L124-L135) now records an EXIT outbound message on the reduce path, which changes the reduced outbound chain every peer computes. Each may be right; none is explained, and the #472 MO1 breaking-change callout is still missing.

> **Fix MO1-FIX**
>
> PR description with one paragraph per behavior change (what, why, which test proves it), plus the `connectToChannel`/`setChannelId` breaking-change note from #472.


## For the record: the #472 TR1 root cause

**For the record: the #472 TR1 failure was not the lobby liability.** From the run-63 log: after the handoff disconnect, peer 2 redialled peer 0 and completed the init handshake; `LobbyMatchingService.onAuthenticatedTransport` closed that transport synchronously inside the `handshakeCompleted` event before `InitHandshakeService` sent its ack; peer 0's ack-timeout then fell back to `verifiedPeerAddressByTransport` and blacklisted peer 2's identity, which had no live transport, so no `Peer disconnected` line was written. Every file on that path is unchanged by #472 and by this PR. TR1 is withdrawn as a #472 blocker; it needs its own fix.

> **Fix MY3-FIX**
>
> Separate card: send the handshake ack before the lobby may close the transport (or defer the lobby close to after `handshakeCompleted` listeners return), or make the ack-timeout disconnect instead of blacklist when the transport never authenticated. Acceptance: the four-peer convergence test 20× green on the pool.

src/rpc/services/spectate/SpectateService.ts
```diff
@@ -215,6 +219,23 @@ class SpectateService extends ARpcService<SpectateServiceRpcMethods> {
             const forkIds = syncPayload.disputeWindows.map(
                 (disputeWindow) => disputeWindow.forkId
             );
+            // Another sync may have finalized this window only in the shared
+            // local EVM. Only chain finality can omit reduction from the multicall.
+            // Read finality first: a later window fetch includes any reduction that
+            // lands between reads, while a false decision safely keeps its calldata.
+            // Values indicate chain-final reduction for each requested fork.
+            const finalizedByFork = new Map<ForkId, boolean>();
```
sh3ll3x3c, line 227: 🔴 **[FR1] Two concurrent syncs on one peer still cross-talk through the shared local EVM, and the honest responder gets blacklisted.** The gate failed the PR's own new test `E2E: BlockQueueManager › wrong-fork blocks › a queue timeout accepts a proved successor fork without excluding the supplier or author` (task 160/2046, `expected 1 to equal 2` on the `succeeded` probe count, log `./logs/run-123/error_E2E-BlockQueueManager.test__…_supplier_or_author.ansi`). The log shows the observer receiving both probe payloads (author and supplier) within the same second and rejecting the author's with `applySyncResponse - rejecting sync { reason: 'reduced fork mismatch' }`, then `Disconnecting and blacklisting peer address` for the author. The rework at [`applySyncResponse` L221-L275](https://github.com/peer3to/state-channels-plus/blob/8cb0c062bbeef2fb305b754c23ff03f818dba906/src/rpc/services/spectate/SpectateService.ts#L221-L275) reads finality from the chain first precisely because "another sync may have finalized this window only in the shared local EVM", but steps 2.3 (`reduceAndFinalize`) and 2.5 (`getDisputeWindows` re-read) still run unserialised against that shared EVM, so the second sync's window persist or reduce lands between the first sync's reduce and its re-read. Proven: the new test fails once in one run on the exact head, with that reject reason, on an honest peer. Inferred: the exact interleaving above (the log has no step-level lines inside `applySyncResponse`).

> **Fix FR1-FIX**
>
> Serialise the local-EVM section of `applySyncResponse` per channel (one mutex around steps 2 to 3, the same way `StateManager` guards state), or make step 2.5 compare against the return value of the sync's own `reduceAndFinalize` call instead of a shared-EVM re-read. Prove it with the failing test in a loop: `yarn test:parallel:distributed --grep "accepts a proved successor fork"` 20× green, and extend `assertSyncWindowReadRace` in `PinnedSyncStaging.ts` to two concurrent probe syncs on one observer asserting neither responder is blacklisted.

src/rpc/services/spectate/SpectateService.ts
```diff
@@ -215,6 +219,23 @@ class SpectateService extends ARpcService<SpectateServiceRpcMethods> {
...
+            await Promise.all(
```
sh3ll3x3c: 🟠 **[SO1] A sync payload now costs the requester one chain call per dispute window, in parallel, sized by the responder.** [`Promise.all(forkIds.map(… isReduceChallengePeriodExpired …))`](https://github.com/peer3to/state-channels-plus/blob/8cb0c062bbeef2fb305b754c23ff03f818dba906/src/rpc/services/spectate/SpectateService.ts#L227-L239) runs before any payload check, and `forkIds` is the length of the peer-supplied `disputeWindows`. Base did one batched read. A Byzantine responder pads the array and gets N concurrent `eth_call`s out of every requester; any RPC failure in that batch lands in the outer `catch` and blacklists the honest responder.

> **Fix SO1-FIX**
>
> Derive finality from the windows `fetchAndPersistOnChainDisputeWindows` already fetches (they carry `reducedResult` and the challenge timestamps), or one multicall, and cap `disputeWindows.length` before the first chain read. Unit case: a payload with 50 windows makes at most one chain read before rejection.

src/rpc/services/spectate/SpectateService.ts
```diff
@@ -234,11 +255,7 @@ class SpectateService extends ARpcService<SpectateServiceRpcMethods> {
                     );
 
                 // 2.3) reduce them if they're not already reduced
-                const isReducedAndFinal =
-                    await diamondStateMachine.localDiamondContract.isReduceChallengePeriodExpired(
-                        channelId,
-                        dw.forkId
-                    );
+                const isReducedAndFinal = finalizedByFork.get(dw.forkId);
```
sh3ll3x3c, line 269: 🟠 **[FO4] Reading reduction finality from the chain changes what "more than one unreduced window" means, and a repeat sync can now blacklist an honest provider.** On base the counter at step 2.4 counted windows not yet reduced in the local EVM, so a second sync on the same peer saw the earlier window as reduced. On head [`isReducedAndFinal = finalizedByFork.get(dw.forkId)`](https://github.com/peer3to/state-channels-plus/blob/8cb0c062bbeef2fb305b754c23ff03f818dba906/src/rpc/services/spectate/SpectateService.ts#L258) is chain finality, so while two consecutive windows await their on-chain reduction (chain lag, a pending reduction tx), every re-sync counts two, hits `more than one unreduced window`, and `rejectSync` blacklists the responder that served a correct payload. The multicall motivation in the comment is right; the counter should not have moved with it.

> **Fix FO4-FIX**
>
> Keep chain finality for deciding which reductions go into the multicall calldata; keep the local-EVM `isReduceChallengePeriodExpired` read for the step 2.4 counter. Unit case in `SpectateService.test.ts`: two windows unreduced on chain, second sync on the same peer, assert no rejection and both reductions in the multicall.

src/rpc/services/spectate/SpectateService.ts
```diff
@@ -608,6 +641,20 @@ class SpectateService extends ARpcService<SpectateServiceRpcMethods> {
                 return undefined;
             }
             const { reduceData, reducedForkId } = computation;
+            latestComputation = computation;
+            computedGenesisTimestamp = (
```
sh3ll3x3c: 🟠 **[TO6] The synthesized successor genesis snapshot's timestamp is never asserted, and it comes from a cached read.** [`computedGenesisTimestamp`](https://github.com/peer3to/state-channels-plus/blob/8cb0c062bbeef2fb305b754c23ff03f818dba906/src/rpc/services/spectate/SpectateService.ts#L645-L649) uses `isKillPeriodExpiredCached`, while the real installer in `ReductionExecutor` uses a fresh `killPeriodEnd`; `isGenesisSnapshotWithoutTimeCheck` on the requester does not cover `timestamp`. No test compares the served timestamp with the installed one.

> **Fix TO6-FIX**
>
> Unit case in `SpectateService.test.ts`: responder serves a computed-but-uninstalled successor genesis, then installs it; assert the served `timestamp` equals the installed snapshot's. Use the fresh read (see CO2).

src/rpc/services/spectate/SpectateService.ts
```diff
@@ -608,6 +641,20 @@ class SpectateService extends ARpcService<SpectateServiceRpcMethods> {
...
+                await stateManager.reductionManager.isKillPeriodExpiredCached(
+                    currentForkId
+                )
+            ).killPeriodEnd;
+            // Reuse the outbound-chain owner before building the proof range.
+            // Background genesis installation persists this same block idempotently.
+            if (computation.reducedOutboundMessageBlock) {
+                stateManager.storage.outboundMessages.store(
```
sh3ll3x3c: 🟠 **[CO2] `generateSyncPayload` forks the reduction owner's post-compute steps.** [L645-L656](https://github.com/peer3to/state-channels-plus/blob/8cb0c062bbeef2fb305b754c23ff03f818dba906/src/rpc/services/spectate/SpectateService.ts#L645-L656) and [L705-L713](https://github.com/peer3to/state-channels-plus/blob/8cb0c062bbeef2fb305b754c23ff03f818dba906/src/rpc/services/spectate/SpectateService.ts#L705-L713) redo what [`ReductionExecutor.prepareLocalCandidate`](https://github.com/peer3to/state-channels-plus/blob/8cb0c062bbeef2fb305b754c23ff03f818dba906/src/stateManager/reduction/ReductionExecutor.ts#L286-L330) does after `compute`: persist the terminal outbound block with `justPersist`, take the kill-period end as genesis timestamp, build `StateSnapshot.from({ forkId, blockHeight: 0, timestamp, snapshotData })`. Two owners, already drifting (cached vs fresh timestamp, TO6).

> **Fix CO2-FIX**
>
> One `ReductionManager.materializeReducedGenesis(forkId, computation)` returning `{ genesisSnapshot, outboundBlock }` used by both the executor and the sync responder; acceptance grep: `StateSnapshot.from({` with `blockHeight: 0` appears once in `src/`.

src/rpc/services/spectate/SpectateService.ts
```diff
@@ -683,22 +752,20 @@ class SpectateService extends ARpcService<SpectateServiceRpcMethods> {
         const latestBlockHeight =
             stateManager.storage.blocks.getNextBlockHeight(forkId) - 1;
 
-        // An above-latest target is unprovable - and must NOT be silently
-        // downgraded to a proof for `latestBlockHeight` (that would forge a
-        // valid-looking proof for a different height). Return undefined so the
-        // caller cuts the requester; we don't serve a different height.
+        // A pinned height is a minimum. Serve the newest proof available, but
```
sh3ll3x3c, line 762: 🟠 **[OO2] Is "requested height is a minimum" the right contract for the block-queue probe?** The probe (`syncAfterInFlight(peer, channelId, forkId, height)`) asks whether a specific queued block is backed; a responder may now answer with a newer proof on a successor fork. The queue then decides "unbacked" from `isBlockUnbackedByLineage` on its own fork, which is what FR1's test exercises.

> **Fix OO2-FIX**
>
> State the probe's contract in `synchronization.md` next to the rewritten `REQ-SYNC-1`: which responses count as "block not backed" versus "cannot tell". Then TO3-FIX gives it a test.

src/stateManager/membership/MembershipService.ts
```diff
@@ -338,8 +353,16 @@ export default class MembershipService {
...
                         try {
-                            sm.storage.forceExit.setForceExit(true);
-                            await sm.disputeManager.dispute(block.forkId);
+                            if (
+                                !(await this.startSelfRemovalDispute(
+                                    block.forkId
+                                ))
+                            ) {
+                                this.logger.warn(
+                                    "Self-removal dispute did not start",
```
sh3ll3x3c, line 362: 🟠 **[FO2] `leaveChannel()` can still hang, one phase later.** The watchdog path is fixed, but [`onBlockCommitted`](https://github.com/peer3to/state-channels-plus/blob/8cb0c062bbeef2fb305b754c23ff03f818dba906/src/stateManager/membership/LeaveChannelService.ts#L125-L129) cancels the watchdog on an authored exit and hands escalation to `startMaybeExitOnChain`, whose two `startSelfRemovalDispute` sites only log when the marker is missing ([L357-L373](https://github.com/peer3to/state-channels-plus/blob/8cb0c062bbeef2fb305b754c23ff03f818dba906/src/stateManager/membership/MembershipService.ts#L357-L373), [L386-L403](https://github.com/peer3to/state-channels-plus/blob/8cb0c062bbeef2fb305b754c23ff03f818dba906/src/stateManager/membership/MembershipService.ts#L386-L403)). If the snapshot post fails and the dispute leaves no marker (`ErrorCantParticipateInDispute`, the FR1 trigger from #472), the phase stays `exit-authored`, no watchdog is armed, and the only resolve path ([L160-L169](https://github.com/peer3to/state-channels-plus/blob/8cb0c062bbeef2fb305b754c23ff03f818dba906/src/stateManager/membership/LeaveChannelService.ts#L160-L169)) needs the signer to disappear on chain, which is what just failed. The client waits with `timeoutMs: null`.

> **Fix FO2-FIX**
>
> Give the boolean a route: either `startMaybeExitOnChain` reports a `false` marker or a thrown error to `leaveChannelService.fail(operation, error)` when a leave is pending, or `onBlockCommitted` re-arms the watchdog on `exit-authored` instead of cancelling it and only `resolve`/`fail` cancel. Test next to `authored exit cancels the recorded watchdog and enters exit-authored`: make the snapshot post fail and the dispute return without a marker, assert `leaveChannel()` rejects within 3× the watchdog.

src/stateManager/membership/MembershipService.ts
```diff
@@ -359,8 +382,16 @@ export default class MembershipService {
...
                     try {
-                        sm.storage.forceExit.setForceExit(true);
-                        await sm.disputeManager.dispute(persistedBlock.forkId);
+                        if (
+                            !(await this.startSelfRemovalDispute(
+                                persistedBlock.forkId
+                            ))
+                        ) {
+                            this.logger.warn(
+                                "Self-removal dispute did not start",
```
sh3ll3x3c, line 401: 🟠 **[TO4] The `startMaybeExitOnChain` warn-only paths have no test.** Neither the snapshot-failure branch nor the slow path ([`MembershipService.ts` L357-L403](https://github.com/peer3to/state-channels-plus/blob/8cb0c062bbeef2fb305b754c23ff03f818dba906/src/stateManager/membership/MembershipService.ts#L357-L403)) is driven to a `false` marker or a throw by any test, which is how FO2 shipped.

> **Fix TO4-FIX**
>
> The FO2 test plus a second case where `postStateSnapshotWait` rejects and the dispute starts, asserting exactly one dispute submission.

contracts/V1/StateChannelDiamondProxy/StateChannelCommon.sol
```diff
@@ -595,8 +596,11 @@ contract StateChannelCommon is StateChannelManagerStorage, StateChannelManagerEv
     }
 
     function _canParticipateInDisputes(bytes32 channelId, address participant) internal view virtual returns (bool) {
-        address[] memory eligibleParticipants = _deriveEligibleParticipantsFromInboundHash(
-            channelId, channelBalances[channelId].latestInboundMessageBlockHash
+        address[] memory eligibleParticipants = _deriveEligibleParticipantsFromInboundHashAndSnapshotParticipants(
```
sh3ll3x3c, line 599: 🟠 **[FO3] The OY2 bound is applied to upload eligibility only; the fraud-proof facet still derives "expected participants" with the unbounded walk.** [`_canParticipateInDisputes`](https://github.com/peer3to/state-channels-plus/blob/8cb0c062bbeef2fb305b754c23ff03f818dba906/contracts/V1/StateChannelDiamondProxy/StateChannelCommon.sol#L598-L606) now passes the snapshot's inbound hash as the lower bound, while [`_deriveExpectedParticipantsForDispute`](https://github.com/peer3to/state-channels-plus/blob/8cb0c062bbeef2fb305b754c23ff03f818dba906/contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol#L769-L777) still calls the `bytes32(0)` variant. A joiner older than the boundary who is not a snapshot participant can no longer upload, but is still counted as an expected participant when disputes are verified, so the two facets answer "who is a participant" differently for the same channel state.

> **Fix FO3-FIX**
>
> Route `_deriveExpectedParticipantsForDispute` through the same bounded derivation with the snapshot boundary as lower bound (or state in a comment why the fraud-proof path must be unbounded, and add that reason to the OY2 decision). Forge case in `DisputeVerificationFacet.t.sol`: an old non-participant joiner is absent from both the eligibility set and the expected-participants set.

test/e2e/E2E-BlockQueueManager.test.ts
```diff
@@ -872,6 +863,128 @@ describe("E2E: BlockQueueManager", function () {
 
         // Supplier != author, so the recorded targets separate them: the timeout
         // asks everyone the entry is attributed to, not just the sender.
+        it("a queue timeout accepts a proved successor fork without excluding the supplier or author", async function () {
```
sh3ll3x3c, line 866: 🔴 **[TR1] The canonical gate fails on a test this PR adds.** `E2E: BlockQueueManager › wrong-fork blocks › a queue timeout accepts a proved successor fork without excluding the supplier or author` at [`E2E-BlockQueueManager.test.ts` L866](https://github.com/peer3to/state-channels-plus/blob/8cb0c062bbeef2fb305b754c23ff03f818dba906/test/e2e/E2E-BlockQueueManager.test.ts#L866-L986), `AssertionError: expected 1 to equal 2` at the `succeeded` assertion, task 160/2046 on `server-10`, exit 1, not rescheduled (a real failure, not starvation); log `./logs/run-123/error_E2E-BlockQueueManager.test__E2E_BlockQueueManager__a_queue_timeout_accepts_a_proved_successor_fork_without_excluding_the_supplier_or_author.ansi`. New in the diff, so it cannot be pre-existing. Cause in FR1.

> **Fix TR1-FIX**
>
> FR1-FIX, then loop the test 20× on the pool before pushing.

test/unit/ValidationService.test.ts
```diff
@@ -240,17 +256,19 @@ describe("Unit: ValidationService", function () {
...
-                .validation.runBlockValidation(encoded)
-                .request();
-
-            // genesis snapshot present -> real fraud proof + dispute on the fork
-            expect(r.resultName).to.equal("DISPUTE");
-            expect(r.disputedForkIds).to.deep.equal([forkId]);
-            expect(r.fraudProofType).to.equal(
-                solProofType(FraudProofType.WrongGenesis)
-            );
+            for (const strategy of ["active", "spectating"] as const) {
```
sh3ll3x3c, line 277: 🟠 **[TO2] The five new "active and spectating" loops only exercise the delegate branch; the observer branch of `wrongGenesis`, `invalidStateTransition`, `objectiveInvalidTimestamp`, `blockForkIsDisputed` and `forgedInboundMessageBlock` is now untested.** In [`ValidationService.test.ts` L259](https://github.com/peer3to/state-channels-plus/blob/8cb0c062bbeef2fb305b754c23ff03f818dba906/test/unit/ValidationService.test.ts#L259) and its siblings the `observer` is `h.getPeer(0)` after `lifecycle.start(3, 0)`, i.e. PARTICIPATING, so the `spectating` iteration hits `isCommittedParticipantStatus` and delegates to the live strategy. Only `doubleSign` has a real observer case (`pending participant records live double-sign fraud and stays pending`). Deleting `this.abort()` from any of those five hooks passes the suite.

> **Fix TO2-FIX**
>
> One `it` per hook with an OPENED/SYNCED spectator (`waitUntilPeerStatus(spectatorIndex, Status.OPENED)` as at L93) asserting `abortCalled` and no dispute submission; keep the participant iteration as the delegate case.

test/fixtures/PinnedSyncStaging.ts
```diff
@@ -0,0 +1,297 @@
...
+    expect(syncResult).to.not.equal(null);
+    const syncPayload = Codec.decode(
+        syncResult!.encodedSyncPayload,
+        Type.SyncPayload
+    );
+
+    const latestFinalizedSnapshot =
+        syncPayload.milestoneSnapshots.at(-1) ??
+        syncPayload.latestForkGenesisSnapshot;
+    expect(Number(latestFinalizedSnapshot.blockHeight)).to.equal(
```
sh3ll3x3c, line 184: 🟠 **[TO3] The rewritten sync-height E2E cases pass if the height argument is ignored.** Base asserted `latestFinalizedSnapshot.blockHeight === requestedHeight` (pinned). Head's [`expectSyncPayloadAboveRequestedHeightWhileAhead`](https://github.com/peer3to/state-channels-plus/blob/8cb0c062bbeef2fb305b754c23ff03f818dba906/test/fixtures/PinnedSyncStaging.ts#L141-L188) asserts only that the payload proves the responder's latest height, which is what an implementation that drops the argument also does; the two cases at [`E2E-Spectate.test.ts` L1205-L1216](https://github.com/peer3to/state-channels-plus/blob/8cb0c062bbeef2fb305b754c23ff03f818dba906/test/e2e/E2E-Spectate.test.ts#L1205-L1216) are therefore identical except for a number the assertion never uses, and the height-0 regression guard is gone.

> **Fix TO3-FIX**
>
> Make the oracle bite on the new rule: stage the responder with a finalized snapshot *between* the requested height and its latest, request a height above that snapshot, and assert the proof's milestones start at or after the requested height and that a request above latest is refused. Keep one case for the minimum-rule and one for the refusal, not two copies.

test/fixtures/SyncedLeaverReductionStaging.ts
```diff
@@ -0,0 +1,89 @@
...
+        await application.release();
+        await leave;
+        expect(await h.control(target).query.getStatus().request()).to.equal(
+            Status.SYNCED
+        );
+        const settlementReads = await h.execOnHost(target, async (sm) => {
+            const contract = sm.stateChannelManagerContract;
+            const original = contract.getParticipants;
+            let reads = 0;
+            Reflect.set(
```
sh3ll3x3c: 🟠 **[TO5] Two new inline `Reflect.set` monkeypatches on the contract while the three TO9 copies were removed.** [`SyncedLeaverReductionStaging.ts` L56-L68](https://github.com/peer3to/state-channels-plus/blob/8cb0c062bbeef2fb305b754c23ff03f818dba906/test/fixtures/SyncedLeaverReductionStaging.ts#L56-L68) and [`DiscoveryRuntimePort.test.ts` L1164-L1176](https://github.com/peer3to/state-channels-plus/blob/8cb0c062bbeef2fb305b754c23ff03f818dba906/test/evm/DiscoveryRuntimePort.test.ts#L1164-L1176) both replace `contract.getParticipants` inside `execOnHost` to count chain reads. test/AGENTS.md asks for a concrete `stub` seam; the slash-recovery triple this PR just added is the pattern.

> **Fix TO5-FIX**
>
> `stub.recordChainMembershipReads()/getChainMembershipReadCount()/restoreChainMembershipReads()` on `StubService`; delete both inline copies.

src/stateManager/validationStrategy/SpectatingValidationStrategy.ts
```diff
@@ -140,13 +138,22 @@ export default class SpectatingValidationStrategy extends AValidationStrategy {
         _conflictingBlock: Block,
         _block: Block
     ): Promise<BlockValidationResult> {
+        if (isCommittedParticipantStatus(this.p2pManager.stateManager.status))
```
sh3ll3x3c, line 141: 🟠 **[CO1] The "committed peer ⇒ live strategy" decision is restated six times inside `SpectatingValidationStrategy`, which turns a strategy into a status-dispatching proxy.** [L141](https://github.com/peer3to/state-channels-plus/blob/8cb0c062bbeef2fb305b754c23ff03f818dba906/src/stateManager/validationStrategy/SpectatingValidationStrategy.ts#L141), L153, L177, L186, L204, L235 each check `isCommittedParticipantStatus(this.p2pManager.stateManager.status)` and forward to `blockValidationStrategy`. `StateManager.getActiveValidationStrategy` already owns that decision. AGENTS.md's rule is that the strategy *is* the pipeline; a strategy that re-decides the pipeline per hook from mutable status is the opposite, and `blockForkIsDisputed` now re-queues the entry bound to the live strategy while the caller passed the spectating one. The AO6a reply promised a replay subclass passed through `validationStrategy`; that is the shape that keeps one decision point.

> **Fix CO1-FIX**
>
> Choose the strategy once at the entry (`getActiveValidationStrategy` for live gossip; a `ReplayValidationStrategy extends BlockValidationStrategy` whose subjective-timestamp hook returns `SUCCESS` for sync replay by a committed peer; the plain spectating strategy for observers) and delete the six status checks. Acceptance: `grep isCommittedParticipantStatus src/stateManager/validationStrategy` returns nothing; `ValidationService.test.ts` keeps both iterations green.

src/stateManager/StateManager.ts
```diff
@@ -475,14 +481,9 @@ class StateManager<
...
     public getActiveValidationStrategy(): AValidationStrategy {
-        return this.getStrategyByStatus(this.status);
-    }
-
-    private getStrategyByStatus(status: Status): AValidationStrategy {
-        if (status === Status.PARTICIPATING) {
-            return this.blockValidationStrategy;
-        }
-        return this.spectatingValidationStrategy;
+        return isCommittedParticipantStatus(this.status)
```
sh3ll3x3c, line 484: 🟠 **[OO1] Should a `PENDING_PARTICIPANT` run the live disputing strategy?** It now does (MO1 item 1). A joiner that is not yet in the snapshot participant set can now be pulled into disputes and can blacklist suppliers on gossip it used to observe passively.

> **Fix OO1-FIX**
>
> Decide and write it into the PR body and `block-confirmation-pipeline.md`. If yes, keep the two `pending participant … double-sign` tests and add the agreement-window case for a pending peer; if no, revert `getActiveValidationStrategy` to `PARTICIPATING` only and move the double-sign handling into the spectating strategy's observer branch.

lukastanisic99:
In reply to [submitted review](https://github.com/peer3to/state-channels-plus/pull/475#pullrequestreview-5131285386):

AR1/AO2 are fixed: the all-report check now resolves all 1,813 declaration rows, both impact modes pass, and the generated reports are stable on repeat refresh. AO3 now gives reduction-executor P9 one owner. AO5b's ten implementation mappings are present; the four existing evidence gaps remain explicitly recorded rather than claimed covered.

MO1: the behavior-change description is drafted locally, including the connectToChannel/setChannelId breaking change. It has not been published. MY3: the handshake acknowledgment ordering is recorded as a separate local follow-up; its implementation and twenty-run acceptance are still outstanding.

The final canonical gate passed 2,058 / 2,058 with no retries. Both typechecks, compile, imports, formatting and the browser worker suite (7/7) pass.

src/rpc/services/spectate/SpectateService.ts
lukastanisic99: Fixed both shared-window reads. A successful local reduction already proves the expected fork through the contract's check. For chain-final windows, the sync now uses its own fetched chain response; retaining a shared-EVM reread there would still allow an older concurrent persist to overwrite the evidence.

The deterministic two-source test reproduces reduced fork mismatch with the old code and passes with the fix. Both syncs accept and neither responder is blacklisted. The queue probe passed twenty consecutive distributed runs; the final full gate passed 2,058 / 2,058.

test/e2e/E2E-BlockQueueManager.test.ts
lukastanisic99: Fixed with FR1. The deterministic concurrent-sync regression proves the overwritten-window failure and the corrected behavior. The queue-timeout case passed twenty consecutive distributed runs (1708–1727), and the final canonical gate passed all 2,058 tasks with no retries.

src/stateManager/membership/MembershipService.ts
lukastanisic99: Fixed. Both authored-exit fallback sites now report a false dispute marker or thrown error to LeaveChannelService. It rejects only the current exit-authored operation for the same fork, so a stale task cannot reject a re-armed leave.

Tests cover fast and slow false/throw paths, slow-path success, and absent/awaiting/stale-fork guards. The leave rejects on fallback failure instead of waiting indefinitely; successful fallback still settles through one dispute.

src/rpc/services/spectate/SpectateService.ts
lukastanisic99: Replaced the parallel per-window finality calls with one multicall.staticCall. Solidity still owns the finality predicate, and the read remains before window persistence. Existing RPC-failure liability stays unchanged. No length cap was added because that needs a protocol decision.

The batching test supplies two copies of a real window, observes one provider call containing both predicates, and now records the exact rejection reason `more than one unreduced window` through a forwarding probe. It does not claim to exercise a valid two-fork lineage.

src/rpc/services/spectate/SpectateService.ts
lukastanisic99: Consolidated reduction genesis materialization on ReductionManager. The executor and sync responder use it, and state application consumes the prepared snapshot instead of rebuilding it. The helper takes computation and genesisTimestamp; the proposed source-fork argument was unused.

Final-dispute construction remains separate. Its event timestamp equals the kill-period end because a threshold-final upload backdates the evidence timestamp; a comment now explains that equality. StateSnapshot.empty also remains separate. Existing computed-successor and installation flows pass.

src/stateManager/membership/MembershipService.ts
lukastanisic99: Added the missing fast failure branches and slow-path cases. The slow cases withhold an exit signature and exercise the agreement-time fallback: false marker and evidence-expired errors reject the leave; successful submission settles it through one dispute. Existing fast snapshot-failure/success coverage remains.

test/unit/ValidationService.test.ts
lukastanisic99: Added real spectator cases for all five hooks. One correction to the proposed oracle: blockForkIsDisputed returns NOT_READY and requeues an observer; it does not abort. The other four hooks return DISPUTE, emit abort and submit no dispute. The new tests assert those distinct outcomes. Existing committed-peer delegation cases remain.

test/fixtures/SyncedLeaverReductionStaging.ts
lukastanisic99: Replaced both inline getParticipants patches with the shared recordChainMembershipReads/getChainMembershipReadCount/restoreChainMembershipReads stub endpoints. The recorder forwards the real chain call. Neither named file contains Reflect.set now, and both affected cases pass.

test/fixtures/PinnedSyncStaging.ts
lukastanisic99: Removed the duplicate height-one E2E and retained the height-zero case with its verification mappings. The existing real-RPC tests still prove acceptance at or above the minimum, refusal above the responder's latest height, and rejection of a supplied proof below the minimum. No milestone-start oracle was added because that is not the protocol rule.

src/stateManager/validationStrategy/SpectatingValidationStrategy.ts
lukastanisic99: The spectating-with-delegation design stays. Luka confirmed that the disputed-fork restore contradicts queue clearing, so the restore is now removed from both live and observer hooks. The live hook still penalizes acknowledged suppliers; other entries stop with NOT_READY and are discarded without scheduling a retry. The extra strategy parameter and its probe field are deleted.

src/rpc/services/spectate/SpectateService.ts
lukastanisic99: No change to the counter. Every sync refreshes windows from the chain, so earlier local-only reductions did not make a later sync's chain-unreduced windows disappear on base. The specification still rejects more than one unreduced window. Chain finality now comes through one batched Solidity call; it still governs calldata inclusion and the counter. The proposed two-unreduced-window acceptance case would contradict the protocol.

src/rpc/services/spectate/SpectateService.ts
lukastanisic99: No timestamp-policy change was needed. The successor snapshot-update simulation checks the genesis timestamp against the chain's kill-period end before installation, so existing computed-successor acceptance already depends on it being correct. Reduction materialization is now shared under CO2. For the separate threshold-final event path, the emitted creation timestamp equals the kill-period end after the contract backdates evidence; that equality is now documented in code.

src/rpc/services/spectate/SpectateService.ts
lukastanisic99: The probe contract remains as specified in block-processing Stage 3: a proved lineage that omits the block means junk; a block on another fork after sync is inconclusive and the source is kept. Minimum-height and verified-successor responses remain accepted by design. No protocol change was made for this comment. The queue-probe case passed twenty consecutive distributed runs after the separate concurrent-sync fix.

contracts/V1/StateChannelDiamondProxy/StateChannelCommon.sol
lukastanisic99: No derivation change. Point-in-time upload eligibility is bounded at the snapshot; historical milestone unanimity remains anchored at the dispute's own inbound hash. Applying the snapshot bound to that historical walk could weaken checks for older milestones. The stricter threshold after a non-slash removal remains the documented pre-existing liveness cost; this pass does not claim to solve it or add the suggested Forge case.

src/stateManager/StateManager.ts
lukastanisic99: Pending participants still use live validation, as explicitly decided and documented. Correction to my earlier rationale: a peer that remains pending does not counter-sign; shouldSignBlock requires PARTICIPATING, and promotion precedes signing. Its on-chain dispute eligibility and the owner decision justify live validation. Existing pending-peer double-sign tests remain; no additional pending agreement-window case was added. The local description draft records the behavior, and nothing was published to GitHub.

sh3ll3x3c (approved):

PR #476, opened 2026-09-07
docs(spec): cut always-paid context, and keep the linker out of generated/

momoshell:
Reading `docs/spec/` cost more than it needed to before a task began.

## Lazy-load the entry point

`docs/spec/AGENTS.md` opened by telling every agent to read `README.md`, `governance.md` and the owning layer README before touching anything — roughly 15KB of instructions paid on every task regardless of what the task was. It now routes each kind of work to the document it actually needs, so a read-only question stops at the owning document instead of four entry files first.

Same content, demoted from always to on demand, with a short rule beside the table to keep it that way: what every task must read before it starts is the one cost every task pays, and a link is cheap where an instruction to read something first is not.

## Keep the linker out of `generated/`

`check-id-links.js --write` rewrites every `.md` it walks, generated reports included — and it cannot parse a generator's nested link. Given ``[[`ID`](target) · 1 plan](target)`` it strips one bracket per run and settles on malformed markup while still reporting zero issues.

`normalizeIdMarkup` and `linkify` now share a filter that skips `generated/`. Those files belong to their generators, and since #470 they aren't even tracked, so `yarn spec:ids:fix` was corrupting untracked local output for no reason.

## Gloss capability

The linker can now carry a definition's own subject in the link text, so a reference is readable without opening its target:

```
[`REQ-RPC-5-CV1R1Y` (Resource bounds)](rpc.md#req-rpc-5-cv1r1y)
```

The subject is read from the definition site, never invented, and only where a definition states one. **The rewrite that applies it across the tree is deliberately not here** — it touches ~170 documents and belongs after the in-flight spec review, so the glosses reflect corrected text.

Two format constraints are encoded and commented, both learned the hard way: the gloss is parenthesised because an ID followed by " — " at line start is how `id-registry.js` recognises a definition site, and subjects containing brackets are skipped because the strip pattern that makes `--write` idempotent stops at the first `)`.

## Not included

An earlier version of this branch also committed refreshed `generated/` reports and added a separate spec workflow. Both are dropped: #470 gitignored `generated/`, and `ci.yml` already has a `spec` job.

Verified on this base: `yarn spec:ids:check` reports 6,617 definitions and 0 issues, `yarn spec:refresh` completes, and a `--write` leaves `generated/` byte-identical.

### DIFF
--- a/docs/spec/AGENTS.md
+++ b/docs/spec/AGENTS.md
@@ -1,7 +1,23 @@
 # Specification maintenance instructions
 
-These rules apply to every file under `docs/spec/`. Read [README.md](./README.md) and
-[governance.md](./governance.md), then the owning layer README and directly linked documents.
+These rules apply to every file under `docs/spec/`. This file is the whole of what every task
+needs; read further only where your task actually reaches:
+
+| Before you                                      | Read                                                          |
+| ----------------------------------------------- | ------------------------------------------------------------- |
+| answer a question about specified behavior      | nothing further — the owning document, then stop              |
+| edit one document inside one layer              | that layer's README                                           |
+| allocate, move, or delete an ID                 | [governance.md](./governance.md)                              |
+| change behavior across layers, or add a subject | [README.md](./README.md) and [governance.md](./governance.md) |
+| record a decision, finding, or approval         | [governance.md](./governance.md)                              |
+
+Reading all four entry documents before every task costs roughly four times this file for context
+almost none of it uses. Follow a link when the work reaches it, not in advance.
+
+Keep that true when editing this file. What every task must read before it starts is the one cost
+paid by every task, so it is the one worth arguing about: a link is cheap, an instruction to read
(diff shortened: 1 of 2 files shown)


lukastanisic99:
I merged Luke's PR that gitignored the generated stuff and added a CI rule - take a look is your CI rule relevant and resolve the conflicts - we can merge this PR, since it's just lazy loading - the one that's stacked on this one, I'll review later today @momoshell

lukastanisic99:
@momoshell I'm not sure how you merged/rebased on top of the latest dispute stuff and how only generated files are a conflict, but it's for sure not good - none of the latest stuff is in here even the CI has the old 1500 tests vs the new 2000 tests + none of the new spec stuff is here your PR that stacks on top of this one also reflects the same miss match 

image

lukastanisic99 (changes requested):
Commented above

PR #479, opened 2026-09-08
fix(storage): bound per-entry queue retention, and the participant union it is sized against

momoshell:
Found while auditing specification claims against the source during the group-3 spec review.

## The defect

`QueueStorage` capped `sourcePeers` and `signatureSources` but not the block's own signature set. Ingress authenticates `blockConfirmation.signedBlock`, **not** the confirmation signatures it carries — so one authenticated peer could resend a single block hash with fresh signatures and grow one entry without bound. The queue holds unvalidated knowledge by design, which is exactly why its structure has to be bounded.

## What it does now

**All three write paths cap retention** — the creating copy, a duplicate merge, and restore. Restore normalizes the entry it is handed, since that object was held by a caller across a dequeue.

**Values must be recoverable, not merely well-shaped.** A length check is not enough: `ethers.Signature.from` accepts an `r` outside the curve order, and even an in-range `r` usually identifies no curve point. Every such value throws out of ethers when validation later derives confirmation signers, so the queue attempts the same recovery its consumer performs and rejects what fails. Cost is bounded per block hash, not per call — a full entry does no crypto, and a dequeue does not refill the allowance.

**Attribution follows retention**, so `signatureSources` never holds a key the block does not.

**Channels have a maximum participant union** (`MAX_CHANNEL_PARTICIPANTS = 256`), enforced at open and at join. The client sizes its retention against it.

## What it does not do

The cap **bounds per-entry memory**. It is not a proof that a needed confirmation is never dropped, and the code, the requirement, the mirror and the audit assessment all say so. Three findings record why:

- **`FIND-SETTLE-1-G2CPV6`** — the maximum is enforced where a union is *proposed*, not everywhere one becomes *authoritative*: consumer-returned genesis participants and snapshot adoption both write unchecked.
- **`FIND-QSTORE-3-1HF4V6`** — retention counts signature bytes, validity counts recovered signers. ECDSA does not require a deterministic nonce, so one participant can produce many valid confirmations and occupy slots the others need.
- **`FIND-QSTORE-2-EVM8T9`** — the Solidity and TypeScript copies of the maximum are unrelated literals and can drift silently.

The first two are protocol decisions and are tracked separately. Retention above the cap is also **first-come** by decision, documented with a test pinning the residual.

## Review

Seven blind review rounds, each verified against source before acting. Several found defects in the previous round's fixes, and the recurring failure was mine: a signature check that did not check what it named, tests that could not fail, an E2E that did not use the path its commit message claimed. Every behaviour added here is mutation-proven — remove the guard, exactly the matching test fails.

Coverage: 39 unit tests, 3 Foundry tests, and an E2E that floods one parked block over the production state-transition RPC and asserts the bound. `yarn spec:refresh` and `check-id-links` clean at 6,642 definitions; specification, implementation mirrors, verification mappings and audit assessment all move with the code.


src/storage/QueueStorage.ts
```diff
@@ -63,7 +83,7 @@ export class QueueStorage {
                 block.allSignatures,
                 options?.senderAddress
             );
-            existingEntry.block.mergeFrom(block);
+            this.mergeBlockCapped(existingEntry, block);
```
lukastanisic99, line 176: **AI-generated review**

🔴 **[FR1] 95% — A third write path still merges signatures uncapped: `restoreEntry`.** The PR caps `queueBlock`'s merge here and the creating copy in `createEntry`, and its second commit says the cap now covers "both write paths". [`restoreEntry` L176](https://github.com/peer3to/state-channels-plus/blob/2df5c87f9285e4929ef14872949920c7ab2dce35/src/storage/QueueStorage.ts#L176) still calls `existing.block.mergeFrom(entry.block)`, which is [`Block.mergeFrom` L247-251](https://github.com/peer3to/state-channels-plus/blob/2df5c87f9285e4929ef14872949920c7ab2dce35/src/models/Block.ts#L247-L251) → `expandSignatures` with no limit. This is the only sanctioned re-queue path ([`BlockQueueManager.restoreQueuedEntry` L268-280](https://github.com/peer3to/state-channels-plus/blob/2df5c87f9285e4929ef14872949920c7ab2dce35/src/stateManager/ingest/BlockQueueManager.ts#L268-L280)), taken on every not-ready outcome. Sequence: entry A is dequeued (≤128 signatures); the flooder queues a fresh copy meanwhile, which creates entry B capped at 128 disjoint signatures; A is restored → B holds up to 256; next cycle the same again → 384, and so on. Each not-ready cycle adds up to 128, so the bound the PR exists to provide does not hold under exactly the adversary it names. The existing test "should merge a restored entry with a copy queued meanwhile" ([L533](https://github.com/peer3to/state-channels-plus/blob/2df5c87f9285e4929ef14872949920c7ab2dce35/test/storage/QueueStorage.test.ts#L533)) asserts the union and would have caught this if written with cap-sized sets.

> **Fix FR1-FIX**
>
> In `restoreEntry`, replace `existing.block.mergeFrom(entry.block)` with `this.mergeBlockCapped(existing, entry.block)` (the timestamp handling is identical). Add a static `it`: queue a copy, dequeue it, queue a copy carrying 128 new signatures, restore the dequeued entry after giving it 128 other signatures, and assert `confirmationSignatures.size <= 128` and `overflowedSources === true`. Acceptance: that test fails on the current head and passes after the change; the three new tests and the two existing restore tests stay green.

src/storage/QueueStorage.ts
```diff
@@ -47,9 +54,22 @@ export class QueueStorage {
             signatureSources: new Map()
         };
         this.trackSource(entry, block.allSignatures, options?.senderAddress);
+        this.capSignatures(entry);
```
lukastanisic99, line 135: **AI-generated review**

🟡 **[FY2] 90% — Attribution is recorded for signatures the entry then discards.** `trackSource` runs first with every signature the copy carried, then `capSignatures` trims the block. `signatureSources` therefore keeps up to 128 keys, some of which are signatures the block no longer holds, and in `queueBlock` the same happens for novel signatures beyond the room ([L81-86](https://github.com/peer3to/state-channels-plus/blob/2df5c87f9285e4929ef14872949920c7ab2dce35/src/storage/QueueStorage.ts#L81-L86)). Harmless for `disconnectPeersForSignatures`, which looks up only signatures still on the block, but it spends the `signatureSources` cap on phantoms and makes the entry inconsistent with `REQ-QSTORE-1`'s "attributes exactly those signatures".

> **Fix FY2-FIX**
>
> Cap first, then attribute only the retained signatures: compute the kept set in `capSignatures`/`mergeBlockCapped` and pass it to `trackSource`. Assert in the flood tests that every key of `signatureSources` is in `block.confirmationSignatures` (plus the original signature).

src/storage/QueueStorage.ts
```diff
@@ -241,6 +261,23 @@ export class QueueStorage {
...
+    // never trips the marker. The on-chain timestamp is not capped -- it is a
+    // single value read from chain observation, never from a gossiped payload.
+    private mergeBlockCapped(entry: QueuedBlockEntry, incoming: Block): void {
+        const held = entry.block.confirmationSignatures;
+        const novel = [...incoming.confirmationSignatures].filter(
+            (signature) => !held.has(signature)
+        );
+        const room = Math.max(QueueStorage.MAX_ENTRY_SIGNATURES - held.size, 0);
+        if (novel.length > room) entry.overflowedSources = true;
+        entry.block.expandSignatures(novel.slice(0, room));
```
lukastanisic99: **Human note**

yes - human needs to take a decision on this;

**AI-generated review**

🟠 **[SO1] 85% — The cap is first-come, so a junk-first flood makes the queue drop honest signatures that arrive later.** **Human assessment needed** Decision: keep the first-come per-entry cap and record the residual, or give each source a per-sender budget of novel signatures so one flooder cannot crowd out honest copies? Evidence: `room` is computed from the held set regardless of who filled it, and novel signatures beyond it are sliced away silently. If an authenticated peer sends 128 junk signatures for a hash before the honest participants' copies arrive, every honest signature is then dropped on merge, and the PR's claim that a real block "never loses a signature it needs to become eligible" ([L36-37](https://github.com/peer3to/state-channels-plus/blob/2df5c87f9285e4929ef14872949920c7ab2dce35/src/storage/QueueStorage.ts#L36-L37)) does not hold. The blast radius is bounded, which is why this is 🟠 rather than 🔴: when the entry is validated, the junk signers are outside the participant union, their source peers are disconnected and blacklisted ([`disconnectPeersForSignatures` L222-236](https://github.com/peer3to/state-channels-plus/blob/2df5c87f9285e4929ef14872949920c7ab2dce35/src/stateManager/ingest/BlockQueueManager.ts#L222-L236)), and the junk is stripped from the block ([`BlockValidationStrategy` L110-113](https://github.com/peer3to/state-channels-plus/blob/2df5c87f9285e4929ef14872949920c7ab2dce35/src/stateManager/validationStrategy/BlockValidationStrategy.ts#L110-L113)), so room frees and re-gossiped honest copies land. The cost is one validation cycle of delay per attacker identity, and pre-profile identities are cheap (the `OQ-34` ban-persistence bullet). The source caps deliberately have the opposite property ("never evicts already-tracked sources", "cannot crowd out an honest source"), so the two caps now differ in kind. Recommended: per-sender budget, because `signatureSources` already attributes every signature to its sender, so "at most N novel signatures per source per entry" is a two-line change and restores the "one flooder cannot displace honest input" property the spec promises for sources. The check that settles it: whether you accept a per-identity one-cycle delay as a documented residual.

> **Fix SO1-FIX**
>
> Conditional on OO1. Per-sender: track `novelBySender: Map<Address, number>` on the entry (or derive from `signatureSources`) and cap novel signatures per sender at `MAX_ENTRY_SIGNATURES / k`, keeping the global cap as the outer bound; add a test "junk-first signature flood from one sender does not displace a later honest copy's signatures". First-come: change the L36-37 comment and `REQ-QSTORE-2` to say honest signatures arriving after an overflow are dropped until the entry is validated and the flooder cut, and add a test that pins that behaviour so nobody relies on the stronger claim.

lukastanisic99 (commented):
**AI-generated review**

🟠 **[XO1] 90% — The specification, the mirror, and the PR text disagree about what the queue caps and what the cap guarantees.** [`REQ-QSTORE-2` at queue.md L39-43](https://github.com/peer3to/state-channels-plus/blob/2df5c87f9285e4929ef14872949920c7ab2dce35/docs/spec/specification/storage/queue.md#L39-L43) says "Per-entry **attribution** retention is capped … never evicts already-tracked sources … A junk-first flood therefore cannot crowd out an honest source". It says nothing about the confirmation-signature set, which this PR now caps with a different property: signatures arriving after the cap are dropped even when honest (see SO1). The mirror row [`QueueStorage.ts.md` L97](https://github.com/peer3to/state-channels-plus/blob/2df5c87f9285e4929ef14872949920c7ab2dce35/docs/spec/implementation/source/src/storage/QueueStorage.ts.md#L97) stays `Covered / None.` and describes only source caps. The code comment at [QueueStorage.ts L36-37](https://github.com/peer3to/state-channels-plus/blob/2df5c87f9285e4929ef14872949920c7ab2dce35/src/storage/QueueStorage.ts#L36-L37) and the PR body claim "a real block never loses a signature it needs to become eligible", which holds only for honest-first arrival. `docs/spec/AGENTS.md` "Maintain all four layers" and root `AGENTS.md` "Specification traceability … in the same pass" both require the spec and mirror to move with the code.

> **Fix XO1-FIX**
>
> Extend `REQ-QSTORE-2` to state the signature cap and its actual property (first-come, or per-sender after OO1), add `REQ-QSTORE-2-VYWJAQ.T1.P4 — unique junk signatures past the cap: marker set, entry bounded, block still dequeueable`, update the `QueueStorage.ts.md` design-decision bullet and conformance row, and reword the L36-37 comment to the guarantee the code gives. Acceptance: `yarn spec:refresh` shows the new permutation and the mirror row cites `MAX_ENTRY_SIGNATURES`.



---

**AI-generated review**

🟠 **[TO1] 85% — No E2E covers the peer-observable behaviour: a signature flood over real gossip, the block still finalising, the flooder cut.** Root `AGENTS.md` requires an E2E for peer-observable guard/punishment semantics. The nearest existing cases are [`E2E-FraudProofsBlockConfirmation` L275 and L343](https://github.com/peer3to/state-channels-plus/blob/2df5c87f9285e4929ef14872949920c7ab2dce35/test/e2e/E2E-FraudProofsBlockConfirmation.test.ts#L275) (one non-participant signature is dropped) and the `E2E-SpectatingAbortDoS` junk-block cases; none sends hundreds of junk confirmation signatures for one hash through `StateTransitionRpcMethods.onBlockConfirmation` and asserts the honest block still applies and the sender is blacklisted. That is also the only test that would show the SO1 crowd-out end to end.

> **Fix TO1-FIX**
>
> Add one static E2E in `E2E-BlockQueueManager.test.ts`: a byzantine peer (via the `byzantine` harness service) re-sends a valid pending block's hash with 300 unique junk signatures before and after the honest confirmations; assert the block reaches the chain state, the flooder is disconnected and blacklisted, and the queued entry's signature count never exceeds 128 (via a `query` projection). Acceptance: passes on head with FR1-FIX, fails if `mergeBlockCapped` is reverted.



---

**AI-generated review**

🟡 **[TY2] 90% — Unit variations still missing after the three new cases.** (a) Restore-path flood (FR1). (b) Boundary: exactly 128 signatures → no marker; 129 → marker and size 128 (both new tests use 300, so the boundary is untested). (c) Junk-first then honest copy: documents whether honest signatures survive (SO1) instead of leaving it implicit. (d) `createEntry` on the sourceless replay path ([`BlockIngestService.onBlockConfirmationStruct` L36-48](https://github.com/peer3to/state-channels-plus/blob/2df5c87f9285e4929ef14872949920c7ab2dce35/src/stateManager/ingest/BlockIngestService.ts#L36-L48)) also trims, which is fine at 128 > participants but unasserted. (e) A capped merge still applies the incoming on-chain timestamp (the timestamp branch of `mergeBlockCapped` has no test). (f) An overflowed entry still dequeues through `tryDequeueAt` with the original signature intact (the new tests check `isBlockQueued` only).

> **Fix TY2-FIX**
>
> Add six static `it` cases in `test/storage/QueueStorage.test.ts` for (a)–(f), each asserting size, marker, and dequeueability as applicable. Acceptance: `yarn test:parallel --grep QueueStorage` green with 31+ cases.



---

**AI-generated review**

🟡 **[TY3] 95% — The three new tests are honest discriminators and follow the harness rules.** Static `it` names, factory-built block plus random 65-byte signatures (junk by design), assertions on the entry itself. The PR body's discriminator claim (flood test fails with `mergeBlockCapped` reverted) was not re-run here, since the tree was in use by the gate; it is consistent with the test bodies. Listed as 🟡 only because the assertion `size <= 128` should be `=== 128` for the 300-signature floods, which pins that the cap keeps the first 128 rather than dropping everything.

> **Fix TY3-FIX**
>
> Tighten the two `at.most(128)` assertions to `equal(128)`.



---

**AI-generated review**

🟡 **[CY1] 90% — Two block-merge implementations now coexist.** [`mergeBlockCapped` L269-279](https://github.com/peer3to/state-channels-plus/blob/2df5c87f9285e4929ef14872949920c7ab2dce35/src/storage/QueueStorage.ts#L269-L279) re-states the timestamp overwrite of [`Block.mergeFrom` L247-251](https://github.com/peer3to/state-channels-plus/blob/2df5c87f9285e4929ef14872949920c7ab2dce35/src/models/Block.ts#L247-L251), and `mergeFrom` stays live for `BlockStorage` ([L65](https://github.com/peer3to/state-channels-plus/blob/2df5c87f9285e4929ef14872949920c7ab2dce35/src/storage/BlockStorage.ts#L65), validated blocks, correctly uncapped) and for `restoreEntry` (FR1, wrongly uncapped). Root `AGENTS.md` "Reuse existing code": one owner per operation.

> **Fix CY1-FIX**
>
> Give `Block.mergeFrom` an optional `{ maxConfirmationSignatures }` returning whether it overflowed, or make `mergeBlockCapped` the single queue-side merge used by both `queueBlock` and `restoreEntry` (FR1-FIX does the latter). Acceptance: `grep -n "mergeFrom(" src/storage/QueueStorage.ts` returns nothing.



---

**Human note**

Yes again no docs, so eaither the lazy loading is not working or cheap on tokens - my agents are still generating docs so hmm...

**AI-generated review**

🟡 **[AY1] 90% — The "Specification traceability" rule was not followed for a `src/` change.** Root `AGENTS.md`: any source or test change that affects specified behaviour updates requirements, implementation subjects, verification traceability, and individual test declarations "in the same pass", then `yarn spec:refresh`. The PR touches neither `docs/spec/implementation/source/src/storage/QueueStorage.ts.md` nor `docs/spec/verification/tests/test/storage/QueueStorage.test.ts.md`. Class layout, static `it` declarations, no mocks, no `console.*`, and comment length are all fine.

> **Fix AY1-FIX**
>
> Apply DO1-FIX and DO2-FIX.



---

**Human note**

Same human comment as for AY1;

**AI-generated review**

🟠 **[DO1] 95% — Required documentation artefacts are missing.** Per `docs/spec/AGENTS.md`: (a) `QueueStorage.ts.md` key design decision and conformance row for the signature cap ([L76](https://github.com/peer3to/state-channels-plus/blob/2df5c87f9285e4929ef14872949920c7ab2dce35/docs/spec/implementation/source/src/storage/QueueStorage.ts.md#L76), [L97](https://github.com/peer3to/state-channels-plus/blob/2df5c87f9285e4929ef14872949920c7ab2dce35/docs/spec/implementation/source/src/storage/QueueStorage.ts.md#L97)); (b) a permutation under `UNIT-TEST-QUEUE-STORAGE-2-K2F40F` ("Caps as markers", [L108](https://github.com/peer3to/state-channels-plus/blob/2df5c87f9285e4929ef14872949920c7ab2dce35/docs/spec/implementation/source/src/storage/QueueStorage.ts.md#L108)) for signature floods; (c) `REQ-QSTORE-2` text and a `.T1.P4` permutation (XO1); (d) Covers rows for the three new declarations in the verification report — after `yarn spec:refresh` they are listed as unaccounted at [verification-coverage L4028-4029](https://github.com/peer3to/state-channels-plus/blob/2df5c87f9285e4929ef14872949920c7ab2dce35/docs/spec/generated/verification-coverage.md) (local, gitignored).

> **Fix DO1-FIX**
>
> Add the four artefacts, map the three new tests (and the FR1/TY2 additions) to the new permutations, run `yarn spec:refresh`, and state in the PR body that the touched documents return to pending review. Acceptance: the refreshed coverage report lists no unaccounted declaration in `test/storage/QueueStorage.test.ts`.



---

**AI-generated review**

🟠 **[DO2] 95% — Inserting the new tests shifted the verification report's line anchors, so `REQ-QSTORE-2.T1.P3` evidence now points at the wrong declaration.** [`QueueStorage.test.ts.md` L42](https://github.com/peer3to/state-channels-plus/blob/2df5c87f9285e4929ef14872949920c7ab2dce35/docs/spec/verification/tests/test/storage/QueueStorage.test.ts.md#L42) maps "junk-first: a flood that fills the cap first…" at `#L212 (line 212)` to `REQ-QSTORE-2-VYWJAQ.T1.P3`; at head [line 212](https://github.com/peer3to/state-channels-plus/blob/2df5c87f9285e4929ef14872949920c7ab2dce35/test/storage/QueueStorage.test.ts#L212) is the new "bounds merged confirmation signatures under a signature flood" and the junk-first test moved to [L275](https://github.com/peer3to/state-channels-plus/blob/2df5c87f9285e4929ef14872949920c7ab2dce35/test/storage/QueueStorage.test.ts#L275). The refreshed coverage report lists the junk-first declaration at L275 as unaccounted, which is the generator saying the anchor is stale. `docs/spec/AGENTS.md`: "Repair a shifted test anchor only when the declaration has one unique match" — it does here.

> **Fix DO2-FIX**
>
> Re-anchor the junk-first row to `#L275 (line 275)` and every following row by its new line; run `yarn spec:refresh` and confirm the junk-first declaration disappears from the unaccounted list.



---

**AI-generated review**

🟠 **[OO1] 85% — First-come per-entry cap, or per-sender novel-signature budget?** **Human assessment needed** Decision: accept that a junk-first flood delays an honest block's signatures by one validation cycle per attacker identity (document it in `REQ-QSTORE-2`), or bound novel signatures per source so one flooder cannot displace honest copies? Evidence and recommendation in SO1. What changes after the decision: `mergeBlockCapped`/`capSignatures`, the L36-37 comment, `REQ-QSTORE-2`, and one unit test either way.

> **Fix OO1-FIX**
>
> Decide; then SO1-FIX.

momoshell:
Thanks — all six are in, and two of them found things I'd have shipped.

**FR1** was real. I proved it before fixing: a dequeue/restore cycle gave `expected 256 to be at most 128`. My commit message had claimed the cap covered "both write paths"; there were three — `createEntry`, `queueBlock`'s merge, and `restoreEntry`.

**TY3** taken — assertions are exact now, which pins that a capped merge keeps the first capful rather than discarding everything.

**XO1** taken, and wider than proposed: `REQ-QSTORE-1` and `REQ-BLOCK-PIPE-5` both needed reconciling, not just `REQ-QSTORE-2`. `REQ-BLOCK-PIPE-5` demanded arrival-order independence *and* per-entry boundedness in the same paragraph, which can't both hold above the bound.

**FY2** — real, but the proposed fix inverts it. "Cap first, attribute the whole copy" records senders for signatures the block dropped, and `unexpectedSignatures` is built from `block.allSignatures`, so those keys are never read while still consuming a bounded map. Attribution now follows retention instead.

**TY2** — four of the six were already covered: both timestamp branches at `QueueStorage.test.ts:67` and `:170`, and `queueBlock(mockBlock)` already exercises sourceless `createEntry`. The proposed "exactly 128 → no marker" also contradicted `queue.md`, which said *"reaching the cap"* — both caps actually mark on *exceeding*, so that sentence was wrong and is fixed.

**SO1** — decided as first-come, documented in `REQ-QSTORE-2` with a test pinning the residual. Two corrections to the framing: the residual is delayed finality until honest signatures are re-offered, not one cycle per identity — dropped signatures are gone, not deferred. And the per-sender budget isn't two lines; it needs sender-aware creation and merge plus new entry state, and is Sybil-bypassable anyway.

**TO1** — done. There's now an E2E that floods one parked block over the production state-transition RPC and asserts the bound.

**What the review didn't reach, and it was the bigger problem:** the cap bounded *cardinality*, which bounds nothing on its own. Confirmation values are never format-checked at ingress, so a capful of near-16MiB strings sat inside the cap. Values must now recover — the same recovery the consumer performs, since a length check accepts an `r` outside the curve order and even an in-range `r` usually has no curve point, and every such value throws in `SignerRecoveryCache`.

Two things the cap still can't promise, both filed rather than papered over: `FIND-SETTLE-1-G2CPV6` (the participant maximum is enforced where a union is proposed, not everywhere one becomes authoritative) and `FIND-QSTORE-3-1HF4V6` (retention counts signature bytes, validity counts recovered signers — so one participant can occupy slots the others need). The PR body now says the cap bounds memory and is not a proof that a needed confirmation survives.

contracts/V1/StateChannelDiamondProxy/Errors.sol
```diff
@@ -3,8 +3,18 @@ pragma solidity ^0.8.8;
...
 error ErrorDuplicateParticipant();
 
+// Upper bound on a channel's participant union, counterpart to the
+// two-participant minimum above. Off-chain agreement needs a signature from
+// every participant, so the union size bounds how many confirmation signatures
+// a valid block carries and therefore what the client must be able to retain;
+// without a maximum on chain that retention bound can only be assumed. The
+// duplicate-participant scan at open is quadratic in the union, so a bound also
+// keeps that loop's gas finite.
+uint256 constant MAX_CHANNEL_PARTICIPANTS = 32;
```
lukastanisic99: put this alongside gasLimit, p2pTime,agrementTime,ChainFallbackTime and other config options and see how it funnels down to the state machine as they all did too; No reason to put in in Errors where as you see you only have errors;

src/storage/QueueStorage.ts
```diff
@@ -29,41 +35,151 @@ export class QueueStorage {
...
+    //
+    // This is a sizing argument, not a proof that a needed signature is never
+    // dropped. Two gaps are open and recorded: the maximum is not enforced on
+    // every path that makes a participant set authoritative
+    // (FIND-SETTLE-1-G2CPV6), and retention counts signature bytes while
+    // validity counts recovered signers, so one participant able to produce
+    // many valid signatures can occupy slots others need
+    // (FIND-QSTORE-3-1HF4V6). Until those are decided this bounds memory,
+    // which is what it was added for.
+    private static readonly MAX_CHANNEL_PARTICIPANTS = 32;
```
lukastanisic99: should not be hardcoded in 2 places - will just lead to bugs - have the chain be authoritative and for this value to be set;

contracts/V1/StateChannelDiamondProxy/Errors.sol
momoshell: Moved it out. It's a constructor arg on the proxy now, stored next to `p2pTime` and the gas limit with a `DEFAULT_MAX_CHANNEL_PARTICIPANTS` fallback, and exposed through `UtilityFacet.getMaxChannelParticipants`. Both `open` and `_processJoinChannel` read the config value instead of the constant, and `LocalDiamond` plus the deployers thread it through.

src/storage/QueueStorage.ts
momoshell: Done — the literal is gone from here. `QueueStorage` takes the maximum as a constructor arg and derives the signature cap from it, and `P2pRuntimeHostRoot` reads `getMaxChannelParticipants()` off the contract right where it already reads the gas limit. The only constant left is a fallback for construction without a chain read (tests, tooling).

One thing worth flagging: a facet exercised outside a constructed proxy now reads the slot as 0 and rejects every union as too large — four `JoinChannelFacet` tests caught it. I seeded the harness with what a proxy would set rather than defaulting a 0 inside the getter, so an unconfigured deployment can't pass silently. Happy to switch to a defensive getter instead if you'd prefer that.

src/stateManager/ingest/BlockQueueManager.ts
```diff
@@ -89,30 +94,68 @@ export default class BlockQueueManager {
...
+            }
+
+            if (options.origin === BlockOrigin.NETWORK) {
+                const eligibility =
+                    await this.stateManager.membershipService.resolveSourceEligibility(
+                        options.senderAddress
+                    );
+                if (eligibility !== SourceEligibility.ELIGIBLE) {
+                    if (eligibility === SourceEligibility.ABSENT) {
+                        await this.stateManager.p2pManager.localRpc.spectateService.sync(
```
momoshell, line 125: `sync()` returns `false` straight away when a sync toward this peer is already running — it never goes to the wire, so the cache below is unchanged and we blacklist. That's fine for a real outsider, but an off-chain-promoted participant (the thing this PR adds) is absent from the chain read too, so it lands here as well: if it gossips two blocks back to back, the second one cuts it before the first sync can prove it belongs.

`syncAfterInFlight` was added for exactly this case, and its comment says a caller that treats `false` as "the peer was cut" should use it instead — the expiry probe at line 640 already does. Could this call switch to it, or is there a reason the ingress path wants the fast `false`?

src/stateManager/eventSync/EventSyncService.ts
```diff
@@ -292,19 +308,98 @@ export default class EventSyncService {
...
+        }
+        await this.eventHandler.onStateSnapshotUpdated(
+            channelId,
+            membership.snapshot,
+            {
+                blockNumber: 0,
+                logIndex: 0
+            }
+        );
+        await this.recoverOnChainSlashes(channelId, undefined, membership);
```
momoshell, line 389: The snapshot is published one call earlier, which sets `onChainEligibility` to its participants, and `membership.slashed` only reaches `knownSlashes` after this function returns. `recoverOnChainSlashes` throws when log recovery runs out of attempts, and `refreshOnChainEligibility` swallows that with a warning — so in that case a participant that is slashed on chain but still listed in the latest snapshot stays cached as eligible, and `resolveSourceEligibility` short-circuits on a positive cache, so nothing ever re-checks it.

Would it work to feed `membership.slashed` into `observeOnChainSlash` before the snapshot publish, so the slash set is always at least as fresh as the eligible set?

src/storage/QueueStorage.ts
```diff
@@ -3,81 +3,174 @@ import Clock from "@/Clock";
...
+                )
+                    continue;
+                supplied = new Set();
+                target.sourcesToSignatures.set(source, supplied);
+            }
+            acceptedSource = true;
+            for (const signature of offered) {
+                if (!supplied.has(signature)) {
+                    // At most maxChannelParticipants signatures per source
+                    if (supplied.size >= this.maxChannelParticipants) continue;
```
momoshell, line 157: The allowance is `maxChannelParticipants` per source and the author's own signature spends one of those slots (pinned by "counts the author first with maximum one"), so one source can carry at most `maxChannelParticipants - 1` confirmations. Meanwhile `BlockIngestService` and `StoredBlockMergeService` both accept signers from `previous ∪ resulting` participants — each snapshot is capped, but their union isn't, so a membership change can make a legitimate copy one signature wider than any single source may hand over.

Is the union provably within the maximum? If it is, a line saying so here would settle it; if not, sizing the allowance against the union instead would.

src/stateManager/block/BlockCommitService.ts
```diff
@@ -83,6 +83,14 @@ export default class BlockCommitService {
         sm.storage.blocks.storeBlock(block, {
             justPersist: options?.strategy instanceof DisputeValidationStrategy
         });
+        if (!(options?.strategy instanceof DisputeValidationStrategy)) {
+            sm.membershipService.publishOffChainEligibility(
```
momoshell, line 87: This publishes `previous ∪ resulting`, while `StateApplicationService` (lines 62 and 124) publishes just the resulting snapshot's participants. Both replace the whole set, so whichever ran last decides, and a peer that's in the union but not in the resulting snapshot — someone who has just left — is eligible or not depending on ordering. Since an `ABSENT` verdict now ends in a blacklist, that difference has teeth.

Should both sites agree on one set, or is the narrower one deliberate for the snapshot path?

docs/spec/specification/open-questions.md
```diff
@@ -66,6 +66,7 @@ the decision.
 Each alternative must identify the black-box cases whose stimulus or oracle changes. Resolution requires an
 update to the owning specification, assumptions/constraints, security analysis, verification plan,
 implementation mirrors, exact test mappings, generated reports, and any invalidated approval fingerprint.
+| [`OQ-SPEC-SPECTATOR-RELAY-1-V6F216`](open-questions.md#oq-spec-spectator-relay-1-v6f216) | Future spectator relaying | Plan | Current queue admission and optional promotion | Future; non-blocking |
```
momoshell, line 69: This row looks like it was meant for the Index table near the top — it's ended up at the bottom of the "Verification impact" prose, two headings below the table, so it renders as a stray table fragment and the index doesn't list the new question. `spec:refresh` passes either way, so nothing catches it.

Move it up into the Index table?

src/storage/QueueStorage.ts
```diff
@@ -3,81 +3,174 @@ import Clock from "@/Clock";
...
+            .filter(([, supplied]) =>
+                [...signatures].some((signature) => supplied.has(signature))
+            )
+            .map(([source]) => source)
+    );
+}
 
-    // Secondary index for efficient queries by coordinates
-    private blocksByCoordinates: Map<CoordinateKey, Set<Hash>> = new Map();
+// Production injects the deployed maximum.
```
momoshell, line 46: The comment block that used to live here explained why the number is safe to size retention against, and pointed at `FIND-QSTORE-3-1HF4V6`. The cap is still here, just per-source now — and the reasoning actually changed, because the author's signature now sits inside the allowance rather than being held separately.

Could the argument come back in some form? Without it the next reader only sees "at most `maxChannelParticipants`" with no way to tell whether that's a real bound or a guess.

docs/spec/audit/open-findings.md
```diff
@@ -43,10 +43,13 @@ Each should become a normal fix + test once confirmed.
...
+| <a id="find-qstore-2-evm8t9"></a>FIND-QSTORE-2-EVM8T9         | _Resolved (2026-09-16)._ The participant maximum is no longer a constant restated in two languages. It is channel configuration: `StateChannelManagerProxy` takes it as a constructor argument, stores it, and exposes it through `UtilityFacet.getMaxChannelParticipants`; both enforcement sites read that value, and the client reads it from the contract at startup and sizes `maxEntrySignatures` from it. There is no second literal left to drift. | No further action for the drift defect. A facet exercised outside a constructed proxy sees the storage slot unset and rejects every union, so harnesses that stand in for a proxy must seed it; no production path reaches a facet without the constructor having run.                                                                                                                                                                                                                                                                 |
```
momoshell, line 52: `FIND-QSTORE-2-EVM8T9` still says the client "sizes `maxEntrySignatures` from it", but that name is gone from the tree — the sizing now runs straight off `maxChannelParticipants`. Worth rewording so the resolved finding still describes the code?

src/stateManager/ingest/BlockQueueManager.ts
```diff
@@ -23,10 +28,9 @@ import P2pEventHooksUtils from "@/utils/P2pEventHooksUtils";
 import { TimeoutManager } from "@/utils/TimeoutManager";
 import type { BlockConfirmationStruct } from "@typechain-types/contracts/V1/types/DataTypes";
 
-export type IngestBlockConfirmationOptions = {
+export type IngestBlockConfirmationOptions = QueueBlockOptions & {
     onChainTimestamp?: Timestamp;
```
momoshell, line 32: `QueueBlockOptions` is a clean discriminated union, but intersecting `onChainTimestamp?` over it means the field also type-checks on the `NETWORK` branch, where line 99 throws it away. Moving it onto the calldata/proof variant would make that impossible to write by mistake — worth doing, or is the current shape keeping a call site simpler?

src/stateManager/ingest/BlockQueueManager.ts
lukastanisic99: Confirmed. A second gossip frame can hit sync's busy return and blacklist an honest newcomer before the first proof finishes. This is already recorded as deferred work in OQ-IMPL-SYNC-IN-FLIGHT-1-WC8385. I would fix the result/waiting behavior in the sync owner, as agreed, rather than add admission tracking.

src/stateManager/eventSync/EventSyncService.ts
lukastanisic99: Confirmed. The pinned read already knows the slash, but that knowledge is published only after event recovery succeeds. A recovery failure can leave the earlier positive snapshot update in the cache. Known slashes need to reach the fast cache before any positive membership from that read.

src/storage/QueueStorage.ts
lukastanisic99: This limit is for off-chain block collection. Each source keeps its own N-signature allowance, including the author, and excess contributions are dropped. We are not adding on-chain checks or changing transition or snapshot validity. Milestone proof unions are not capped by this accounting rule.

src/stateManager/block/BlockCommitService.ts
lukastanisic99: The narrower snapshot-path set is not the intended behavior for the leave block. The human engineer confirmed that a participant who just left may still supply that block after its resulting snapshot is adopted. We will preserve its previous/resulting union for source eligibility so both paths accept that late copy without blacklisting the sender for having left. The existing restriction on relaying later blocks remains.

docs/spec/specification/open-questions.md
lukastanisic99: Confirmed. The spectator-relay row is outside the Index table. It should move into that table without changing the question ID or its detailed entry.

src/storage/QueueStorage.ts
lukastanisic99: A short current-policy comment makes sense: each source gets N distinct values, including the author, so N sources contribute at most N² values. The old shared-cap argument should stay removed; it would not prove that all previous/resulting signer unions fit.

docs/spec/audit/open-findings.md
lukastanisic99: Confirmed. The resolved finding should name maxChannelParticipants and the N-source/N-value quotas. It should describe production queue injection rather than claim no fallback literal exists anywhere.

src/stateManager/ingest/BlockQueueManager.ts
lukastanisic99: Agreed as a type cleanup. Network timestamps are already ignored, so this is not a demonstrated trust-boundary bug. The options type can restrict onChainTimestamp to calldata/proof origins and keep validationStrategy common.

PR #481, opened 2026-09-12
feat: worker protocols as services over one rpc router

daiagi:
replaces #466. same goal, peer rpc model: one router per realm, transports vary, root type is the contract.

- `RpcRouter`: the core `P2PManager` already had, one concrete class. `P2PManager` extends it, sets 6 policy fields. no abstract base, no `PortRpcRouter`
- far side typed by the remote root -> `remoteRpc`. no manifests. a router holding a single transport sends a targetless request to it
- log links = the port transports. `WorkerLinks`, `logControl.ts` gone
- both `protocol.ts`, `HostCore`, every `case "..."` switch gone
- stateless services built from their methods class, no shell files

src vs #461 +103 (#466 +647).

two #466 bugs caught by the browser gates: bare `Buffer` in `RpcRouter` -> every browser peer frame dropped; webrtc bridge port minted only on the threaded path. fixed, pinned.

breaking: `p2pSigner.setIsLeader` / `disconnectFromPeers` now `Promise<void>`. await them.

carries all #461 fixes promised on #466. `docs/spec/tools/report-change-impact.js` (a deleted source counts as mapped once its report is gone) is its own commit, review it separately.

follow-ups, not here:

| what | where |
|---|---|
| sdk worker ends on a fatal (report, flush, reject client, dispose, drain) | next PR in the stack |
| log receiver: merge-on-write, one compressed stream per thread | follow-up |
| log receiver: raw per-thread stream endpoint | after the above |


lukastanisic99:
Will take a look later - playing around with some ideas around this, hopefully will be done today and then will do a PR and we can discuss.

PR #480, opened 2026-09-10
chore: pin forge v1.2.3 and check solidity formatting in ci

daiagi:
`dispute` had 4 `.sol` files `forge fmt` wanted to rewrite. the pre-commit hook only formats files staged in a normal commit; merge commits skip it, so anything arriving via a merge is never checked.

forge 1.1.0 / 1.2.3 / 1.3.0 all emit identical output for this repo; 1.8.1 reformats 18 files. the exact version is the only thing that holds.

- `.forge-version` = `v1.2.3`, README installs from it
- ci lint job: `foundry-toolchain` at v1.2.3 + `forge fmt --check`
- the 4 files reformatted; whitespace only, all 55 artifacts byte-identical with `bytecode_hash=none`

`forge fmt --check` exit 0; mangled `JoinChannelFacet.sol` -> exit 1 naming it; `forge test` 112 passed; `yarn test:parallel:distributed` 2064 passing, 0 failing.

PR #484, opened 2026-09-13
feat(contracts): carry the failed comparison in every remaining custom error

sh3ll3x3c:
## Summary

Phase 2 of the Solidity error-argument work (card 377; Phase 1 was #449). Every remaining custom error that guards a comparison now carries the operands of the check that failed, so a revert alone says what the chain held versus what the caller sent.

- ~35 errors gained arguments at ~45 construction sites across the proxy, `StateChannelCommon`, and seven facets. Naming follows Phase 1: expected/required side first (`expected*`/`actual*`, `current*`/`submitted*`), `block.timestamp` is always `currentTimestamp`, identity args use the struct field names.
- Predicate guards (helper returns `bool`) carry the identity keys that locate the state plus any value the helper already returns. `_isEvidencePeriodExpired` and `_isReduceChallengePeriodExpired` now return `(bool, uint256 periodEnd)` like `_isKillPeriodExpired`, so period errors carry `(periodEnd, currentTimestamp)` without re-deriving the formula at call sites.
- One cause, one name. `ErrorInvalidStateSnapshot` covered four different comparisons and is split into `ErrorNotGenesisSnapshot`, `RaceConditionGenesisTimestampNotAvailable` + `ErrorSnapshotGenesisTimestampMismatch`, `ErrorSnapshotDataForkMismatch`, and `ErrorInvalidStateSnapshotHash`. The threshold-signature site of `ErrorJoinChannelInvalidSignature` becomes `ErrorJoinChannelConfirmationNotThresholdSigned` (the client `switch` in `MembershipService` learns the new name); the single-commitment lookup in `_killDispute` becomes `ErrorDisputeCommitmentNotFound`.
- Seven empty-array / zero checks stay bare. Twelve declarations are deleted: ten never constructed, the split one, and `ErrorDisputeStateMachineJoiningFailed` together with the dead `_applyJoinChannelToStateMachine` that was its only constructor. `RaceConditionErrorName` drops the deleted name.
- Client handling keys on error names only; no 4-byte selector is hardcoded anywhere, so `LoggerUtils.getCustomEvmErrorMetadata` decodes every new argument with no per-error code.
- Solidity is formatted with forge v1.2.3, the version #480 pins. Three files were not 1.2.3-clean on `dispute`, so their diffs carry a few reflow hunks.

### Judgment calls

- **Error arguments are evaluated eagerly.** The design assumed `require(cond, Err(args))` evaluates `args` only on failure under via-IR (Phase 1 measured cheap in-scope args, which cannot tell lazy from eager). The review disproved it from `forge inspect … ir-optimized`. Rule now: an argument the condition did not already compute is raised with `if (!cond) revert Err(args);`. The only affected site was `ErrorNotGenesisSnapshot` (StateSnapshotFacet runtime 12,133 → 12,044 B once fixed). Recorded in the `Errors.sol` design entry.
- `ErrorJoinChannelInvalidSignature(expectedSigner, actualSigner)` still folds "wrong signer" and "invalid signature" into one `require`; when the signature is invalid the recovered address may be zero. Left as one error; splitting it is a follow-up if it matters in practice.
- `ErrorOutboundMessageBalanceMismatch` and `CantWithdrawMoreThanDeposits` report `Balance.amount` only, not the opaque `Balance.data`.
- `AStateMachine.joinChannel` lost its only on-chain caller with the dead helper; the external wrapper stays (state-machine public surface) and the architecture view says so.
- `_isEvidencePeriodExpired` is now evaluated before two cheaper checks in `uploadDispute`'s existing-window branch (a tuple return cannot sit in a boolean expression); a few extra SLOADs on that path, same semantics.
- `LocalDiamond.sol` is not forge-1.2.3-clean at base; it is left to #480 rather than reformatted here. `forge fmt --check` fails only on that file.
- `StateChannelManagerProxy.sol.md` has stale line anchors already at base; not touched here.

## Review record

- **Adversarial review round 1** (cross-company, gpt-6-astra @ max, read-only session over the full delta): 0 critical, 0 high, 3 medium, 3 low. All six verified with executed probes and accepted: M1 eager evaluation of the `ErrorNotGenesisSnapshot` hash; M2 argument-bearing revert expectations used identical values for the two compared operands, so a swapped argument order could not fail them (three site classes); M3 the atomic-deposit test asserted a harness-manufactured payload instead of the proxy's real deposit loop; L1 the malicious-snapshot e2e accepted any deposit amount below the withdrawal; L2 one spec mirror still named a deleted error; L3 stale source-report anchors. Fixed in d4a5dae1 / d5981dab.
- **Fix-verify** (independent): every fix mutation-checked on the pool. Swapping the kill-period operands, the threshold/signature counts, the genesis fork ids, and `joinChannels[i].participant → msg.sender` each turned exactly the intended tests red and nothing else; the M1 regression rebuilds the facet at 12,133 B vs 12,044 B fixed. Reachability of the new production-path deposit test: `JoinChannelFacet.joinChannel → _processJoinChannel → self-call → StateChannelManagerProxy.depositAssetsComposable loop → revert`. Round 2 not run: one-round policy, fix-verify CLEAN.
- Maintainability coverage (reviewer-produced, single scope, all six categories checked): duplication of decisions, file cohesion, function complexity, module boundaries, abstraction quality, change shape — no findings; artifact `findings.json` in the review record.
- **Implementation review** (pre-PR): approve with fixes, one blocker: a discovery runtime-port fixture passed an empty argument list for an error that now takes two values, hidden because only the forge tier and a five-file grep had run. Fixed in 492b475a / ba48a7de together with: the spectate e2e now asserts the decoded error name for the new join-channel abort case; `RaceConditionDisputeKillPeriodNotExpired` sites require the window first and revert `RaceConditionDisputeWindowNotOpen` (one Foundry test each); the two hand-assembled `{name, args}` logs go through `LoggerUtils.getCustomEvmErrorMetadata`; two operand orders and two eager storage-read arguments corrected. The operand-asserting tests for the nineteen pre-existing errors whose tests only checked the name landed in the changes below.

## Changes after review

- Four rejections now carry the compared sets instead of their sizes, so the payload names the member that differed and still says something when the counts agree: `RaceConditionOnChainSlashes(channelId, disputeSlashes, onChainSlashes)`, `ErrorDisputeCommitmentNotAvailable(channelId, forkId, committedDisputeHashes, submittedDisputeHashes)`, `ErrorJoinChannelConfirmationNotThresholdSigned(participant, thresholdParticipants, signers)`, and `ErrorDisputeStateMachineInboundProcessingFailed(..., stateMachineStateHash)` with the hash of the state the inbound walk was seeded with.
- Each derived operand is computed only on the failure branch. The join and inbound sites move from `require` to `if (!cond) revert`; the other two were already plain reverts.
- One owner per derived operand: `_disputeCommitmentHash` / `_disputeCommitmentHashes` in `DisputeUtils` (used by `areDisputesCommitted` and every commitment push or lookup, which also shrinks `LocalDiamond` by 948 B), and `_eip191SignedHash` on `UtilityFacet` shared by `verifyThresholdSigned`, `retrieveSignerAddress` and the new `retrieveSignerAddresses`. The new helper stays off the routed surface like the other stateless utility helpers, with its exclusion in the selector-routing fixture. Proxy size unchanged at 12,729 B.
- Every widened error now has a test that reaches the real revert and matches the whole payload with hand-built expected values and distinguishable operands: the four above plus the nineteen pre-existing operand gaps (manager proxy open and calldata guards, join on a disputed fork, snapshot fork/height/milestone/withdrawal guards, the common inbound, exit-balance, repeated-block, kill-period and reduced-window guards, and the dispute-facet deadline, timeout-calldata and unavailable-genesis sites). New Foundry files `DisputeFraudProofFacetPayloads.t.sol` and `StateSnapshotFacetSameFork.t.sol`.
- The spectate force-join case now observes the restored `SYNCED` status and the cleared join submission height, not only the terminal `OPENED`; the reduced-result commit obligation moves to the `StateChannelCommon` report that owns the helper.

## Deferred findings

None. Two pre-existing items reported for their owners: `LocalDiamond.sol` formatting (#480) and stale anchors in `StateChannelManagerProxy.sol.md` at base.

## Test Plan

- Contract sizes (`forge build --force --sizes`): every production facet and the proxy under 24,576 B. DisputeFraudProofFacet 22,914 → 23,100 B (1,476 B headroom), DisputeVerificationFacet 20,413 → 20,747 B, StateSnapshotFacet 11,785 → 12,044 B, JoinChannelFacet 7,284 → 7,444 B, proxy 12,604 → 12,729 B. `LocalDiamond` and `DisputeExpiryGuardHarness` were already over on `dispute`.
- Foundry: every `vm.expectRevert(bytes4)` on a changed error became a full `abi.encodeWithSelector` expectation with operands from the test's own fixture; each new error name has a test that reaches the real revert (new `StateSnapshotFacetUpdateFork.t.sol`, plus cases in `DisputeVerificationFacet.t.sol`, `JoinChannelFacet.t.sol`, `StateChannelManagerProxyDeposit.t.sol`). The Phase 1 via-IR cheatcode quirk did not reappear.
- TS: `OpenChannel.test.ts` hand-written `error X()` fragments replaced by the real ABI; `UniversalDeployment`, `StateChannelManagerBinding`, `E2E-MaliciousUpdateSnapshot`, `E2E-Spectate` updated to assert the decoded arguments by name.
- Distributed pool at HEAD: canonical full gate `yarn test:parallel:distributed` 2064 passing / 0 failing after the fixture fix (one stale Foundry task-count fixture failed first, fixed, its file rerun 23/23; two starvation-signature tests passed on their scheduled rerun); `--forge-only` 11/11; `--grep "E2E: Malicious updateSnapshot|stateChannelManager binding|Universal Deployment|E2E: Spectate Service|Open Channel - MathStateChannel"` 56/56. Typecheck node + browser, `lint:imports`, `spec:refresh`, `spec:ids:check`, `spec:impact` (incl. `--base 77ef395a`), `git diff --check` clean.
- After review, at 031ca31e: `--grep "^(E2E: Spectate Service|E2E: Join channel race conditions|ContractCaller and ContractErrors|Unit: DisputeManager|E2E: Dispute Manager)"` 114/114 (five starvation-signature reschedules, all passed); `--forge-only` 13/13 covering all thirteen Foundry files. After the fixture sharing at e3ddcecb: `--forge-only` 13/13 and `--grep "ContractCaller and ContractErrors"` 7/7. `forge build`, both typechecks, `lint:imports`, `forge fmt --check` (except the pre-existing `LocalDiamond.sol`), `spec:refresh`, `spec:ids:check`, `spec:impact` clean.
- Spec: 18 maintained docs mirrored (Errors.sol design record, facet and helper source reports, architecture views, `evmErrorHandler` union mirror, verification reports incl. the new Foundry file). They reappear as pending in the generated review report until re-approved.


sh3ll3x3c:
**2026-09-15 — merged `dispute` into the branch (c776e448)**

The branch went CONFLICTING after #485 (internal services) and #461 (crash logs) landed on `dispute`. Resolved by merge, not rebase, so the reviewed commits keep their SHAs.

- 8 conflicting files: `JoinChannelFacet.sol` (dispute only reformatted the threshold-signature `require`; the PR's `ErrorJoinChannelConfirmationNotThresholdSigned` version wins), the `E2E-MaliciousUpdateSnapshot` import block, and six spec reports where the PR's re-anchored error signatures met the #485 report rewrites. Spectate report links follow the `src/rpc/network/services` move; anchors re-computed from the merged files.

Gates on the merge commit: `forge fmt --check`, `yarn compile`, both tsc builds, `lint:imports`, `spec:refresh`, `spec:impact` clean. Distributed runs: focused Mocha tier over the touched suites 228 passing, 0 failing; full forge tier 11 suites passing, 0 failing.

contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol
```diff
@@ -382,7 +384,7 @@ contract DisputeFraudProofFacet is StateChannelCommon {
             if (!found) return _valid(dispute.input.disputer);
         }
 
-        revert RaceConditionOnChainSlashes();
+        revert RaceConditionOnChainSlashes(dispute.input.channelId, disputeSlashes.length, onChainSlashes.length);
```
lukastanisic99: since this will rarely trigger (gas is less of a problem + in cases where it triggers it's probably a staticCall anyway (eth_call)), better to add the actuall arrays than the lengths, so the agent has enough context to deduct 'why is this address in or isn't in' - like this 5!=6 isn't usefull much;

contracts/V1/StateChannelDiamondProxy/DisputeVerificationFacet.sol
```diff
@@ -200,11 +200,17 @@ contract DisputeVerificationFacet is StateChannelCommon {
         DisputeData storage disputeData = disputeData[channelId];
         DisputeWindow storage disputeWindow = disputeData.disputeWindowMap[disputes[0].input.forkId];
         //require all disputes are part of commitment
-        require(areDisputesCommitted(disputeWindow, disputes), ErrorDisputeCommitmentNotAvailable());
+        // `if (!...) revert` because the committed count is a storage read
+        // `areDisputesCommitted` makes internally and the caller does not (P4).
+        if (!areDisputesCommitted(disputeWindow, disputes)) {
+            revert ErrorDisputeCommitmentNotAvailable(
+                channelId, forkId, disputeWindow.evidence.disputeCommitments.length, disputes.length
```
lukastanisic99: again would be more usefull to return the array of dispute hashes so at least we know which hash is missing is or is not commited and the agent can figure out which dispute is that;

contracts/V1/StateChannelDiamondProxy/DisputeVerificationFacet.sol
```diff
@@ -256,9 +262,14 @@ contract DisputeVerificationFacet is StateChannelCommon {
...
         }
 
-        // require(_canParticipateInDisputes(channelId, msg.sender), ErrorCantParticipateInDispute());
         // require that provided disputes correspond to committed set
-        require(areDisputesCommitted(disputeWindow, disputes), ErrorDisputeCommitmentNotAvailable());
+        // `if (!...) revert` because the committed count is a storage read
+        // `areDisputesCommitted` makes internally and the caller does not (P4).
+        if (!areDisputesCommitted(disputeWindow, disputes)) {
+            revert ErrorDisputeCommitmentNotAvailable(
+                channelId, forkId, disputeWindow.evidence.disputeCommitments.length, disputes.length
```
lukastanisic99: -||-

contracts/V1/StateChannelDiamondProxy/JoinChannelFacet.sol
```diff
@@ -57,27 +62,38 @@ contract JoinChannelFacet is StateChannelCommon {
...
         // Check threshold from the current eligibility set
         address[] memory thresholdParticipants = _getOnChainThresholdSet(channelId);
         (bool isValid,) = UtilityFacet(utilityFacetAddress).verifyThresholdSigned(
             thresholdParticipants, sjc.encodedJoinChannel, joinChannelConfirmation.signatures
         );
-        require(isValid, ErrorJoinChannelInvalidSignature());
+        require(
+            isValid,
+            ErrorJoinChannelConfirmationNotThresholdSigned(
+                jc.participant, thresholdParticipants.length, joinChannelConfirmation.signatures.length
```
lukastanisic99: -||-, better to have the actuall addresses than lengths

contracts/V1/StateChannelDiamondProxy/StateChannelCommon.sol
```diff
@@ -540,7 +540,15 @@ contract StateChannelCommon is StateChannelManagerStorage, StateChannelManagerEv
         for (uint256 i = 0; i < inboundMessageBlocks.length; i++) {
             for (uint256 j = 0; j < inboundMessageBlocks[i].messages.length; j++) {
                 bool success = stateMachineImplementation.processInboundMessage(inboundMessageBlocks[i].messages[j]);
-                require(success, ErrorDisputeStateMachineInboundProcessingFailed());
+                require(
+                    success,
+                    ErrorDisputeStateMachineInboundProcessingFailed(
+                        i,
```
lukastanisic99, line 547: also hash(encodedStateMachineState) so we can understand at which state the operation failed

lukastanisic99 (commented):
**AI-generated review**

🟠 **[XO1] 95% — The new membership coverage credit exceeds the test assertions.** The new P14 requires restoring `SYNCED`, clearing the force-join marker, and aborting ([definition](https://github.com/peer3to/state-channels-plus/blob/c776e448059f266cb5bb850bfc1d197ca34914f9/docs/spec/implementation/source/src/stateManager/membership/MembershipService.ts.md#L90)). Its credited test reads only final `OPENED` and `isDisposed` ([assertions](https://github.com/peer3to/state-channels-plus/blob/c776e448059f266cb5bb850bfc1d197ca34914f9/test/e2e/E2E-Spectate.test.ts#L1022)). The two cleanup calls are separate from abort, and [abort](https://github.com/peer3to/state-channels-plus/blob/c776e448059f266cb5bb850bfc1d197ca34914f9/src/stateManager/StateManager.ts#L283) sets `OPENED` itself. Removing the cleanup calls would not invalidate those assertions. The new Covers assignment therefore hides two unobserved effects.

> **Fix XO1-FIX**
>
> Observe the cleanup effects through the public lifecycle/storage surface, or split P14 so this case receives only the terminal-abort credit and the untested cleanup remains an explicit gap.



---

**AI-generated review**

🟠 **[TO1] 90% — Commitment-set counts are untested at both callers.** [contracts/V1/StateChannelDiamondProxy/DisputeVerificationFacet.sol:206](https://github.com/peer3to/state-channels-plus/blob/c776e448059f266cb5bb850bfc1d197ca34914f9/contracts/V1/StateChannelDiamondProxy/DisputeVerificationFacet.sol#L206): Neither the challenge nor finalize rejection asserts the channel, fork, committed count, and submitted count. The new single-missing-commitment test exercises a different error.

> **Fix TO1-FIX**
>
> Exercise both public callers with unequal committed/submitted counts and assert the full payload.



---

**AI-generated review**

🟠 **[TO2] 90% — Inbound failure indices have no production assertion.** [contracts/V1/StateChannelDiamondProxy/StateChannelCommon.sol:545](https://github.com/peer3to/state-channels-plus/blob/c776e448059f266cb5bb850bfc1d197ca34914f9/contracts/V1/StateChannelDiamondProxy/StateChannelCommon.sol#L545): No test asserts the new block index, message index, participant, and message type returned when state-machine processing returns false.

> **Fix TO2-FIX**
>
> Fail at distinct nonzero block/message indices and assert all four fields from the production loop.



---

**AI-generated review**

🟠 **[TO3] 90% — Withdrawal failure indices have no production assertion.** [contracts/V1/StateChannelDiamondProxy/StateSnapshotFacet.sol:152](https://github.com/peer3to/state-channels-plus/blob/c776e448059f266cb5bb850bfc1d197ca34914f9/contracts/V1/StateChannelDiamondProxy/StateSnapshotFacet.sol#L152): The new withdrawal-failed payload has no assertion identifying the exact failed message.

> **Fix TO3-FIX**
>
> Make the withdrawal return false at distinct block/message indices and assert both indices and its participant.



---

**AI-generated review**

🟡 **[TY4] 90% — The processed-participant count is untested.** [contracts/V1/StateChannelDiamondProxy/StateChannelManagerProxy.sol:229](https://github.com/peer3to/state-channels-plus/blob/c776e448059f266cb5bb850bfc1d197ca34914f9/contracts/V1/StateChannelDiamondProxy/StateChannelManagerProxy.sol#L229): No assertion verifies that this count comes from successful processed joins.

> **Fix TY4-FIX**
>
> Use fewer than two successful joins and assert the processed count, including a partial deposit outcome.



---

**AI-generated review**

🟡 **[TY5] 90% — Duplicate-calldata payload fields are untested.** [contracts/V1/StateChannelDiamondProxy/StateChannelManagerProxy.sol:168](https://github.com/peer3to/state-channels-plus/blob/c776e448059f266cb5bb850bfc1d197ca34914f9/contracts/V1/StateChannelDiamondProxy/StateChannelManagerProxy.sol#L168): The existing ContractErrors case checks the name only. It cannot catch a wrong stored commitment, fork, transaction count, or participant.

> **Fix TY5-FIX**
>
> Extend the real duplicate-post case to assert all four operands.



---

**AI-generated review**

🟡 **[TY6] 90% — Wrong-author operand order is untested.** [contracts/V1/StateChannelDiamondProxy/StateChannelManagerProxy.sol:172](https://github.com/peer3to/state-channels-plus/blob/c776e448059f266cb5bb850bfc1d197ca34914f9/contracts/V1/StateChannelDiamondProxy/StateChannelManagerProxy.sol#L172): The existing wrong-author test checks only the error name.

> **Fix TY6-FIX**
>
> Assert distinct expected author and actual sender in the real rejection.



---

**AI-generated review**

🟡 **[TY7] 90% — Join snapshot hashes are untested.** [contracts/V1/StateChannelDiamondProxy/JoinChannelFacet.sol:56](https://github.com/peer3to/state-channels-plus/blob/c776e448059f266cb5bb850bfc1d197ca34914f9/contracts/V1/StateChannelDiamondProxy/JoinChannelFacet.sol#L56): The stale-snapshot E2E checks the name and lifecycle outcome, but not the current/submitted hashes.

> **Fix TY7-FIX**
>
> Assert both distinct hashes from the stale snapshot failure.



---

**AI-generated review**

🟡 **[TY8] 90% — The disputed-join identity payload is untested.** [contracts/V1/StateChannelDiamondProxy/JoinChannelFacet.sol:74](https://github.com/peer3to/state-channels-plus/blob/c776e448059f266cb5bb850bfc1d197ca34914f9/contracts/V1/StateChannelDiamondProxy/JoinChannelFacet.sol#L74): No source-produced assertion verifies the rejected channel and fork.

> **Fix TY8-FIX**
>
> Trigger the disputed-fork join rejection and assert channelId and forkId.



---

**AI-generated review**

🟡 **[TY9] 90% — Failed fork traversal has no payload assertion.** [contracts/V1/StateChannelDiamondProxy/StateSnapshotFacet.sol:47](https://github.com/peer3to/state-channels-plus/blob/c776e448059f266cb5bb850bfc1d197ca34914f9/contracts/V1/StateChannelDiamondProxy/StateSnapshotFacet.sol#L47): The new current/target fork operands have no test oracle.

> **Fix TY9-FIX**
>
> Use a valid target genesis that cannot be reached through finalized reductions and assert distinct current and target forks.



---

**AI-generated review**

🟡 **[TY10] 90% — Old snapshot heights have no source-produced payload assertion.** [contracts/V1/StateChannelDiamondProxy/StateSnapshotFacet.sol:66](https://github.com/peer3to/state-channels-plus/blob/c776e448059f266cb5bb850bfc1d197ca34914f9/contracts/V1/StateChannelDiamondProxy/StateSnapshotFacet.sol#L66): Injected reduction errors exercise client handling but do not prove the compared heights.

> **Fix TY10-FIX**
>
> Reject a stale snapshot and assert current and submitted heights with unequal values.



---

**AI-generated review**

🟡 **[TY11] 90% — Invalid milestone counts have no payload assertion.** [contracts/V1/StateChannelDiamondProxy/StateSnapshotFacet.sol:70](https://github.com/peer3to/state-channels-plus/blob/c776e448059f266cb5bb850bfc1d197ca34914f9/contracts/V1/StateChannelDiamondProxy/StateSnapshotFacet.sol#L70): No assertion pins the fork and proof/snapshot lengths used for invalid milestone evidence.

> **Fix TY11-FIX**
>
> Use a failing milestone input with distinguishable collection lengths and assert all fields.



---

**AI-generated review**

🟡 **[TY12] 90% — Outbound balance mismatch operands are untested.** [contracts/V1/StateChannelDiamondProxy/StateChannelCommon.sol:564](https://github.com/peer3to/state-channels-plus/blob/c776e448059f266cb5bb850bfc1d197ca34914f9/contracts/V1/StateChannelDiamondProxy/StateChannelCommon.sol#L564): The over-withdrawal E2E covers a later total-balance error, not this embedded-exit/message comparison.

> **Fix TY12-FIX**
>
> Assert participant plus distinct encoded-exit expected amount and message actual amount at this guard.



---

**AI-generated review**

🟡 **[TY13] 90% — Duplicate inbound persistence identity is untested.** [contracts/V1/StateChannelDiamondProxy/StateChannelCommon.sol:265](https://github.com/peer3to/state-channels-plus/blob/c776e448059f266cb5bb850bfc1d197ca34914f9/contracts/V1/StateChannelDiamondProxy/StateChannelCommon.sol#L265): No payload assertion pins the channel and already-persisted block hash.

> **Fix TY13-FIX**
>
> Persist then repeat the same block through the production helper and assert both identity fields.



---

**AI-generated review**

🟡 **[TY14] 90% — Challenge expiry operands are untested.** [contracts/V1/StateChannelDiamondProxy/DisputeVerificationFacet.sol:213](https://github.com/peer3to/state-channels-plus/blob/c776e448059f266cb5bb850bfc1d197ca34914f9/contracts/V1/StateChannelDiamondProxy/DisputeVerificationFacet.sol#L213): The helper now returns a deadline but no source-produced assertion checks this caller’s deadline/current-time order.

> **Fix TY14-FIX**
>
> Challenge after expiry and assert the calculated challenge end and a later current timestamp.



---

**AI-generated review**

🟡 **[TY15] 90% — Not-yet-expired kill deadlines are untested at both sites.** [contracts/V1/StateChannelDiamondProxy/StateChannelCommon.sol:645](https://github.com/peer3to/state-channels-plus/blob/c776e448059f266cb5bb850bfc1d197ca34914f9/contracts/V1/StateChannelDiamondProxy/StateChannelCommon.sol#L645): The new tests cover expired kill windows and missing windows. They do not pin the active-window deadline pair here or in the WrongGenesis handler.

> **Fix TY15-FIX**
>
> Exercise both active-window rejection sites with unequal deadline/current time and assert the pair.



---

**AI-generated review**

🟡 **[TY16] 90% — Already-reduced fork operands are untested.** [contracts/V1/StateChannelDiamondProxy/StateChannelCommon.sol:648](https://github.com/peer3to/state-channels-plus/blob/c776e448059f266cb5bb850bfc1d197ca34914f9/contracts/V1/StateChannelDiamondProxy/StateChannelCommon.sol#L648): Synthetic client errors do not verify original, existing reduced, and submitted reduced forks.

> **Fix TY16-FIX**
>
> Drive the common helper against an already-reduced window and assert three distinguishable forks.



---

**AI-generated review**

🟡 **[TY17] 90% — Upload evidence expiry has no production deadline assertion.** [contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol:93](https://github.com/peer3to/state-channels-plus/blob/c776e448059f266cb5bb850bfc1d197ca34914f9/contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol#L93): The runtime-port fixture constructs the error itself; it cannot prove the upload calculates the right evidence end.

> **Fix TY17-FIX**
>
> Exercise the upload deadline guard and assert the computed end and current time.



---

**AI-generated review**

🟡 **[TY18] 90% — On-chain slash counts have no production assertion.** [contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol:387](https://github.com/peer3to/state-channels-plus/blob/c776e448059f266cb5bb850bfc1d197ca34914f9/contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol#L387): DisputeManager tests inject this error to check consumption, leaving its channel and two count operands unverified.

> **Fix TY18-FIX**
>
> Reach the Solidity slash guard with distinguishable counts and assert the full payload.



---

**AI-generated review**

🟡 **[TY19] 90% — Unexpected-calldata identity is untested in both timeout branches.** [contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol:517](https://github.com/peer3to/state-channels-plus/blob/c776e448059f266cb5bb850bfc1d197ca34914f9/contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol#L517): Client error injection cannot prove the fork, height, participant, and stored commitment selected here or at the second construction site.

> **Fix TY19-FIX**
>
> Exercise both timeout proof branches and assert their complete calldata identity payloads.



---

**AI-generated review**

🟡 **[TY20] 90% — Unavailable-genesis identity is untested outside updateFork.** [contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol:494](https://github.com/peer3to/state-channels-plus/blob/c776e448059f266cb5bb850bfc1d197ca34914f9/contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol#L494): The new snapshot test covers updateFork, but the two dispute-proof construction sites and the WrongGenesis handler select their own origin/fork operands.

> **Fix TY20-FIX**
>
> Assert the identity triple at each remaining source site, using distinct origin and target forks.



---

**AI-generated review**

🟡 **[DY1] 95% — The new shared-helper test obligation is owned by the wrong source report.** New P22 sits under [DisputeVerificationFacet](https://github.com/peer3to/state-channels-plus/blob/c776e448059f266cb5bb850bfc1d197ca34914f9/docs/spec/implementation/source/contracts/V1/StateChannelDiamondProxy/DisputeVerificationFacet.sol.md#L105), but its test calls a test-only wrapper directly into [StateChannelCommon._commitToDisputeReducedResult](https://github.com/peer3to/state-channels-plus/blob/c776e448059f266cb5bb850bfc1d197ca34914f9/contracts/V1/StateChannelDiamondProxy/StateChannelCommon.sol#L631). It executes no production DisputeVerificationFacet entry point. The single-file UNIT ownership rule requires the obligation to belong to the common source report.

> **Fix DY1-FIX**
>
> Allocate the obligation under StateChannelCommon and update the test Covers assignment without renumbering surviving IDs.

sh3ll3x3c:
Decision (2026-09-16), following the five inline comments.

All five accepted. The errors carry the compared data, not counts:
- `RaceConditionOnChainSlashes`: both slash arrays.
- `ErrorDisputeCommitmentNotAvailable` (both sites): the committed dispute hashes and the hashes of the submitted disputes.
- `ErrorJoinChannelConfirmationNotThresholdSigned`: the threshold participant addresses and the recovered signer addresses.
- `ErrorDisputeStateMachineInboundProcessingFailed`: adds `keccak256(encodedStateMachineState)`.

Two of these sites still use `require(cond, Error(args))`. solc 0.8.34 via IR evaluates the error arguments before the condition, so array copies and the state hash would run on every successful join and every inbound message. Both sites switch to `if (!cond) revert`, the same rule the other sites in this PR already follow.

The operand assertions from the review (TO1-TO3, TY4-TY20) go in the same fix round, written against the new signatures. XO1 and DY1 are spec-report fixes and land with them.

contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol
sh3ll3x3c: Agreed. `RaceConditionOnChainSlashes` will carry `disputeSlashes` and `onChainSlashes` themselves instead of their lengths. This site is already a plain `revert`, so nothing changes on the happy path.

contracts/V1/StateChannelDiamondProxy/DisputeVerificationFacet.sol
sh3ll3x3c: Agreed. The error will carry `disputeWindow.evidence.disputeCommitments` and the `keccak256(abi.encode(dispute))` of each submitted dispute, so the missing or extra hash falls out by comparing the two arrays. The submitted hashes are computed only on the failure path; this site is already `if (!...) revert`.

contracts/V1/StateChannelDiamondProxy/DisputeVerificationFacet.sol
sh3ll3x3c: Same payload as the challenge site above: committed hashes and submitted hashes. Both callers get the same shape.

contracts/V1/StateChannelDiamondProxy/JoinChannelFacet.sol
sh3ll3x3c: Agreed. The error will carry the threshold participant addresses and the signer addresses recovered from the submitted signatures. This site is a `require`, and solc 0.8.34 via IR evaluates the arguments before the condition, so it switches to `if (!isValid) revert` so the array copies and the recovery happen only on failure.

contracts/V1/StateChannelDiamondProxy/StateChannelCommon.sol
sh3ll3x3c: Agreed. It will add `keccak256(encodedStateMachineState)`, the state set before the loop, next to the block and message indices. This `require` sits inside the per-message loop, and solc evaluates its arguments eagerly, so the site switches to `if (!success) revert` so the hash is computed only on failure.

contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol
lukastanisic99: @sh3ll3x3c I don't see any changes here

sh3ll3x3c: Landed in d8dc2903: `RaceConditionOnChainSlashes(channelId, disputeSlashes, onChainSlashes)` carries both address arrays. Test with the full payload in d20de0a9 (`DisputeVerificationFacet.t.sol`, `test_handleDisputeOnChainSlashesNotSubset_listedSlashesAreSubsetRevertsCarryingBothSlashSets`).

contracts/V1/StateChannelDiamondProxy/DisputeVerificationFacet.sol
sh3ll3x3c: Landed in d8dc2903: `ErrorDisputeCommitmentNotAvailable(channelId, forkId, committedDisputeHashes, submittedDisputeHashes)`. The per-dispute hash now has one owner, `_disputeCommitmentHash` in `DisputeUtils`, used by `areDisputesCommitted` and every commitment push or lookup. Submitted hashes are computed only on the failure branch. Test in d20de0a9 substitutes one dispute at equal length and matches both arrays.

contracts/V1/StateChannelDiamondProxy/DisputeVerificationFacet.sol
sh3ll3x3c: Same payload as the challenge site, landed in d8dc2903. Test in d20de0a9 submits an extra dispute through `challengeDisputeReduction` and matches both hash arrays.

contracts/V1/StateChannelDiamondProxy/JoinChannelFacet.sol
sh3ll3x3c: Landed in d8dc2903: `ErrorJoinChannelConfirmationNotThresholdSigned(participant, thresholdParticipants, signers)`. Signers come from the new `UtilityFacet.retrieveSignerAddresses`, which shares the EIP-191 digest with `verifyThresholdSigned`, and is called only on the failure branch; the site is now `if (!isValid) revert`. Tests in d20de0a9: the existing shortfall case matches the exact arrays, and a new case supplies as many signatures as the threshold has members with one from a non-member, which the counts could not tell apart.

contracts/V1/StateChannelDiamondProxy/StateChannelCommon.sol
sh3ll3x3c: Landed in d8dc2903: `ErrorDisputeStateMachineInboundProcessingFailed` appends `keccak256(encodedStateMachineState)` of the seed state, and the site is `if (!success) revert` so the hash is computed only on failure. Tests in d20de0a9 and b4f69079 match the hash and the block and message indices at distinct non-zero positions.

sh3ll3x3c:
2026-09-17: the review fixes are pushed, head 031ca31e (four commits on top of c776e448).

- d8dc2903: the four rejections carry the compared sets instead of their sizes; join and inbound sites move to `if (!cond) revert`; one owner for the commitment hash (`DisputeUtils`) and for the EIP-191 digest (`UtilityFacet`).
- d20de0a9: full-payload tests for those four errors, Foundry and the spectate e2e.
- b4f69079: operand assertions from production reverts for the nineteen pre-existing errors whose tests only checked the name; two new Foundry files.
- 031ca31e: the spectate force-join case observes the restored `SYNCED` and the cleared join height; the reduced-result commit obligation moves to the `StateChannelCommon` report.

Pool: focused Mocha 114/114, `--forge-only` 13/13. Local checks clean. PR body updated with a "Changes after review" section.

PR #485, opened 2026-09-13
Internal RPC services

lukastanisic99:
# Human

Since we cannot use ordinary shared-memory multithreading for our JavaScript objects in V8 as we would in other languages, we use workers. Each worker has its own separate environment, like a separate process, so we have to pass messages between them. That gives us the same message-passing problem we already solved for communication between peers. We can piggyback on our existing custom RPC solution, including its type safety, to handle internal communication and reduce complexity.

Complexity is unavoidable, but we can structure it well in a couple of places that we understand once and hopefully never look at again. Creating services should be easy, trivial. Services share common creation and disposal functionality, and we can compose them. When thinking about services, we should focus on their internal logic and how to write them as deep modules. We should not have to keep thinking about how to wire them together or deal with the same problems we already solved.

We already had a solution at the network level, but network and internal are logically and semantically different problems with different trust models. Both transports are like a telephone line, sending bytes to the other side. Internal transport can also transfer ownership of things such as ports. Network transport has addresses associated with network peers; internal connections do not use that same addressing model.

Our network peers are currently homogeneous, all running the same global RPC service. Internal services piggyback on that work but support heterogeneous systems, with services composed as building blocks.

Roots are counterparts to our main network RPC service, but they are heterogeneous, so we have multiple roots composing different services. They provide thread entry points and inherit shared base functionality: implicit connections, parent and children, disposal and shared loggers + more can be added in the future.

`createRoot` handles all the wiring, and remote control hides the shared complexity so you can use it as a black box. You can add and compose as many services as you want without caring about the wiring. One function selects a new thread or the same thread, so you can quickly spin up more threads to partition further. This gives us a clear, intuitive way to create roots as microservices, building blocks reused across the heterogeneous stack.

Heterogeneous network peers might be possible later, but for now we have the network/internal split.

# Agent 

## Design

- **Shared routing, separate trust boundaries.** Common code registers requests before sending, settles responses, dispatches endpoints and rejects pending work when a transport closes. Network routing keeps peer admission, framing and blacklist policy. Internal transports expose runtime services without registering them as peer-facing services. Typed proxies derive their arguments and results from endpoint declarations.
- **One handle per remote root.** `RemoteRoot` owns access to the typed connection, transport state, diagnostics and disposal. Roots keep one connection map; parent and child relationships are views of it. Common creation handles attachment, readiness and startup failure. Platform adapters own Node and browser worker operations.
- **Communication readiness is separate from application setup.** `setupP2pRuntime` owns signers, deployment and application assembly. The client root owns host communication, and the SDK host owns its executor and fallback WebRTC bridge. A ready communication port does not imply deployment has finished.
- **Disposal follows ownership.** Root disposal recursively disposes children and releases local resources and logger descendants. Abort closes the owning runtime in both placements. Cleanup helpers attempt every step in order and preserve the first failure. Concurrent disposal calls share completion; later quiescence calls drain new work.
- **Worker exit follows a disposal handshake.** A self-disposing child requests acknowledgement before exiting. Parent-requested disposal attempts its response before final shutdown. When the parent closes a worker port, it marks shutdown expected first, preventing a normal exit from being reported as fatal. Unexpected exits still report their cause and reject pending requests.
- **Loggers upload; services coordinate.** A logger family shares its store and one optional logger-service reference through `sharedResources`. The service tracks stores, not every child logger. Standalone loggers still upload without a service. An upload triggers other attached stores and sends generation-based gossip over runtime connections; coalescing suppresses repeated uploads. Store limits remain finite, and the last logger releases shared upload resources.
- **Fallback WebRTC belongs to each host.** Each SDK host owns its bridge factory, so multiple inline hosts cannot replace each other's bridge. The worker bridge retains callback services and forwards negotiation to the main-thread broker, including when the client itself runs inside a worker.

## Benefits

- Startup, request settlement, error handling and cleanup fixes apply across runtime types and placements.
- Root ownership makes abort and failed setup release workers, transports, logger stores and crash listeners consistently.
- Typed calls reduce protocol drift, and request diagnostics retain the actual wire request ID.
- Store-based gossip avoids walking logger trees or maintaining an acknowledged distributed collection round.
- Real worker fault tests cover the default client → SDK worker → executor worker path. Browser worker and WebRTC gates now run in CI.

## Tradeoffs and compatibility

- Inline execution now uses asynchronous ports and full runtime disposal. It remains supported, but it does not provide worker isolation or equivalent performance under load.
- Acknowledged disposal adds a round trip. It deliberately has no request timeout while a live child finishes cleanup; transport loss must settle pending work.
- Upload gossip is best-effort coordination, not proof that every root uploaded successfully. It assumes gossip normally arrives much faster than the coalescing window. Generations suppress stale triggers, and newly connected services synchronize their counters.
- The WebRTC bridge keeps an extra local negotiation hop to preserve callback ownership and attachment ordering.
- Shared infrastructure has a wider impact when changed, so lifecycle tests alone are insufficient: the full distributed suite and both browser gates remain necessary.
- This changes the public API: consumers must use `ANetworkRpcService` instead of the removed `ARpcService` alias. The old public executor-factory and worker-start APIs are also removed. Inline disposal now closes the runtime and later calls reject.
- The refactor does not claim a measured performance improvement. Recorded per-test measurements were within noise of the prior implementation, and production code grew by 267 physical lines against the baseline. The goal is clearer ownership and fewer duplicate implementations.

## Validation

Final distributed run-474 passed **2,400 tests**, with zero failures and seven recovered starvation cases. The browser worker gate passed **14 cases**; direct and nested-worker WebRTC passed. Both TypeScript builds, compilation, import lint, formatting, whitespace and specification checks passed.

The final runs include regressions for busy-parent disposal, lost-parent cleanup, unexpected SDK and executor exits, disposal during pending calls, repeated quiescence and logger resource release. Earlier full runs exposed three test setup races; those now use explicit ordering or existing shared staging without increasing timeouts or changing their behavior assertions. Specification and verification reports accompany the implementation.


src/rpc/router/ARpcRouter.ts
```diff
@@ -0,0 +1,192 @@
...
+        return setTimeout(callback, timeoutMs);
+    }
+
+    protected cancelRpcTimeout(timeout: ReturnType<typeof setTimeout>): void {
+        clearTimeout(timeout);
+    }
+
+    protected restoreRpcError(error: unknown): Error {
+        return new Error(
+            typeof error === "string" ? error : "RPC request failed on the peer"
```
daiagi, line 92: 3 routers where 1 do. `ARpcRouter` = 3 abstract (`rpcRoot`, `defaultRequestTimeoutMs`, `admitRpcResponse`) + 3 override hooks (`scheduleRpcTimeout`, `cancelRpcTimeout`, `restoreRpcError`), then `NetworkRpcRouter` + `InternalRpcRouter` fill them.

what differs per realm = timer, default timeout, same-sender check, error decode, bad-frame policy. all values/fns -> inject into one concrete router. string vs object frame stay on transport (`onRpc` vs `onRpcFrame`).

not about lines (384 here, 407 in #481 `RpcRouter`). about one class to read instead of hook chain across 3 files. see `src/rpc/RpcRouter.ts` on `feat/worker-rpc-v2`.

src/rpc/internal/AInternalRpcService.ts
```diff
@@ -0,0 +1,57 @@
...
+    public abstract createRPCMethods(sender: InternalTransport): TMethods;
+
+    protected prepareError(error: unknown): unknown {
+        return serializeError(error);
+    }
+
+    protected responseSendFailed(error: unknown): void {
+        // The transport reports uncaught dispatch failures to the owning root.
+        throw error;
+    }
```
daiagi, line 24: `prepareError` and `responseSendFailed` never overridden, no service in `src` touch them -> hooks with one value. inline both into `runRPC`:

```ts
prepareError: (error) => serializeError(error),
reply: (response) => {
    try {
        sender.sendRpcResponse(response);
    } finally {
        // Disposal must finish even when the parent cannot receive its reply.
        this.afterResponse(rpc, sender);
    }
}
```

`afterResponse` keep, `RuntimeLifecycleService` + `SdkSetupService` override it.

src/rpc/internal/AInternalRpcRoot.ts
```diff
@@ -0,0 +1,215 @@
...
+        const handlerExecutionContext = context?.handlerExecutionContext;
+        this.startContext = context;
+        this.router = new InternalRpcRouter(this, timeoutMs);
+        if (handlerExecutionContext) {
+            const onMessage = this.router.onMessage.bind(this.router);
+            this.router.onMessage = (message, sender) =>
+                handlerExecutionContext.runHandler(() =>
+                    onMessage(message, sender)
+                );
+        }
```
daiagi, line 75: ctor overwrite `this.router.onMessage` with wrapped copy. get why both wraps exist (this one -> ALS peer context for responses/detached work, `AInternalRpcService.ts` L40 -> stamp before error serialized). but reassigning method on router instance from outside = hidden. router take wrap as ctor arg instead?

src/rpc/internal/AInternalRpcService.ts
lukastanisic99: Agreed. Neither internal-service hook has an override, and the send-failure hook only rethrows. The implementation now serializes directly and sends directly inside the existing try/finally. afterResponse stays because lifecycle and failed setup depend on it, including when sending the reply fails.

src/rpc/internal/AInternalRpcRoot.ts
lukastanisic99: Agreed: the router should accept the inbound context explicitly instead of having onMessage replaced after construction. The implementation moves only the outer wrapper into the router and keeps the service-level wrapper, because endpoint errors must receive their peer stamp before serialization.

src/rpc/router/ARpcRouter.ts
lukastanisic99: I'm not convinced that composition is better than inheritance here, since this entire PR is about creating a clear mental split between Network and Internal. When you see one, you should have the full information.

Composition, in my opinion, makes the split less clear and makes it dependent on the passed object. It's less of a deep module and more intertwined with different things, where you have to think “how was this constructed, with which arguments?” instead of clearly reading “NetworkRpcRouter” and that on its own telling you everything you need to know. It's verbose enough and isolated enough.

daiagi (approved):
poker-web break after merge: `src/index.ts` L12-13 / L62-63 rename `ARpcService` -> `ANetworkRpcService` and `ARpcMethods` -> `ANetworkRpcMethods`. PR body only mention first one.

poker-web link sdk (`link:../state-channels-plus`) and extend both old names on `develop` + `backend-cleanup`:

- `src/backend/zk/rpc/join/ZkJoinRpcMethods.ts`
- `src/backend/zk/rpc/join/ZkJoinService.ts`
- `src/backend/zk/rpc/negotiation/ZkOpenChannelNegotiationRpcMethods.ts`
- `src/backend/zk/rpc/negotiation/ZkOpenChannelNegotiationService.ts`

-> compile fail there. mechanical rename, just coordinate.

PR #478, opened 2026-09-08
docs(spec): record timeout calldata and milestone finality findings

daiagi:
two hard-checked dispute findings. docs only, no fix.

- `FIND-TIMEOUT-1`: timeout dispute lose calldata race (junk calldata only; valid calldata ingest via chain event, fork advance, no issue) -> `RaceConditionDisputeTimeoutCalldataPosted` unhandled in `DisputeManager.dispute`, marker roll back, no re-arm. honest peers never dispute again, fork stuck. reproduced in harness, both disputers initiate, zero commit.
- `FIND-DISPUTE-2`: `_isLastMilestoneFinalByEveryone` read live snapshot participants at call time, contradict `REQ-FIN-7`. JOIN consumed by same-fork snapshot inside kill period flip final -> not final, honest disputer slashed. foundry probe show slash through real `applyDisputeFraudProofs`. full snapshot-post ordering asserted from code, not run e2e yet.
- added `REQ-FIN-7.T1.P18` permutation for widened set.
- `DisputeFraudProofFacet.sol.md` conformance row -> Contradicts. fixed two stale sentences (security-assessment "historical walk", `StateChannelCommon.sol.md` "one reader").

PR #486, opened 2026-09-15
(fix) freeze membership during dispute kill period

daiagi:
fixes `FIND-DISPUTE-2` from #478.

story.  alice, bob, carol, eve are the participants. one earlier snapshot post already landed -> chain pruned the old inbound join records.
- eve leaves. her exit snapshot is fully signed, not posted yet
- bob disputes. last milestone eve never signed, `postedAuditingData = false` -> bob owes auditing data, slashable
- alice sends `DisputeLastMilestoneNotFinalAndNoAuditingData`
- bob posts eve's signed exit first (anyone can post it) -> chain set shrinks to alice, bob, carol -> milestone reads final -> alice's proof invalid -> **alice slashed**, bob keeps his dispute

cause: `_isLastMilestoneFinalByEveryone` judges a committed dispute against the live chain participant set and live slashes, not what existed when the dispute committed. join side is the mirror: a join adopted mid-dispute widens the set and slashes an honest disputer.

fix:
- `_updateStateSnapshot` refuses adoption onto a fork whose kill period is open (`RaceConditionSnapshotDuringKillPeriod`), same-fork and fork update. last require, so the refusal proves every other check passed
- finality read subtracts only the dispute's own `onChainSlashes` + slashes recorded before the window opened
- upload refuses an anchor below the consumed inbound height
- sdk: reducer lands the reduce alone, spectator treats the refusal as a race (no blacklist), leaver whose exit and fallback are both refused waits for settlement instead of rejecting (`FIND-LEAVE-3`)

proof: `test_leaverExitDuringKillPeriod_honestChallengerNotSlashed` and `test_joinConsumedDuringKillPeriod_honestDisputerNotSlashed` on the routed diamond. both fail with the freeze removed (honest side slashed).

follow-up: `FIND-LEAVE-4` (leave watchdog fallback still rejects on evidence-expired).


docs/spec/specification/protocol-model/finality.md
```diff
@@ -134,6 +134,14 @@ Finality arrives by exactly one of three routes:
   (`common adjudication logic`),
   and finalizes the dispute window immediately
   (`DisputeManagerFacet._isDisputeThresholdFinal`).
+- On-chain (dispute last-milestone probe, no posted data): the expected set is the chain's snapshot
```
lukastanisic99: what is weird is that the previous PR #478 added this:

'Two readers keep the unbounded walk: `reduce` for slash eligibility, because after a reduction is mined the snapshot lists only the survivors and a late reducer must still...' and that is not here in this file nor does this file mark any delition - not sure how it was rebased. It does however have the new features that were merged in `src` - weird;

contracts/V1/StateChannelDiamondProxy/StateSnapshotFacet.sol
```diff
@@ -89,6 +89,13 @@ contract StateSnapshotFacet is StateChannelCommon {
             ),
             ErrorOutboundMessageBlocksInvalid()
         );
+        DisputeWindow storage disputeWindow = disputeData[channelId].disputeWindowMap[newSnapshot.forkId];
+        (bool killPeriodExpired, uint256 killPeriodEnd) = _isKillPeriodExpired(disputeWindow, _getEvidenceTime());
+        // a fork's membership is frozen while its committed disputes can still be killed
+        require(
+            !_isDisputeWidnowCreated(disputeWindow) || killPeriodExpired,
```
lukastanisic99: This `!_isDisputeWidnowCreated(disputeWindow)` is enough - once a fork is disputed, a snapshot update can't be submited for it, the kill period expiring doesn't change that - a new forkId has to be provided; 

A better name would maybe be:
RaceConditionSnapshotUpdateDisputedFork

contracts/V1/StateChannelDiamondProxy/Errors.sol
```diff
@@ -120,9 +120,11 @@ error RaceConditionJoinChannelSnapshotMismatch();
...
     bytes32 submittedInboundMessageBlockHash, bytes32 onChainInboundMessageBlockHash
 );
+error RaceConditionSnapshotDuringKillPeriod(uint256 killPeriodEnd, uint256 currentTimestamp);
 error RaceConditionForceInboundJoinForkDisputed();
 error ErrorDisputeThrottled(address disputer, uint256 throttleExpiry, uint256 currentTimestamp);
 error ErrorDuplicateSelectorRegistration(bytes4 selector);
 error ErrorRouteTargetHasNoCode(bytes4 selector, address target);
 
 error RaceConditionDisputeWindowNotOpen(bytes32 channelId, bytes32 forkId);
+error RaceConditionDisputeAnchorBehindSnapshot(uint256 consumedInboundHeight, uint256 disputeInboundHeight);
```
lukastanisic99: I thought I already did this one before, but it seems I didn't, so good that you noticed it - better name would be:
RaceConditionDisputeInboundNotLatest

Also groupd all the RaceCondition* errors together, so we can easily see what we need withouth thinking should we look somewhere else

contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol
```diff
@@ -47,6 +47,13 @@ contract DisputeManagerFacet is StateChannelCommon {
             _canParticipateInDisputes(dispute.input.channelId, msg.sender),
             ErrorCantParticipateInDispute(dispute.input.channelId, msg.sender)
         );
+        uint256 consumedInboundHeight =
+            stateSnapshots[dispute.input.channelId].snapshotData.latestInboundMessageBlockHeight;
+        // a dispute built before a same-fork advance would be judged against a set it never saw
+        require(
+            dispute.input.lastInboundMessageBlockHeight >= consumedInboundHeight,
```
lukastanisic99: '>' is not possible, so only '==' is relevant, but it's better if you target the specific hash `input.lastInboundMessageBlockHash == ...Hash`

contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol
```diff
@@ -772,9 +772,21 @@ contract DisputeFraudProofFacet is StateChannelCommon {
...
+            _derivePendingParticipantsFromInboundHash(
+                channelId, dispute.input.latestInboundMessageBlockHash, bytes32(0)
+            )
+        );
+        // slashes before the window opened were visible to every dispute in it; later ones count only if committed
+        DisputeWindow storage window = disputeData[channelId].disputeWindowMap[dispute.input.forkId];
+        uint256 slashedBefore = _isDisputeWidnowCreated(window) ? window.evidence.creationTimestamp : block.timestamp;
+        address[] memory slashes = UtilityFacet(utilityFacetAddress).concatAddressArraysNoDuplicates(
+            dispute.input.onChainSlashes, _getOnChainSlashedParticipantsUpToTimestamp(channelId, slashedBefore - 1)
         );
```
lukastanisic99: this function is only used in `_isLastMilestoneFinalByEveryone` and it makes sense for you to want a deterministic outcome for a historic check that can be checked through time vs the `isEligible` which is a real-time check on the current chain state and it can change; 

Better to rename this one and canParticipateInDisputes into something more explicit so their intentions and difference is clear from reading the function name; 

Also, for this one, I don't think you need to bother and do the complex operation of caluclatin the on-chain slashes before the window opened, you can just use dispute.input.onChainSlashes as the deterministc set. The disputer self picks it based on their perception and based on that perception they caculate did the milestone reach finality or not; That way this is always deterministic withouth any race condition and is simple; If the disputer didn't see something, they have to submit a higher threshold, which is fine and if a disputer lies and included more addresses they get slashed from onChainSlashes not subset fraud proof;  

So thechincally this is a 4th threshold category in our spec - historic treshold, excluding on-chain slashes since it commits to the set at a specific point in time, unlike the previous category which is totally of-chain and doesn't have an on-chain achor to bound the set; 

Edit: Forgot to say +including pending participants, but that's also a deterministic set since dispute.input.latestInboundBlockHash commits to the enclosed set;

docs/spec/specification/disputes/disputes.md
```diff
@@ -409,7 +409,7 @@ forced inclusion and dispute remedies remain available after operational failure
 
 **<a id="req-dis-1-xaj1va"></a>`REQ-DIS-1-XAJ1VA`.** A dispute MUST state at least one of the five valid inputs (timeout, valid on-chain slashes, self-removal, forced inbound inclusion, or an accepted existing-window state contribution); fraud proofs are not a dispute input.
 
-**<a id="req-dis-2-pkvz7e"></a>`REQ-DIS-2-PKVZ7E`.** Upload is limited to eligible disputers: `disputer == msg.sender`, dispute-eligible, ≤1 dispute per window per participant, throttled to one upload per `evidenceTime` per sender.
+**<a id="req-dis-2-pkvz7e"></a>`REQ-DIS-2-PKVZ7E`.** Upload is limited to eligible disputers: `disputer == msg.sender`, dispute-eligible, ≤1 dispute per window per participant, throttled to one upload per `evidenceTime` per sender, and anchored at or above the chain's consumed inbound height (a dispute built before a same-fork advance is refused and must be rebuilt).
```
lukastanisic99: has to be exactly at the tip (e.g. latestHash == latestHash) - I also added it to the watchtowers PR a few days ago

docs/spec/specification/enforcement/snapshot-adoption.md
```diff
@@ -29,7 +29,7 @@ channel and the only writer of the canonical snapshot after opening.
...
+| Same-fork advance                       | Any submitter carrying the proof | Same fork; strictly newer snapshot; milestone finality proof verifies ([state-proofs.md](../disputes/state-proofs.md) via the [proof-verification module](./proof-verification.md)); all pending inbound blocks consumed by the new snapshot; the current fork's dispute kill period is not open                                                                                                                                               | Snapshot replaced; outbound range processed; housekeeping may clear consumed inbound and stale dispute data. |
```
lukastanisic99: fork is not disputed! kill period expiring does not allow a disputed fork to update the snapshot later

docs/spec/specification/enforcement/snapshot-adoption.md
```diff
@@ -68,20 +74,24 @@ Everything that leaves the channel passes here, so the adversarial submissions a
 gapped outbound ranges (linkage verification), double-release across overlapping batches
 (prune-then-process with at-most-once release), over-withdrawal (the deposits cap), regression or
 sideways adoption (monotonicity and the two proof paths), and admission erasure (the inbound-
-consumption gate). Housekeeping that clears consumed data must never clear anything a still-open
+consumption gate). A membership change adopted onto a fork while a committed dispute's kill period is open, by an advance or
+a successor-fork update, would change the participant set that dispute's finality read is judged against, turning an honest disputer or
+an honest prover into a slashable party; the kill-period freeze
```
lukastanisic99: explain this one in one of my comments already

docs/spec/specification/peer-communication/synchronization.md
```diff
@@ -123,7 +123,8 @@ Ordered verification; any failure aborts the sync with no partial effect:
     make a newcomer adopt an undercollateralized snapshot
     ([cross-layer-messages.md](../settlement/cross-layer-messages.md) §6).
 11. **Simulate adoption.** Verify, without any on-chain transaction, that the implied snapshot
-    advance would succeed on-chain; a simulated revert aborts.
+    advance would succeed on-chain; a simulated revert aborts, except a refusal that only defers adoption
+    because the target fork's kill period is open.
```
lukastanisic99, line 127: In the V1 sync design we just aborted the channel if the fork was disputed, since that was 'easy' - not sure is it easy to 'recover' during a dispute window - didn't think much, but we'll have to revisit spectating at some point;

docs/spec/specification/protocol-model/finality.md
```diff
@@ -134,6 +134,14 @@ Finality arrives by exactly one of three routes:
...
+  participants at the dispute's commitment, joiners recorded at or below the dispute's committed inbound
+  anchor, minus the slashes the dispute itself lists and the on-chain slashes recorded before the
+  dispute's window opened. The chain refuses adopting any snapshot onto the disputed fork while its kill
+  period is open ([`REQ-ENFSNAP-4-ESP98F` (Kill-period freeze)](../enforcement/snapshot-adoption.md#req-enfsnap-4-esp98f)) and refuses uploads anchored below its
+  consumed inbound height ([`REQ-DIS-2-PKVZ7E`](../disputes/disputes.md#req-dis-2-pkvz7e)). The participant
+  half therefore cannot change during the kill period, a slash landing after the window opened moves the
+  verdict only if the dispute lists it, and a member slashed before the window opened is not expected.
```
lukastanisic99: already commented on all of this

lukastanisic99 (commented):
**AI-generated review**

🟡 **[XY1] 95% — The facet report promises an SDK retry that is absent.** [The report](https://github.com/peer3to/state-channels-plus/blob/7b215fcb9cf4afe71fe28dfa5c57d418656ce44f/docs/spec/implementation/source/contracts/V1/StateChannelDiamondProxy/DisputeManagerFacet.sol.md#L34) says the SDK rebuilds and resubmits after `RaceConditionDisputeAnchorBehindSnapshot`. [The handlers](https://github.com/peer3to/state-channels-plus/blob/7b215fcb9cf4afe71fe28dfa5c57d418656ce44f/src/disputeManager/DisputeManager.ts#L207) contain no such branch; the retries below cover timeout timing and refreshed slashes. [The proposed retry test](https://github.com/peer3to/state-channels-plus/blob/7b215fcb9cf4afe71fe28dfa5c57d418656ce44f/test/unit/DisputeManager.test.ts#L1059) is skipped as unreachable. This is a documentation contradiction, not evidence of a demonstrated live retry failure.

> **Fix XY1-FIX**
>
> Describe the implemented admission refusal and the stated unreachability rationale. Remove the claim that this retry exists; if retry is intended, implement and prove it separately.



---

**AI-generated review**

🟡 **[XY2] 95% — A freeze refusal does not prove that every adoption check passed.** [The spectate conformance row](https://github.com/peer3to/state-channels-plus/blob/7b215fcb9cf4afe71fe28dfa5c57d418656ce44f/docs/spec/implementation/source/src/rpc/network/services/spectate/SpectateService.ts.md#L119) calls the freeze the last check and treats it as proof that every check passed. But [the next operation](https://github.com/peer3to/state-channels-plus/blob/7b215fcb9cf4afe71fe28dfa5c57d418656ce44f/contracts/V1/StateChannelDiamondProxy/StateSnapshotFacet.sol#L99) applies withdrawals, and [that operation](https://github.com/peer3to/state-channels-plus/blob/7b215fcb9cf4afe71fe28dfa5c57d418656ce44f/contracts/V1/StateChannelDiamondProxy/StateSnapshotFacet.sol#L146) can still fail with `ErrorWithdrawalFailed` or `CantWithdrawMoreThanDeposits`. The [facet report](https://github.com/peer3to/state-channels-plus/blob/7b215fcb9cf4afe71fe28dfa5c57d418656ce44f/docs/spec/implementation/source/contracts/V1/StateChannelDiamondProxy/StateSnapshotFacet.sol.md#L44) likewise overstates guaranteed validity. Spectate does verify milestones, outbound proofs and the final balance invariant before this call; I have not established an invalid-proof acceptance exploit. The confirmed issue is the stronger claim about checks and eventual adoption.

> **Fix XY2-FIX**
>
> State exactly which validation stages precede the refusal. Separate proof acceptance from successful withdrawal execution and eventual adoption; update the source comment and both reports without claiming all later execution must succeed.



---

**Human note**

This one is red, but I didn't verify it - not sure if it's relevant in adition to the other human comments I made

**AI-generated review**

🔴 **[FR1] 90% — A predecessor snapshot can still change finality during its reduced fork’s kill period.** The guard reads only `disputeWindowMap[newSnapshot.forkId]` in [_updateStateSnapshot](https://github.com/peer3to/state-channels-plus/blob/7b215fcb9cf4afe71fe28dfa5c57d418656ce44f/contracts/V1/StateChannelDiamondProxy/StateSnapshotFacet.sol#L92), but finality derives participants from [_getSnapshotParticipants(channelId)](https://github.com/peer3to/state-channels-plus/blob/7b215fcb9cf4afe71fe28dfa5c57d418656ce44f/contracts/V1/StateChannelDiamondProxy/DisputeFraudProofFacet.sol#L778), which reads [the channel-wide current snapshot](https://github.com/peer3to/state-channels-plus/blob/7b215fcb9cf4afe71fe28dfa5c57d418656ce44f/contracts/V1/StateChannelDiamondProxy/StateChannelCommon.sol#L111). Consider chain snapshot F, an expired/reduced F → G, and an open dispute window on G before adoption. This is a supported state: [the new integration staging](https://github.com/peer3to/state-channels-plus/blob/7b215fcb9cf4afe71fe28dfa5c57d418656ce44f/test/fixtures/ReducedForkKillPeriodStaging.ts#L43) creates it, and [asserts the old snapshot remains](https://github.com/peer3to/state-channels-plus/blob/7b215fcb9cf4afe71fe28dfa5c57d418656ce44f/test/fixtures/ReducedForkKillPeriodStaging.ts#L90). A valid same-fork update of F checks F’s expired window and can land while G is still open. A previously signed F membership change therefore changes the participant set read by G’s already-committed dispute; consuming/pruning inbound joins can change that read too. The new anchor guard runs only at upload and cannot prevent a later snapshot change. This leaves the central stability guarantee unfulfilled and allows the finality predicate to change while a proof can still land. This conclusion comes from the source path; I did not execute a reproduction.

> **Fix FR1-FIX**
>
> Protect every snapshot mutation that can affect an open dispute’s participant read, including predecessor advancement while a successor window is open, or store the required participant basis immutably at admission. Add a routed-diamond regression: keep F on chain, open G, attempt a valid membership-changing same-fork F update, and prove G’s finality verdict and fraud-proof result cannot change. Cover both removal and consumed-join changes. The existing unrelated-fork test must distinguish an unrelated fork from this reduced successor.



---

**AI-generated review**

🟡 **[RY1] 90% — The new leaver fixture can leave a reduction held after an assertion fails.** [The fixture](https://github.com/peer3to/state-channels-plus/blob/7b215fcb9cf4afe71fe28dfa5c57d418656ce44f/test/fixtures/HonestLeaverDisputeStaging.ts#L254) obtains `leaverReduction` before its first phase. That phase’s [finally](https://github.com/peer3to/state-channels-plus/blob/7b215fcb9cf4afe71fe28dfa5c57d418656ce44f/test/fixtures/HonestLeaverDisputeStaging.ts#L310) releases only rebuild/send; the reduction release belongs to the later phase at [the second finally](https://github.com/peer3to/state-channels-plus/blob/7b215fcb9cf4afe71fe28dfa5c57d418656ce44f/test/fixtures/HonestLeaverDisputeStaging.ts#L374). A timeout or assertion in the first phase skips the second phase and its release, which can leave work parked during teardown and hide the original failure.

> **Fix RY1-FIX**
>
> Put both phases under an outer finally that always releases the reduction hold, while preserving the existing phase-specific releases.



---

**AI-generated review**

🟡 **[TY1] 90% — The evidence-expired leave tests do not prove recovery from a real expired window.** [The fixture](https://github.com/peer3to/state-channels-plus/blob/7b215fcb9cf4afe71fe28dfa5c57d418656ce44f/test/fixtures/AuthoredLeaveFailureStaging.ts#L19) injects `RaceConditionDisputeEvidencePeriodExpired` at send. Its [fast branch](https://github.com/peer3to/state-channels-plus/blob/7b215fcb9cf4afe71fe28dfa5c57d418656ce44f/test/fixtures/AuthoredLeaveFailureStaging.ts#L115) later opens a fresh self-removal as a stand-in; the slow branch stops after observing `awaiting-settlement` and `didIDispute === false`. This proves error classification, but not the real expired-window workflow or its two settlement outcomes. The source [restarts leave on a new fork](https://github.com/peer3to/state-channels-plus/blob/7b215fcb9cf4afe71fe28dfa5c57d418656ce44f/src/stateManager/membership/LeaveChannelService.ts#L191) when the signer remains a participant, so the absent outcome assertion is not proof of a hang.

> **Fix TY1-FIX**
>
> Stage an existing committed window whose evidence period really expires before fallback. Assert the leave promise resolves if reduction removes the signer, and that retained membership restarts leave and eventually completes. Keep these distinct from the explicitly deferred watchdog finding.



---

**AI-generated review**

🟠 **[DO1] 95% — 74 exact-test anchors are incorrect in four changed reports.** The audit checked 282 declaration rows at head and found 74 incorrect anchors: DiscoveryRuntimePort 59, MilestoneFinalityFreeze 5, selfRemoval 5 and SpectateService 5. The same check on the base found zero incorrect anchors among 257 rows. For example, [the exact-target snapshot row](https://github.com/peer3to/state-channels-plus/blob/7b215fcb9cf4afe71fe28dfa5c57d418656ce44f/docs/spec/verification/tests/test/unit/SpectateService.test.ts.md#L34) points to [the new dispute-opens case](https://github.com/peer3to/state-channels-plus/blob/7b215fcb9cf4afe71fe28dfa5c57d418656ce44f/test/unit/SpectateService.test.ts#L194), while its named test begins at [line 266](https://github.com/peer3to/state-channels-plus/blob/7b215fcb9cf4afe71fe28dfa5c57d418656ce44f/test/unit/SpectateService.test.ts#L266). Some errors are one-line shifts; others assign obligations to a different body. The complete correction list is saved locally in `evidence/anchor-audit.json`.

> **Fix DO1-FIX**
>
> Repair both the links and displayed line numbers using full declaration identity, not the leaf name alone. Recheck all changed verification reports and regenerate the inverse reports. Preserve each test’s actual coverage rather than moving its credit to whichever declaration occupies the old line.



---

**AI-generated review**

🟡 **[DY2] 95% — Two new conformance links resolve to nonexistent reports.** [The spectate row](https://github.com/peer3to/state-channels-plus/blob/7b215fcb9cf4afe71fe28dfa5c57d418656ce44f/docs/spec/implementation/source/src/rpc/network/services/spectate/SpectateService.ts.md#L119) uses `../../../evm/EvmDiamondStateMachine.ts.md`, which resolves under `src/rpc/evm`; it needs `../../../../evm/...`. [The reduction row](https://github.com/peer3to/state-channels-plus/blob/7b215fcb9cf4afe71fe28dfa5c57d418656ce44f/docs/spec/implementation/source/src/stateManager/reduction/ReductionExecutor.ts.md#L88) climbs one level too far with `../../../../contracts/...`; it needs `../../../contracts/...`. These are new broken targets, separate from inherited link problems.

> **Fix DY2-FIX**
>
> Correct the two relative targets and validate their resolved files. Keep conformance links aimed at the implementation reports.

daiagi:
The same-fork advance is refused permanently once its fork is disputed (`_isForkDisputed`), matching the existing join freeze. The fork-update adoption keeps its kill-period version, since that target fork can still be disputed on its own. Both share one error, `RaceConditionSnapshotUpdateDisputedFork(channelId, forkId, killPeriodEnd, currentTimestamp)`.

The finality read subtracts only `dispute.input.onChainSlashes`. No timestamp bound was needed: `DisputeOnChainSlashesNotSubset` already slashes a disputer who lists an address that was not actually slashed. Renamed `_deriveExpectedParticipantsForDispute` to `_getHistoricThresholdSet` and `_canParticipateInDisputes` to `_canParticipateInDisputesNow`, to separate the deterministic historic set from the live eligibility check.

The upload anchor is compared by exact equality to the chain's inbound head (`channelBalances`), not `>=` against the snapshot-consumed tip. A refused upload is not retried: the marker rolls back. `RaceConditionDisputeAnchorBehindSnapshot` is renamed `RaceConditionDisputeInboundNotLatest`. Top-ups are refused on a disputed fork now too, the same gate joins already had.

`RaceCondition*` errors are grouped together in `Errors.sol`.

PR #488, opened 2026-09-16
Run the browser gates in the parallel and distributed runner

momoshell:
Closes ["Run browser tests in distributed runner"](https://trello.com/c/JBhVjU3Y).

The browser smokes were standalone Node scripts that only ran when invoked by
hand, which is how their breakage survived unnoticed through hundreds of gates.
They are a third runner tier now, discovered beside the Mocha and forge tiers.

One task per `test/browser/run-*.mjs` gate — a gate boots a Vite server, its own
Hardhat node and one Chromium, so a task per `node:test` case would relaunch that
stack per case. Gates take no warm slot, and reach a worker through a new
`browser-test` Hardhat task, the same indirection the forge tier uses. New flags:
`--browser-only`, `--no-browser`, `--browser-test-pattern`.

## Worker hosts must update before this merges

A worker runs the runner from its own checkout, so an un-updated worker can't
execute a `browser` task — against the live pool it failed *fatally*,
quarantining every worker mid-run. `DISTRIBUTED_PROTOCOL_VERSION` is bumped
13 → 14 with the tier vocabulary, so that is now a clear message before any task
is assigned:

```
Ignoring worker server-9: Distributed worker protocol mismatch:
orchestrator requires 14, worker host provides 13.
Update and restart the worker host or rebase this branch.
```

Each host: pull this branch → restart `yarn test:parallel:server` → rebuild the
runner image, which now installs the Chromium matching the Playwright version
`yarn.lock` resolves.

## Verification

Both gates pass on a real distributed worker (2 passing, 301s). That run found
two failures a local run cannot, fixed here: a worker gets a fresh `HOME` and
never downloads browsers, so a gate needs `PLAYWRIGHT_BROWSERS_PATH` (the image
sets it, an `unsafe-host` worker must export it); and the slowest scenario,
~39s idle against a 45s in-page budget, timed out at 45.04s under worker load —
the budget is 120s now.

With Docker: the image builds and Chromium launches inside a container created
with the environment's own flags. It launches there even with its own sandbox
on, so `SCP_BROWSER_CONTAINED` is insurance for hosts where AppArmor or a
restricted userns blocks it.

Not reachable from a Mac: a full run on the **docker** backend, which needs a
volume driver enforcing `size`. The first Linux worker closes that gap —
`yarn test:parallel:docker:self-check` confirms its environment first.

Plus 51 tier tests; both tsc projects, `lint:imports`, prettier and
`spec:ids:check` clean, `spec:refresh` changed no tracked file.

## One question for review

The CI `browser` job now duplicates what the gate covers. I left it as a
pre-rollout safety net and as the only check running Chromium outside the
container — drop it if that's not worth paying twice.


scripts/e2e-parallel/shared/browserTaskDiscovery.js
```diff
@@ -0,0 +1,55 @@
...
+const { globSync } = require("glob");
+const {
+    BROWSER_TEST_TASK,
+    DEFAULT_BROWSER_TEST_PATTERN
+} = require("./browserConfig");
+const { escapeRegex, sanitizeFileName } = require("./taskDiscovery");
+const { TASK_RUNNERS } = require("./taskRunners");
+
+function isBrowserGateFile(filePath) {
+    return path.extname(filePath) === ".mjs";
```
lukastanisic99: **AI-generated review**

🟠 **[FO1] 95% — A custom browser glob can report success without running a browser test.** The filter accepts every `.mjs` file once a supplied pattern replaces the default. For example, `--browser-only --browser-test-pattern browser/sdkRuntimeServer.mjs` selects [a module that only exports helpers](https://github.com/peer3to/state-channels-plus/blob/a477f33d61013133efbcece51ed44e2669931bfc/test/browser/sdkRuntimeServer.mjs#L15); loading it runs no assertions. A broad `browser/**` pattern includes helper modules as extra passing tasks. This violates [the documented run-*.mjs gate boundary](https://github.com/peer3to/state-channels-plus/blob/a477f33d61013133efbcece51ed44e2669931bfc/scripts/e2e-parallel/shared/browserConfig.js#L4) and defeats the empty-selection check. This conclusion is from static inspection; the example was not executed.

> **Fix FO1-FIX**
>
> Intersect custom/shared patterns with the supported browser gate naming boundary, or require an explicit gate manifest. A helper-only selection must fail as containing no runnable gates; a broad browser pattern must select only the real gates.

scripts/e2e-parallel/shared/browserTaskDiscovery.js
```diff
@@ -0,0 +1,55 @@
+const path = require("path");
+const { globSync } = require("glob");
+const {
+    BROWSER_TEST_TASK,
+    DEFAULT_BROWSER_TEST_PATTERN
+} = require("./browserConfig");
+const { escapeRegex, sanitizeFileName } = require("./taskDiscovery");
```
lukastanisic99: **AI-generated review**

🟡 **[KY1] 100% — Browser discovery imports an unused escapeRegex helper.** The new module never calls `escapeRegex`; only `sanitizeFileName` is used. Symbol and caller searches found no need for this import.

> **Fix KY1-FIX**
>
> Remove only escapeRegex from the destructuring import. Keep sanitizeFileName and the shared helper’s real Mocha/Forge callers.

test/scripts/e2eParallelBrowserTasks.test.ts
```diff
@@ -0,0 +1,795 @@
...
+    for (const root of [
+        ...scratchRoots.splice(0),
+        ...scratchLogDirs.splice(0)
+    ]) {
+        fs.rmSync(root, { recursive: true, force: true });
+    }
+});
+
+/** A throwaway gate tree: `<root>/browser/<name>.mjs` for each given name. */
+function writeGateTree(gateNames: string[], extraFiles: string[] = []) {
```
lukastanisic99: **AI-generated review**

🟡 **[AY2] 100% — Reusable test setup remains inside the test file.** [scratchRoot](https://github.com/peer3to/state-channels-plus/blob/a477f33d61013133efbcece51ed44e2669931bfc/test/scripts/e2eParallelBrowserTasks.test.ts#L206), [writeGateTree](https://github.com/peer3to/state-channels-plus/blob/a477f33d61013133efbcece51ed44e2669931bfc/test/scripts/e2eParallelBrowserTasks.test.ts#L224), and [writeGate](https://github.com/peer3to/state-channels-plus/blob/a477f33d61013133efbcece51ed44e2669931bfc/test/scripts/e2eParallelBrowserTasks.test.ts#L235) serve multiple cases. [The test-file rule](https://github.com/peer3to/state-channels-plus/blob/a477f33d61013133efbcece51ed44e2669931bfc/test/AGENTS.md#L188) requires reusable staging in a discoverable helper or fixture.

> **Fix AY2-FIX**
>
> Move the shared filesystem/process setup and cleanup bookkeeping into a focused fixture under test/fixtures. Keep direct literal it declarations in the test file.

lukastanisic99 (commented):
**AI-generated review**

🟠 **[XO1] 100% — The browser launch policy still treats host execution as supported.** The engineer clarified that all tests must run inside the container. The [browser CI job](https://github.com/peer3to/state-channels-plus/blob/a477f33d61013133efbcece51ed44e2669931bfc/.github/workflows/ci.yml#L91) currently installs Chromium on the GitHub runner VM and invokes the [browser package scripts directly](https://github.com/peer3to/state-channels-plus/blob/a477f33d61013133efbcece51ed44e2669931bfc/.github/workflows/ci.yml#L119); its Docker command only copies out the Forge executable. The job therefore runs these tests outside the test-runner container. The [new helper](https://github.com/peer3to/state-channels-plus/blob/a477f33d61013133efbcece51ed44e2669931bfc/test/browser/chromiumLaunch.js#L8) also documents a host fallback and claims that Chromium keeps its own sandbox there, although [its options](https://github.com/peer3to/state-channels-plus/blob/a477f33d61013133efbcece51ed44e2669931bfc/test/browser/chromiumLaunch.js#L15) leave that sandbox disabled by default ([Playwright launch options](https://playwright.dev/docs/api/class-browsertype#browser-type-launch-option-chromium-sandbox)). The host job predates this PR, but it cannot count as evidence for the newly required container path. The earlier recommendation to enable Chromium’s sandbox on the host is withdrawn: that would not satisfy container-only execution.

> **Fix XO1-FIX**
>
> Keep the browser CI job for now, as requested in OO1, but route its test execution through the Docker-backed runner/image. Remove the documented host fallback from supported verification, and check that the actual test processes execute inside the container. Correct the helper comment and its test; Chromium’s own sandbox is a separate setting and is not a replacement for the required container.



---

**AI-generated review**

🟠 **[TO2] 95% — Unit coverage does not exclude helper modules under custom globs.** [The helper-exclusion case](https://github.com/peer3to/state-channels-plus/blob/a477f33d61013133efbcece51ed44e2669931bfc/test/scripts/e2eParallelBrowserTasks.test.ts#L272) uses only the default pattern. [The shared-pattern case](https://github.com/peer3to/state-channels-plus/blob/a477f33d61013133efbcece51ed44e2669931bfc/test/scripts/e2eParallelBrowserTasks.test.ts#L333) supplies a TypeScript file, so it checks only the extension boundary and misses FO1.

> **Fix TO2-FIX**
>
> Add separate static cases for a broad shared browser glob and a helper-only browser pattern. Assert that only gate entry points are discovered and that helper-only selection reports no runnable gates.



---

**AI-generated review**

🟠 **[TO3] 95% — Unit coverage never verifies when browser warm-up runs.** [The build cases](https://github.com/peer3to/state-channels-plus/blob/a477f33d61013133efbcece51ed44e2669931bfc/test/scripts/e2eParallelBrowserTasks.test.ts#L652) compare a constant and package-script text. They never exercise [the new warm-up loop](https://github.com/peer3to/state-channels-plus/blob/a477f33d61013133efbcece51ed44e2669931bfc/scripts/test-e2e-parallel.js#L299), so they do not prove one build before admission, no build when no browser tasks survive filtering, or no local build in distributed mode.

> **Fix TO3-FIX**
>
> Exercise the runner entry point with an owned temporary project whose real build command records a marker. Assert one marker before task admission and no marker for excluded, grep-empty, dry-run, and distributed paths.



---

**AI-generated review**

🟠 **[TO4] 95% — Unit coverage omits an unavailable browser build command.** No test reaches [the spawn-error branch](https://github.com/peer3to/state-channels-plus/blob/a477f33d61013133efbcece51ed44e2669931bfc/scripts/e2e-parallel/shared/taskRunners.js#L52) through browser warm-up. The direct gate tests use an existing Node executable; a missing gate file does not exercise a missing build executable.

> **Fix TO4-FIX**
>
> Run the build boundary with an unavailable executable in an owned temporary environment. Assert a nonzero runner exit, the browser-specific installation guidance, and no admitted gate.



---

**AI-generated review**

🟠 **[TO5] 95% — Unit coverage omits a nonzero browser build exit.** [The extracted build-failure branch](https://github.com/peer3to/state-channels-plus/blob/a477f33d61013133efbcece51ed44e2669931bfc/scripts/e2e-parallel/shared/taskRunners.js#L58) can stop all task admission, but the new suite never invokes it with a build that exits unsuccessfully.

> **Fix TO5-FIX**
>
> Use a real temporary build command that exits with a fixed nonzero code. Assert that the diagnostic preserves that code and the runner does not start a browser task.



---

**AI-generated review**

🟠 **[TO6] 95% — Unit coverage omits a browser build killed by a signal.** [The null-status branch](https://github.com/peer3to/state-channels-plus/blob/a477f33d61013133efbcece51ed44e2669931bfc/scripts/e2e-parallel/shared/taskRunners.js#L60) formats a signal separately from an exit code. [The existing signal case](https://github.com/peer3to/state-channels-plus/blob/a477f33d61013133efbcece51ed44e2669931bfc/test/scripts/e2eParallelBrowserTasks.test.ts#L389) covers the asynchronous gate runner, not this synchronous build path.

> **Fix TO6-FIX**
>
> Use an owned build process that terminates itself by signal. Assert that browser warm-up reports the signal as a failure and admits no task.



---

**AI-generated review**

🟠 **[TO7] 95% — The extracted worker-fork environment lacks absent-input and precedence coverage.** [Its one case](https://github.com/peer3to/state-channels-plus/blob/a477f33d61013133efbcece51ed44e2669931bfc/test/scripts/e2eParallelBrowserTasks.test.ts#L681) supplies both browser variables and one module path. It does not verify [the extracted builder](https://github.com/peer3to/state-channels-plus/blob/a477f33d61013133efbcece51ed44e2669931bfc/scripts/e2e-parallel/distributed/remoteEnvironment.js#L55) when browser variables are absent, when the source already has HOME/NODE_PATH, or when multiple node paths must preserve their order. These values decide both isolation and module lookup in the real worker.

> **Fix TO7-FIX**
>
> Add direct cases for absent browser variables, source HOME overridden by the environment home, and runner/project/source NODE_PATH order. Keep the unrelated-secret exclusion assertion.



---

**AI-generated review**

🟠 **[TO8] 95% — Integration coverage bypasses the new distributed browser execution path.** [Gate tests call runBrowserGate directly](https://github.com/peer3to/state-channels-plus/blob/a477f33d61013133efbcece51ed44e2669931bfc/test/scripts/e2eParallelBrowserTasks.test.ts#L354); [the admission test replaces runTaskImpl](https://github.com/peer3to/state-channels-plus/blob/a477f33d61013133efbcece51ed44e2669931bfc/test/scripts/e2eParallelBrowserTasks.test.ts#L461); [the environment case checks an object](https://github.com/peer3to/state-channels-plus/blob/a477f33d61013133efbcece51ed44e2669931bfc/test/scripts/e2eParallelBrowserTasks.test.ts#L681). The successful [browser CI job](https://github.com/peer3to/state-channels-plus/actions/runs/35094089271/job/104788270538) runs standalone package scripts. None of these tests proves task transport → guest worker → [Hardhat browser-test](https://github.com/peer3to/state-channels-plus/blob/a477f33d61013133efbcece51ed44e2669931bfc/tasks/browserTest.ts#L23) → Chromium inside the runner image. The PR body explicitly says a full Docker-backend run is still missing.

> **Fix TO8-FIX**
>
> Add a browser-only integration run through a real distributed worker using the runner image. Verify real Chromium success, one deliberately failing gate reported as failed with its output, and the image browser path surviving the guest/worker boundary.



---

**AI-generated review**

🟠 **[TO9] 90% — No integration case verifies browser cancellation and recovery.** The new gate adds Vite, an owned node, and Chromium below [task cancellation](https://github.com/peer3to/state-channels-plus/blob/a477f33d61013133efbcece51ed44e2669931bfc/scripts/e2e-parallel/shared/runTask.js#L92). The new tests never cancel that real tree or verify a later browser task can run. Generic process tests cannot establish cleanup of Chromium and the gate-owned servers.

> **Fix TO9-FIX**
>
> Cancel a real browser task after its page and servers are ready; assert its node, Vite and Chromium processes/ports are gone, then run a following gate successfully. Preserve failure output through the distributed result path.



---

**AI-generated review**

🟠 **[TO10] 100% — The sandbox test does not prove the required container execution.** [The test named keeps Chromium’s own sandbox outside that container](https://github.com/peer3to/state-channels-plus/blob/a477f33d61013133efbcece51ed44e2669931bfc/test/scripts/e2eParallelBrowserTasks.test.ts#L782) only expects `headless: true`, which leaves Chromium’s sandbox disabled. More importantly, the engineer’s required boundary is the test-runner container; returning launch options proves nothing about whether the test process is inside it. This is the weak assertion behind XO1.

> **Fix TO10-FIX**
>
> Replace the host-sandbox claim with a test of the supported container launch policy. Add real container execution evidence through the Docker-backed runner, including the launch helper and guest environment; do not treat SCP_BROWSER_CONTAINED or an options-object assertion alone as proof of isolation.



---

**AI-generated review**

🟠 **[AO1] 100% — The new tests replace real collaborators with mocks.** [The scheduler case replaces task execution](https://github.com/peer3to/state-channels-plus/blob/a477f33d61013133efbcece51ed44e2669931bfc/test/scripts/e2eParallelBrowserTasks.test.ts#L461), and [three launch cases replace chromium.launch](https://github.com/peer3to/state-channels-plus/blob/a477f33d61013133efbcece51ed44e2669931bfc/test/scripts/e2eParallelBrowserTasks.test.ts#L728). This conflicts with [the no-mocks rule](https://github.com/peer3to/state-channels-plus/blob/a477f33d61013133efbcece51ed44e2669931bfc/test/AGENTS.md#L164). The launch success mock returns a string, so it cannot detect incorrect Playwright launch behavior. Similar older Forge tests do not exempt this new file.

> **Fix AO1-FIX**
>
> Use real owned child processes for scheduler admission and real Playwright launches for launch policy/error cases. Keep simple pure-object tests for environment assembly; move reusable process/browser setup into fixtures.



---

**AI-generated review**

🔴 **[DR1] 100% — All 14 browser test links were left at their old declaration lines.** The timeout refactor shifts every declaration in [the worker browser gate](https://github.com/peer3to/state-channels-plus/blob/a477f33d61013133efbcece51ed44e2669931bfc/test/browser/run-worker-contract-executor.mjs#L258), but its [verification table](https://github.com/peer3to/state-channels-plus/blob/a477f33d61013133efbcece51ed44e2669931bfc/docs/spec/verification/tests/test/browser/run-worker-contract-executor.mjs.md#L20) is unchanged. For example, the performance case moved 248 → 258 and the nested crash-log case 462 → 475. All 14 rows are stale; some old lines now land inside a different test. Regeneration reports the current declarations as unreferenced, so the old Covers assignments no longer establish exact evidence. [The specification workflow](https://github.com/peer3to/state-channels-plus/blob/a477f33d61013133efbcece51ed44e2669931bfc/docs/spec/AGENTS.md#L132) requires exact declaration links in the same change and rejects new structural gaps.

> **Fix DR1-FIX**
>
> Update all 14 declaration anchors and displayed line numbers by unique test name, retaining only Covers assignments that the bodies still prove. Refresh the overview to include distributed execution, run spec:refresh, and confirm these declarations retain their intended mappings. Leave engineer approval to the engineer.



---

**Human note**

For now keep it and later we can remove it once things are stable;

**AI-generated review**

🟢 **[OO1] 100% — The engineer decided to keep the browser CI job temporarily.** **ADDRESSED as a decision.** The preserved Human note says to keep it until things are stable. The engineer also specified that all tests must run inside the container, so keeping [the job](https://github.com/peer3to/state-channels-plus/blob/a477f33d61013133efbcece51ed44e2669931bfc/.github/workflows/ci.yml#L91) means changing its execution path as described in XO1. This no longer needs a human decision; removal can be reconsidered after container-backed execution is stable.

> **Fix OO1-FIX**
>
> No further decision required. Preserve the temporary job and apply XO1-FIX to its execution path.

momoshell:
**TO9** — done in `2ab163ca`. Two cases run a real gate through the Hardhat task, wait for it to report its Chromium and port, cancel it, then check the browser is gone and the port is free, and that a following gate still runs. Replacing every group kill with a child-only kill fails both, so they bite.

**TO10 / XO1 wording** — fair hit, fixed in `4088d651`. The helper claimed Chromium's sandbox "cannot start" in the container; it can — it launched with the sandbox on under `--read-only`, `--cap-drop=ALL` and no-new-privileges. The container is the isolation boundary, and `SCP_BROWSER_CONTAINED` is for hosts where AppArmor or a restricted userns won't allow the sandbox. Comment, README and both test names say that now.

**TO8 / the rest of XO1** — I can't verify these from a Mac. The full path (transport → guest worker → `browser-test` → Chromium in the image) needs the docker backend, and the environment won't create its volume on Docker Desktop: `quota size requested but no quota support`. The driver has to enforce `size`, which macOS doesn't do. What I could check, I did: the image builds and Chromium launches inside a container with the environment's own flags.

@lukastanisic99  Since we need bringing up the first Linux worker anyway, I'd rather run that evidence there than assert it from here.
Same for routing the CI job through the image: plain `docker run` would work on a GitHub runner, but it's only observable by pushing and watching CI, and once the pool runs the gates that job's only value is as a non-container control. My suggestion is both land in the follow-up that decides whether the job stays at all.

The rest of the panel is addressed across `d5c268bf`, `2ab163ca` and `4088d651`: FO1 (the `run-*.mjs` boundary is a rule now, so a helper-only pattern reports no runnable gates), DR1 (declaration anchors regenerated by name, two declarations that never had a row got one), AO1 (launch cases drive real Playwright in a child process; the scheduler case keeps `runTaskImpl`, the seam `runScheduler` exposes and `e2eParallelWorkerScheduler.test.ts` already uses), AY2 (staging moved to `test/fixtures/distributed/browserGateTrees.ts`), TO2/TO4/TO5/TO6/TO7, and TO3 via `resolveWarmUps`.

PR #474, opened 2026-09-06
fix(p2p): stop discovery from re-dialing peers a disconnect meant to drop

sh3ll3x3c:
## Summary

Hyperswarm keeps re-dialing any peer that still shares a joined topic until its peer info is banned; `socket.destroy()` only pauses it. Reproduced on a local DHT testnet with the installed Hyperswarm 4.3.5: after a destroy the peer redialed in 2.6 s; after `peerInfo.ban(true)` there was no reconnection in 45 s. So a `P2PManager.disconnectConnection` on a peer that shares a live topic is not a disconnect (Trello: "P2pManager.disconnectConnection is not correctly used in some cases").

This change:

- adds a named soft ban owned by `ProfileManager` (`banReconnect` / `allowReconnect` / `isReconnectBanned` / `banTransportReconnect`): the discovery handle is banned and the handshake refuses the identity, but nothing is excluded and the ban lifts with its cause; a blacklist implies it and survives lifting it; the WebRTC release paths respect it
- makes `P2PManager` track the discovery keys it observes; channel close and `disconnectFromPeers` now leave every key before closing transports (`disconnectFromPeers` is async), and `connectToChannel` re-observes
- makes the handshake refuse a reconnect-banned signer without exclusion, and ban the arriving handle of a blacklisted signer so a rotated discovery key does not bypass the exclusion
- makes lobby cleanup leave the topic before closing session transports, and session-bans non-selected, abusive, and late-joining peers until the topic is left
- makes the local test discovery backend skip banned peers, mirroring Hyperswarm

Spec layer: `REQ-AUTH-4` gains P4–P7, `REQ-LOBBY-9` gains P12–P15, new `REQ-UPG-7` (ending participation stops observing every key first, and nothing is admitted through discovery while no key is observed) with P1–P5; implementation mirrors, verification reports, and generated reports are updated.

## Judgment calls

- The blacklist and its sites are unchanged, including the post-proof ack-timeout exclusion (`REQ-AUTH-4` mandates it). No TTL; reputation stays future work.
- `restoreHolepunchReachability` has no "no live WebRTC" clause; the existing upgrade-policy tests require the release on a current-WebRTC close, so the live-WebRTC guard lives in `allowReconnect` instead.
- A peer that authenticates on the topic after a handoff is session-banned like a non-selected peer (found by the e2e wave; without it the peer loops until cleanup).
- `getJoinedDiscoveryKeys` has no production caller; it exists for the harness query.
[...]

PR #490, opened 2026-09-16
fix(p2p): make every disconnect carry an explicit policy and leave the lobby topic before cutting losers

sh3ll3x3c:
## Summary

Hyperswarm re-dials any peer that still shares a joined topic with us unless that peer's info is banned, and closing a transport never banned it. So a plain `disconnectConnection` on a peer that still shared a topic was not a disconnect, and the call sites never said what they meant. Trello: "P2pManager.disconnectConnection is not correctly used in some cases".

This change makes the decision explicit and puts it in one place:

- `P2PManager.disconnectConnection(transport, policy)` takes a required `DisconnectPolicy`, an ordered four-tier ladder:
  - `ALLOW` closes the connection; the peer may reconnect and nothing is banned.
  - `allowRetry(n)` closes and counts the close against the peer's session bound. The close that reaches `n` is applied as `SUSPEND` instead.
  - `SUSPEND` closes, bans the peer's Hyperswarm info and refuses its reconnect for the rest of this session. Nothing is written to its profile, so the bar dies with the session.
  - `BLACKLIST` has the same immediate effect plus the verdict recorded on the profile, which travels with it and is only lifted by an explicit unblacklist.

  Expected and upgrade closes pass `ALLOW` from `NetworkTransport.afterClose`, so no close inherits another caller's decision. The blacklist helpers delegate to the same switch. `ProfileManager` stays the single owner of all ban state: the recorded blacklist, the session suspension set, the session retry counters, and every Hyperswarm `ban()` write.
- Every call site was audited and now states its tier, with its current outcome preserved. Objective protocol faults (bad signatures, invalid state transitions, malformed frames, negotiation step violations, spectate misuse) stay `BLACKLIST`. Lifecycle closes and non-attributable failures are `ALLOW`: expected and upgrade closes, and a remote negotiation abort. Malformed frames and unknown service or method names at the router keep blacklisting as before.
- The handshake's timing filter takes the two new tiers, since a skewed clock or a slow link is an environment fault rather than misconduct. The request-time skew check, the response round-trip check and the response-timestamp check now state `allowRetry(InitHandshakeService.TIMING_RETRY_LIMIT)` (currently 3, one constant so the bound moves in one place); the post-proof ack timeout and the completed-guard expiry state `SUSPEND`. The request-time and response-timestamp checks previously blacklisted on the first failure. The ack timeout also gains its missing close: it used to exclude by an address that often has no profile yet and never closed the transport at all.
- The lobby now leaves its discovery topic before it closes the non-selected transports at handoff, and keeps the matched transport. Previously the topic was left only in cleanup after the whole channel-open negotiation, so every loser was re-dialed and re-cut for the entire negotiation. This is a plausible source of the known lobby ack-timeout flake.
- A session-scoped do-not-rematch set in `LobbyMatchingService` replaces the blacklist for peers that vanish mid-commit, abandon a reservation, match into an already open channel, or burn an opening window. Selection, pick, and commit admission consult it; it clears when the lobby session ends and survives a failed negotiation's cleanup.

Not in this change: persistence. Nothing in the SDK persists across restarts today, so a durable blacklist needs a new node and browser storage layer. That is a follow-up, together with storing the exclusion reason.

## Judgment calls

- **Enum plus one factory, not a policy object.** Only `ALLOW_RETRY` carries a value, so `DisconnectPolicy` is a discriminated union over a `DisconnectTier` enum with the three constant tiers exported as plain values and the bounded tier built by `allowRetry(n)` — existing `DisconnectPolicy.ALLOW` / `DisconnectPolicy.BLACKLIST` call sites keep their spelling, and nothing more general is built for four fixed values.
- **One retry counter per peer, not per check.** The counter is keyed by normalized address, the way `ProfileManager` already keys peers, and is shared by every call site that states the bounded tier, so a peer cannot spread its allowance across different checks. The consequence is that two sites naming different bounds still draw on the same counter; that is accepted, and the handshake's three timing checks deliberately share one bound. An unproven transport has no identity to count, so it only closes.
- **A session bar is manager state; a blacklist is profile state.** Suspension lives in a set on `ProfileManager` and writes nothing to the profile, which is what makes it session-scoped and what the unit tests use as their oracle: hand the same profile to a second manager and the blacklist follows it while the bar does not. The set is also checked independently of any profile, so a barred identity dialling back on a fresh handle is still refused.
- **The lobby keeps its own set.** `SUSPEND` is a connection-layer bar; "do not select this peer again in this lobby" is a matching fact with a lifetime the lobby defines. The two are not merged.
- **DEF-5 is respected.** The initial-sync timeout in `SpectateService` and the fork-dispute ack timeout in `IsForkDisputedService` keep blacklisting the selected peer, per the accepted DEF-5 disposition of 2026-08-31. The synchronization and dispute-acknowledgment failure tables are unchanged for those rows.
- **`OpenChannelNegotiationService` committed-peer loss** used to call `profileManager.blacklistPeer` directly, bypassing `P2PManager` and never closing anything. It now goes through the lobby exclusion path like its siblings.
- **Remote abort split from protocol failure.** A polite `abort` is a protocol message, not a violation; it now ends the attempt as a lobby exit. Step violations still exclude.
- **Leaving the lobby topic is one-sided.** It stops our dials. A non-selected peer that still observes the topic can dial us until it leaves too; each such connection completes the handshake and is closed with reconnect allowed. That is the accepted cost of not banning peers for lifecycle events. The e2e cases assert what the local discovery backend can prove: we stop dialing non-selected peers during the negotiation.
- **Fork-dispute duplicate acknowledgment stays excluded.** Two concurrent duplicates from one peer reach the inner check that the sequential guard would have caught; the peer must not be able to pick a milder penalty by timing.
- **`unblacklistPeer` has no production caller.** Left as is; a way out of an exclusion belongs with persistence. It does not lift a session bar it did not place.
- **Set lifetime.** Failed negotiations call `cleanup()` too; clearing the set there would restore the livelock, so it clears only when a match resolves or the lobby completes.

## Test plan

- `yarn tsc --noEmit -p tsconfig.json` and `-p tsconfig.browser.json`: clean. `yarn lint:imports`: clean.
- New and updated tests (no mocks, real transports through the harness):
  - `test/transport/HolepunchBanPolicy.test.ts`: the policy cases now walk the whole ladder — ALLOW leaves no exclusion or ban; BLACKLIST excludes and bans; expected close is ALLOW; ALLOW on WebRTC still releases the Holepunch fallback; SUSPEND bans and bars without recording a verdict; retry-tier closes below the bound leave the peer re-admittable; the close that reaches the bound bars and bans it; a barred peer's reconnect is refused; a second manager holding the same profile forgets the bar but keeps the blacklist. The thirteen pinned ban-sequence cases are unchanged.
  - `test/rpc/guards/HandshakeCompletedGuard.test.ts`: the two waiter-expiry cases now read the session bar instead of the profile verdict; the non-negotiating case still reads the verdict.
  - `test/e2e/E2E-InitHandshake.test.ts`: a peer with a skewed request time is closed and reconnects for two rounds without any penalty, then is barred and disconnected on the round that reaches the bound; the response-timeout and response-timestamp cases now assert neither a blacklist nor a bar.
  - `test/rpc/lobbyMatching/LobbyMatchingService.test.ts`: excluded peer refused on pick, kept out of the candidate set, set cleared at session end and kept across a released handoff, topic left before a non-selected close, selected transport kept.
  - `test/e2e/E2E-LobbyMatching.test.ts`: no peer re-dialed during negotiation on plain and targeted topics; both sides excluded after commitment silence and not re-matched; remote abort treated as a lobby exit.
  - `test/rpc/openChannelNegotiation/OpenChannelNegotiationService.test.ts`: lost committed peer excluded instead of blacklisted; former "blacklisted" expectations flipped.
- Mutation checks, each in a clean checkout with one core line reverted and its proving test run alone on the pool, then restored: leave-before-cut order, do-not-rematch admission, expected-close policy, and the remote-abort split each go red on exactly their test and green on restore. The ban-policy suite's thirteen pinned sequences are unchanged.
- Distributed pool, touched suites (earlier round): `--grep` over every touched suite (P2PManager, ProfileManager ban policy, transports, lobby matching unit and e2e, channel negotiation, init handshake unit and e2e, completed guard, RPC router, fork-dispute e2e): 195 tasks, all passing after two fixture fixes to the new lobby staging (one lobby e2e case recovered from starvation on its second run, which is the runner's normal handling).
- Full gate `yarn test:parallel:distributed` on d949a742: 2414 passing, 1 failing, 5 tasks recovered from starvation on their second run (the runner's normal handling). The one failure, a spectate join case this change does not touch, failed with every peer's runtime child disposed at once and passes 2/2 in isolation on the pool. An earlier full run on an intermediate head had a single failure in the ping-service probe case caused by a router change that was then dropped from this PR.
- Distributed pool for the tier work: `--grep "(ProfileManager Holepunch ban policy|P2PManager disconnect policy|HandshakeCompletedGuard|E2E: Init Handshake)"`: 45 passing, 0 failing, no starved or retried tasks (`logs/run-18`, wall 152s).
- Spec: REQ-RPC-6 now states the four-outcome ladder, including the counted outcome's one-bound-counter-per-peer rule; REQ-AUTH-4 separates a clock fault, which spends a bound before it bars, from malformed or invalid-signature traffic, which still records a verdict. New REQ-LOBBY-10 for the session-scoped do-not-rematch set, REQ-LOBBY-9 owns leaving the topic before the non-selected closes; REQ-NEG-4 splits step violations from remote abort. Implementation mirrors updated for `DisconnectPolicy` (added), `P2PManager`, `ProfileManager`, `PeerProfile`, the handshake service and endpoints, and the completed guard; verification reports re-anchored for the ban-policy, completed-guard and init-handshake suites, and their test declarations follow. `yarn spec:refresh` is deterministic (second run byte-identical), `yarn spec:ids:check` reports 0 issues, `yarn spec:impact --base origin/dispute` exits 0 with no unmapped paths. Newly planned permutations that have no test yet are tracked as gaps (the already-open and burned-window negotiation cases assert only "not blacklisted", not the bar). OQ-34's failure-outcome-consistency bullet is resolved; sync timeout and dispute-ack silence stay outside the policy per DEF-5. The approvals register is untouched and awaits engineer review.


sh3ll3x3c:
@lukastanisic99 this replaces #474. It does what the card asked, nothing more.

`disconnectConnection` now takes a required policy: ALLOW (close, peer can reconnect) or BLACKLIST (close and exclude). Every call site was audited and states its tier. Faults stay BLACKLIST. Timeouts, clock skew, guard expiry, and a polite negotiation abort become ALLOW, because a timeout is not proof of fault.

The lobby now leaves its topic before it cuts the non-selected peers and keeps the matched connection, as you proposed. Peers that drop mid-commit or burn an opening window go into a session-scoped do-not-rematch set instead of the blacklist. No suspension tier.

Persistence is not in this PR. There is no storage layer in the SDK yet, so a durable blacklist and its reason are a follow-up.

One thing that differs from your note: REQ-AUTH-4 no longer excludes a verified-but-silent peer. Sync timeout and dispute-ack silence still blacklist per DEF-5.

src/rpc/network/services/initHandshake/InitHandshakeRpcMethods.ts
```diff
@@ -71,7 +72,12 @@ class InitHandshakeRpcMethods extends ANetworkRpcMethods<InitHandshakeService> {
                     reason: "request time outside agreement window"
                 }
             );
-            this.p2pManager.disconnectAndBlacklistPeer(this.senderTransport);
+            // Clock skew is an environment fault, not misbehaviour: refuse the
+            // handshake and close, but do not blacklist the peer.
+            this.p2pManager.disconnectConnection(
+                this.senderTransport,
+                DisconnectPolicy.ALLOW
```
lukastanisic99: This can lead to an infinite loop even between honest peers if their perception of time is too skewed;

src/rpc/network/services/initHandshake/InitHandshakeService.ts
```diff
@@ -121,7 +122,10 @@ class InitHandshakeService extends ANetworkRpcService<InitHandshakeRpcMethods> {
                         ? `handshake response not received in time: ${error.message}`
                         : "handshake response not received in time"
             });
-            this.p2pManager.disconnectConnection(transport);
+            this.p2pManager.disconnectConnection(
+                transport,
+                DisconnectPolicy.ALLOW
```
lukastanisic99: again here, a byzantine peer just needs to not respond and we can try again in loops forever + this is all errors not just timeout

src/rpc/network/services/initHandshake/InitHandshakeService.ts
```diff
@@ -181,7 +185,10 @@ class InitHandshakeService extends ANetworkRpcService<InitHandshakeRpcMethods> {
                 agreementTimeSeconds: agreementTime,
                 reason: "response RTT outside agreement window"
             });
-            this.p2pManager.disconnectConnection(transport);
+            this.p2pManager.disconnectConnection(
+                transport,
+                DisconnectPolicy.ALLOW
```
lukastanisic99: same here

src/rpc/network/services/initHandshake/InitHandshakeService.ts
```diff
@@ -199,7 +206,12 @@ class InitHandshakeService extends ANetworkRpcService<InitHandshakeRpcMethods> {
                 agreementTimeSeconds: agreementTime,
                 reason: "response timestamp outside agreement window"
             });
-            this.p2pManager.disconnectAndBlacklistPeer(transport);
+            // Same class as the RTT check above: a skewed clock is an
+            // environment fault, so close without punishing the peer.
+            this.p2pManager.disconnectConnection(
+                transport,
+                DisconnectPolicy.ALLOW
```
lukastanisic99: This service was designed as a filter to not engage further with peers that have too skewed clocks of that have too large latency - I agree that blacklisting here is too aggresive but just allowing an infinite amount of attempts is also not the solution since we can go in circles even between honest peers; 

This is the same problem of 'we need some middle ground' - that middle ground is missing; That middle ground should be done once and just used alongise these other policies and not have every service implement its own custom sulution to it;

sh3ll3x3c:
Decision (2026-09-16), following the four handshake comments.

Agreed on all four. An ALLOW close on a peer that still shares a topic is followed by a redial, so the skew and timeout sites need a bound, and that bound belongs in one place, not in each service.

This PR shrinks. The handshake commit (the skew, RTT and ack-timeout sites, the guard expiry, and their tests and spec entries) comes out, and those sites go back to the `dispute` behaviour. #490 then carries only the explicit policy at every call site, the lobby leave-before-cut order, and the do-not-rematch set.

The handshake filter moves to a separate PR on `dispute`: one session-scoped SUSPEND tier owned by `ProfileManager` and applied through `P2PManager`, next to ALLOW and BLACKLIST, used at the four handshake sites and the guard expiry. The catch in `InitHandshakeService` is split so a timeout, a remote refusal and a transport close are handled as what they are. Link follows when it is open.

src/rpc/network/services/initHandshake/InitHandshakeRpcMethods.ts
sh3ll3x3c: Agreed. This site goes back to the `dispute` behaviour in this PR. The bounded SUSPEND tier for it lands in the follow-up PR, see the decision comment on the PR.

src/rpc/network/services/initHandshake/InitHandshakeService.ts
sh3ll3x3c: Agreed on both points. This catch receives the router timeout, a remote error (including the peer's own skew refusal), a transport close, and a send failure, and logs all of them as a timeout. It is unchanged from `dispute` in this PR. The follow-up PR splits it by cause and bounds the retries with SUSPEND.

src/rpc/network/services/initHandshake/InitHandshakeService.ts
sh3ll3x3c: Same as above: back to the `dispute` behaviour here, SUSPEND in the follow-up PR.

src/rpc/network/services/initHandshake/InitHandshakeService.ts
sh3ll3x3c: Agreed. One central session-scoped SUSPEND tier, owned by `ProfileManager` and applied through `P2PManager`, next to ALLOW and BLACKLIST. No service gets its own version. It lands in the follow-up PR on `dispute` together with the handshake sites. This PR keeps only the two tiers and the lobby fix.

sh3ll3x3c:
Follow-up PR with the session-scoped SUSPEND tier for the handshake filter: https://github.com/peer3to/state-channels-plus/pull/491

sh3ll3x3c:
Correction to the decision above (2026-09-16). The follow-up PR is error handling only: https://github.com/peer3to/state-channels-plus/pull/492 makes the handshake request catch branch on the router's failure cause and log each as what it is, with the tiers unchanged from `dispute`. The middle ground for the handshake filter is not decided by that PR; it stays an open design question here for @lukastanisic99. #491 was opened by mistake with a tier bundled in and is closed.

src/rpc/network/services/initHandshake/InitHandshakeRpcMethods.ts
sh3ll3x3c: Bounded now (5a09287e): this site states `allowRetry(InitHandshakeService.TIMING_RETRY_LIMIT)`. Each skewed request closes and counts against the peer for the session; the close that reaches the bound (3) suspends the peer, so two honest peers with skewed clocks stop after three rounds instead of looping.

src/rpc/network/services/initHandshake/InitHandshakeService.ts
sh3ll3x3c: Same bound here: `allowRetry(TIMING_RETRY_LIMIT)`, counted per peer for the session, suspended on the third. Handling by error cause is in #492 and stays separate from the tier.

src/rpc/network/services/initHandshake/InitHandshakeService.ts
sh3ll3x3c: Same: `allowRetry(TIMING_RETRY_LIMIT)`.

src/rpc/network/services/initHandshake/InitHandshakeService.ts
sh3ll3x3c: The middle ground is in now, once, in 5a09287e: `DisconnectPolicy` is the ladder ALLOW / allowRetry(n) / SUSPEND / BLACKLIST. `ProfileManager` owns the session suspension set and the per-peer retry counters next to the persisted blacklist; `P2PManager.disconnectConnection` applies the tier. SUSPEND bans the Hyperswarm info and refuses reconnects for this session only, nothing written to the profile; BLACKLIST is SUSPEND plus the persisted record. This site and the post-proof ack timeout use SUSPEND, the three timing checks use allowRetry(3) with the bound in one constant. Tests in 79f840b9 walk the whole ladder and the three-round handshake case.

PR #491, opened 2026-09-16
fix(rpc): suspend instead of blacklist on handshake timing failures

sh3ll3x3c:
## Summary

The init handshake is a filter against peers whose clocks disagree with ours or
whose latency does not fit the agreement window. Blacklisting them was too
aggressive — a skewed clock is an environment fault, not proof of misbehaviour —
but a plain close is not enough either: an honest pair with skewed clocks keeps
redialling each other, and a silent byzantine peer keeps redialling us.

This adds the missing middle ground once, centrally.

**A session-scoped suspension, owned by `ProfileManager`.** `suspendPeer(transport | address)`
marks the profile and bans its Hyperswarm peer info exactly as `blacklistProfile`
does, so the peer is not redialled for the rest of this runtime. It is a separate
fact from the blacklist: `PeerProfile.isSuspended` sits next to `isBlackListed`,
a suspended peer is never blacklisted, and there is no release path — a future
persisted blacklist inherits nothing from it. Admission refuses a suspended
identity the same way it refuses a blacklisted one, and re-applies the matching
exclusion to the arriving transport so its fresh peer-info handle is banned too.
Every site that releases a Holepunch ban keeps it while a suspension holds.
`P2PManager.disconnectAndSuspendPeer` is the single entry point, so no service
assembles its own variant.

**The handshake sites that move from blacklist to suspend:** the request-time skew
check in `InitHandshakeRpcMethods`, the response round-trip and response-timestamp
bounds in `InitHandshakeService`, the post-proof ack timeout, and the
`HandshakeCompletedGuard` expiry. The ack timeout was the worst of these — it
called `blacklistPeer(address)` on an address that has no profile yet, so it
neither punished nor closed anything; it now closes and suspends by transport.
Malformed input and an undecodable signature still blacklist.

**The request leg branches on why it failed, not on message text.** `ARpcRouter`
attaches a small named union to the standard `Error` `cause` of every rejection
it creates — `request-timeout`, `remote-error`, `transport-closed`,
`send-failed` — and `InitHandshakeService` reads it back: a timeout or a remote
refusal leaves a peer to suspend, an already-closed transport or a failed send
leaves nothing, and each is logged as what it is rather than all as
"response-timeout".

**The refusal now reaches the peer before the close.** The skew branch used to
disconnect and then throw, so the caller saw a bare disconnect instead of our
error response and had nothing to act on. The throw now happens first and the
close-and-suspend runs on the next macrotask, after the reply is on the wire, so
the other side suspends us too and neither end redials.

## Judgment calls

- **Suspension is not persisted and has no unsuspend.** It is a statement about
  this runtime, not a verdict, so it deliberately does not join the blacklist's
  (future) persistence story. `unblacklistPeer` therefore keeps the peer-info ban
  while a suspension holds.
- **`InitHandshakeService`'s pre-verification exclusion check gained a suspended
  branch of its own** rather than folding suspension into the existing
  blacklisted branch. The blacklisted branch does a plain close; for a suspended
  signer that would leave the new connection's handle unbanned and the peer
  looping, which is the bug being fixed, so the suspended branch re-applies the
  suspension. Blacklist behaviour there is unchanged.
- **The deferred close uses the existing `TimeoutManager` at zero delay.** The RPC
  reply is written from the rejection's own microtask chain, so a macrotask is
  the first point that is guaranteed to run after it, and `TimeoutManager` is the
  repo's disposal-aware scheduler rather than a raw `setTimeout`.
- **`tsconfig.json` gains `lib`, but only `es2022.error`.** The target stays
  `es2020`; this adds the type for the standard `new Error(msg, { cause })`
  option and nothing else, rather than widening the whole library surface.
- **`RelayAdmissionProbe` and the guard probe gained `profileSuspended` /
  `firstSuspended` / `originalSuspended` fields** so every timeout assertion reads
  the two flags apart. A test that only checked "not blacklisted" would pass on a
  plain close and miss the suspension entirely.

## Test Plan

Gates: `yarn tsc --noEmit -p tsconfig.json`, `yarn tsc --noEmit -p tsconfig.browser.json`,
`yarn lint:imports`, `yarn spec:refresh`, `yarn spec:ids:check`, `yarn spec:impact --base <merge-base>`
— all clean.

`yarn test:parallel:distributed --grep "InitHandshake|Init Handshake|HandshakeCompletedGuard|ProfileManager|P2PManager"`
— **94 passing, 0 failing** (wall 300s).

New unit coverage, `test/rpc/initHandshake/InitHandshakeSuspension.test.ts`, drives the
real handshake over a recording Holepunch socket so each case is a genuine wire
exchange:

- a verified peer whose ack never arrives is closed, banned and suspended, and is
  not blacklisted;
- an unanswered challenge (request timeout) suspends;
- a refused challenge (remote error response) suspends;
- a pending request rejected as a closed transport suspends nothing — no ban, no
  close;
- a skewed inbound request has its error response on the wire *before* the socket
  is destroyed. The socket records how many frames it had written at destroy
  time, so closing first turns this red.

`test/transport/HolepunchBanPolicy.test.ts` adds the `ProfileManager` side:
suspending an unauthenticated profile bans its handle without blacklisting it,
and a suspended identity dialling back on a new connection is refused and
re-banned at admission.

`test/rpc/guards/HandshakeCompletedGuard.test.ts` now asserts the expired-deferral
transport is suspended and *not* blacklisted; the never-negotiating branch still
blacklists, so swapping the two turns both red.

`test/e2e/E2E-InitHandshake.test.ts` asserts the peer-observable outcome for the
request-skew, no-response and response-skew cases: closed and suspended, not
blacklisted, with the observer's status unchanged.

Specification: `REQ-AUTH-4` now separates the penalty (still requiring proof)
from the session-scoped withdrawal that every timing failure produces, and
requires the refusal to precede the close. Implementation mirrors, conformance
rows, component test obligations and verification reports follow for every
touched file and test.


sh3ll3x3c:
@lukastanisic99 this is the follow-up from #490: the handshake filter gets the central middle ground you asked for. Based on `dispute`, independent of #490 except for a small overlap in `P2PManager` and `ProfileManager`.

sh3ll3x3c:
Closing: this bundled a new tier with the error-handling fix. The error-handling fix alone is https://github.com/peer3to/state-channels-plus/pull/492. The tier question stays open on #490.

PR #492, opened 2026-09-16
fix(rpc): handle each handshake request failure by its cause

sh3ll3x3c:
## Summary

The catch around the init-handshake challenge request received every kind of
rejection — a deadline, an error response from the peer, a transport that closed
mid-request, a send that never left — and logged and handled all of them as
`response-timeout`. Two of those readings were wrong, and one acted on a
connection that was already gone.

`ARpcRouter` now carries the kind of failure on each rejection it produces, as a
small named union (`request-timeout`, `remote-error`, `transport-closed`,
`send-failed`) that `getRpcRequestFailureCause` reads back — no error subclass
and no message matching. The errors the router creates itself carry their kind
on the standard `Error` `cause`. A transport close rejects with the reason its
caller supplied, unchanged in identity, class and fields, because callers close
with restored or typed errors that the caller still owns; those reasons are
recorded in a weak set the kind is read back from instead.

The handshake initiator branches on that cause. A timeout, a remote error, and a
failed send still close the connection exactly as before. A transport that was
already closed when the request settled is left alone — there is nothing to
close. Each cause is logged as what it was, so a refusal no longer reads as a
timeout. None of these failures is attributable, so none punishes the peer.

`tsconfig.json` gains `es2022.error` in `lib` for the standard `cause` option.

## Test Plan

- New unit suite drives a real Holepunch transport through each way the request
  can end without a response, asserting the close (or its absence) and that no
  ban or blacklist is recorded.
- New router case reads back the cause of one real rejection per kind and checks
  no pending entry or timer is left behind.
- Both tsc projects, `yarn lint:imports`, `yarn spec:refresh`, and
  `yarn test:parallel:distributed --grep "InitHandshake|NetworkRpcRouter"`
  (17 passing).
- The router case also asserts a transport close rejects the pending caller with
  the supplied `Error` object itself:
  `yarn test:parallel:distributed --grep "InternalTransport|NetworkRpcRouter|InitHandshake"`
  (25 passing).


sh3ll3x3c:
Closing. #490 now resolves the bounded tier to a plain ALLOW when the transport is already closed, so a closed transport no longer counts as a handshake timeout, which was the harm this change addressed. Timeout, remote error and send failure all take the same retry tier on dispute, so branching on the failure cause would not change any tier decision. The only remaining difference is the reason text in the handshake log; if that is wanted it is a small change in the catch block on dispute with no router change.

